PR #689, opened 2026-04-15
fix: CLI hard errors for invalid inputs, silent ignores, and startup ordering

morozsm:
## Summary

CLI audit revealed 18 issues across 5 categories. All fixed in a single commit.

**Critical (crashes/silent fails):**
- `discover --serial-only --lan-only` → `IndexError` (empty tasks list indexed); now returns clear error
- `proxy` failure → raw Python traceback; now caught with `Error: proxy failed: …`

**Startup order:**
- `serve`: port conflict now detected _before_ radio connects (was after); matches `web` behavior

**Input validation:**
- `$ICOM_PORT` / `$ICOM_SERIAL_BAUDRATE` non-numeric: traceback at argparse build time → clear `Error: $VAR must be an integer`
- `freq "14.074.000"`: `could not convert string to float` → `invalid frequency — use Hz/kHz/MHz`
- `date "2024-01"`: `not enough values to unpack` → `expected YYYY-MM-DD`
- `time "25:99"`: no range check → validates hour 0–23, minute 0–59
- `power 999`: no range check → validates 0–255
- `scope --output /bad/dir/scope.png`: path validated _before_ expensive capture (was after)
- `--tls-cert` / `--tls-key`: file existence checked before server starts

**Silent ignores:**
- `--backend serial --host X`: host silently dropped → prints `Warning: --host is ignored`
- `--backend lan --serial-port X`: serial-port silently dropped → prints `Warning: --serial-port is ignored`

**Audio / web:**
- `audio bridge` without `[bridge]` extra: generic ImportError → `install: pip install 'icom-lan[bridge]'`
- `audio bridge`: removed dead duplicate `CAP_AUDIO` check (unreachable code)
- `web --rigctld`: startup failure swallowed silently → caught with explicit error

**Minor:**
- Presets: web-only keys (`web_bridge`, `web_rigctld`) now skipped for `serve` command
- `--timeout` help text clarified: "Connection and auto-discovery timeout"
- `_cmd_discover` type annotation: `Radio | None` (always called with `None`)

## Test plan

- [ ] `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — zero failures
- [ ] `uv run ruff check src/icom_lan/cli.py` — clean
- [ ] `icom-lan discover --serial-only --lan-only` → clear error, exit 1
- [ ] `ICOM_PORT=abc icom-lan status` → clear error, exit 1
- [ ] `icom-lan freq "14.074.000"` → actionable error with Hz/kHz/MHz hint
- [ ] `icom-lan serve` when port 4532 is busy → error before radio connects

### DIFF
--- a/src/icom_lan/cli.py
+++ b/src/icom_lan/cli.py
@@ -65,6 +65,18 @@ def _get_env(name: str, default: str = "") -> str:
     return os.environ.get(name, default)
 
 
+def _env_int(name: str, default: int) -> int:
+    """Read an integer from an environment variable; exit with a clear error if non-numeric."""
+    val = os.environ.get(name, "")
+    if not val:
+        return default
+    try:
+        return int(val)
+    except ValueError:
+        print(f"Error: ${name} must be an integer (got {val!r})", file=sys.stderr)
+        sys.exit(1)
+
+
 # ---------------------------------------------------------------------------
 # Auto-discovery helpers
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


morozsm:
### Code review

Found 1 issue:

**`_cmd_scope`: writability probe deletes an existing output file**

The path validation touches then immediately unlinks the file to test writability. If the user points `--output` at an existing file (e.g. re-running scope capture to overwrite a previous result), `touch()` succeeds silently and `unlink()` **deletes the original before capture starts**. If the capture subsequently fails, the file is gone with no recovery.

The same pattern was used correctly in `_cmd_audio_rx` (ln 1503–1504, pre-existing), but there the output is always a fresh file. For scope it's a regression because the default output is `scope.png` — a path users are likely to reuse.

Fix: replace the touch/unlink probe with a directory write-permission check:

```python
import os
if not os.access(out_path.parent, os.W_OK):
    print(f"Error: output directory is not writable: {out_path.parent}", file=sys.stderr)
    return 1
```

https://github.com/morozsm/icom-lan/blob/f20ae4c3cc01f7a99feee520a2f04a68994008ad/src/icom_lan/cli.py#L2183-L2191

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

morozsm:
Fixed in 6ec2e99: replaced `touch()`/`unlink()` with `os.access(out_path.parent, os.W_OK)`.

Issue #1385, opened 2026-05-03
epic: diagnostic data collection — bundle generator + extension point + opt-in upload


morozsm:
## Goal

Provide a one-click "Send report" UX for icom-lan users that collects all information typically useful for debugging any icom-lan issue, packages it as a redacted bundle, and (optionally, opt-in only) uploads it to a maintainer-operated triage service.

This is the **A** subsystem of a 4-subsystem feature; sibling work (Pro client, DO backend, AI triage) happens in morozsm/icom-lan-pro.

## Architecture

- **Always-on diagnostic logging** via `SafeRotatingFileHandler` at DEBUG level (~15 MiB rotation cap, 5 MiB × 3 files). Best-effort: any I/O failure during init or emit is silently swallowed; the application never crashes because of diagnostic logging.
- **`DiagnosticContributor` protocol** — modular contributors collect data into a shared `BundleContext`-aware bundle directory.
- **Built-in contributors**: `system`, `invocation`, `radio`, `audio`, `logs`, `state`, `errors`, `dependencies`, `config`.
- **Extension point**: setuptools entry point group `icom_lan.diagnostics` + runtime register helper. icom-lan-pro plugs in here for its own contributors (Tauri logs, Rust crashes, RC-28, DSP).
- **Bundle assembly**: ZIP file with per-contributor subdirectories, `manifest.json`, schema versioning (`icom-lan-bundle-v1`).
- **Three trigger surfaces** all call the same `build_bundle()` API:
  - CLI: `icom-lan diagnose [--upload]`
  - Web UI: "Send report" button + consent dialog → POST `/api/v1/diagnose/send`
  - Pro Tauri UI: own button → POST `/api/local/v1/diagnose/send` → internal call to same `build_bundle()`
- **Opt-in upload**: anonymous POST to a configurable endpoint (default `https://reports.msmsoft.net/v1/diagnostics/upload`; override via `ICOM_LAN_REPORT_ENDPOINT` env var). Preview + explicit consent before send. No automatic background submission, ever.
- **PII redaction**: regex-based scrubbers for absolute paths, IPv4/IPv6, credentials, tokens, signing keys. Applied by built-in contributors; extension contributors can call the same shared utilities.

## Bundle metadata schema (`icom-lan-bundle-v1`)

```json
{
  "schema_version": "icom-lan-bundle-v1",
  "submission_id": "uuid-v4",
  "app": { "name": "icom-lan", "version": "0.20.0", "build_id": "..." },
[...]

PR #1386, opened 2026-05-03
docs(#1385): diagnostic data collection design spec

morozsm:
## Summary

Comprehensive design document for the open-core (icom-lan) side of the cross-repo diagnostic data collection feature. Captures every decision made during the brainstorming session before implementation begins.

## Scope

This spec covers subsystem **A** of a 4-subsystem feature (A=icom-lan, B=Pro client, C=DO backend, D=AI triage). B/C/D live in icom-lan-pro and have their own epic + sub-issues there.

## What's in the spec

- **Architecture overview** with data flow.
- **Components** — `SafeRotatingFileHandler`, `DiagnosticContributor` protocol, discovery (entry points + runtime register), built-in contributors, PII redaction, bundle assembler, HTTP upload client, CLI, Web UI handler + frontend.
- **Bundle layout & manifest schema** — `icom-lan-bundle-v1`, required vs optional fields.
- **Trigger surfaces** — CLI (interactive default + flags), Web UI (Settings panel + modal), Pro Tauri (delegates to open-core REST endpoint).
- **Privacy invariants** — no automatic submission, no first-run prompts, contact opt-in only, headless mode unchanged.
- **Error handling** — fail-safe logging, per-contributor try/except, typed upload errors.
- **Testing strategy** — unit + integration + privacy-invariant fuzz + Web UI e2e.
- **Open-core boundary** — no Pro imports; extension point is pure protocol + entry points.
- **Cross-repo coordination** — links to Pro epic, contract, sub-issues.
- **Sub-issues breakdown** — 15 atomic implementation tasks (all within autonomous-pipeline guardrails).
- **Out of scope** — error overlay, crash hook, AI triage, Pro Tauri UI, multi-language.

## Key design decisions

1. **Always-on rotating file logging** at DEBUG level (~15 MiB cap) via `SafeRotatingFileHandler` that swallows all I/O errors silently.
2. **Maximalist data collection with auto-redaction** — paths, IPs, credentials, tokens regex-scrubbed. LAN passwords / audio PCM / .env / keychain dropped entirely (never masked).
3. **Setuptools entry points** for Pro discovery — zero-config plug-in pattern (pip install Pro → contributors auto-loaded).
4. **Interactive CLI by default** when on TTY, full flag set for scripted use.
5. **Web UI: Settings panel + modal** with form / preview / result screens. Error overlay deferred to v2.
6. **Generic upload client with `request_signer` hook** — open-core stays anonymous; Pro injects Authorization header via the hook.
7. **Default endpoint hardcoded** to `https://reports.msmsoft.net/v1/diagnostics/upload`; overridable via `ICOM_LAN_REPORT_ENDPOINT`.
8. **No PII auto-collection** — `contact.email` / `contact.callsign` are opt-in user-supplied at submission time only.

## Cross-repo

- Tracking issue: #1385
- Pro epic (B/C/D): morozsm/icom-lan-pro#583
- Contract: morozsm/icom-lan-pro#584 (POST /v1/diagnostics/upload, Required vs Optional fields)

## Next steps

After approval:

1. Invoke `writing-plans` skill to produce an atomic implementation plan from this spec.
2. File 15 sub-issues in icom-lan from §12 of the spec.
3. Run sub-issues through the autonomous pipeline (`/solve-issue N` per task).

## Test plan

- [x] Spec self-review pass: placeholder scan, internal consistency, scope check, ambiguity check.
- [ ] Reviewer feedback: please flag anything unclear, missing, or over-scoped.

### DIFF
--- a/docs/architecture/open-core-policy.md
+++ b/docs/architecture/open-core-policy.md
@@ -59,6 +59,52 @@ SWR, and similar meter readings surfaced in `radio_poller.py` and the
 not analytics. It does not leave the user's machine and is not subject to
 this rule.
 
+### Carve-out — user-initiated diagnostic support reports
+
+A diagnostic report workflow that lets a user **explicitly send a bundle of
+local logs and state** to a maintainer-operated triage endpoint is
+permitted. Telemetry is forbidden because it is automatic, ambient, and
+invisible; a support report is the opposite of those things — the user
+authored it, knows what's in it, and chose to send it.
+
+A diagnostic report mechanism in open-core must satisfy **all** of these:
+
+- **No automatic, background, recurring, first-run, or silent collection or
+  submission.** Logs may be written locally on a rotating file basis (see
+  `SafeRotatingFileHandler` in `icom_lan.diagnostics`), but submission to a
+  remote endpoint is never automatic.
+- **The user must explicitly start the workflow every time.** No "always send
+  reports" toggle, no install-time consent that grants ongoing permission, no
+  background queue. Each report = one user gesture (CLI command, Web UI
+  button, Pro Tauri click).
+- **CLI defaults never upload.** `icom-lan diagnose` (no flags) builds and
+  saves locally. `--upload` is opt-in. With a TTY, the final consent prompt
+  defaults to "save locally" (`[y/N]` for upload), so accidentally hitting
+  Enter cannot transmit anything.
+- **Headless / non-TTY mode never prompts and never auto-uploads.** Scripted
+  use must explicitly pass `--upload` to send.
+- **Destination endpoint is visible before upload.** The full URL (including
+  any `ICOM_LAN_REPORT_ENDPOINT` override) appears in the preview shown to
+  the user.
+- **Bundle is always available locally.** The user can always choose "save
+  locally" instead of uploading; the ZIP file is the same.
+- **Contact fields are opt-in only.** Email and callsign are never
+  auto-populated from environment variables, system identity, prior reports,
+  or any ambient source — only typed by the user at submission time.
+- **The endpoint behaviour is documented as a public contract.** The shape
+  of the anonymous request lives in `docs/contracts/diagnostic-bundle-v1.md`,
+  so the open-core upload client builds against a non-proprietary spec.
+
+This carve-out does **not** allow:
+- Crash reporters that send automatically on uncaught exceptions.
+- "Send anonymous usage stats" toggles.
+- Rate-limited periodic uploads.
+- First-run consent dialogs that authorise future automatic sends.
+- Pre-filling of contact fields from any source.
+
+If a future feature does not fit cleanly inside this carve-out, the default
+is the broader §2 rule: no telemetry, no exceptions.
+
 ---
 
 ## 3. Headless mode is sacred
--- a/docs/plans/2026-05-03-diagnostic-data-collection-design.md
+++ b/docs/plans/2026-05-03-diagnostic-data-collection-design.md
@@ -115,7 +115,8 @@ def configure_diagnostic_logging() -> None:
         )
         handler.setLevel(logging.DEBUG)
         handler.setFormatter(_DIAGNOSTIC_FORMATTER)
-        logging.getLogger().addHandler(handler)
+        # Attach to the `icom_lan` logger, NOT root — see "Logger scope" below.
+        logging.getLogger("icom_lan").addHandler(handler)
     except Exception as exc:
         sys.stderr.write(
             f"icom-lan: diagnostic logging disabled: {exc}\n"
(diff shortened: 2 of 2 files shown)


morozsm:
Reviewed the spec with the open-core/no-telemetry policy in mind. I don't think the feature needs to be de-scoped: helping users send a diagnostic report is useful, and the intended model is already explicit/user-initiated rather than telemetry.

I would tighten the spec in a few concrete places so the intent is hard to misread later:

1. Add a small carve-out to `docs/architecture/open-core-policy.md` for **user-initiated diagnostic support reports**. Something along these lines:

   - telemetry remains forbidden: no automatic, background, recurring, first-run, or silent collection/submission;
   - a diagnostic report workflow is allowed only when the user explicitly starts it;
   - upload requires an explicit action every time;
   - CLI/headless defaults never upload;
   - the destination endpoint is visible before upload;
   - the user can save the bundle locally;
   - contact fields are never auto-filled from env/config/system identity.

   This keeps the policy honest without turning it into bureaucracy. The current policy wording is stricter than this spec, so a reviewer can reasonably read the upload client as conflicting with it.

2. Change the CLI confirmation from `Send to <endpoint>? [Y/n]` to `[y/N]`, and make Enter save locally. Upload can still be easy, but it should not be the default keystroke for a privacy-sensitive report.

3. Clarify that `icom-lan diagnose` without `--upload` builds/saves locally. A non-CLI user still gets the Web UI button; the CLI just should not surprise headless/scripted users.

4. For Web UI send/save endpoints, add one sentence about abuse resistance. Since the web server can run on `0.0.0.0` with no auth token, the diagnostics `send` endpoint should not be triggerable by another LAN client just by posting `{preview_id, consent: true}`. Reasonable options: require API auth for diagnostics upload when not bound to localhost, add a same-origin/CSRF-style guard, or disable upload from unauthenticated remote web sessions.

5. Reword “no automatic data collection” to distinguish local bounded diagnostic logs from submission. The design is really “no automatic submission, ever”. If always-on logs stay, scope them to the `icom_lan` logger/app entry points rather than root logger/library import, so we don't accidentally capture unrelated application logs when icom-lan is imported as a library.

6. The `request_signer(aiohttp.ClientRequest)` hook is probably too coupled to aiohttp internals, and aiohttp is currently only a dev dependency. A simpler public contract like an async header provider / request metadata signer would be easier for Pro to implement without making aiohttp part of the extension API.

Net: I support the direction. I would just make the policy/spec explicitly say “user-commanded support report is allowed; telemetry is still forbidden”, then adjust the default interactions so that interpretation remains true in code.

morozsm:
Addressed all six points in commit cd9770e3.

| # | Point | Resolution |
|---|---|---|
| 1 | Open-core policy carve-out | Added "Carve-out — user-initiated diagnostic support reports" section to `docs/architecture/open-core-policy.md` §2 with the full bullet list (no automatic/background/first-run/silent; explicit user gesture every time; CLI defaults never upload; consent default is save-locally; endpoint visible before upload; bundle always available locally; contact fields opt-in only; endpoint behaviour documented as a public contract). Plus an explicit "this carve-out does NOT allow" list (auto crash reporters, "send anonymous stats" toggles, periodic uploads, first-run consent dialogs that authorise future sends, contact pre-fill). |
| 2 | CLI default `[Y/n]` → `[y/N]` | Fixed in §4.8. Pressing Enter saves locally; only an explicit `y` triggers upload. |
| 3 | Clarify CLI default behaviour | `icom-lan diagnose` without `--upload` builds and saves locally — never prompts, never uploads. Added: non-TTY with `--upload` but without `--no-confirm` also saves locally (confirmation cannot be obtained without a TTY → we refuse to silently transmit). |
| 4 | Web UI abuse resistance | §4.9 now specifies same-origin enforcement on `send`/`save`/`DELETE preview`, preview-bound CSRF token in `X-Diagnostic-CSRF` header, loopback exception for `Origin`, and inheritance of existing API auth on remote-bound deployments. Net: cross-origin LAN page can't drive a CSRF-style upload, and a misconfigured 0.0.0.0 deployment with API auth stays protected. |
| 5 | Logger scope + wording | `SafeRotatingFileHandler` attaches to `logging.getLogger("icom_lan")`, not root — host-app loggers stay out of our file when icom-lan is used as a library. Added a paragraph distinguishing "no automatic submission" (the policy) from "no local bounded logs" (which are fine). |
| 6 | `request_signer(aiohttp.ClientRequest)` → header provider | Replaced with `HeaderProvider = Callable[[], Awaitable[dict[str, str]]]`. Pro returns a dict of headers, fully decoupled from icom-lan's HTTP client choice. On 401 we call the provider again before retrying — Pro's implementation can refresh on call without external coordination. |

Direction unchanged; the model is "user-commanded support report is allowed; telemetry is still forbidden", and the defaults now make that interpretation hold in code as well as in prose.

Issue #52, opened 2026-02-28
Status packet parsing: 5s delay when civ_port=0, no retry logic


morozsm:
## Bug
When radio returns `civ_port=0, audio_port=0` in status packet, the code waits ~5 seconds before falling back to defaults (50002/50003).

## Debug log
```
20:05:11,683 Status: civ_port=0, audio_port=0
20:05:16,781 CI-V port not in status, using default 50002
```

5 seconds wasted waiting for a retry that never comes. This adds to connect time and compounds with other timeouts.

## Expected
- If status returns port=0, immediately fall back to defaults (no 5s wait)
- Or: request status again once before fallback
- Log at WARNING level when using fallback ports

## Related
#49

## Definition of Done
- [ ] Regression test written (must fail without fix)
- [ ] Fix implemented
- [ ] Test passes
- [ ] Full test suite passes

Issue #19, opened 2026-02-26
enhance(hamlib): structured logging + guardrails for rigctl server


morozsm:
## Goal
Improve operability and safety diagnostics for Hamlib TCP clients.

## Scope
- Structured per-command logs:
  - `timestamp`, `client_ip`, `cmd`, `args`, `duration_ms`, `rprt`.
- Configurable log level.
- Basic guardrails:
  - connection limit,
  - optional command rate limiting/throttling,
  - graceful handling of malformed input.

## Acceptance Criteria
- Logs are machine-readable and useful for troubleshooting client behavior.
- Guardrails do not break normal `rigctl`/WSJT-X usage.
- Tests for malformed input and overload edge cases.

## Dependencies
- Depends on server skeleton issue.

morozsm:
Status update (v0.6.6):
- Circuit breaker + poller stats logging implemented
- Lazy poller lifecycle (start/stop with clients)
- Transition hold windows for packet mode switches
- Client connect/disconnect/timeout logging

Remaining:
- Configurable log levels per subsystem (e.g. --log-level rigctld=DEBUG)
- Structured JSON log output option
- Rate-limit log spam for repeated timeout patterns

PR #1512, opened 2026-05-15
ci(gates): migrate Agent Review and Rebrand gates to Mac mini runner

morozsm:
## Summary

Routes the two gate workflows in this repo (Agent Review Gate and
Rebrand grep gate) to the new Mac mini self-hosted Linux ARM64 runner
instead of `ubuntu-latest`. Comment refresh on both files pointing at
the operator doc.

## Why

This repo is public, so `ubuntu-latest` minutes are currently free, but
the migration is worth doing for two reasons:

1. **Parity.** All four `rigplane-*` repos converge on the same gate
   tier, so there is one routing token (`gate`) and one operator doc to
   reason about. PR #52 in `rigplane-tower` and the parallel migration
   PR #19 in `rigplane-www` already moved.
2. **Pickup speed.** GitHub-hosted `ubuntu-latest` jobs queue for tens of
   seconds before the runner is provisioned; the Mac mini container
   picks up in single-digit seconds for the API-only Agent Review Gate
   and sub-second for the Rebrand grep gate.

## Scope

Two workflow files, each one-line `runs-on` swap plus a comment:

- `.github/workflows/agent-review-gate.yml`: `ubuntu-latest` →
  `[self-hosted, linux, gate]`. API-only workflow.
- `.github/workflows/rebrand-gate.yml`: `ubuntu-latest` →
  `[self-hosted, linux, gate]`. Single `git grep` + bash filter.

No other workflows are touched: `quick.yml`, `full.yml`, `publish.yml`,
and `docs.yml` keep their existing runners — they need the build/test
toolchains and the Mac mini gate-tier image deliberately does not
include them. Phase 3 will introduce a separate `rigplane-ci-build-runner`
image and migrate those workloads.

## Verification

- `rp-gate-core` container is live on the Mac mini, runner
  `mm-gate-core` is online and registered against this repo with labels
  `[self-hosted, Linux, ARM64, gate, mac-mini]`.
- Container has `bash` 5.1, `git` 2.34.1, `grep` available (verified via
  `docker exec rp-gate-core which git bash grep`) — sufficient for the
  Rebrand grep gate's `actions/checkout@v6` and
  `.github/scripts/check-rebrand-allowlist.sh`.
- `uv run pytest`, `uv run ruff`, and import-linter not run — change is
  workflow YAML only, no Python imports affected.

## Layer / boundary check

CI workflows live in `.github/workflows/`, outside the `src/rigplane/`
layered packages. No layering matrix impact. No public/private boundary
move — this is internal CI plumbing.

## Open-core policy check

No telemetry, no headless-sacred violation, no Pro-vs-public hollowing.
Workflow YAML only.

### DIFF
--- a/.github/workflows/agent-review-gate.yml
+++ b/.github/workflows/agent-review-gate.yml
@@ -20,7 +20,10 @@ jobs:
   agent-review:
     name: Update Agent Review Gate status
     if: github.event_name == 'pull_request' || github.event.issue.pull_request
-    runs-on: ubuntu-latest
+    # Self-hosted Mac mini gate-tier runner. API-only workflow via
+    # actions/github-script -- no compile/test toolchain needed. See
+    # rigplane-tower/docs/development/mac-mini-self-hosted-runners.md.
+    runs-on: [self-hosted, linux, gate]
     timeout-minutes: 2
 
     steps:
--- a/.github/workflows/rebrand-gate.yml
+++ b/.github/workflows/rebrand-gate.yml
@@ -21,7 +21,10 @@ concurrency:
 
 jobs:
   grep-gate:
-    runs-on: ubuntu-latest
+    # Self-hosted Mac mini gate-tier runner. Sub-second git-grep + bash
+    # filter, no toolchain needed. See
+    # rigplane-tower/docs/development/mac-mini-self-hosted-runners.md.
+    runs-on: [self-hosted, linux, gate]
     timeout-minutes: 2
 
     steps:


morozsm:
Agent Review: PASS

## Scope

Two workflow files only; both swap `runs-on: ubuntu-latest` for `[self-hosted, linux, gate]` and add a clarifying comment pointing at `rigplane-tower/docs/development/mac-mini-self-hosted-runners.md`.

- `.github/workflows/agent-review-gate.yml` (line 26): API-only via `actions/github-script@v7` — no toolchain needed.
- `.github/workflows/rebrand-gate.yml` (line 27): single `git grep` + bash filter via `.github/scripts/check-rebrand-allowlist.sh` — only needs `git` and `bash`, both present in the gate-tier image.

## Cross-repo consistency

Label set `[self-hosted, linux, gate]` matches `rigplane-tower/.github/workflows/agent-review-gate.yml` line 26 exactly. Same `timeout-minutes: 2`. Comment style aligned.

## CLAUDE.md compliance

- **CI workflows section**: `quick.yml`, `full.yml`, `publish.yml` are explicitly the cost-tiered trio; this PR touches none of them, so the matrix/billing-aware policy is untouched. Gate workflows are orthogonal to that tier.
- **Architecture / Layer boundaries**: workflow YAML is not part of `import-linter` layer matrix; no `src/` import surface touched.
- **Open-core constraints**: no telemetry, headless code, public SDK, or `local-extensions/` boundary touched. Nothing sacred.

## Safety

- No secrets referenced; permissions block unchanged (`contents: read`, `issues: read`, `pull-requests: read`, `statuses: write`).
- `actions/github-script@v7` and `actions/checkout@v6` versions unchanged.
- Concurrency groups unchanged.
- Allowlist script (`check-rebrand-allowlist.sh`) is self-contained pure bash with `set -euo pipefail`, uses `git grep` only — no network, no toolchain dependency, runs sub-second.
- Rebrand allowlist already includes `.github/workflows/rebrand-gate.yml` (line 79), so the comment edit referencing the brand pattern indirectly is fine; the new comment text introduces no `icom[-_]lan` / `IcomLanError` token.

## Risk

Low. Worst case if the self-hosted runner pool is offline: both gates queue and eventually time out, blocking PR merges — same observable failure mode as a GitHub-hosted outage. Rollback is a one-line revert per file.

Recommend merge once required checks are green.

morozsm:
Post-merge smoke test: validate gate routing to mm-gate-core.

PR #1543, opened 2026-05-20
fix: support PCM audio fallback for WebView clients

morozsm:
## Summary
- add PCM16 TX microphone fallback for WebView clients without WebCodecs
- let browser RX clients request PCM16 when Opus decode is unavailable
- preserve profile-selected stereo codec behavior while keeping global-default mono fallback covered
- ignore local .agents workspace artifacts

## Tests
- uv run ruff check src/ tests/
- npm run check
- npm run build
- npm run test
- uv run pytest tests/ -q --tb=short --ignore=tests/integration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ config/
 .firecrawl/
 .playwright-mcp/
 .copilot-tmp/
+.agents/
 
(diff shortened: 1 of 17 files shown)


morozsm:
Agent Review: PASS

PR #1546, opened 2026-05-21
[codex] avoid spurious CI-V reconnect recovery

morozsm:
## What changed

When a CI-V soft reconnect is requested while the transport is already open and data is still flowing, treat the reconnect as a no-op, clear transient recovering state, refresh the watchdog timestamp, and emit a structured `civ.soft_reconnect.noop` log.

The normal recovery path still runs when the transport is stalled.

## Verification

Agent-run checks:

- `uv run pytest tests/test_radio_coverage.py -k soft_reconnect` -> 6 passed
- `uv run pytest tests/test_civ_rx_coverage.py -k watchdog` -> 10 passed
- `uv run pytest tests/test_radio_coverage.py tests/test_civ_rx_coverage.py` -> 254 passed
- `uv run ruff check src/rigplane/runtime/_control_phase.py tests/test_radio_coverage.py` -> passed
- `uv run pytest` -> 5980 passed, 112 skipped, 3 deselected, 9 xfailed, 1 xpassed

### DIFF
--- a/src/rigplane/runtime/_control_phase.py
+++ b/src/rigplane/runtime/_control_phase.py
@@ -455,7 +455,34 @@ class ControlPhaseRuntime:
         """Reconnect CI-V transport using existing control session."""
         h = self._host
         if h._civ_transport is not None:
-            logger.warning("soft_reconnect: CI-V transport already open")
+            now = time.monotonic()
+            last_data = getattr(h, "_last_civ_data_received", None)
+            transport_open = (
+                getattr(h._civ_transport, "_udp_transport", None) is not None
+            )
+            data_flowing = isinstance(last_data, (int, float)) and (
+                now - float(last_data)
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Rationale: the already-open CI-V soft_reconnect path now only no-ops when the UDP transport is open and recent CI-V data is flowing; stale already-open state remains marked as recovering, preserving the watchdog recovery path. The watchdog escalation path force-cleans CI-V before reconnect, so this guard does not block that recovery flow.

Checks run:
- uv run pytest tests/test_radio_coverage.py -k soft_reconnect
- uv run pytest tests/test_civ_rx_coverage.py -k watchdog
- uv run ruff check src/rigplane/runtime/_control_phase.py tests/test_radio_coverage.py
- uv run pytest tests/test_radio_coverage.py tests/test_civ_rx_coverage.py

PR #1561, opened 2026-05-22
Document PortAudio callback ring intent

morozsm:
## Summary
- documents the PortAudio TxStream callback ring contract for WSJT-X/BlackHole-style consumers
- calls out that write() enqueues quickly, the callback clocks playback, overflow drops old queued audio, and underrun fills silence

Refs rigplane/rigplane-pro#955.

## Tests
- uv run ruff check src/rigplane/audio/backend.py
- uv run pytest tests/test_audio_backend.py -q


morozsm:
Agent Review: PASS

PR #1571, opened 2026-05-22
chore(docs): add IndexNow key and post-deploy ping

morozsm:
## Summary

Adds Microsoft IndexNow integration for `rigplane.dev`:

1. Publishes the site key at `https://rigplane.dev/4f4bcd28cb8d426ca58ec08b5035541e.txt` by dropping the file into `docs/`. mkdocs already copies static assets verbatim into `site/` — verified locally that `mkdocs build --strict` includes the file alongside `CNAME`/`sitemap.xml`.
2. Adds a new `indexnow` job to `.github/workflows/docs.yml` that runs after `deploy` succeeds, fetches the freshly-published `sitemap.xml` from `rigplane.dev`, and POSTs the full URL list to `https://api.indexnow.org/IndexNow`. Current sitemap has 84 URLs — well within IndexNow's 10000-per-request limit.

## Why a literal key, not a secret

IndexNow keys are public by design — the key file is served at a publicly-resolvable URL so consumers can verify ownership. Storing the same value as a GitHub secret would be theater while making it harder to keep the workflow value and the published file in sync. The two are deliberately co-located in this PR so a future rotation requires editing both in one diff.

## Failure modes

- 4xx response from IndexNow surfaces as a workflow `::warning::` without failing the deploy. Lets a one-off bad request not block doc rollouts.
- 5xx response fails the job loud — those mean IndexNow itself is down and we want to know.
- Sitemap fetch retries 5 times with 10s delay to absorb GitHub Pages CDN propagation latency after a fresh deploy.

## Test plan

- [x] `uv run mkdocs build --strict` locally — key file ends up at `site/4f4bcd28cb8d426ca58ec08b5035541e.txt` with correct contents.
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on `docs.yml` — YAML valid.
- [x] sitemap.xml URL count: 84 (well under 10000).
- [ ] First post-merge deploy will trigger the new `indexnow` job; verify in the Actions log that the POST returns 200/202.

## Boundary

Open-core / docs-tooling — no proprietary impact. The IndexNow key itself is public by spec, so no secret material is added.


morozsm:
Agent Review: PASS

Reviewed against PR head `c424f3f6` covering both commits (`5a979d6c` initial + `c424f3f6` newline strip).

**IndexNow spec correctness:** POST payload has `host`, `key`, `keyLocation`, `urlList`; `Content-Type: application/json; charset=utf-8`; `urlList` filtered to `https://rigplane.dev/` only via `startswith` guard.

**Key file:** Exactly 32 bytes, no trailing newline after `c424f3f6` (`wc -c` = 32, `xxd` confirms ASCII ends at offset 0x1f with `541e`, no `0a` byte). Matches IndexNow's exact-match expectation for strict consumers (Bing).

**Workflow safety:** `indexnow` job is `needs: deploy`, which is `needs: build` — no path runs the ping if either build or deploy fails.

**No secret leakage:** IndexNow key is public by spec; no other sensitive material in the diff.

**Failure handling:** 4xx → `::warning::` annotation in Actions UI, non-fatal; 5xx → step failure. Right trade-off — deploy is the critical path, IndexNow is best-effort.

**mkdocs `exclude_docs`:** Lists `EXTENDED_PROTOCOL_RESEARCH.md`, `architecture/`, `contracts/`, `operations/`, `parity/`, `plans/`, `internals/rfc-audio-v1-mini.md`. The new `.txt` matches none — verified locally that `mkdocs build --strict` copies it to `site/4f4bcd28cb8d426ca58ec08b5035541e.txt` alongside `CNAME` and `sitemap.xml`.

**Trigger ordering:** `docs/` is in the `on: push` `paths` filter, so the first post-merge push will build+deploy the key file first, then `indexnow` fires after `deploy` completes. No chicken-and-egg.

**CDN headroom:** 30s sleep + `curl --retry 5 --retry-delay 10` totals ~80s. Empirically sufficient for GitHub Pages CDN propagation.

Issue #1576, opened 2026-05-23
Align docs and package metadata with Hamlib provider strategy


morozsm:
## Context

The Hamlib provider and assisted discovery strategy changes RigPlane's public docs posture:

- Hamlib is a replaceable backend adapter for long-tail CAT coverage.
- RigPlane capabilities remain the public product/platform contract.
- Native providers remain valuable where they give richer control, audio, discovery, lifecycle, or diagnostics.
- RigPlane should avoid promising universal auto-detection for serial CAT radios.

Related Linear project: https://linear.app/morozsm/project/hamlib-provider-and-assisted-discovery-db43161177b0/issues

## Scope

Update public core documentation and metadata so it no longer positions RigPlane as categorically "no Hamlib" or implies every new radio is just a TOML profile.

## Surfaces to keep aligned

- `README.md`
- `pyproject.toml` project description / PyPI-facing metadata
- `docs/index.md`
- `docs/guide/radios.md`
- `docs/guide/rig-profiles.md`
- `docs/guide/wsjtx-setup.md`
- Future Hamlib provider docs when implementation begins

## Acceptance criteria

- README and docs describe native providers plus planned Hamlib-backed long-tail CAT coverage.
- Xiegu X6100 and Lab599 TX-500 are represented as profile/candidate radios with assisted discovery and Hamlib-backed fallback planned.
- Rig profile docs distinguish native profile work from Hamlib provider/model metadata work.
- WSJT-X docs keep the client-facing `Hamlib NET rigctl` setup without claiming there is no Hamlib provider/dependency anywhere in the architecture.
- Public/private boundary is clear: generic provider contract, discovery candidates, safe probing, and docs belong in core; managed bundling/setup/diagnostics workflow belongs in Pro.

## Boundary

Classification: public-core candidate with boundary-review.

morozsm:
Related public website copy tracking issue: https://github.com/rigplane/rigplane-www/issues/91

morozsm:
Follow-up audit found and aligned additional core docs beyond the initial scope: docs/api/rigctld.md, docs/guide/cli.md, docs/guide/ic7610-usb-setup.md, docs/guide/ic7300-usb-setup.md, docs/guide/ic705-usb-setup.md, and docs/internals/contributing.md. Remaining grep hits are either LAN/native-provider-specific or explicitly describe avoiding bespoke long-tail CAT backends.

morozsm:
[...]

PR #1583, opened 2026-05-23
Align docs with Hamlib provider strategy

morozsm:
## Summary
- Reframes public core docs around native providers plus planned Hamlib-backed CAT coverage.
- Removes absolute no-Hamlib positioning from README, docs index, rigctld/WSJT-X docs, and radio support pages.
- Updates TX-500/X6100 and adding-radio guidance so long-tail support routes through provider-path choice instead of assuming bespoke native backend work.

## Validation
- uv run mkdocs build --strict
- git diff --check

Tracks rigplane/rigplane-core#1576.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,12 @@
 > [rigplane.dev/migrate](https://rigplane.dev/migrate).
 
 **rigplane** is a multi-vendor radio control library and Web UI — Python
-asyncio core plus a self-contained browser front-end. It speaks Icom CI-V
-(LAN UDP and USB), Yaesu CAT, and Kenwood-style CAT, with stable backends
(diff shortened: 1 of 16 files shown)


morozsm:
Agent Review: PASS

Reviewed head 4e8dea74bf7360c54780292d9b4b4e15b14ff7fa. Docs and package metadata align with the Hamlib strategy: native providers stay direct where they add value, Hamlib is framed as planned long-tail CAT provider coverage underneath RigPlane, and the client-facing rigctld endpoint is kept separate from any Hamlib-backed provider. The changed public docs do not expose private strategy details or overstate the Hamlib work as already shipped. Existing CI checks are green.

PR #1610, opened 2026-05-27
feat(MOR-164): raw CI-V pipe API for Hamlib A1 bridge

morozsm:
## Summary
Implements **MOR-164**: a narrow, first-class **raw CI-V pipe** seam on the Icom backend so a future *transparent* Hamlib A1 bridge can use RigPlane purely as a CI-V byte transport. Replaces the MOR-159 prototype monkeypatch of `_route_civ_frame`.

## API (Icom `CoreRadio`)
- `await radio.send_civ_raw_fire_and_forget(frame: bytes)` — fire-and-forget TX; **no response matcher, no timeout on a bare ACK** (the external CAT master, e.g. Hamlib, owns request/response matching).
- `sub = radio.add_raw_civ_listener(cb)` → `RawCivSubscription` (`.close()` / context manager). The callback receives the **exact on-wire inbound frame bytes**, including bare ACK frames like `FE FE E0 98 FB FD`.

## Behavior / policy
- Inbound tap lives in the CI-V RX pump (`_civ_rx.deliver_raw_civ`), using raw `frame_bytes` (pre-`_route_civ_frame`).
- **Transceive broadcasts (`to_addr == 0x00`) are filtered out** of the raw stream; only frames from this radio addressed to the controller (`0xE0`) are delivered.
- Listener exceptions are isolated (one bad listener cannot break routing).

## Documented limitations (for the bridge-daemon follow-up)
- RigPlane's own poller/readback traffic is **not yet auto-quiesced** during a raw-pipe session — the bridge consumer must avoid competing reads (scoped ownership deferred).
- Inbound bytes are reconstructed from the parsed `CivFrame` in the pump — byte-exact for freq/mode/PTT/ACK; raw pre-parse capture for exotic frames (scope `0x27`) left as future work.

## Tests
`tests/test_raw_civ_pipe.py` (7): ACK delivery, data-frame delivery, transceive filtering, foreign-address filtering, subscription removal (+ idempotent close), listener-failure isolation, and fire-and-forget transmit-without-waiting/timeout.

## Validation
- `pytest --ignore=integration`: **5829 passed**, 57 skipped, 9 xfailed.
- `ruff check src/ tests/`: clean · `lint-imports`: 4 contracts kept.
- PR-gate mypy `mypy --strict src/rigplane/web`: **clean**. (This change adds **0** errors to `mypy src/`; the pre-existing `mypy src/` debt is tracked in a separate issue.)

## Scope
[...]

PR #1612, opened 2026-05-27
feat(MOR-166): local Hamlib A1 bridge runner (slice 1)

morozsm:
## Summary
**MOR-166 slice 1** — a local Hamlib A1 **bridge runner**. Adds `src/rigplane/hamlib_bridge.py`: a transparent CI-V byte transport between a stock Hamlib `rigctld` and a RigPlane radio, built on the **MOR-164 raw CI-V pipe** (#1610). RigPlane owns the real transport; Hamlib stays the sole CAT master.

```
rigctl client → rigctld (front) → -r <host>:<back-port> → HamlibBridge
    ↳ TX: radio.send_civ_raw_fire_and_forget(frame)
    ↳ RX: radio.add_raw_civ_listener(cb) → frame written back to rigctld
```

## What it does
- `HamlibBridge.open_transport()` binds a back-side TCP listener and registers the raw CI-V listener; `spawn_rigctld()` launches stock rigctld with the right argv (`-m`, `-r host:back_port`, `-t front`, `-c civaddr`, `-C` conservative conf: `retry=0,timeout=3000,…,ptt_type=RIG`).
- Transparent pipe: rigctld's raw CI-V frames → `send_civ_raw_fire_and_forget`; radio's inbound frames → written back to rigctld. CI-V frame reassembly across reads.
- Per-frame trace/audit records (`BridgeFrame`, both directions) via an optional `on_frame` sink + a bounded in-memory deque.
- Narrow `RawCivPipe` Protocol → decoupled from the concrete Icom backend, fully testable.

## Scope (slice 1 only)
Internal module — **no UI, no Pro, no assisted discovery, no credential handling**. Deferred: session ownership / poller quiescing (slice 2), non-Icom serial rigs (slice 3).

## Tests
`tests/test_hamlib_bridge.py` (6): pipe TX (multi-frame in one write + frame split across reads), RX of a bare ACK, both-direction trace records, exact `rigctld` argv, rigctld spawn via mock, and subscription cleanup on stop. No hardware / no real rigctld in CI.

## Manual validation (real hardware, not in CI)
Ran the module against the **real IC-7610 over LAN** (creds from BWS, never logged): `F 14100000`→`RPRT 0`, independent RigPlane readback confirms `freq=14100000 / mode=CW`, radio restored to baseline, no TX. 14 CI-V frames each way.

## Gates
`pytest --ignore=integration`: **5847 passed**. `ruff` clean · `lint-imports` 4 kept · PR-gate `mypy --strict src/rigplane/web` clean (module adds 0 to `mypy src/`; pre-existing debt = MOR-165). Import-linter unaffected (top-level module, no layer changes).

> Implementation PR — requires an **independent agent review** (`Agent Review: PASS`) before merge; not self-reviewed.

### DIFF
--- a/src/rigplane/hamlib_bridge.py
+++ b/src/rigplane/hamlib_bridge.py
@@ -1,0 +1,273 @@
+"""Local Hamlib A1 bridge runner — MOR-166 slice 1.
+
+A transparent CI-V byte transport between a stock Hamlib ``rigctld`` and a
+RigPlane radio, built on the MOR-164 raw CI-V pipe. RigPlane owns the real
+transport; Hamlib stays the sole CAT master.
+
+    rigctl client → rigctld (front) → ``-r <host>:<back-port>`` → HamlibBridge
+        ↳ TX: radio.send_civ_raw_fire_and_forget(frame)
+        ↳ RX: radio.add_raw_civ_listener(cb) → frame written back to rigctld
+
+This is an internal module: no UI, no Pro workflow, no assisted discovery, and
+no credential handling (the caller owns the already-connected radio). Session
+ownership / poller quiescing (slice 2) and non-Icom serial rigs (slice 3) are
+intentionally out of scope here.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import shutil
+import time
+from collections import deque
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any, Protocol
+
+logger = logging.getLogger(__name__)
+
+# Conservative defaults for the Hamlib side: let RigPlane own pacing, never let a
+# bare ACK look like a timeout, force CAT PTT (so PTT rides the data path, not
+# RTS/DTR which cannot cross TCP), and disable Hamlib's state cache.
+DEFAULT_RIGCTLD_CONF = (
+    "retry=0,timeout=3000,post_write_delay=0,write_delay=0,cache_timeout=0,ptt_type=RIG"
+)
+
+_CIV_PREAMBLE = b"\xfe\xfe"
+_CIV_TERMINATOR = 0xFD
+
+
+@dataclass(slots=True)
+class BridgeFrame:
+    """One CI-V frame crossing the bridge — a trace/audit record.
+
+    ``direction`` is relative to the radio: ``"tx"`` = Hamlib → radio (a frame
+    rigctld emitted, forwarded to the real radio), ``"rx"`` = radio → Hamlib.
+    CI-V frames carry no credentials, so the raw bytes are safe to record.
+    """
+
+    t_ms: float
+    direction: str
+    frame: bytes
+
+    @property
+    def hex(self) -> str:
+        return self.frame.hex(" ")
+
+
+class RawCivPipe(Protocol):
+    """Minimal radio surface the bridge needs — the MOR-164 raw CI-V pipe."""
+
+    async def send_civ_raw_fire_and_forget(self, frame: bytes) -> None: ...
+
+    def add_raw_civ_listener(self, callback: Callable[[bytes], Any]) -> Any: ...
+
+
+class HamlibBridge:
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-166 slice 1 (local Hamlib A1 bridge runner). I verified the diff against PR head (commit 2861e6ec), read both new files in full, and independently exercised the edge cases rather than trusting the test suite.

## Acceptance criteria — all met
1. **Back-side listener + rigctld argv** — `open_transport()` binds an ephemeral TCP listener (port 0) and exposes it via `back_port`. `rigctld_argv()` emits `-m <model> -r <host>:<back-port> -t <front> [-c <civaddr>] -C <conf>` in correct order; `-c` is omitted when civaddr is None and present (decimal `152` for `0x98`) when set. Verified both paths manually.
2. **MOR-164 raw CI-V pipe** — TX via `send_civ_raw_fire_and_forget` (fire-and-forget, exception-isolated per frame at hamlib_bridge.py:247-249); RX via `add_raw_civ_listener` with a synchronous callback `_on_radio_frame`. Confirmed against the real `CoreRadio.deliver_raw_civ` (_civ_rx.py:599-622) that listeners are invoked synchronously from the CI-V pump, so a sync callback doing a non-blocking `writer.write()` is the correct shape — it never awaits drain and never blocks the pump. The local `RawCivPipe` Protocol matches the real radio surface exactly.
3. **Trace/audit** — `BridgeFrame` records with bounded `deque(maxlen=trace_maxlen)`; optional `on_frame` sink wrapped in try/except so a raising sink can never break the pipe (hamlib_bridge.py:216-219). CI-V frames carry no credentials; raw bytes recorded as documented.
4. **No-hardware CI** — tests use a fake raw-pipe radio + direct TCP client; rigctld spawn checked via mock. No real hardware or rigctld required. Confirmed.
5. **Narrow/internal** — module imports stdlib only (no rigplane imports), defines its own Protocol, no UI/Pro/discovery/credential coupling. No new import-linter layer needed.

## Independent defect hunt
- **`_extract_frames` reassembly** — verified: multi-frame-per-read, frame split across reads (buffered), leading noise before `FE FE` (dropped), noise-only chunk ending in `0xFD` (dropped, buffer advances), trailing partial (retained). Buffer mutation (`del buf[:end+1]`) is correct. The only truncation case is a payload byte == `0xFD`, which CI-V reserves as terminator and never emits — not a real-world defect.
- **RX with no client** — `_writer is None` and `is_closing()` guarded; emit with no client connected does not crash and still records the frame. Verified.
- **Lifecycle** — `back_port` before open raises `RuntimeError`; `open_transport()` idempotent (no double-subscribe, guard precedes subscribe); `stop()` idempotent; subscription closed, server closed, proc terminated then killed on 3s timeout. Verified.

## Minor non-blocking nits (not required for slice 1)
- `spawn_rigctld` opens the `stderr_path` file handle without closing it in `stop()` (comment says lifetime tied to proc); parent fd leaks until GC. Optional path, unused by default/tests.
- `start()`/`spawn_rigctld()` are not guarded against double-call (a second `start()` would leak the first rigctld via `_proc` overwrite). Out of scope for single-use `async with` usage.

## Gates (all green)
- `uv run pytest tests/test_hamlib_bridge.py -q` -> 6 passed
- `uv run pytest tests/test_raw_civ_pipe.py tests/test_civ_rx_coverage.py -q` -> 138 passed
- `uv run pytest tests/ --ignore=tests/integration` -> 5847 passed, 57 skipped, 0 failures
- `uv run ruff check src/ tests/` -> All checks passed
- `uv run lint-imports` -> 4 contracts kept, 0 broken
- `uv run mypy --strict src/rigplane/web` -> Success, 21 files (PR gate)
- `uv run mypy src/rigplane/hamlib_bridge.py` -> Success, adds no new mypy errors

Correct, in-scope, well-tested, gates green.

PR #1627, opened 2026-05-28
feat: add response-capable raw CI-V transactions

morozsm:
## Summary

- Adds scoped raw CI-V transactions via `POST /api/v1/civ/transaction` with explicit `expect` modes: `none`, `ack`, and `data`.
- Adds a CI-V ownership guard so response-capable transactions quiesce cooperating pollers and conflict cleanly with external CAT sessions.
- Reuses the existing CI-V RX pump and `CivRequestTracker`, including deterministic NAK handling for data transactions and stale ACK backlog protection.
- Keeps `/api/v1/commands` and `/api/v1/commands/batch` fire-and-forget; batch transaction support remains deferred.
- Updates web API docs, Web UI guide, changelog, and internals notes.

Closes #1620.
Closes #1621.
Closes #1622.
Closes #1623.
Closes #1626.
Part of #1619.

## Verification

- `uv run pytest tests/test_raw_civ_transaction.py tests/test_web_command_batch_http.py tests/test_hamlib_bridge.py tests/test_poller.py -q --tb=short` → 76 passed
- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → 5883 passed, 57 skipped, 3 deselected, 9 xfailed, 1 xpassed
- `uv run pytest tests/ -q --tb=short` → 6113 passed, 112 skipped, 3 deselected, 9 xfailed, 1 xpassed
- `uv run ruff check src/ tests/ docs/CHANGELOG.md` → passed
- `uv run mypy src/rigplane/runtime/radio.py src/rigplane/runtime/_civ_rx.py src/rigplane/web/server.py src/rigplane/web/web_routing.py` → passed
- `uv run mkdocs build --strict` → passed; existing MkDocs/material warning and excluded-link info messages only

## Review

Independent sub-agent review found and fixed initial matcher blockers, then re-reviewed current diff with `Agent Review: PASS`.

PR #1630, opened 2026-05-28
feat(MOR-170): distinct X6200 profile + discovery hwid disambiguation vs IC-705

morozsm:
## What

Closes the Xiegu X6200 / Icom IC-705 misidentification at the shared CI-V
address `0xA4` that wedged the operator's X6200 during real-hardware testing
(see [MOR-170](https://linear.app/morozsm/issue/MOR-170/) for the full
diagnostic trail).

## Background

Both rigs default to CI-V address `0xA4` (Hamlib `rigs/icom/xiegu.c`
`x6100_priv_caps`; Radioddity *Xiegu X6200 CI-V Implementation V1.0.6*, page 4
verbatim: *"the preamble for all commands is FE FE **A4** 00"*). The existing
address-only auto-detect classified every X6200 as IC-705 and drove it with
the IC-705 profile. On real hardware the rig displayed a flicker per CAT burst
and entered an unresponsive state after 3-4 sequential `rigplane status`
invocations. WSJT-X on the same rig over the same USB cable does NOT cause
this — Hamlib ships a dedicated `RIG_MODEL_X6200` (reuses `x6100_priv_caps`)
with a narrower `X108G_LEVELS` / `X108G_FUNCS` envelope.

### `rigs/x6200.toml` (new)

Near-copy of `rigs/x6100.toml` with:
- `civ_addr = 0xA4` (X6200/IC-705 collision — Radioddity PDF + Hamlib)
- `default_baud = 19200` (SERIAL-B per PDF page 4 — not 115200, which is
  SERIAL-A terminal speed)
- `[commands]` drawn strictly from PDF Table 1 (parts 1-5)
- **`power_on` / `power_off` INTENTIONALLY OMITTED** — CI-V `0x18` is not
  in the X6200 documented command set and is the leading suspect for the
  wedge if the IC-705 profile path ever sends it on the X6200
- Header comment cross-references both the PDF (with firmware version) and
  Hamlib `xiegu.c` (with line numbers) as joint sources of truth

### `src/rigplane/backends/discovery.py`

New module-level helper `_resolve_xiegu_x6200_override(address, port)` and a
one-line call in `discover_serial_radios` after the address-based lookup.
At address `0xA4`, a port whose USB hwid matches the X6200 fingerprint
(`VID:PID=1A86:55D2` — Nanjing QinHeng / WCH CH342 — or product name
containing `"USB Dual_Serial"`) classifies as `xiegu_x6200` instead of
`icom_ic705`. All other 0xA4 ports keep the existing IC-705 behaviour.
Other addresses are unaffected.

### `tests/test_discovery.py`

Three regression tests in `TestDiscoverSerialRadios`:
1. X6200 fingerprint at 0xA4 → `model="X6200"`, `profile_id="xiegu_x6200"`.
2. Non-Xiegu hwid at 0xA4 → `model="IC-705"`, `profile_id="icom_ic705"` —
   existing IC-705 owners unaffected.
3. Product-name-only fallback (no VID/PID surfaced by pyserial) → X6200.

## Verification

- `uv run pytest tests/ --ignore=tests/integration`: **5947 passed, 0 failures**
- `uv run mypy --strict src/rigplane/web` (the CI gate): clean
- `uv run ruff check src/ tests/`: clean
- New TOML round-trips through `tomllib.load`; rig_loader picks up 8 profiles
  (was 7) including "X6200"

**Not verified on live hardware in this PR** — live X6200 smoke is gated on
operator OK after merge (the radio just recovered from a wedge during the
diagnostic session; defensive posture is to ship the test-verified fix first,
then run the live smoke with the new profile).

## Pre-existing main state (NOT this PR)

`uv run mypy src/` reports one error introduced by MOR-169 (#1627):
`src/rigplane/runtime/_civ_rx.py:397: error: Redundant cast to "bytes"`.
Reproduced on `origin/main` with this PR's changes stashed — not a regression
from this PR. `mypy src/` is not a CI gate (only the strict `web` gate is),
so it does not block this PR; flagging for a follow-up.

## Out of scope (separate follow-up tickets — see MOR-170 thread)

1. **CI-V-level disambiguation via `0x1D 0x19`** (Xiegu model ID probe,
   returns `0x6200` on this rig per PDF page 9). Firmware-authored,
   hwid-independent, survives future USB-chip changes.
2. **Connection-lifecycle audit** for stray `0x18` sends — guard behind a
   per-profile capability flag so future conservative profiles can't trip
   the same wedge.
3. **`rigs/x6100.toml` civ_addr** — currently `0x70`, Hamlib says `0xA4` for
   the same rig family. Needs validation on actual X6100 hardware before
   changing; operator only has X6200.

### DIFF
--- a/rigs/x6200.toml
+++ b/rigs/x6200.toml
@@ -1,0 +1,346 @@
+# Rig Configuration — Xiegu X6200
+#
+# Xiegu X6200 ultra-portable HF/50MHz transceiver.
+# Protocol: CI-V (subset of standard Icom CI-V, similar to X6100 family).
+#
+# Sources of truth (cross-referenced):
+#   1. Radioddity "Xiegu X6200 CI-V implementation V1.0.6" (firmware V1.0.6).
+#      All commands below come from the documented Table 1 (parts 1-5),
+#      pages 5-9 of that PDF.
+#   2. Hamlib master ``rigs/icom/xiegu.c`` — ``x6200_caps`` reuses the same
+#      ``x6100_priv_caps`` struct (line ~690), confirming X6100 and X6200
+#      share the same CI-V personality at the protocol level even though
+#      they're separate Hamlib ``RIG_MODEL_*`` entries.
+#
+# History: MOR-170. Filed after the IC-705 profile (which X6200 was
+# misidentified as by default — both default to CI-V address 0xA4) drove
+# the radio's display into a flicker → wedge state on real hardware.
+# The X6200 documented CI-V envelope is narrower than IC-705's, and
+# crucially does NOT include the CI-V power command 0x18; this profile
+# therefore omits ``power_on`` / ``power_off`` entirely so the runtime
+# cannot send 0x18 on behalf of an X6200.
+
+[radio]
+id = "xiegu_x6200"
+model = "X6200"
+civ_addr = 0xA4
+receiver_count = 1
+has_lan = false
+has_wifi = true
+# PDF page 4: "For CAT control, a speed of 19.200 bps on the SERIAL-B port
+# is used." SERIAL-A at up to 115200 is the terminal port, not CAT.
+default_baud = 19200
+
+[protocol]
+type = "civ"
+address = 0xA4
+
+[capabilities]
+# Aligned to Hamlib's ``X108G_LEVELS`` / ``X108G_FUNCS`` envelope
+# (xiegu.c lines 52, 54) intersected with what the Radioddity X6200 PDF
+# explicitly documents as supported by firmware V1.0.6. Intentionally
+# narrower than the IC-705 envelope:
+#   - no ``power_control`` (no documented CI-V 0x18)
+#   - no ``scope`` (no documented 0x27 family)
+#   - no ``filter_width`` SET (PDF table 5 documents widths but encoding
+#     is the segmented BCD index family; mirror x6100.toml's "feature not
+#     advertised until tables are confirmed on hardware" stance — see
+#     issue #1159 referenced in x6100.toml)
+features = [
+    "audio",
+    "af_level",
+    "rf_gain",
+    "squelch",
+    "attenuator",
+    "preamp",
+    "nb",
+    "nr",
+    "notch",
+    "pbt",
+    "tx",
+    "split",
+    "vox",
+    "compressor",
+    "cw",
+    "break_in",
+    "rit",
+    "xit",
+    "tuner",
+    "meters",
+    "repeater_tone",
+    "tsql",
+    "data_mode",
+    "scan",
+    "dial_lock",
+    "agc",
+]
+
+# PDF page 10 (Table 3 — Mode, Data & filter bandwidth) documents:
+# LSB, LSB-D, USB, USB-D, AM, CW, NFM, CWR. The ``-D`` variants are the
+# data-mode flavours selected via byte 2 of mode commands 0x26 0x00/0x01.
+[modes]
+list = ["USB", "LSB", "CW", "CW-R", "AM", "FM", "RTTY", "DIGI"]
+
+[filters]
+list = ["FIL1", "FIL2", "FIL3"]
+
+[vfo]
+scheme = "ab"
+
+# PDF page 5: ``0x11 0x00`` (off) / ``0x11 0x01`` (on) — binary toggle.
+[attenuator]
+values = [0, 1]
+
+# PDF page 8: ``0x16 0x02 0x00`` / ``0x16 0x02 0x01`` — binary toggle.
+[preamp]
+values = [0, 1]
+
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of `9a39b187` against `origin/main`. All checks below were re-run on the worktree, not trusted from the PR body.

## Scope
- Exactly 3 files changed (`git show --stat HEAD`): `rigs/x6200.toml` (+346), `src/rigplane/backends/discovery.py` (+36), `tests/test_discovery.py` (+102). No foreign content; `src/rigplane/runtime/_civ_rx.py` and `runtime/radio.py` (MOR-169 territory on a neighbour branch) are untouched.
- Worktree is on the correct branch with no uncommitted changes after my mutation experiment was reverted.

## rigs/x6200.toml
- `civ_addr = 0xA4` matches Radioddity PDF page 4 (`FE FE A4 00 ...`).
- `default_baud = 19200` matches PDF page 4 ("a speed of 19.200 bps on the SERIAL-B port is used").
- Every opcode in `[commands]` is sourced from Table 1 (parts 1-5, pages 5-9) or is a standard Icom CI-V freq/mode opcode (`0x03/0x04/0x05/0x06`) already used by Hamlib's `x6100_priv_caps` and the existing `rigs/x6100.toml`. Opcode set: `{0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x0F, 0x11, 0x14, 0x15, 0x16, 0x19, 0x1A, 0x1C, 0x1D, 0x25, 0x26, 0xFA, 0xFB}`. No outliers.
- **`0x18` is fully absent** — verified by scanning every value in the parsed TOML. `power_on`/`power_off` are not declared and the comment block explicitly forbids re-adding them without hardware re-validation. This is the load-bearing guard for the wedge motivation in MOR-170.
- TOML round-trips through `tomllib.load`; `discover_rigs(Path('rigs'))` returns 8 profiles (was 7) with X6200 listed at `civ_addr=0xA4` alongside IC-705 — matches PR body.

## discovery.py override
- Address guard is the first check (`if address != 0xA4: return None`) — other addresses (0x94 IC-7300, 0x98 IC-7610, 0xA2 IC-9700) cannot be perturbed.
- VID/PID match (`0x1A86:0x55D2`, WCH CH342) and product-name fallback (`"USB Dual_Serial"`) are independent OR branches; either alone resolves to X6200. Returning `None` falls through to the existing `CIV_PROFILE_MAP` mapping, preserving IC-705 behaviour for non-Xiegu hwids.
- All downstream `RadioDiscoveryResult` fields (vid/pid/product/serial_number/usb_audio) are still populated unconditionally — override only mutates `model` and `profile_id`.

## Tests
- All 3 new tests use the shared 0xA4 wire response (`_IC705_RESPONSE`) so the override path is what differentiates the outcome, not the CI-V payload. The IC-705 negative test correctly uses an Icom-looking VID (`0x0C26`) with `product="IC-705"` — does not match the X6200 fingerprint.
- **Mutation check performed**: temporarily forced `_resolve_xiegu_x6200_override` to always return `None`. The two positive tests (`test_civ_addr_0xA4_with_xiegu_hwid_resolves_to_x6200`, `test_civ_addr_0xA4_xiegu_product_name_fallback`) failed with `assert 'IC-705' == 'X6200'`; the IC-705 negative test still passed. This confirms the override is the load-bearing logic, not coincidence. Override restored, `git diff` clean.

## Full verification (re-run on worktree)
- `uv run --no-sync pytest tests/test_discovery.py -q`: 67 passed.
- `uv run --no-sync pytest tests/ -q --tb=short --ignore=tests/integration`: 5947 passed, 7 skipped, 11 xfailed, 1 xpassed — matches PR body.
- `uv run --no-sync mypy --strict src/rigplane/web`: clean (21 files).
- `uv run --no-sync ruff check src/ tests/`: clean.
- `uv run --no-sync lint-imports`: 4 contracts kept, 0 broken.

## Risk notes (not blocking)
- The product-name match string `"USB Dual_Serial"` is specific to the WCH CH342 composite enumeration — low false-positive risk, but if Xiegu ever ships a different USB chip the override silently stops firing and the rig classifies as IC-705 again. The TODO in the docstring (CI-V `0x1D 0x19` probe-based disambiguation) is the right long-term fix and is correctly listed as out-of-scope for this PR.
- The toml includes `set_freq=0x05`, `get_mode=0x04`, `set_mode=0x06` which are not explicitly in Table 1 of the V1.0.6 PDF (only `0x03` Get Frequency appears on page 4 as an example). They are standard Icom CI-V opcodes that Hamlib's `x6100_priv_caps` (shared by X6200) treats as supported, and the existing `rigs/x6100.toml` already uses them. Acceptable precedent; the file comments document the source-of-truth chain honestly.

Defensive posture (omitted `0x18`, narrower capability envelope, hwid-based disambiguation with address guard) directly addresses the wedge symptom described in MOR-170. Live hardware smoke-test is appropriately deferred to post-merge per the PR body.

PR #1632, opened 2026-05-28
fix(MOR-172): drop LAN OpenClose packets on serial CI-V wire (X6200 wedge)

morozsm:
## What

Fixes [MOR-172](https://linear.app/morozsm/issue/MOR-172/) — the Xiegu X6200 CI-V wedge isolated to the malformed 4-byte frame `FE FE 00 FD` that the serial transport stack emitted on every session close.

## Root cause

`_send_open_close_on_transport` in `runtime/_control_phase.py` builds a 22-byte LAN-style binary OpenClose packet (type marker `0x01C0` at offset 0x10) for every session open/close. On serial, `SerialCivTransport.send_tracked` handed that packet to `_unwrap_civ_frame`, which misread the type marker's high byte as `payload_len = 1` and extracted the single close-flag byte at offset 0x15. The serial codec then wrapped that single byte as `FE FE 00 FD` — a malformed CI-V frame (no source, no command, addressed to PC `0x00`) that wedged the X6200's CI-V parser for ~5-10 seconds.

Confirmed on live X6200 hardware (operator session 2026-05-28): display blanks for ~1s timed exactly with the malformed-frame TX; subsequent CI-V transactions timeout until natural recovery. WSJT-X over the same USB cable never sees this because it holds one long-lived serial session and never triggers the OpenClose pass-through.

### `src/rigplane/backends/icom7610/drivers/serial_session.py`

Two-layer defence:

**Fix A** — `_unwrap_civ_frame` rejects packets carrying the OpenClose type marker `0x01C0` at offset 0x10–0x11 (LE u16). DATA packets pass through unchanged whether they come from this module's `_wrap_civ_frame` (`pkt[0x10] = 0x00`) or from `runtime/_civ_rx.py::_wrap_civ` (`pkt[0x10] = 0xC1`); the discriminator is the **exact type-marker value**, not "byte 0x10 ≠ 0x00" (an earlier draft of this fix made that mistake and silently dropped every real CI-V command).

**Fix B** — `SerialCivTransport.send_tracked` short-circuits on the same type marker at the higher layer before unwrap is attempted. Belt and braces — either guard alone fixes the bug; both together make the intent self-documenting at both layers and resistant to future refactors on either side.

### `tests/test_serial_session.py` (new)

7 regression tests covering both directions of the defence plus the bug introduced and fixed during this PR's iteration:

1. `test_send_tracked_drops_openclose_close_packet` — Fix B, close direction
2. `test_send_tracked_drops_openclose_open_packet` — Fix B, open direction
3. `test_send_tracked_forwards_real_civ_frame` — Fix B does not over-shoot
4. `test_unwrap_civ_frame_returns_empty_on_openclose_packet` — Fix A
5. `test_unwrap_civ_frame_returns_empty_on_zero_payload_data_packet` — original branch
6. `test_unwrap_civ_frame_extracts_real_civ_payload` — DATA from `_wrap_civ_frame` (`pkt[0x10]=0x00`)
7. `test_unwrap_civ_frame_extracts_runtime_wrap_civ_payload` — DATA from `runtime._civ_rx::_wrap_civ` (`pkt[0x10]=0xC1`); locks in the "match marker exactly, not 'non-zero'" lesson

**Mutation testing** (manual, in worktree): reverting Fix A alone breaks exactly test #4; reverting both fixes additionally breaks tests #1 and #2. Each test is load-bearing for the protection it covers.

## Verification

- `uv run pytest tests/ --ignore=tests/integration`: **5957 passed, 0 failures** (+7 new tests; +5 previously-failing `test_icom7610_serial_radio.py` tests that were briefly broken by the wrong-discriminator first draft of Fix A — now back to green with the correct discriminator).
- `uv run mypy --strict src/rigplane/web` (CI gate): clean.
- `uv run ruff check src/ tests/`: clean.

**Not verified on live hardware in this PR** — gated next step: operator runs `rigplane status` 3x consecutive against the real X6200 with display in view; success criterion = no flicker, no wedge, all three reads return data. If it passes, MOR-172 closes.

## Out of scope

Why the `_send_open_close → no-op` override on `_IcomSerialRadioBase` doesn't actually reach this code path remains unresolved — the wire-tap evidence in the MOR-172 investigation shows `OpenClose(close) sent` is logged on serial despite the override. This PR works around it at the serial-transport layer; the architectural cleanup belongs in MOR-173 (single source of truth) once that lands.

### DIFF
--- a/src/rigplane/backends/icom7610/drivers/serial_session.py
+++ b/src/rigplane/backends/icom7610/drivers/serial_session.py
@@ -13,6 +13,12 @@ _CIV_HEADER_SIZE = 0x15
 _CLIENT_ID = 0x00010001
 _RADIO_ID = 0x00000098
 
+# Type marker at offset 0x10-0x11 (little-endian u16) for LAN-style OpenClose
+# session-control packets emitted by ``_send_open_close_on_transport`` in
+# ``runtime/_control_phase.py``. These packets have no CI-V payload and must
+# never be forwarded onto a serial CI-V wire — see MOR-172.
+_LAN_OPENCLOSE_TYPE_MARKER = 0x01C0
+
 
 class _LifecycleCivLink(CivLink, Protocol):
     """CivLink with connect/disconnect lifecycle required by serial session."""
@@ -29,13 +35,33 @@ class _LifecycleCivLink(CivLink, Protocol):
 
 
 def _unwrap_civ_frame(packet: bytes) -> bytes:
-    """Extract CI-V frame payload from shared-core UDP envelope."""
+    """Extract CI-V frame payload from shared-core UDP envelope.
+
+    Session-control packets such as OpenClose share the envelope layout
+    but stamp the LAN type marker ``0x01C0`` at offset 0x10 (LE u16) and
+    carry no CI-V payload. Note that the marker's high byte spills into
+    offsets 0x11-0x12, so a payload-length-only check misreads it as
+    ``payload_len = 1`` and extracts the trailing close-flag byte — which
+    the serial codec then wraps as a malformed CI-V frame ``FE FE 00 FD``
+    that wedged the Xiegu X6200's CI-V parser on every session close
+    (MOR-172). Reject the marker explicitly.
+
+    DATA packets can come from either :func:`_wrap_civ_frame` (byte 0x10
+    = 0x00) or from runtime ``_civ_rx._wrap_civ`` (byte 0x10 = 0xC1);
+    both layouts must pass through. Hence we gate on the OpenClose
+    marker exactly, not on "byte 0x10 not zero".
+    """
     if len(packet) <= _CIV_HEADER_SIZE:
         return b""
+    if (
+        len(packet) >= 0x12
+        and int.from_bytes(packet[0x10:0x12], "little") == _LAN_OPENCLOSE_TYPE_MARKER
+    ):
+        return b""
     payload_len = int.from_bytes(packet[0x11:0x13], "little", signed=False)
-    start = _CIV_HEADER_SIZE
     if payload_len <= 0:
-        return packet[start:]
+        return b""
+    start = _CIV_HEADER_SIZE
     end = start + payload_len
     if end > len(packet):
         return packet[start:]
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of `codex/mor-172-civ-wedge-fix` @ `c5d619d2`.

## Scope (verified)
2 files only:
- `src/rigplane/backends/icom7610/drivers/serial_session.py` (+50/-4)
- `tests/test_serial_session.py` (new, +182)

## Discriminator correctness (verified)
Both Fix A (`_unwrap_civ_frame`) and Fix B (`SerialCivTransport.send_tracked`) gate on the **exact** marker:
```python
int.from_bytes(packet[0x10:0x12], "little") == 0x01C0
```
This matches `runtime/_control_phase.py:826` `struct.pack_into("<H", pkt, 0x10, 0x01C0)`.

- DATA packets from serial-session `_wrap_civ_frame` (`pkt[0x10] = 0x00`, line 80) pass through.
- DATA packets from `runtime/_civ_rx.py:_wrap_civ` (`pkt[0x10] = 0xC1`, line 1611) pass through.
- Only the LAN OpenClose marker `0x01C0` is rejected.

The previous iteration's bug (gating on "`pkt[0x10] != 0x00`" which falsely matched `0xC1` DATA packets) is **not** present. `test_unwrap_civ_frame_extracts_runtime_wrap_civ_payload` explicitly constructs a packet with `pkt[0x10] = 0xC1` and asserts the CI-V frame is extracted — exactly the regression test that was missing before.

## Mutation testing (performed in worktree)

| Mutation | Tests failing |
|---|---|
| Fix A → `if False` | `test_unwrap_civ_frame_returns_empty_on_openclose_packet` (1 fail, 6 pass) |
| Fix B → `if False` | none (0 fail, 7 pass) |
| Both → `if False` | `test_send_tracked_drops_openclose_close_packet`, `test_send_tracked_drops_openclose_open_packet`, `test_unwrap_civ_frame_returns_empty_on_openclose_packet` (3 fail, 4 pass) |

Observation: Fix B is true belt-and-braces — Fix A alone is sufficient because `send_tracked` calls `_unwrap_civ_frame`, gets `b""`, and the `if not frame: return` short-circuits before reaching the link. Tests #1/#2 verify the end-to-end behavior (correct), not Fix B's existence specifically. The commit message acknowledges this honestly ("reverting Fix A alone breaks exactly … reverting both Fix A and Fix B additionally breaks both …"). Acceptable: defence-in-depth at two layers with a single observable contract.

After mutations, file restored cleanly (`git diff` empty before final verification).

## Verification (all green)
- `uv run --no-sync pytest tests/test_serial_session.py -q` — **7 passed**
- `uv run --no-sync pytest tests/test_icom7610_serial_radio.py -q` — **20 passed** (these were broken by the previous Fix-A iteration; now green)
- `uv run --no-sync mypy --strict src/rigplane/web` — **Success: no issues found in 21 source files**
- `uv run --no-sync ruff check src/ tests/` — **All checks passed**
- `uv run --no-sync pytest tests/ -q --tb=short --ignore=tests/integration` — **5957 passed, 7 skipped, 3 deselected, 11 xfailed, 1 xpassed**

## Minor note (non-blocking)
Comment in `send_tracked` at lines 167-168 says "_unwrap_civ_frame already short-circuits non-DATA packets … (it gates on byte 0x10)". With the new implementation it gates on the exact 16-bit marker `0x01C0`, not "byte 0x10 alone". Documentation drift, not a behaviour bug.

Cleared to merge once required checks are green and PR is non-draft.

Issue #1651, opened 2026-05-28
Data: initial Core real-radio matrix templates


morozsm:
Parent: #1645
Depends on: #1646

## Summary

Create initial Core hardware validation matrix templates for the radios planned for local testing.

## Initial templates

- Yaesu FTX-1
- Lab599 TX-500
- Icom IC-7300
- Xiegu X6200
- Optional when evidence is available: IC-7610, IC-705, IC-9700

## Scope

Each template should define:

- model aliases and expected backend/transport choices;
- discovery expectations;
- default baud/host/port assumptions where relevant;
- safe frequency windows or operator-provided safe frequency variables;
- expected supported capabilities;
- skipped/unsupported capabilities with reasons;
- manual operator prompts for reverse state sync and guarded controls;
- audio/scope expectations and guardrails;
- recovery checks.

## Acceptance criteria

- Templates exist for FTX-1, TX-500, IC-7300, and X6200.
- Templates are data-only and do not encode proprietary Pro behavior.
- Templates are validated by schema tests from #1646.
- Templates can drive the runner from #1647 without hand-editing code.
- Missing or unknown capability decisions are explicit as `manual_required` or `unsupported_pending_evidence`, not silently assumed.

morozsm:
Closing as implemented. Initial data-only templates exist for FTX-1, TX-500, IC-7300, and X6200 under docs/validation/templates, and template/schema coverage exists in tests. Further radio-specific evidence belongs in validation follow-ups under #1645/#1482.

Issue #1647, opened 2026-05-28
Feature: Core hardware validation runner and artifacts


morozsm:
Parent: #1645
Depends on: #1646

## Summary

Add a local, opt-in Core hardware validation runner that executes a matrix against a real radio and writes a machine-readable artifact plus a human summary.

This must run without Pro. The goal is to isolate Core backend/protocol/API behavior from packaged app and UI behavior.

## Scope

- Add a local command or script for running a matrix against a selected model/backend/transport.
- Start/connect Core headlessly and exercise public Core surfaces where applicable.
- Emit JSON/YAML artifact with Core version, commit, radio target, transport, safety flags, level results, per-check evidence, and logs path.
- Keep CI mock-backed; real-radio runs are skipped unless explicitly configured.
- Support resumable/partial runs so a failed scope/audio check does not erase earlier CAT evidence.
- Generate a concise issue payload when a check fails.

## Required first checks

- Discovery/probe evidence.
- Connect/disconnect lifecycle.
- `/api/v1/runtime`, `/api/v1/state`, `/api/v1/capabilities` snapshots.
- Basic rigctld `f` and `m` where rigctld is enabled.
- Safe command dispatch through HTTP command endpoint.
- Optional ordered batch check for safe commands.

## Acceptance criteria

- One documented command can run a Core hardware validation matrix locally.
- The runner can run against at least one serial target without Pro.
- Output artifact captures pass/fail/skip/unsupported and enough raw evidence to debug failures.
- Failures do not hang indefinitely; timeouts are bounded and reported per check.
- Existing tests cover runner planning and artifact serialization without hardware.

## Non-goals

- No Pro profile persistence or packaged app checks.
- No TX/PTT writes by default.
- No hosted upload or private support integration.

morozsm:
Implemented by #1653 — hardware runner with read/write+restore diagnostics, verified live on a real X6200.

morozsm:
Closing as implemented by #1653 and subsequent validation-matrix work. Core now has the gated local validation runner, hardware execution path, JSON/human artifacts, bounded per-check results, and mock-backed tests. Any stale docs or follow-up validation checks should be handled as focused follow-up issues.

Issue #1646, opened 2026-05-28
Spec: Core real-radio validation schema and safety policy


morozsm:
Parent: #1645

## Summary

Define the Core real-radio validation matrix contract before implementation. The matrix must produce repeatable local evidence that a radio/backend is supported through Core, without relying on Pro UI behavior.

## Scope

- Define the YAML/JSON matrix schema for a radio validation run.
- Define levels, result IDs, statuses, evidence fields, safety gates, and operator metadata.
- Define the default safe mode: RX/read-only plus explicitly safe set/readback operations.
- Define how tests declare required capabilities, optional capabilities, unsupported/skipped behavior, and manual operator steps.
- Define artifact naming, raw logs, sanitized issue payloads, and summary output.
- Define how Core distinguishes failures in discovery, transport, command execution, readback, state publishing, rigctld compatibility, audio, and scope/waterfall.

## Acceptance criteria

- A documented matrix schema exists under Core docs or internals docs.
- Status vocabulary is explicit: pass, fail, skip, unsupported, manual_required, blocked.
- Safety policy forbids TX/PTT writes by default and requires an explicit operator flag for any TX-adjacent action.
- The schema can represent the reported field observations from #1645, including reverse state sync failure, PRE/ATT failures, Notch failure, mode/filter failure, and audio/scope absence.
- The schema supports per-radio allowlists so FTX-1, TX-500, IC-7300, X6200, and IC-7610 can differ without hardcoded assumptions.

## Test plan

- Add schema validation tests using fixture matrix artifacts.
- Include at least one fixture with mixed pass/fail/unsupported results.
- Include a fixture that represents a failed reverse state sync check.

morozsm:
Addressed by #1652 (Core validation matrix schema + safety contract).

Issue #1645, opened 2026-05-28
Epic: real-radio support validation matrix


morozsm:
## Goal

Create an end-to-end real-radio validation matrix for Core so we can run a repeatable local test suite and make an evidence-backed statement that a radio/backend is supported.

This is broader than one-off hardware smoke tests. The output should become a durable support certification workflow: automated where safe, semi-automated where hardware interaction is required, and manual only where radio/operator safety or UI observation demands it.

## Initial target radios

Planned near-term hardware:

- Yaesu FTX-1
- Lab599 TX-500
- Icom IC-7300
- Optional/additional: Xiegu X6200, IC-7610, IC-705/IC-9700 when available

## Core boundary

Core owns public/open support evidence:

- backend/protocol correctness;
- discovery/probing and candidate metadata;
- command/readback behavior;
- rigctld compatibility where Core serves it;
- generic audio/scope behavior where exposed by Core;
- raw transaction behavior;
- machine-readable validation artifacts useful to downstream users.

Pro owns packaged-app UX, profile persistence, managed local runtime, product workflows, WSJT-X launch/profile handling, and support evidence packaging.

### Level 0: Static profile contract

Automated, no hardware.

- Rig profile loads and schema validates.
- Model registry resolves `--model` aliases and CI-V/CAT addresses.
- Default serial baud and backend type are explicit.
- Capability map is internally consistent: unsupported commands are not exposed.
- Public docs list supported transport(s), required cables/network settings, and safety notes.

### Level 1: Safe discovery/probe

Hardware required, read-only/default-safe.

- Serial/network candidate is discovered with model, transport, port/host, baud, USB metadata when available.
- Discovery does not wedge or alter persistent radio settings.
- Wrong-port handling is explicit and bounded.
- Probe artifacts include command, model, firmware/version if available, transport, selected baud, and errors.

### Level 2: Basic control contract

Hardware required, safe read/write with operator-approved non-TX actions.

- Connect/disconnect lifecycle is clean.
- Initial state resolves: frequency, mode, filter/VFO when supported.
- Set frequency to a safe nearby value and read back.
- Set mode/filter where supported and read back.
- UI-facing state changes propagate via WebSocket/state API.
- No PTT/TX action occurs.

### Level 3: Capability-specific matrix

Hardware required, mostly automated with per-radio allowlist.

- PTT read status, but TX write remains guarded.
- Split/VFO operations where supported.
[...]

PR #1653, opened 2026-05-28
feat: hardware validation runner — read/write diagnostics with restore (#1647)

morozsm:
## Summary

Implements the **hardware-execution path** of `rigplane validate` (#1647): connect a radio, run the matrix, emit a machine-readable artifact + human summary with real per-check verdicts, evidence, and timestamps. Also expands the per-radio templates (#1651) with RX-control checks. Builds on the schema/dry-run base from #1646. Part of epic #1645.

Verified end-to-end against a **real Xiegu X6200** over serial (see evidence below).

## Behavior

- **Default = read/write with automatic restore.** Each RX-safe control runs a read-modify-verify-restore cycle: read original → write a different value → read back (did it react?) → restore the original (in a `finally` that never raises) → control-read. This is the only way to detect "control doesn't react" faults.
  - no reaction → `fail` / `readback` (the headline diagnostic)
  - timeout / NAK → `fail` / `command_execution`; connect/auth → `fail` / `transport`
  - capability absent → `unsupported`; restore failure recorded as evidence
- **`--read-only`** disables all writes (every write check becomes `skip`).
- **Safety:** TX/PTT and the antenna tuner are **never auto-actuated**. `tx.ptt`/`tuner.tune` stay `blocked` without `--tx-allowed`/`--tuner-allowed` and `manual_required` (read-only observation) when authorized. Fail-closed authorization gate.
- **Timestamps:** every check carries ISO-8601 UTC `started_at`/`finished_at`; the artifact carries `generated_at`; one INFO log line per check for log correlation. Schema stays v1 (additive optional fields).

## Live X6200 evidence (read/write, default safety)

`pass 6 · fail 8 · unsupported 2 · manual_required 1 · blocked 2`

| check | verdict | corroborates field report |
|---|---|---|
| freq.write | pass (21.200→21.201→restore) | freq write works |
| nb / nr / agc | pass | NB/NR/AGC work |
| preamp / attenuator | fail/command_execution — "no cmd29 route 0x16/0x02 … 0x11" | PRE/ATT don't react (root cause surfaced) |
| notch | fail/command_execution — timeout | Notch doesn't work |
| mode | fail/readback — wrote LSB, readback USB | mode/filter don't change |
| filter_width | unsupported (not in X6200 profile) | — |
| audio / scope | manual_required / unsupported | no audio/waterfall |
| tuner / tx | blocked (not actuated) | — |
| rf_gain / af_level / rit / xit | fail (readback/timeout) | flagged for follow-up (analog-level exact-equality vs quantization) |

## Verification

- `uv run pytest tests/ -q --ignore=tests/integration` → **5968 passed**, 57 skipped, 11 xfailed.
- `uv run mypy src/` clean for new code; `ruff check` clean; `lint-imports` 4 kept / 0 broken (hardware.py imports only core+validation+stdlib; CLI keeps hardware imports deferred).
- Independent agent review: **PASS** (verified restore-in-finally, no-TX/no-tuner actuation, no-react→fail/readback, --read-only skips writes, template declarations match each rig TOML, schema back-compat).

## Known follow-up (not in this PR)

Analog level controls (rf_gain/af_level) use exact-equality on readback; the X6200 quantizes the 0–255 scale (wrote 200, read 198), so a control that *did* react reports `fail/readback`. A tolerance-based comparison for analog levels is a sensible fast-follow.

## Out of scope

TX keying, tuner tune cycle, audio/scope capture, capability bug fixes, Pro code, new dependencies.

### DIFF
--- a/docs/validation/templates/ftx1.json
+++ b/docs/validation/templates/ftx1.json
@@ -30,6 +30,102 @@
       "tx_adjacent": false
     },
     {
+      "check_id": "mode.set",
+      "capability": "",
+      "level": 2,
+      "declaration": "supported",
+      "summary": "Set operating mode via CI-V write (filter coupling).",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "filter_width.set",
+      "capability": "filter_width",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Set DSP IF filter width via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "rf_gain.set",
+      "capability": "rf_gain",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Set RF gain level via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "af_level.set",
+      "capability": "af_level",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Set AF output level via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "preamp.set",
+      "capability": "preamp",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Set preamp state via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "attenuator.set",
+      "capability": "attenuator",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Set attenuator state via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "notch.set",
+      "capability": "notch",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Toggle manual notch via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "nb.set",
+      "capability": "nb",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Toggle noise blanker via CI-V write.",
+      "tx_adjacent": false
+    },
+    {
+      "check_id": "nr.set",
+      "capability": "nr",
+      "level": 3,
+      "declaration": "supported",
+      "summary": "Toggle noise reduction via CI-V write.",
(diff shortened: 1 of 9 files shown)


morozsm:
**Agent Review: PASS**

Independent review (implementation agent did not self-review) for head `ca58aa20cdcdfed3948f866ea9167554cf28ae65`.

Verified against source:
- **No TX/tuner actuation**: grep of `validation/` shows zero `set_ptt`/tune/key-TX calls; `tx.ptt`/`tuner.tune` only read `radio_state`; fail-closed `_is_authorized` gates them (blocked without flags, manual_required when authorized). `set_ptt.assert_not_called()` asserted.
- **Restore-in-finally**: `_read_modify_verify_restore` restores `original` in a `finally` wrapped in try/except; a restore failure becomes `restore_error` evidence and never escapes; original restored even when write/readback fails. Mirrored for `mode.set` (mode + filter).
- **RMVR mapping**: no-reaction → fail/readback; reacted-but-not-restored → fail/readback; timeout → fail/command_execution; conn/auth → fail/transport. make_changed values stay in range (AgcMode 1/3, preamp 0↔1, bool flips, level deltas).
- **--read-only** disables all writes (write checks → skip; no `set_*` reached).
- **Templates**: every non-empty capability ∈ KNOWN_CAPABILITIES; declarations match each `rigs/<model>.toml` (zero "supported but absent"); x6200 filter_width / tx500 notch+agc+xit / ftx1 agc+xit = unsupported_pending_evidence.
- **Schema back-compat**: SCHEMA_VERSION=1; timestamps optional; PR1 fixtures still validate. Layering: hardware.py imports only core+validation+stdlib; CLI keeps hardware imports deferred.

Checks: ruff clean, mypy clean for new code, lint-imports 4/0, pytest full suite 5968 passed. Additionally verified end-to-end on a real X6200 (artifact: pass 6 / fail 8 / unsupported 2 / manual_required 1 / blocked 2; field-report faults reproduced). May merge once required checks are green.

PR #1654, opened 2026-05-28
feat: drive RF/AF/preamp/att/NB/NR via rigctld client + tolerance-aware validation readback

morozsm:
## Summary

Extends the shared `RigctldClientRadio` backend to drive **RF gain, AF, preamp, attenuator, noise blanker, noise reduction** via the rigctl level/func protocol (`l/L`, `u/U`), and adds **tolerance-aware readback** to the validation RMVR engine for analog controls. Foundation for `rigplane validate --provider hamlib` (follow-up PR) and a fix for the analog-level false-fail seen in native runs.

## Changes

- **`RigctldClientRadio`** (additive): `get/set` for `rf_gain`, `af_level`, `preamp`, `attenuator` (+level), `nb`, `nr`. Scale conversion 0–255 ↔ Hamlib 0.0–1.0 float (rounded, clamped 0–255); preamp 0/1/2 and attenuator bool map to representative dB. Six new capability tags + registered commands. Unsupported level/func on a given Hamlib model surfaces as `CommandError` (honest fail, not crash). `freq/mode/ptt/vfo` untouched.
- **`fake_rigctld.py`**: simulates `l/L/u/U` statefully (unknown name → ENIMPL) so the backend is testable without hardware.
- **Tolerance-aware RMVR** (`validation/hardware.py`): analog controls compare within a tolerance (rf_gain/af ±3 on the 0–255 scale, filter_width ±50 Hz, rit ±10 Hz) instead of exact equality, so scale quantization isn't a false "did not react". Bool/enum/mode and **frequency stay exact**. Evidence still records exact original/changed/readback.

## Verification

- `uv run pytest tests/ -q --ignore=tests/integration` → **5975 passed**, 57 skipped, 11 xfailed.
- ruff clean; mypy clean for changed files; `lint-imports` 4 kept / 0 broken.
- Independent agent review: **PASS** (additive-only backend confirmed; rigctl verbs match the repo's own handler; scale clamp safe at bounds; freq stays exact; KNOWN_CAPABILITIES contains all six tags).

## Known follow-up (out of scope)

The expanded capability set now flows into the web layer when a station runs `rigplane web --radio-backend rigctld`. The on/off controls work, but UI sub-controls `set_nb_level/nr_level/nb_depth/nb_width` are not implemented by the rigctld client and would raise `AttributeError` reachable from the UI — caught by the poller's broad handler (degraded per-command, not a crash). Gate those sub-controls on method presence in a follow-up.

### DIFF
--- a/src/rigplane/backends/rigctld_client/radio.py
+++ b/src/rigplane/backends/rigctld_client/radio.py
@@ -17,6 +17,20 @@ _SUPPORTED_COMMANDS = {
     "set_ptt",
     "get_vfo_slot",
     "set_vfo_slot",
+    "get_rf_gain",
+    "set_rf_gain",
+    "get_af_level",
+    "set_af_level",
+    "get_preamp",
+    "set_preamp",
+    "get_attenuator",
+    "set_attenuator",
+    "get_attenuator_level",
+    "set_attenuator_level",
+    "get_nb",
+    "set_nb",
+    "get_nr",
+    "set_nr",
 }
 
 
@@ -77,7 +91,7 @@ class RigctldClientRadio:
 
     @property
     def capabilities(self) -> set[str]:
-        caps = {"tx"}
+        caps = {"tx", "rf_gain", "af_level", "preamp", "attenuator", "nb", "nr"}
         if self._vfo_supported:
             caps.add("vfo")
         return caps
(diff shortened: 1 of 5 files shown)


morozsm:
**Agent Review: PASS**

Independent review (implementation agent did not self-review) for head `85f1ee7dd375e1f6917b5f3e70ef45e7c46548f2`.

Verified against source:
- **Additive-only backend**: existing freq/mode/data_mode/ptt/vfo methods unchanged; only new level/func methods, tags, and `_SUPPORTED_COMMANDS` entries added. The one existing capabilities assertion was updated.
- **Wire-protocol correctness**: `l/L RF|AF|PREAMP|ATT`, `u/U NB|NR` match the repo's own rigctld handler/protocol naming; GET parses the right shape; unsupported/error → `CommandError` (matrix records fail/command_execution, not a crash).
- **Scale clamp safety**: `_level_255_to_float`/`_float_to_level_255` clamp to [0,255] both directions (tested at -10/999 and -1.0/2.0); no out-of-range write path. preamp 0/1/2↔dB and att bool↔dB reversible on the on/off transition.
- **Tolerance**: applied only to rf_gain/af_level/filter_width/rit; **frequency, bool, enum, mode stay exact**; comparator is symmetric/integer-safe; evidence records exact values (off-by-2 test asserts exact 98/200/198/96 via a real RadioState double, no MagicMock).
- **KNOWN_CAPABILITIES** contains all six tags.

Checks: ruff clean, mypy clean for changed files, lint-imports 4/0, full suite 5975 passed. Non-blocking follow-up noted in the PR (web sub-control gating for the rigctld backend). May merge once required checks are green.

PR #1657, opened 2026-05-28
fix(MOR-184): set mode via CI-V 0x26 selected-mode (X6200); IC-7610 unchanged

morozsm:
## Summary

Fixes **MOR-184**: on the single-RX Xiegu X6200, setting mode did nothing. The runtime emitted a bare `0x06 <mode>` mode-set, which the X6200 **ignores** (no-op); it requires the full mode tuple. Routes MAIN `set_mode` through CI-V **0x26** selected-mode (`26 00 <mode> <data_mode> <filter>`) for rigs whose profile declares the command.

This is also the *correct* command for data modes (FT8 = USB-D), and it's exactly what Hamlib and rigplane's own state poller already use — confirmed by a live CI-V trace (Hamlib drives the X6200 via `26 00 …`).

## Design — data-driven, no per-rig code

- New builder `commands.set_selected_mode` (0x26 0x00 + mode/data_mode/filter).
- `_set_mode_main` routes via 0x26 **iff** `RadioProfile.set_mode_via_selected` is true; that field is **derived** from the profile's `[commands]` table (`"set_selected_mode" in commands`) — not a hand-authored quirk flag. data_mode/filter are filled from current state so only the mode changes.
- Rigs that don't declare the command (incl. **IC-7610**) keep the **byte-identical** `0x06` path — zero flagship behavior change.

## Hardware-verified (both radios, live)

- **X6200 (serial):** `validate --provider native` → `mode.set` **pass** (USB→LSB, readback LSB, restored). Previously fail/readback "did not react".
- **IC-7610 (LAN):** `set_mode_via_selected=False` (gate off); `set_mode` USB→LSB→readback LSB→restored — **unchanged 0x06 path, no regression**.

## Verification

- New tests: `set_selected_mode` builder byte-lock; X6200 emits the 0x26 frame; **IC-7610 regression-lock** (emits unchanged 0x06, gate off); data_mode/filter fill (mode-only change).
- `uv run pytest tests/ -q --ignore=tests/integration` → **6003 passed**. ruff/mypy clean for changed files; lint-imports 4 kept / 0 broken.
- Independent agent review pending.

## Scope / out of scope

MAIN `set_mode` only. `get_mode` (0x04) untouched (works on both radios). RIT/XIT readback = MOR-179 (write-only, → MOR-180). att/preamp = MOR-178 (merged). No rig TOML changes; mock_server untouched.

### DIFF
--- a/src/rigplane/commands/__init__.py
+++ b/src/rigplane/commands/__init__.py
@@ -78,6 +78,7 @@ from .freq import (
     parse_selected_mode_response,
     set_freq,
     set_frequency,
+    set_selected_mode,
 )
 
 # --- mode.py ---
@@ -563,6 +564,7 @@ __all__ = [
     "parse_selected_freq_response",
     "get_selected_mode",
     "get_unselected_mode",
+    "set_selected_mode",
     "parse_selected_mode_response",
     "_CMD_SELECTED_FREQ",
     "_CMD_SELECTED_MODE",
--- a/src/rigplane/commands/freq.py
+++ b/src/rigplane/commands/freq.py
@@ -146,6 +146,29 @@ def get_unselected_mode(
     return build_civ_frame(to_addr, from_addr, _CMD_SELECTED_MODE, data=bytes([0x01]))
 
 
+def set_selected_mode(
+    mode: Mode,
+    data_mode: int,
+    filter_index: int,
+    to_addr: int,
+    from_addr: int = CONTROLLER_ADDR,
+) -> bytes:
+    """Build a "set selected receiver mode" CI-V command (0x26 0x00).
+
+    Frame layout: ``FE FE <to> <from> 26 00 <mode> <data_mode> <filter> FD``.
+    The leading ``0x00`` data byte selects the active receiver (mirroring
+    :func:`get_selected_mode`); ``mode`` / ``data_mode`` / ``filter`` set the
+    full tuple so a mode-only change preserves the radio's current data-mode
+    and filter.
+    """
+    return build_civ_frame(
+        to_addr,
+        from_addr,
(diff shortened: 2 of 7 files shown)


morozsm:
**Agent Review: PASS**

Independent review (implementation agent did not self-review) for head `238ada0a3521acf8e64f8ecac6f32a67e2d58b62`.

- **Data-driven gate, defaults False**: `RadioProfile.set_mode_via_selected` derived (`rig_loader.py:176`) from `"set_selected_mode" in self.commands` — not a hand-authored quirk bool. X6200 declares it → True; IC-7610 does not → False. No rig TOML touched.
- **IC-7610 zero-regression**: the `else` branch is byte-identical to the original `set_mode(..., receiver=RECEIVER_MAIN)` 0x06 call; cache tail unchanged. Regression-lock test asserts gate False + frame `fe fe 98 e0 06 00 fd` + no `26 00` — non-vacuous. Confirmed live: IC-7610 set_mode still works over LAN.
- **Builder matches the hardware-proven frame**: `set_selected_mode` → `FE FE <to> E0 26 00 <mode> <data_mode> <filter> FD` (the 0x00 is the receiver-select data byte, not a sub); matches `26 00 00 00 01`. Plain frame, not cmd29 — no CLAUDE.md conflict.
- **Mode-only change**: data_mode from current MAIN state, filter from arg→`_filter_width`→default 1 (never 0). Scope clean (get_mode/freq/att/rit/SUB/poller/mock_server untouched).
- ruff/lint-imports clean; changed files mypy-clean; full suite **6003 passed**.

Hardware-verified live: X6200 `mode.set` now PASS (USB→LSB→restored); IC-7610 (gate off) unchanged and working. May merge once required checks are green.

PR #1667, opened 2026-05-29
fix(test): include notch in X6200 write_only_capabilities assertions

morozsm:
**Repairs red `main`.** MOR-207 (#1665) added `"notch"` to `rigs/x6200.toml` `[validation].write_only_controls`, so the X6200 profile now exposes `{rit, xit, notch}`. Two CLI-forwarding tests in `test_validate_hamlib_provider.py` still asserted the old `{rit, xit}` set and have been failing on `main` since #1665 merged — that PR's review ran only `tests/test_rig_loader.py`, not the full suite, so the cross-file assertion break was missed.

Fix: update both assertions (`test_run_hardware_forwards_write_only_capabilities`, `test_run_hardware_hamlib_forwards_write_only_capabilities`) to `frozenset({"rit", "xit", "notch"})`.

Test-only, 1 file. `pytest tests/test_validate_hamlib_provider.py` → 10 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1669, opened 2026-05-29
fix(MOR-182): harden rigctld-client response alignment over the bridge

morozsm:
## MOR-182 — harden Hamlib bridge response alignment for level/func SETs

`validate --provider hamlib` runs the matrix through `HamlibBridge` + the rigctld client. On a flaky X6200, level/func SET acks came back **desynchronized** — a stray GET-shaped value line left in the socket buffer by a late/out-of-band frame was consumed as the next command's reply (`L AF 0.784` read `0.0392157`; `U NB 1` read `0`), surfacing as noisy `command_execution` fails.

### Root cause
At the **rigctld-client ↔ socket layer** (`RigctldTransport`), not the bridge (which is an intentional transparent byte pipe). Commands are already lock-serialized, so it's not interleave: each transaction read exactly its expected lines and left extras in the `StreamReader` buffer, where the **next** transaction consumed them.

### Fix (3 files, all inside the existing per-connection lock)
- **`backends/rigctld_client/transport.py`**:
  - `_drain_stale()` — before each `query()`/`command()`, non-blockingly discard any bytes already buffered from a prior transaction, so a leftover line can't be read as this command's reply.
  - `command()` **RPRT re-sync** — if the read line isn't `RPRT`-shaped, bounded non-blocking re-read for the real `RPRT <n>` (handles a stray line arriving within the transaction window). A legitimate negative RPRT (`l RF` → `RPRT -11`) is RPRT-shaped → still raises via `_raise_rprt`.
  - `query()` gains only the pre-drain (its replies are value-shaped).
- **`tests/fake_rigctld.py`** — `extra_lines` knob to inject a stray preceding line, deterministically reproducing the desync.
- **`tests/test_rigctld_client_backend.py`** — regression tests: stray-preceding-line drained (`L AF`→`0.0392157`), leftover-between-transactions discarded (`U NB 1`→`0`), GET still reads value after drain (no hang), negative RPRT still raises; existing serialization test stays green.

Implementation note: `asyncio.wait_for(reader.read(), timeout=0)` fires before reading buffered data on this event loop, so the drain uses `timeout=0.001`.

No bridge or production-radio change.

### Gates
- `pytest tests/test_rigctld_client_backend.py` → all pass (incl. `test_transport_serializes_requests`)
- full suite (no integration) → **6125 passed** (rebased on the #1667 main fix)
- ruff / mypy (`transport.py`) / lint-imports → clean

Final confidence needs a live X6200 native-vs-hamlib re-run (orchestrator, post-merge, serial port free); the unit tests prove the desync fix deterministically without hardware.

Linear: MOR-182. Unblocks reliable MOR-201 (Generator B) execution + MOR-202 comparison.

### DIFF
--- a/src/rigplane/backends/rigctld_client/transport.py
+++ b/src/rigplane/backends/rigctld_client/transport.py
@@ -3,10 +3,13 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 
 from ...exceptions import CommandError
 from ...exceptions import ConnectionError as RadioConnectionError
 from ...exceptions import TimeoutError as RadioTimeoutError
+
+_LOGGER = logging.getLogger(__name__)
 
 _ERROR_HINTS = {
     -1: "invalid parameter",
@@ -64,12 +67,29 @@ class RigctldTransport:
         except OSError:
             pass
 
+    async def _drain_stale(self) -> None:
+        """Discard any unread bytes left in the socket buffer from a prior
+        transaction (e.g. a late/out-of-band frame the bridge injected) so the
+        next command reads only its own reply."""
+        reader = self._reader
+        if reader is None:
+            return
+        while True:
+            try:
+                chunk = await asyncio.wait_for(reader.read(4096), timeout=0.001)
+            except (asyncio.TimeoutError, TimeoutError):
+                return
+            if not chunk:
+                return  # EOF
+            _LOGGER.debug("rigctld transport: drained %d stale bytes", len(chunk))
+
     async def query(self, command: str, *, response_lines: int) -> list[str]:
         """Send a command and read a fixed number of response lines."""
         if response_lines <= 0:
             raise ValueError("response_lines must be > 0")
 
         async with self._lock:
+            await self._drain_stale()
             await self._write_line(command)
             lines: list[str] = []
             for _ in range(response_lines):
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-182 (rigctld-client response desync fix). Reviewed adversarially; I did not author this PR.

**Scope (3 files, +128/-0):** `src/rigplane/backends/rigctld_client/transport.py` (the fix), `tests/fake_rigctld.py` (`extra_lines` injection knob), `tests/test_rigctld_client_backend.py` (regression tests). Confirmed no `hamlib_bridge.py` change, no production-radio change, no stray edits. The fix is correctly client-level, leaving the bridge as the intentional transparent pipe.

**Fix correctness:**
- `_drain_stale()` is called at the START of both `query()` and `command()`, inside `async with self._lock`. Non-blocking (`reader.read(4096)`, `timeout=0.001`); terminates on TimeoutError (empty buffer, both `asyncio.TimeoutError` and `TimeoutError` caught) and on `b""` (EOF). No hang path.
- `command()` re-sync loop is bounded: `for _ in range(_MAX_RESYNC - 1)` = 3 iterations max. RPRT-shaped lines (including a negative `RPRT -11`) break immediately and flow to `_parse_rprt`/`_raise_rprt`. A lone malformed response (nothing buffered behind it) hits the `timeout=0.001` TimeoutError → break, leaving the malformed `line` so `_parse_rprt` still raises its normal CommandError. No infinite loop, no silent pass.
- `query()` only pre-drains (no shape re-sync) — correct, GET replies are values.
- All changes stay inside the per-connection lock; public signatures unchanged.

**Concurrency:** the drain runs before the write while holding the lock, which serializes transactions. It cannot consume another transaction's in-flight reply (none is, the lock is held) nor the command's own reply (nothing written yet). Reasoning holds.

**Regression proof:** reverted just transport.py to origin/main in a worktree — `test_command_drains_stray_preceding_line` and `test_leftover_line_discarded_between_transactions` FAIL with exactly the reported bug (`malformed status ... '0.0392157'` and `... '0'`), and pass with the fix. `test_negative_rprt_still_raises` confirms `l RF`→`RPRT -11` still raises; `test_get_reads_value_after_drain` and `test_transport_serializes_requests` confirm no happy-path/serialization regression. The `extra_lines` fake knob injects bytes before the response as intended.

**Gates (temp worktree, Python 3.13):**
- `pytest tests/test_rigctld_client_backend.py` → 17 passed in 0.14s (no hang)
- `pytest tests/ --ignore=tests/integration` → 6125 passed, 7 skipped, 11 xfailed, 1 xpassed in 154s (no hang)
- `ruff check` (3 files) → clean
- `mypy` transport.py → clean
- `lint-imports` → 4 contracts kept

No bug found (no hang, no infinite loop, no race). Approving.

PR #1677, opened 2026-05-29
fix(MOR-219): resolve X6200 USB audio via CDC-ACM topology

morozsm:
## Summary

Fixes **MOR-219** — the browser RX "LIVE" stream was **silent for the Xiegu X6200**. Root cause: the macOS USB-audio topology resolver could not link the X6200's CDC-ACM serial port to its USB audio input.

`_extract_tty_suffix()` matched only `usbserial-…` (FTDI/CP210x), not the `usbmodem…` (CDC-ACM) ports used by the X6200's WCH CH342 dual-serial bridge. With no TTY suffix, IORegistry lookup never ran → topology resolution returned `None` → audio fell back to silence.

## Change (generic, topology-anchored — no per-profile audio field)

- **`_extract_tty_suffix()`** now matches both `usbserial-XXXX` and `usbmodemXXXX`, so CDC-ACM ports reach IORegistry hub-prefix matching. The X6200's C-Media `USB Audio Device` already matches the existing `USB Audio (?:CODEC|Device)@` node scan, so it links by physical hub.
- **`_is_usb_audio_codec()`** and **`usb_driver._USB_NAME_PATTERNS`** recognise the C-Media identity (`c-media`/`cmedia`) as a robust-identity fallback for Linux/Windows, where topology resolution is not yet implemented (tracked in **MOR-222**).
- Module docstring updated: macOS covers FTDI/CP210x + CDC-ACM; Linux/Windows use name/identity fallback + the existing optional `[usb]` override escape hatch (no new per-profile field).

## Tests

Added regression tests in `tests/test_usb_audio_resolve.py`: `usbmodem` suffix extraction, single-X6200 topology resolution to the C-Media device, X6200-alongside-Icom serial-location lookup, C-Media name recognition, and name-fallback preference. Existing IC-7610/IC-7300/FTX-1 resolution unchanged.

## Gate

- `pytest` (non-integration): **6226 passed**, 7 skipped, 11 xfailed
- `ruff check` + `ruff format --check`: clean
- doctests in `_usb_resolve.py`: pass
- mypy: no new errors (pre-existing baseline only)

## Hardware verification (real X6200, macOS, 2026-05-29)

- TTY suffix now extracts `58910181093` (previously `None`).
- Topology confirmed: audio `locationID 0x01122000` ↔ serial CH342 `0x01124000` → both hub prefix `0x0112`.
[...]

PR #1680, opened 2026-05-29
fix(MOR-224): accept CDC-ACM serial ports in discovery (X6200 / CH342)

morozsm:
## Summary

Fixes **MOR-224** — the Xiegu X6200 was **not detected** during serial discovery on Linux and Windows; its port was hard-skipped before any CI-V probe ran. Same release vehicle as the X6200 audio fix #1677 (MOR-219). **Blocks rigplane-pro beta.7.**

## Root cause (confirmed, corrects the issue hypothesis)

`_is_candidate()` accepted a port only when the OS **device path** contained the substring `"usb"`. USB serial radios don't all carry `"usb"` in their device name:
- macOS: `/dev/cu.usbserial-*` / `usbmodem*` — has `"usb"` ✅
- Linux FTDI/CP210x: `/dev/ttyUSB*` — has `"usb"` ✅
- **CDC-ACM bridges: `/dev/ttyACM*` (Linux) and `COMx` (Windows) — no `"usb"`** ❌

The X6200's WCH CH342 is a CDC-ACM device, so it (and in fact **any** USB serial radio on Windows COMx) was filtered out at enumeration. macOS was unaffected only because `usbmodem` happens to contain `"usb"`.

This is **not** the MOR-170 hwid-disambiguation override: that helper runs only *after* a successful CI-V probe and never skips a port. The skip is the long-standing device-name-substring filter (present since the filter's first commit); the X6200 is simply the first CDC-ACM radio to expose it.

## Change

`_is_candidate()` now anchors on USB **identity**: accept when pyserial surfaces a USB vendor id, or the hwid string contains `"USB VID:PID="`, falling back to the device-name substring. Bluetooth/virtual exclusions still take precedence; IC-705 vs X6200 disambiguation untouched. Generic, cross-platform, no per-radio special-casing.

## Tests

X6200 on Linux `ttyACM*` and Windows `COMx` accepted by vid; generic USB radio on Windows COM accepted; CDC-ACM accepted by hwid when vid absent; bluetooth-with-vid still excluded (name precedence); `ttyACM` enumerates as a candidate. Existing exclusions / `"usb"`-name cases unchanged.

## Gate

- `pytest` (non-integration): **6208 passed**, 57 skipped, 11 xfailed

## Hardware verification

- Real X6200 (macOS): discovery still resolves to `xiegu_x6200` (count=1, addr 0xA4 @ 19200) — no regression.
- Linux/Windows verified via mock ports through the real `_is_candidate` (I have only macOS hardware). Manual checklist: plug X6200 into Linux (`ttyACM*`) and Windows (`COMx`), run discovery, confirm model `X6200`; confirm other USB radios still found; confirm a genuine IC-705 is not mislabeled.

## Scope / follow-up

- Override still keys on USB hwid fingerprint (`1A86:55D2` / `"USB Dual_Serial"`) — robustness via CI-V model-ID probe tracked in **MOR-226** (was MOR-170 "out of scope").
- Open-core; Pro pins core → needs a core release + re-pin to unblock beta.7.

### DIFF
--- a/src/rigplane/backends/discovery.py
+++ b/src/rigplane/backends/discovery.py
@@ -821,14 +821,25 @@ def _normalize_usb_audio_metadata(value: object) -> dict[str, object] | None:
 
 
 def _is_candidate(port: object) -> bool:
-    """Return True if *port* is a plausible Icom serial connection.
+    """Return True if *port* is a plausible USB-connected radio serial port.
 
-    Inclusion criteria:
-    - Device path contains ``"usb"`` (case-insensitive)
+    Inclusion criteria (any one is sufficient):
+    - The OS reports a USB vendor id for the port (``port.vid`` is set), or
+    - The hardware id string contains ``"usb"`` (e.g. ``USB VID:PID=...``), or
+    - The device path contains ``"usb"`` (case-insensitive).
 
-    Exclusion criteria:
+    Exclusion criteria (checked first, take precedence):
     - Device path contains ``"bluetooth"`` (case-insensitive)
     - Device path contains ``"debug"``, ``"wlan"``, or ``"spi"`` (virtual ports)
+
+    Anchoring on the USB *identity* rather than only the device-path substring
+    matters cross-platform: USB serial radios do not all carry ``"usb"`` in
+    their OS device name. macOS uses ``/dev/cu.usbserial-*`` / ``usbmodem*``
+    (has ``"usb"``) and Linux ``/dev/ttyUSB*`` for FTDI/CP210x bridges, but
+    CDC-ACM composite bridges enumerate as ``/dev/ttyACM*`` on Linux and
+    ``COMx`` on Windows — neither contains ``"usb"``. The Xiegu X6200's WCH
+    CH342 is exactly such a CDC-ACM device, so the old device-name-only filter
+    hard-skipped it on Linux/Windows before any CI-V probe could run (MOR-224).
 
     Args:
         port: A ``serial.tools.list_ports_common.ListPortInfo`` object.
@@ -847,4 +858,9 @@ def _is_candidate(port: object) -> bool:
         if kw in device_lower:
             return False
 
+    if getattr(port, "vid", None) is not None:
+        return True
+    hwid_lower = str(getattr(port, "hwid", "") or "").lower()
+    if "usb" in hwid_lower:
+        return True
     return "usb" in device_lower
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of `codex/mor-224-x6200-discovery` (MOR-224). Verified:

- **Root cause confirmed.** Old `_is_candidate()` returned only `"usb" in device_lower`, so a CDC-ACM bridge enumerating as `/dev/ttyACM*` (Linux) or `COMx` (Windows) — neither containing "usb" — was skipped before any CI-V probe. The WCH CH342 in the X6200 is exactly such a device.
- **Fix correctness.** Exclusions run first: the `_BLUETOOTH_KEYWORDS` and `_VIRTUAL_KEYWORDS` loops `return False` before the new vid/hwid accepts (discovery.py:853-866), so name-based exclusions take precedence even when a vid is present. New accepts: `port.vid is not None` → True, else `"usb"` in hwid string → True, else fall back to the original device-name substring.
- **Cross-platform behavior (REPL-verified).** Linux `/dev/ttyACM0` vid=0x1A86 → True; Windows `COM3` vid=0x1A86 → True; macOS `usbmodem` (no vid) → True; Linux `/dev/ttyUSB0` hwid `USB VID:PID=...` → True; `/dev/ttyS0` (no vid, no usb) → False. Exclusion-precedence: bluetooth/wlan/debug ports WITH a vid set → still False. Edge: `vid=0x0000` → True (intentional; exclusions already filtered virtual/BT first), and a `/dev/ttyACM0` with no vid and no hwid → False (no false-positive on bare ACM).
- **Design intent.** Accepting any USB-vid port is deliberate — discovery then probes and non-radios fail the CI-V handshake. Confirmed it does not blanket-accept virtual ports.
- **Tests meaningful, not tautological.** 6 new `_is_candidate` cases (incl. the bluetooth-with-vid precedence guard and the hwid-fallback-when-vid-absent case) + 1 `enumerate_serial_ports` integration case asserting device/vid/pid pass-through. `_make_port` `**metadata` helper unchanged.
- **No scope creep.** `git diff main...HEAD --stat`: only `src/rigplane/backends/discovery.py` (+20/-4) and `tests/test_discovery.py` (+63). Single commit. No changes to the override resolver, factory, profiles, or the MOR-219 audio resolver (only docstring text mentions Xiegu).
- **Checks green.** `pytest tests/test_discovery.py` 73 passed; `pytest -k "discovery or x6200"` 114 passed, 2 xfailed; `ruff check src/ tests/` clean; `ruff format --check` both files already formatted. Layering confined to backends/discovery, no new imports.

PR #1682, opened 2026-05-29
feat(MOR-226): disambiguate X6200 vs IC-705 by CI-V model ID

morozsm:
## Summary

Resolves **MOR-226** — makes the Xiegu X6200 vs Icom IC-705 disambiguation (shared CI-V address 0xA4) robust at the **protocol level**, instead of relying solely on the exact USB hwid fingerprint introduced in MOR-170.

## Why

MOR-170 distinguished the two radios only by `vid==0x1A86 && pid==0x55D2` / product `"USB Dual_Serial"` (WCH CH342). That is fragile: an X6200 behind a different/future USB bridge — or one whose VID/PID the OS doesn't surface and whose product string differs — was **misclassified as IC-705** and driven with the wrong CI-V envelope (the exact display-wedge MOR-170 was filed to prevent). The `rigs/x6200.toml` TODO already flagged this: *"probe this in discovery rather than relying on hwid fingerprint."*

## Change

When the USB hwid is **inconclusive** at 0xA4, probe the Xiegu-only opcode **`0x1D 0x19`** (documented for the X6200 — Radioddity CI-V V1.0.6 PDF p.9; already declared in `rigs/x6200.toml` as `get_xiegu_model_id`). A valid reply positively identifies a Xiegu; an IC-705 has no such opcode and NAKs → stays IC-705.

- hwid fingerprint remains the **fast path** (no extra serial I/O when it matches).
- The extra `0x1D 0x19` is sent only to 0xA4 ports that don't match the X6200 fingerprint. Safe — standard Icom CI-V NAKs unknown commands rather than wedging.
- New `probe_xiegu_model_id()` + `_is_xiegu_model_id_reply()` mirror the existing CI-V probe/parse style; wired into `discover_serial_radios` only at 0xA4 when the hwid override is inconclusive. MOR-170 hwid/product-name paths unchanged.

## Tests

Reply-parsing (valid / NAK / wrong-echo / short / no-preamble / command-echo-skip), the probe (valid→True, NAK→False, timeout, open-failure, writer-close), and discover-level X6200-via-CI-V promotion + IC-705-NAK-stays-IC-705.

## Gate

- `pytest` (non-integration): **6237 passed**, 57 skipped, 11 xfailed
- `mypy src/rigplane/backends/discovery.py`: clean

## Hardware verification

- **Real X6200 (macOS):** `0x1D 0x19` returns model id `0x6200`; `probe_xiegu_model_id` → True; discovery resolves `xiegu_x6200`. ✅
- **IC-705 negative path (NAK → stays IC-705):** unit-tested; not validated on hardware (no IC-705 available). Manual checklist: connect an IC-705, run discovery, confirm it is still classified IC-705 (and that the radio NAKs `0x1D 0x19` cleanly, no wedge). Also confirm an X6200 behind a non-CH342 bridge promotes to X6200.

## Scope

Completes the X6200 follow-ups alongside MOR-219 (#1677) and MOR-224 (#1680). Open-core; needs a core release + re-pin for rigplane-pro beta.7.

### DIFF
--- a/src/rigplane/backends/discovery.py
+++ b/src/rigplane/backends/discovery.py
@@ -40,6 +40,7 @@ __all__ = [
     "probe_serial_civ",
     "probe_serial_kenwood_cat",
     "probe_serial_yaesu_cat",
+    "probe_xiegu_model_id",
     "rank_hamlib_probe",
 ]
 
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-226 (X6200 vs IC-705 disambiguation at shared CI-V 0xA4 via the Xiegu-only `0x1D 0x19` model-ID probe). I did not write this code.

**Verified — checks:**
- `uv run pytest tests/test_discovery.py -q` → 86 passed (1.28s).
- `uv run pytest tests/ -k "discovery or x6200 or xiegu"` → 132 passed, 2 xfailed.
- `uv run ruff check src/ tests/` → all checks passed; `ruff format --check` → both files already formatted.
- `uv run mypy src/rigplane/backends/discovery.py` → success, no issues.

**Verified — `_is_xiegu_model_id_reply` truth table (REPL):**
| input | result |
|---|---|
| `FE FE E0 A4 1D 19 62 00 FD` (valid) | True |
| `FE FE E0 A4 FA FD` (NAK) | False |
| `FE FE E0 A4 19 00 01 05 FD` (19 00 transceiver-ID) | False |
| `FE FE E0` (short, len 3) | False |
| 8 zero bytes (no preamble) | False |
| command echo only (dest A4, no `FE FE E0`) | False |
| echo + reply | True |
| preamble truncated at len 5 (`frame[5]` would IndexError) | False (guarded by `len(frame) < 6`) |
| valid reply with no `FD` terminator | True (opcode is the discriminator; model bytes empty) |
| leading junk then valid frame | True |

The `len(frame) < 6` guard correctly protects the `frame[4]`/`frame[5]` indexing; NAK is rejected at `frame[4] == 0xFA`; the `19 00` ID frame is rejected by the explicit `frame[4]==0x1D and frame[5]==0x19` check; the command echo lacks the `FE FE E0` controller preamble and is correctly skipped.

**Verified — integration gating:**
- Probe fires ONLY when `xiegu_override is None and civ.address == 0xA4`. The X6200 hwid fast-path (VID/PID `1A86:55D2` or product `USB Dual_Serial`) sets `xiegu_override` in `_resolve_xiegu_x6200_override` *before* the guard, so it provably skips the probe. The two MOR-170 fast-path tests each queue only ONE fake frame; the suite completing in 1.28s with no hangs confirms the probe is not invoked there (an erroneous invocation would block on the empty reader queue until timeout).
- Genuine IC-705 (non-Xiegu hwid that NAKs) → stays IC-705 / `icom_ic705` (test `test_civ_addr_0xA4_civ_nak_stays_ic705`).
- Inconclusive-hwid X6200 → promoted to X6200 via the probe (test `test_civ_addr_0xA4_civ_model_id_resolves_to_x6200`).
- Non-0xA4 radios (IC-7610) skip the probe via the address gate — no behavior change.

**Verified — frame & robustness:**
- `_XIEGU_MODEL_ID_CMD = FE FE A4 E0 1D 19 FD` — correct CI-V addressing (to=0xA4 radio, from=0xE0 controller).
- `probe_xiegu_model_id` returns False on open failure, timeout, NAK, and unexpected data; writer is always closed in `finally`. Discover-level tests correctly model the two-read sequence (CI-V probe consumes the first queued frame, model-ID probe the second).

**Scope:** `git diff main...HEAD --stat` → only `src/rigplane/backends/discovery.py` and `tests/test_discovery.py` (+274 LOC), single commit. No changes to `_resolve_xiegu_x6200_override`, factory, profiles, or the audio resolver. Layering intact; `probe_xiegu_model_id` exported in `__all__`. Tests are meaningful, not tautological.

**Documented limitation (not a blocker):** the IC-705 negative path is unit-tested but not hardware-validated — no IC-705 available. Accepted.

**Minor (non-blocking, no fix required):** the read loop re-checks the deadline both in the `while` condition and via `remaining <= 0`; harmless redundancy. A valid reply missing its `FD` terminator is accepted (model bytes default empty) — acceptable since the opcode match is the discriminator.

LGTM — may merge once required checks are green.

PR #1683, opened 2026-05-29
fix(MOR-223): derive TX/tuner authorization from the registry, not the mutable template flag

morozsm:
## MOR-223 — registry-derived authorization (defense-in-depth)

The TX/tuner authorization gate trusted the **mutable** template flag `entry.tx_adjacent`, so a hand-edited/mutated template with `tx.ptt`/`tuner.tune` set to `tx_adjacent=False` would BYPASS the gate. This makes the gate derive its safety class from the **immutable registry** (`get_spec(check_id).kind == CheckKind.TX_ADJACENT_BLOCKED`), matching the ADR §4.2/§10 claim that authorization is re-applied "regardless of the template flag". Hardening — no current exploit path (the MOR-206 override layer already clamps the flag; novel ids are inert).

### Change
- `validation/runner.py`: new `_registry_gated_capability(entry)` (immutable registry capability for a TX_ADJACENT_BLOCKED check) + `_is_safety_gated(entry)` (template flag OR registry); `_is_authorized` uses the registry capability for the tuner-vs-tx split (a mutated `entry.capability` can't dodge it); dry-run gate uses `_is_safety_gated`.
- `validation/hardware.py`: Pre-gate 1 in `_run_one_check` uses `_is_safety_gated` (blocked-result body unchanged).
- No import cycle (registry imports only schema/core).

### Tests (`tests/test_authz_registry_gated.py`, 6)
mutated `tx.ptt`/`tuner.tune` (tx_adjacent=False) → BLOCKED without authorization; tuner uses its own gate (tx_allowed alone doesn't clear it); mutated `capability` can't shift the split; non-gated `rf_gain.set` unaffected; hardware pre-gate blocks with `set_ptt` never called; canonical tx_adjacent=True regression.

### Gates
- `uv run pytest tests/ -q` — 6523 passed, 0 failures (full suite)
- ruff / lint-imports — clean; `mypy src/` — no new errors

### DIFF
--- a/src/rigplane/validation/hardware.py
+++ b/src/rigplane/validation/hardware.py
@@ -50,7 +50,7 @@ from rigplane.core.radio_protocol import (
 )
 from rigplane.core.types import AgcMode
 from rigplane.validation.registry import CheckKind, CheckSpec, ValueRule, get_spec
-from rigplane.validation.runner import _is_authorized
+from rigplane.validation.runner import _is_authorized, _is_safety_gated
 from rigplane.validation.schema import (
     CapabilityDeclaration,
     CapabilityDeclarationEntry,
@@ -213,7 +213,7 @@ async def _run_one_check(
 ) -> CheckResult:
     """Execute a single template entry, applying universal pre-gates first."""
     # Pre-gate 1: authorization for TX-adjacent checks.
-    if entry.tx_adjacent and not _is_authorized(entry, safety):
+    if _is_safety_gated(entry) and not _is_authorized(entry, safety):
         return _base_result(
             entry,
             CheckStatus.BLOCKED,
--- a/src/rigplane/validation/runner.py
+++ b/src/rigplane/validation/runner.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
+from rigplane.validation.registry import CheckKind, get_spec
 from rigplane.validation.schema import (
     CapabilityDeclaration,
     CapabilityDeclarationEntry,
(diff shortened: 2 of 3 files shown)


morozsm:
**Agent Review: PASS**

Independent adversarial review (reviewer ≠ implementer), HEAD `c7133c8e`, verified by execution:

- **Diff scope** — 3 files (`runner.py`, `hardware.py`, new test); no uv.lock.
- **Layer/no cycle** — lint-imports 4/0; runner→registry within `validation/`, registry never imports runner.
- **Bypass CLOSED (reviewer's own `python -c`)** — mutated `tx.ptt` (tx_adjacent=False) → BLOCKED w/o auth, cleared by tx_allowed; mutated `tuner.tune` → BLOCKED, **tx_allowed alone does NOT clear it**, only tuner_allowed; capability-dodge (`tuner.tune` + `capability="tx"`) still requires tuner_allowed (registry capability wins).
- **No over-blocking** — `rf_gain.set` `_is_safety_gated` False (SKIP, not blocked); `tx.ptt` gated regardless of flag value.
- **hardware pre-gate** — Pre-gate 1 uses `_is_safety_gated`, first statement before any actuation; MagicMock run confirms `set_ptt` never called when unauthorized.
- **Tests** — 6, meaningful (BLOCKED-vs-SKIP + tuner/tx split + real `set_ptt.assert_not_called()`); pass.
- **Full gates** — `pytest tests/` 6523 passed / 0 failures; ruff clean; lint-imports 4/0; `mypy src/` 0 new errors.

No template construction actuates tx/tuner without authorization. PASS — may merge once required checks are green.

PR #1685, opened 2026-05-29
refactor(MOR-230): pair USB audio by identity, not positional order

morozsm:
## Problem

The macOS USB-audio resolver (`src/rigplane/audio/_usb_resolve.py`, `_resolve_macos` step 7) paired a serial CI-V port to its audio device by **positional index**:

- sort the ioreg audio-node `locationID`s → find the serial prefix's index (`pair_idx`)
- pick `usb_inputs[pair_idx]` / `usb_outputs[pair_idx]` from the flat sounddevice enumeration order.

This assumes the sounddevice enumeration order matches the `locationID`-sorted order **and** that every radio contributes exactly one input entry and one output entry. It **desyncs for mixed-vendor / mixed-shape sets** — e.g. a single duplex C-Media "USB Audio Device" (the Xiegu X6200: ONE device with both input + output channels) next to a split-pair Icom "USB Audio CODEC" (separate input-only + output-only entries). The flat input/output lists no longer line up with the sorted prefixes, so each port resolves to the **other** radio's device. Correct only for single-radio and homogeneous-Icom setups.

## Fix — pair by identity, not position

Select the sounddevice entry that corresponds to the matched audio **device's own identity**, not the i-th USB-audio device in enumeration order:

1. `_find_audio_codec_entries(ioreg_text)` now returns `(product_name, locationID)` pairs (the IORegistry node name is exactly the CoreAudio / sounddevice device name).
2. `_cluster_usb_audio_devices(devices)` groups the enumerated USB-audio entries into per-physical-device clusters: a **duplex** entry (in + out) stands alone; an adjacent **same-named** out-only + in-only pair merges into one split-pair cluster. Each cluster carries its product name.
3. `_pair_audio_device_for_location(...)` selects the matched audio device (the entry sharing the serial port's hub prefix) and picks the sounddevice cluster whose **product name + same-name rank** match it.

The pairing logic is factored into reusable, platform-neutral helpers the future Linux/Windows resolvers (MOR-228/229) can call once they can enumerate `(name, locationID)` audio entries + sounddevice clusters. Public entry point `resolve_audio_for_serial_port(...)` signature is unchanged.

### Design assumption

`sounddevice` exposes no `locationID` field (verified on hardware), so a direct `locationID → sounddevice index` link does not exist. CoreAudio also does **not** enumerate USB audio devices in `locationID` order across vendors (a C-Media device can enumerate before an Icom CODEC with a lower `locationID`) — so a global location-rank correlation is unsafe. The CoreAudio **product name** *is* the IORegistry node name, making it the strongest available identity link; within one vendor/name, remaining ambiguity (e.g. two identical Icom CODECs) is resolved by enumeration order, which for same-model devices tracks `locationID` order — preserving validated homogeneous-multi-radio behaviour. No radio is special-cased by name or VID.

## Tests

- New `TestResolveMixedVendorShapes`: a duplex C-Media "USB Audio Device" (in=1/out=2, prefix `0x1423`) + split-pair Icom "USB Audio CODEC" (out-only + in-only, prefix `0x0111`), enumerated in an interleaved order that desyncs the old flat positional logic. Asserts the X6200 usbmodem port → the C-Media duplex device (rx=1/tx=1) and the Icom usbserial port → the CODEC split pair (rx=3/tx=2), and that the two radios never share a device. Confirmed these fail against the old positional code (each port resolved to the other radio).
- All existing cases stay green: single-radio, two/three homogeneous Icom, Yaesu, X6200.

## Gate

- `uv run pytest tests/test_usb_audio_resolve.py -q` — 51 passed
- `uv run pytest tests/ -q -k "audio or usb or resolve"` — 870 passed, 13 skipped
- `uv run ruff check src/ tests/` — clean; `ruff format --check` on both touched files — clean
- `uv run mypy src/rigplane/audio/_usb_resolve.py` — clean
- `uv.lock` churn reverted

### DIFF
--- a/src/rigplane/audio/_usb_resolve.py
+++ b/src/rigplane/audio/_usb_resolve.py
@@ -17,7 +17,14 @@ rather than fragile, collision-prone device-name strings.
 3. Find all USB audio devices (``USB Audio CODEC``, ``USB Audio Device``) in
    IORegistry with their ``locationID``.
 4. Match audio devices sharing the same hub prefix as the serial port.
-5. Map matched locations to ``sounddevice`` device indices by positional order.
+5. Map the matched audio device to ``sounddevice`` indices by **identity**:
+   group the enumerated USB-audio entries into per-device clusters (a duplex
+   entry stands alone; an output-only entry adjacent to a same-named
+   input-only entry forms one split-pair cluster), then select the cluster
+   whose product name and same-name rank match the IORegistry device. This
+   survives mixed-vendor / mixed-shape sets (a single duplex C-Media device
+   next to a split-pair Icom CODEC), where a flat positional index over
+   inputs/outputs would desync (MOR-230).
 
 **Fallback**: When IORegistry is unavailable, ``ioreg`` is missing, or the
 platform is not macOS, falls back to name-based matching (see
@@ -135,14 +142,14 @@ def _resolve_macos(
 
     serial_prefix = serial_location >> 16
 
-    # 4. Find all USB Audio CODEC locationIDs
-    audio_locations = _find_audio_codec_locations(ioreg_text)
-    if not audio_locations:
+    # 4. Find all USB Audio CODEC (name, locationID) pairs
+    audio_entries = _find_audio_codec_entries(ioreg_text)
+    if not audio_entries:
         logger.warning("usb-audio-resolve: no USB Audio CODEC devices found in ioreg")
         return None
 
     # 5. Check if any audio device shares our hub prefix
-    matching = [loc for loc in audio_locations if (loc >> 16) == serial_prefix]
+    matching = [loc for _name, loc in audio_entries if (loc >> 16) == serial_prefix]
     if not matching:
         logger.warning(
             "usb-audio-resolve: no audio devices match prefix %#06x for %s",
@@ -166,39 +173,19 @@ def _resolve_macos(
 
     devices = list(sd.query_devices())
 
-    # Collect all USB Audio CODEC input/output device indices (ordered)
-    usb_inputs: list[int] = []
-    usb_outputs: list[int] = []
-    for idx, dev in enumerate(devices):
-        if _is_usb_audio_codec(dev.get("name", "")):
-            if _safe_int(dev.get("max_input_channels")) > 0:
-                usb_inputs.append(idx)
-            if _safe_int(dev.get("max_output_channels")) > 0:
-                usb_outputs.append(idx)
-
-    # 7. Determine positional index: sorted unique prefixes → pair index
-    unique_prefixes = sorted(set(loc >> 16 for loc in audio_locations))
-    try:
-        pair_idx = unique_prefixes.index(serial_prefix)
-    except ValueError:
+    # 7. Pair by identity, not positional order across flat input/output lists.
+    pair = _pair_audio_device_for_location(devices, audio_entries, serial_location)
+    if pair is None:
         logger.warning(
-            "usb-audio-resolve: prefix %#06x not in audio prefixes %s",
+            "usb-audio-resolve: could not pair an audio device for prefix %#06x "
+            "(serial loc %#010x, audio entries %s)",
             serial_prefix,
-            [f"{p:#06x}" for p in unique_prefixes],
+            serial_location,
+            [(n, f"{loc:#010x}") for n, loc in audio_entries],
         )
         return None
 
-    if pair_idx >= len(usb_inputs) or pair_idx >= len(usb_outputs):
-        logger.warning(
-            "usb-audio-resolve: pair index %d out of range (inputs=%d, outputs=%d)",
-            pair_idx,
-            len(usb_inputs),
-            len(usb_outputs),
-        )
-        return None
-
-    rx_idx = usb_inputs[pair_idx]
-    tx_idx = usb_outputs[pair_idx]
+    rx_idx, tx_idx = pair
 
     logger.info(
         "usb-audio-resolve: %s → prefix %#06x → RX device [%d], TX device [%d]",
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: BLOCKED

Independent review on a fresh clone of `codex/MOR-230-usb-audio-identity-pairing` (not the implementing worktree). The refactor is well-structured and the mixed-vendor test genuinely catches the old bug, but there is a concrete falsy-zero defect in the new clustering helper that can break audio resolution for a single Icom radio. Blocking on that.

## Bug: device index `0` is dropped during split-pair merge

`src/rigplane/audio/_usb_resolve.py:264`

```python
clusters.append((pend_name, pend_rx or rx, pend_tx or tx))
```

`pend_rx or rx` / `pend_tx or tx` use the truthiness `or` idiom to fill the missing half. But a `sounddevice` index of `0` is falsy, so when the half being merged sits at enumeration index 0, `0 or None` evaluates to `None` and the index is lost.

Reproduced on this branch (single Icom "USB Audio CODEC" split pair with playback at index 0, capture at index 1 — realistic when no built-in output precedes the USB codec, e.g. a headless Mac mini):

```
clusters: [("USB Audio CODEC", 1, None)]   # tx=0 dropped
pair:     None                              # _pair_..._for_location returns None -> resolution FAILS
```

Capture-first ordering at index 0 yields `("USB Audio CODEC", None, 1)` (rx=0 dropped) the same way. `_pair_audio_device_for_location` then hits its "incomplete audio cluster" guard and returns `None`, so `_resolve_macos` returns `None` and USB audio does not resolve at all.

Why the suite misses it: every fixture (`_make_mock_sd_xiegu`, `_make_mock_sd_cmedia_duplex_plus_icom_split`) places a "Built-in Speaker" at index 0, so no USB-audio entry is ever at index 0. The duplex path (line 250, `(name, idx, idx)`) is unaffected because it does not use `or`; only the split-pair merge is broken.

Risk: a previously working single-IC-7610 setup can silently lose audio resolution after this refactor depending on CoreAudio enumeration order — a regression vs. the pre-MOR-230 positional code, which used explicit list indexing (`usb_outputs[pair_idx]`) with no falsy-zero hazard.

### Required fix
Replace the `or`-merge with explicit None checks, e.g.:

```python
merged_rx = rx if pend_rx is None else pend_rx
merged_tx = tx if pend_tx is None else pend_tx
clusters.append((pend_name, merged_rx, merged_tx))
```

Add a regression test with a USB split pair (and a duplex) at sounddevice index 0 — no built-in device ahead of it — asserting a complete `(rx, tx)` cluster and successful resolution.

## What I verified (otherwise sound)
- Diff scope: only `src/rigplane/audio/_usb_resolve.py` + `tests/test_usb_audio_resolve.py`. No cross-layer leakage; imports are stdlib-only.
- Mixed-vendor test is non-tautological: reverting just the src to `main` makes `TestResolveMixedVendorShapes::test_x6200_resolves_to_cmedia_duplex` and `::test_icom_resolves_to_codec_split_pair` FAIL (old positional code cross-wires X6200->idx3, Icom->idx1). Restored, all pass.
- Test counts: `tests/test_usb_audio_resolve.py` 51 passed; full `-k "audio or usb or resolve"` 870 passed, 13 skipped.
- `ruff check` clean, `ruff format --check` clean, `mypy src/rigplane/audio/_usb_resolve.py` clean.
- Identical-name parity (two same-model radios): same-name-rank logic resolves correctly when split halves are grouped per device AND not at index 0 (the index-0 case is the bug above). The doc rationale (name + enumeration-order rank, no locationID in sounddevice) is reasonable.
- Reusability for MOR-228/229: `_pair_audio_device_for_location` and `_cluster_usb_audio_devices` take plain `devices` + `(name, locationID)` entries and contain no macOS-specific calls, so they are cleanly reusable by future resolvers (good, once the bug is fixed).

### Lower-priority observations (not blocking, worth a look)
- `_cluster_usb_audio_devices` adjacency heuristic: if two same-name split devices are enumerated interleaved (pb, pb, cap, cap) the clusters come out mis-paired/incomplete. Not observed on real hardware (USB topology order keeps halves adjacent), but the helper silently produces wrong clusters rather than detecting it. Fine to defer if real enumeration guarantees adjacency.

Please fix the index-0 merge, add the regression test, push, and rerun checks.

morozsm:
Agent Review: PASS

Re-review of PR #1685 (MOR-230, head `6bdf75c7`) — supersedes the prior BLOCKED. The index-0 blocker is **resolved and verified**.

## Blocker fix — verified
`_cluster_usb_audio_devices` (src/rigplane/audio/_usb_resolve.py:268-270) now coalesces split-pair halves with explicit `is None` checks:
```python
merged_rx = pend_rx if pend_rx is not None else rx
merged_tx = pend_tx if pend_tx is not None else tx
```
The old `pend_rx or rx` / `pend_tx or tx` that dropped a valid falsy index 0 is gone. No other falsy-index hazard in the pairing path: the duplex branch (:250) uses `idx` directly, and `_pair_audio_device_for_location` guards with `rx is None or tx is None` (:347), not truthiness.

## Regression test — non-tautological (confirmed)
`TestResolveSplitPairAtIndexZero` (tests/test_usb_audio_resolve.py:767) models a split CODEC at indices 0/1 (output-only at the falsy index 0, no built-in audio) and asserts `tx_device_index == 0`. Swapping in the buggy parent (`ed6ddc81`, `pend_tx or tx`), the test FAILS exactly as the blocker predicts — `tx=None` → incomplete cluster → resolve returns None. With the fix it passes. (Note: reverting against `main` is not a valid tautology check here — main predates the clustering code and uses a different positional path; the meaningful baseline is the buggy clustering commit, against which the test correctly fails.)

## Gate — all green
- `pytest tests/test_usb_audio_resolve.py` → **52 passed**
- `pytest -k "audio or usb or resolve"` → 871 passed, 13 skipped
- `ruff check src/ tests/` → clean; `ruff format --check` on both files → already formatted
- `mypy src/rigplane/audio/_usb_resolve.py` → clean
- diff scope → still only the two files (resolver + its tests); mixed-vendor intent (`TestResolveMixedVendorShapes`) intact

## Non-blocking (acknowledged, not required for this PR)
The interleaved `pb,pb,cap,cap` same-name adjacency edge from the prior review remains; unlikely on real hardware and acceptable to leave documented.

Verdict: PASS — may merge once required checks are green and this PASS is fresh for the current head.

PR #1689, opened 2026-05-29
fix(MOR-237): stabilize X6200 CH342 serial session + CI-V RX framing

morozsm:
## MOR-237 — Xiegu X6200 over CH342 USB serial: session flapping + short-BCD CI-V errors

Two related faults observed live (`~/Library/Logs/rigplane-pro/rigplane-managed.log`) on the X6200 over the WCH CH342 dual-serial bridge (`/dev/cu.usbmodem58910181093` @ 19200). The fix also hardens the shared Icom serial path generally — no X6200 port hardcoding.

### Root causes

**1. CI-V RX framing — truncated frequency frames hit the BCD decoder.**
The X6200 / a flaky serial line delivered short frequency frames parsed as `cmd=0x00 sub=0x00 data=` (0 bytes) and `cmd=0x00 ... data=0000` (2 bytes). `cmd 0x00` (unsolicited freq) and `0x03` (GET-freq response) route into `parse_frequency_response` → `bcd_decode`, which requires exactly 5 bytes and raised `ValueError: BCD data must be exactly 5 bytes, got 0/2`. The exception was caught but logged on every poll cycle (~6980 debug lines in the captured log), once in `_update_state_cache_from_frame` and once in `_update_radio_state_from_frame`.

**2. Serial stability — spurious reconnect loop with traceback flood.**
`_serial_civ_watchdog_loop` calls `soft_reconnect()` the instant `serial_session.ready` is false. A single recoverable read hiccup flips `healthy → False`; the watchdog then disconnects the working session and tries to reopen. When the CH342 device node briefly vanished during macOS USB renumbering (`SerialException: ... No such file or directory`), `soft_reconnect` failed, logged a full-traceback `WARNING`, and retried every `_SERIAL_WATCHDOG_RETRY_S` (0.5s) — forever (`Failed to reconnect serial session ...` repeated indefinitely, interleaved with `Connected to X6200 ...`).

### Fix

- **CI-V framing (`runtime/_civ_rx.py`):** validate freq-frame payload is exactly `_FREQ_BCD_LEN` (5) bytes before decoding, in both the legacy `_update_state_cache_from_frame` cmd 0x00/0x03 branch and the `_handle_freq` dispatch handler. Short/partial/over-long freq frames are skipped with a single DEBUG line — the 5-byte BCD decoder never sees a malformed payload. No exception, no state corruption.
- **Serial stability (`backends/_icom_serial_base.py`):** the watchdog now tracks consecutive reconnect failures, applies a **capped exponential backoff** (`_SERIAL_WATCHDOG_RETRY_S` doubling up to `_SERIAL_WATCHDOG_RETRY_MAX_S = 5.0s`), and **demotes repeated identical failures from WARNING+traceback to DEBUG**. The first failure of a run is still a WARNING with traceback for diagnosis; a vanished/transient port is then retried slowly and quietly and recovers (counter resets) when the node returns. Genuine disconnects still trigger an immediate first reconnect — the existing recovery test is unchanged.

This stays within the existing layering (`runtime/` framing, `backends/` serial session) and does not touch the X6200 audio/discovery/disambiguation work from MOR-219/224/226.

### Tests
- New `tests/test_civ_rx_short_frame_robustness.py`: short (0/2/3/4/6-byte) freq frames for cmd 0x00 and 0x03 — asserts no raise, no state corruption, no `BCD data must be exactly` log, and that a valid 5-byte frame after a short one still decodes.
- `tests/test_icom7610_serial_radio.py`: added a capped-exponential-backoff unit test and a transient-open-failure (`FileNotFoundError` "port gone" on calls 2–5, recover on 6) test asserting **at most one WARNING** for the whole outage. Existing `test_serial_watchdog_retries_after_transient_soft_reconnect_failure` still passes.

### Gate
- `uv run pytest -k "serial or civ or x6200 or backend"` — pass
- `uv run pytest tests/ --ignore=tests/integration` — **6335 passed**, 0 failed
- `uv run ruff check` + `ruff format --check` on changed files — clean
- `uv run mypy src/` — no new errors in touched files (the one `_civ_rx.py:403 redundant-cast` is the pre-existing baseline, confirmed against `origin/main`)
- `uv.lock` churn reverted

### Manual hardware-validation checklist (could not run — no physical X6200)
1. Connect the X6200 via USB (CH342 dual-serial); confirm macOS enumerates `/dev/cu.usbmodem…091` (SERIAL-A) and `…093` (SERIAL-B).
2. Start the managed station against the X6200 serial backend @ 19200 baud.
3. Tail `~/Library/Logs/rigplane-pro/rigplane-managed.log` and confirm:
   - no recurring `BCD data must be exactly 5 bytes` lines,
   - no `Failed to reconnect serial session … No such file or directory` reconnect loop (at most a single WARNING if the node briefly drops, then quiet DEBUG backoff, then `CI-V data resumed` / `Connected to X6200`),
   - stable freq/mode display while spinning the VFO and changing modes.
4. Physically unplug/replug (or trigger a USB renumber) and confirm the session recovers without a traceback flood.

### Uncertainty
- The captured short frames are `cmd=0x00` with 0/2 data bytes — clearly **truncated**, not a valid alternate shape (a 4-byte X6200 freq would log as `data=00400714`, never seen). The fix treats them as partial and skips them, which is correct for truncation. If the X6200 ever emits a genuinely valid non-5-byte freq frame on real hardware, that would surface as a steady stream of `skipping short freq frame` DEBUG lines with a frozen frequency — the hardware run in the checklist will confirm freq still tracks (it should, since the valid 5-byte frames are unaffected).
- Whether the truncation originates from the flaky serial (partial reads) or an embedded `0xFD`-byte mis-frame in `SerialFrameCodec`/`iter_civ_frames` could not be confirmed without the device; the `_civ_rx` guard is robust to both. Stabilizing the serial session (fault #2) should also reduce the rate of partial frames feeding fault #1.

Independent review pending.

### DIFF
--- a/src/rigplane/backends/_icom_serial_base.py
+++ b/src/rigplane/backends/_icom_serial_base.py
@@ -83,6 +83,12 @@ class _IcomSerialRadioBase(CoreRadio):
     _DEFAULT_MODEL: str = ""
     _SERIAL_WATCHDOG_INTERVAL_S = 0.2
     _SERIAL_WATCHDOG_RETRY_S = 0.5
+    # Cap for the exponential backoff applied after repeated reconnect failures
+    # (e.g. the USB serial device node briefly disappearing during macOS device
+    # renumbering on a CH342 bridge — MOR-237). Without a backoff the watchdog
+    # retried every _SERIAL_WATCHDOG_RETRY_S and flooded the log with full
+    # tracebacks twice a second while the port was gone.
+    _SERIAL_WATCHDOG_RETRY_MAX_S = 5.0
 
     def __init__(
         self,
@@ -534,6 +540,11 @@ class _IcomSerialRadioBase(CoreRadio):
         self._civ_data_watchdog_task = None
 
     async def _serial_civ_watchdog_loop(self) -> None:
+        # Number of consecutive failed soft-reconnect attempts. Drives a capped
+        # exponential backoff and demotes repeated, identical failures from a
+        # WARNING-with-traceback to a quiet DEBUG so a transient/vanished port
+        # does not flood the log (MOR-237).
+        consecutive_failures = 0
         try:
             while True:
                 await asyncio.sleep(self._SERIAL_WATCHDOG_INTERVAL_S)
@@ -547,20 +558,55 @@ class _IcomSerialRadioBase(CoreRadio):
                     self._civ_recovering = False
                     self._conn_state = RadioConnectionState.CONNECTED
                     self._last_civ_data_received = time.monotonic()
+                    consecutive_failures = 0
                     continue
 
                 self._civ_stream_ready = False
                 self._civ_recovering = True
                 try:
                     await self.soft_reconnect()
-                except Exception:
-                    logger.warning(
-                        "serial-civ-watchdog: soft reconnect failed",
-                        exc_info=True,
+                    consecutive_failures = 0
+                except Exception as exc:
+                    consecutive_failures += 1
+                    if consecutive_failures == 1:
+                        # First failure of a run: surface it once, with the
+                        # full traceback for diagnosis.
+                        logger.warning(
+                            "serial-civ-watchdog: soft reconnect failed (%s); "
+                            "retrying with backoff",
+                            exc,
+                            exc_info=True,
+                        )
+                    else:
+                        # Subsequent identical failures (e.g. port still gone):
+                        # keep the log quiet, just note the count.
+                        logger.debug(
+                            "serial-civ-watchdog: soft reconnect still failing "
+                            "(attempt %d): %s",
+                            consecutive_failures,
+                            exc,
+                        )
+                    await asyncio.sleep(
+                        self._serial_watchdog_retry_delay(consecutive_failures)
                     )
-                    await asyncio.sleep(self._SERIAL_WATCHDOG_RETRY_S)
         except asyncio.CancelledError:
             pass
+
+    def _serial_watchdog_retry_delay(self, consecutive_failures: int) -> float:
+        """Capped exponential backoff for repeated reconnect failures.
+
+        The first retry uses ``_SERIAL_WATCHDOG_RETRY_S``; each subsequent
+        failure doubles the delay up to ``_SERIAL_WATCHDOG_RETRY_MAX_S`` so a
+        long-absent port is retried slowly and quietly instead of twice a
+        second with a full traceback (MOR-237).
+        """
+        exponent = max(0, consecutive_failures - 1)
+        multiplier: float = float(2**exponent)
+        delay: float = float(self._SERIAL_WATCHDOG_RETRY_S) * multiplier
+        cap: float = float(self._SERIAL_WATCHDOG_RETRY_MAX_S)
+        if delay > cap:
+            return cap
+        return delay
 
     # ------------------------------------------------------------------
     # Internal helpers
--- a/src/rigplane/runtime/_civ_rx.py
+++ b/src/rigplane/runtime/_civ_rx.py
@@ -49,6 +49,12 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 CIV_HEADER_SIZE = 0x15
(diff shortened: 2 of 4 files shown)


morozsm:
Agent Review: PASS

Independent review (fresh clone, branch `codex/mor-237-x6200-serial-stability`). I did not write this code.

## Decisive question — reader/reassembly: the framer is SOUND; this is correctly a tolerate-and-resync fix, not a papered-over reassembly bug

I traced the full RX path and empirically reproduced the ground-truth split against the actual codec:

- Reader is `SerialFrameCodec.feed()` in `src/rigplane/backends/icom7610/drivers/serial_civ_link.py`. It **buffers across `read()` calls** (a persistent `bytearray`, emits a frame only on finding `0xFD`).
- Flow: `SerialCivLink.receive()` → `SerialCivTransport.receive_packet()` → `_civ_rx_loop` (`runtime/_civ_rx.py:710`) → `iter_civ_frames` + `parse_civ_frame` → `_route_civ_frame`.

Feeding the authoritative wire frame `FE FE E0 A4 03 00 40 07 14 00 FD` split exactly as the ground truth describes:
- cut after the cmd byte ("got 0"): `feed(FE FE E0 A4 03)` → `[]`; `feed(00 40 07 14 00 FD)` → one complete frame. Reassembled correctly.
- cut after 2 BCD bytes ("got 2"): `feed(FE FE E0 A4 03 00 40)` → `[]`; rest → one complete frame. Reassembled correctly.

So a **partial `read()` does NOT produce a short CivFrame** — the codec absorbs it. The only way a `cmd=0x00 data= len=0` frame reaches `_civ_rx` is a genuinely *complete-but-short* frame at the codec, i.e. line corruption that injects a spurious `0xFD`/`FE FE` into the stream: `feed(FE FE E0 A4 03 FD)` → emits `fefee0a403fd` → `parse_civ_frame` → `cmd=0x03 data= len=0`. That matches the "got 0/got 2" log exactly. The codec then resyncs on the next SOF, so a later complete 5-byte frame lands and freq updates normally — no stall.

**Conclusion:** there is no reassembly bug to fix. The reader already reassembles partial reads correctly; the residual short frames are irrecoverable transient line corruption that any framer would surface. `_civ_rx` skip-and-resync is the correct, sufficient handling, and the watchdog session-stabilization reduces the corruption rate. A reader-level change is NOT required to call MOR-237 resolved.

## Verification
- Diff scope: exactly `runtime/_civ_rx.py`, `backends/_icom_serial_base.py`, `tests/test_civ_rx_short_frame_robustness.py`, `tests/test_icom7610_serial_radio.py`. Within guardrails. No touch to X6200 audio/discovery (MOR-219/224/226) or other backends.
- Tests are meaningful (verified by reverting each src file to origin/main):
  - short-frame robustness: **10 failures** without the guard (BCD ValueError / state corruption), green with it.
  - watchdog quiet-on-transient-outage: baseline emits **4 WARNINGs**; fix asserts ≤1 and passes. Backoff test fails on baseline (no `_SERIAL_WATCHDOG_RETRY_MAX_S`).
  - existing `test_serial_watchdog_retries_after_transient_soft_reconnect_failure` still passes — fast first reconnect (`retry_delay(1)==base==0.5s`) and counter reset on recovery (lines 561/568) preserved; cap 5.0s sane.
- `uv run pytest -k "serial or civ or x6200 or backend"`: 761 passed (1 unrelated flake — `test_web_server::test_info_and_state_flow_from_serial_mock_radio`, a TimeoutError under parallel load; passes in isolation).
- `uv run pytest tests/ --ignore=tests/integration`: **6335 passed**, 0 failed.
- ruff check + format --check on changed files: clean.
- mypy src/: the only error in a touched file is the pre-existing `_civ_rx.py` redundant-cast on the unchanged `_civ_frame_to_bytes` helper (confirmed on origin/main). No new errors.

Minor non-blocking note: WARNING is keyed on `consecutive_failures==1`, so a flapping port re-warns per failure-run — acceptable/desirable visibility.

Sound fix at the right layer. Approving.

Issue #425, opened 2026-03-30
IC-7610 web runtime: intermittent CI-V watchdog reconnect loop drops live scope/waterfall


morozsm:
## Summary
On a real IC-7610, the web runtime can come up successfully, stream live hardware scope/waterfall for a short time, then lose CI-V data and enter a reconnect loop:

- `civ-data-watchdog: no CI-V data for 2.xs`
- `OpenClose failed for 5.1s, triggering soft_reconnect`
- `Disconnected`
- `Soft reconnect ...`
- scope/waterfall disappear in the UI

This is separate from the audio FFT layout work. The scope UI may be correct, but the live transport becomes unstable and drops the continuous stream.

## Reproduction
1. Start current web server against a real IC-7610.
2. Open the web UI.
3. Observe that the main hardware waterfall may appear briefly.
4. Within seconds/minutes it can disappear as the runtime reconnects.

## Observed logs
Typical sequence:
- `Connected to ... (control=50001, civ=50002)`
- `scope: enable queued after radio_reconnect`
- `civ-data-watchdog: no CI-V data for 2.1s, requesting data start`
- `OpenClose failed for 5.1s, triggering soft_reconnect`
- `Disconnected`
- reconnect cycle repeats

## Impact
- Main spectrum/waterfall vanish mid-session
- Audio FFT verification becomes unreliable because the shared radio runtime is unstable
- Browser WS clients churn as reconnects happen

## Investigation areas
- watchdog threshold too aggressive for IC-7610 live scope sessions
- false positive on "no CI-V data" while scope stream is actually transitioning
- interaction between scope enable/re-enable and CI-V Open/Close
- reconnect path may be tearing down a still-usable CI-V transport
- potential interaction with multiple concurrent WS channels (control + scope + audio-scope)

## Acceptance criteria
- Web UI can stay connected to a real IC-7610 for an extended smoke test without waterfall disappearing
- No repeated soft reconnect loop under normal idle/operator usage
- Watchdog only reconnects on real transport failure, not transient scope enable gaps

PR #1694, opened 2026-05-30
fix(MOR-239): resume AudioContext on LIVE gesture so RX audio plays

morozsm:
## Summary

Fixes **MOR-239** — the last blocker for rigplane-pro beta.7 LIVE audio. On desktop (Tauri WKWebView) the RX AUDIO **LIVE / Browser audio stream** produced **no sound** and a **blank Audio Scope**, even though the backend was streaming real PCM16 MONO frames over the audio WS (confirmed: USB capture RMS −40.5 dBFS, server log `audio: sent frame #1 … #425`).

## Confirmed root cause

Frontend, not capture/backend. Under the WKWebView / mobile-Safari **autoplay policy** an `AudioContext` starts (and can re-enter) `suspended`. `RxPlayer.playPcm16()` / `decodeOpus()` then **silently `return`ed for every frame** when `ctx.state === 'suspended'`, so nothing was ever scheduled → no audio. The async `resume().catch(() => {})` in `start()` was the only resume attempt and was easily lost.

Mono routing was re-checked and is **not** the cause: a 1-ch buffer up-mixes to `splitter[0] → mainGain → mainPanner → destination`, and `mainGain` is audible under the default `focus='both'`.

## Fix

- **Resume inside the user gesture** — `start()` resumes synchronously in the LIVE-button click stack (`onclick → onMonitorModeChange('live') → AudioManager.startRx → RxPlayer.start`), the only place WKWebView honours `resume()`. Idempotent when re-invoked.
- **Re-nudge resume** on every `audio_start` / WS (re)connect (`AudioManager.onopen`).
- **Resume on focus/visibility regain** — WKWebView re-suspends a backgrounded webview; `visibilitychange` + window `focus` listeners re-attempt resume (attached on `start`, torn down on `stop`).
- **No more silent drops** — `_ensureRunning()` re-attempts `resume()` and emits a throttled `console.warn` when frames are dropped due to a suspended context, so the state is diagnosable. Frames flow as soon as the context reaches `running`.

## Tests / gates (in worktree, off latest `main`)

- `npm test` — **1847 passed** (4 new MOR-239 tests: suspended→resume+warn, resume-then-play, focus-regain resume + listener teardown, mono→audible-MAIN routing)
- `npm run lint` — clean
- `npm run check` (svelte-check + tsc) — **0 errors / 0 warnings**
- `npm run build` — success

## Manual hardware-validation checklist (cannot run desktop WebView here)

1. Launch rigplane-pro desktop (Tauri) build with this branch's frontend; connect the X6200 (USB audio + serial).
2. Open the web UI, click RX AUDIO **LIVE**.
3. Confirm **audible RX audio** from the host output.
4. Confirm the **Audio Scope shows a live waveform**.
5. Background the window / switch tabs, then refocus → audio resumes (no manual re-click).
6. With DevTools open, verify **no** repeating `RxPlayer: AudioContext suspended …` warnings during normal playback (a few at startup before the gesture resolves are acceptable).

> Independent agent review pending — implementation agent cannot self-review.

### DIFF
--- a/frontend/src/lib/audio/__tests__/rx-player.test.ts
+++ b/frontend/src/lib/audio/__tests__/rx-player.test.ts
@@ -248,3 +248,79 @@ describe('RxPlayer jitter bounds (#1363)', () => {
     p.stop();
   });
 });
+
+describe('RxPlayer suspended-context recovery (MOR-239)', () => {
+  it('start() resumes a context that is created suspended (WKWebView autoplay)', () => {
+    ctx.state = 'suspended';
+    const p = new RxPlayer();
+    p.start();
+    expect(ctx.resume).toHaveBeenCalled();
+    p.stop();
+  });
+
+  it('drops the frame while suspended but re-attempts resume and warns (not silent)', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    ctx.state = 'suspended';
+    const p = new RxPlayer();
+    p.start();
+    ctx.resume.mockClear();
+
+    p.feed(pcm16(480));
+
+    // Frame can't play on a suspended ctx, but instead of a silent return
+    // we re-attempt resume() and surface a warning so it's observable.
+    expect(ctx.createBufferSource).not.toHaveBeenCalled();
+    expect(ctx.resume).toHaveBeenCalled();
+    expect(warn).toHaveBeenCalled();
+    p.stop();
+    warn.mockRestore();
+  });
+
+  it('plays frames once the context has resumed to running', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    ctx.state = 'suspended';
+    const p = new RxPlayer();
+    p.start();
+    p.feed(pcm16(480));           // dropped — suspended
+    expect(ctx._lastSrc).toBeNull();
+
+    ctx.state = 'running';        // resume() honoured by the gesture
+    p.feed(pcm16(480));           // now audible
+    expect(ctx.createBufferSource).toHaveBeenCalled();
+    expect(ctx._lastSrc.start).toHaveBeenCalled();
+    p.stop();
+    warn.mockRestore();
+  });
+
+  it('re-attempts resume on focus/visibility regain (WKWebView re-suspend)', () => {
+    ctx.state = 'suspended';
+    const p = new RxPlayer();
+    p.start();
+    ctx.resume.mockClear();
+    window.dispatchEvent(new Event('focus'));
+    expect(ctx.resume).toHaveBeenCalled();
+    p.stop();
+    // After stop the listener must be gone — no further resume calls.
+    ctx.resume.mockClear();
+    window.dispatchEvent(new Event('focus'));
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-239 (desktop/WKWebView RX audio + blank Audio Scope). Reviewed the full diff and read all three changed files in full, plus the LIVE-button gesture chain and the AudioManager WS lifecycle.

## Gates (re-run locally, all green)
- `npm test` → 1847 passed / 105 files, 0 failures
- `npm run check` → 0 errors, 0 warnings
- `npm run lint` → clean

## Correctness — gesture attribution holds
- LIVE click path: `command-bus.ts` `onMonitorModeChange('live')` → `audioManager.startRx()` → `rxPlayer.start()` → `_resume()` → `ctx.resume()`, all synchronous with no `await` interposed before the resume. The `resume()` runs inside the gesture call stack, which is exactly what WKWebView/mobile-Safari autoplay policy requires. The earlier-built graph and `_resume()` both run before any WS work in `startRx()`.
- Re-nudge points are correct and idempotent: `start()` on an existing context early-returns to a bare `_resume()` (no double-build, no graph re-wire), so the `ws.onopen` re-invoke and the `audio_start` re-invoke are safe. Context persists across WS reconnects (only `stopRx` calls `rxPlayer.stop()`).

## Regressions — none found
- Listener lifecycle is clean: `_attachResumeListeners` is idempotent (`_resumeListenersAttached` guard), uses a stable arrow-property handler, and `_detachResumeListeners` removes the same reference. `stop()` detaches. The test proves no post-stop resume fires (focus event after stop → `resume` not called). No leak, no double-attach.
- `stop()` during reconnect is safe: `stopRx` sets `_rxEnabled=false` before `stop()`/`maybeDisconnect`→`close()` clears the reconnect timer; a late reconnect `onopen` sees `_rxEnabled=false` and won't re-start.
- Opus path behavior preserved: `_ensureRunning()` gates before decoder creation (same as the prior suspended early-return), so no half-configured decoder.

## Warn-throttle
`_droppedSuspendedFrames <= 3 || % 200 === 0` with reset-to-0 on running. Sound; counter growth while permanently suspended is harmless. Replaces the silent frame-drop that made the bug undiagnosable — good observability win.

## Mono routing
The claim (mono PCM16 → splitter output[0]=MAIN, output[1]=SUB silent) is correct per WebAudio: a 1-ch source into a discrete-interpretation `ChannelSplitter(2)` up-mixes to ch0=signal, ch1=silence, and default `focus='both'`/`mainGainDb=0` → MAIN gain = 1. Minor nit (non-blocking): the test mock splitter is a stub with no channel semantics, so the mono test only verifies `connect(preGain)` + `mainGain>0`, not the splitter routing itself — the inline comment slightly overstates what is mechanically proven. The behavioral assertions are still valid and non-tautological (suspended-drop, resume-on-running, focus-after-stop tests genuinely exercise state transitions).

Scope is within guardrails (3 files), no new abstractions, layering respected (audio package only). Mergeable once required checks are green.

PR #1695, opened 2026-05-30
fix(MOR-241): dispatch audio FFT frames to /api/v1/audio-scope (fix blank scope on non-hw-scope radios)

morozsm:
## Root cause

`AudioFftScope.on_frame()` (`src/rigplane/audio/fft_scope.py`) is a **single-slot** callback setter. `WebServer.__init__` called it twice for non-hardware-scope radios:

1. `self._audio_fft_scope.on_frame(self._broadcast_audio_scope)` → `/api/v1/audio-scope`
2. (only when `CAP_SCOPE` absent) `self._audio_fft_scope.on_frame(self._broadcast_scope)` → `/api/v1/scope`

The second call **overwrote** the first, so `_broadcast_audio_scope` became dead code and `/api/v1/audio-scope` received **zero FFT frames** → the AUDIO SCOPE panel rendered blank on the X6200. IC-7610 (`CAP_SCOPE` present) skipped step 2 and was unaffected.

## The fix

Register a single `_dispatch_audio_fft_frame(frame)` callback that fans out:

- always feeds `_broadcast_audio_scope` (`/api/v1/audio-scope`);
- only for radios **without** a hardware scope (`CAP_SCOPE not in radio.capabilities`) also feeds `_broadcast_scope` (`/api/v1/scope`), since those radios derive their main panadapter from RX audio.

The PCM tap registration (`set_pcm_tap(... feed_audio)`) is unchanged. One source file touched (`server.py`); `fft_scope.py` and the single-slot engine are untouched.

## Compatibility (IC-7610 byte-identical)

For hardware-scope radios `_dispatch_audio_fft_frame` calls **only** `_broadcast_audio_scope` and never `_broadcast_scope` — `/api/v1/scope` stays sourced exclusively from the real hardware scope. No behavior change for that class.

## Tests

- `tests/test_audio_scope_dispatch.py` (new, hardware-free):
  - `test_x6200_audio_scope_receives_fft_frames` — reuses the MOR-236 mono-USB X6200 pattern (real `Ic705SerialRadio` + `WebServer`), feeds synthetic PCM16 through the broadcaster PCM tap, asserts ≥1 `ScopeFrame` on the audio-scope handler **and** that `/api/v1/scope` is still fed. Confirmed to **FAIL on the unfixed code** (0 audio-scope frames) and **PASS after the fix**.
  - `test_hardware_scope_radio_audio_fft_only_on_audio_scope` — IC-7610-class non-regression: audio FFT reaches `/api/v1/audio-scope` but never `/api/v1/scope`.
- `tests/integration/test_x6200_audio_scope_smoke.py` (new, **OFF by default**): real X6200 serial + USB RX audio, asserts a live FFT frame arrives within 5s. Double-gated — under `tests/integration/` (ignored by the default `--ignore=tests/integration` run), `serial_integration` marker, and `RIGPLANE_HW_SMOKE=1` env opt-in. Verified skipped in a normal run.

## Gate results

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **6347 passed, 7 skipped, 11 xfailed, 1 xpassed, 0 failures**
- `uv run ruff check src/ tests/` → clean; `uv run ruff format src/ tests/` → clean
- `uv run mypy src/` → no new errors in `server.py`; 6 pre-existing baseline errors remain in `backends/yaesu_cat/radio.py` and `runtime/_civ_rx.py` (unrelated)

Independent review pending.

### DIFF
--- a/src/rigplane/web/server.py
+++ b/src/rigplane/web/server.py
@@ -413,10 +413,12 @@ class WebServer:
         _has_scope = (CAP_SCOPE in radio.capabilities) if radio is not None else False
         if radio is not None and _has_audio:
             self._audio_fft_scope = AudioFftScope(fft_size=2048, fps=20, avg_count=2)
-            self._audio_fft_scope.on_frame(self._broadcast_audio_scope)
+            # AudioFftScope.on_frame() is a single-slot setter: register one
+            # dispatch method that fans out to BOTH broadcasters. Registering
+            # twice would clobber the first callback (MOR-241).
+            self._audio_fft_scope.on_frame(self._dispatch_audio_fft_frame)
             if not _has_scope:
-                # No hardware scope — audio FFT also feeds /api/v1/scope
-                self._audio_fft_scope.on_frame(self._broadcast_scope)
+                # No hardware scope — audio FFT also feeds /api/v1/scope.
                 self._audio_broadcaster.set_pcm_tap(self._audio_fft_scope.feed_audio)
             logger.info(
                 "Audio FFT scope available (has_audio=%s, has_hw_scope=%s)",
@@ -608,6 +610,24 @@ class WebServer:
             )
             if self._radio is not None:
                 self._set_scope_data_callback(self._broadcast_scope)
+
+    def _dispatch_audio_fft_frame(self, frame: Any) -> None:
+        """Fan out an audio FFT frame to the relevant scope channels.
+
+        Single registered callback for :class:`AudioFftScope` (whose
+        ``on_frame`` is a single-slot setter). Always feeds the dedicated
+        ``/api/v1/audio-scope`` channel. For radios WITHOUT a hardware scope
+        the same frame also drives ``/api/v1/scope`` (the main panadapter),
+        since those radios derive their spectrum from RX audio. Hardware-scope
+        radios (e.g. IC-7610) keep ``/api/v1/scope`` sourced exclusively from
+        the real scope, so the audio FFT never touches it (MOR-241).
+        """
+        self._broadcast_audio_scope(frame)
+        _has_scope = (
+            CAP_SCOPE in self._radio.capabilities if self._radio is not None else False
+        )
+        if not _has_scope:
+            self._broadcast_scope(frame)
 
     def _broadcast_scope(self, frame: Any) -> None:
         """Broadcast scope frame to all registered handlers.
--- a/tests/integration/test_x6200_audio_scope_smoke.py
+++ b/tests/integration/test_x6200_audio_scope_smoke.py
@@ -1,0 +1,100 @@
+"""Hardware-gated smoke test for MOR-241 — X6200 AUDIO SCOPE delivers frames.
+
+Brings up the real ``/api/v1/audio-scope`` wiring against a physical Xiegu
+X6200 (serial CI-V + USB RX audio) and asserts that at least one FFT
+:class:`ScopeFrame` reaches the audio-scope channel within a few seconds.
+
+OFF BY DEFAULT. This test is double-gated and is skipped in a normal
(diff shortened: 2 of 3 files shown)


morozsm:
Agent Review: PASS

Independent adversarial review of MOR-241 (codex/mor-241-audio-scope-dispatch). Confirmed the fix is correct, scoped, and well-guarded by tests.

## Scope
Exactly 3 files changed (+388/-3): `src/rigplane/web/server.py` (only source file), `tests/test_audio_scope_dispatch.py`, `tests/integration/test_x6200_audio_scope_smoke.py`. Engine `src/rigplane/audio/fft_scope.py` is untouched (`git diff` empty) — no second callback slot added; `on_frame` remains single-slot (`self._callback = callback`, server.py:163-169 of fft_scope), which is exactly the clobber the fix routes around.

## Correctness
- Root cause confirmed against origin/main: `__init__` previously called `on_frame(_broadcast_audio_scope)` then, for non-hw-scope radios, `on_frame(_broadcast_scope)` — the second assignment clobbered the first, so `/api/v1/audio-scope` got 0 frames. Fix registers one `_dispatch_audio_fft_frame` and keeps the `set_pcm_tap` line byte-identical inside the same `if not _has_scope` branch.
- Hardware-scope non-regression: `_dispatch_audio_fft_frame` (server.py:614-631) calls `_broadcast_audio_scope` unconditionally and `_broadcast_scope` ONLY when `CAP_SCOPE not in self._radio.capabilities`. For IC-7610 (`_has_scope=True`) the audio FFT never touches `/api/v1/scope`; that channel stays sourced exclusively from the real hardware scope. Verified by the sibling test asserting `main_handler.frames == []`.
- Predicate parity: dispatch uses `CAP_SCOPE in self._radio.capabilities` (line 627), identical to the `_has_scope` local (line 413) and to the pre-existing tap-disable check (line 670-674). `CAP_SCOPE == "scope"` confirmed. `self._radio = radio` (line 387) is the same object. Note: a separate `runtime_capabilities()`-based `_supports_scope` helper exists and can diverge for a radio that tags "scope" without implementing `ScopeCapable`, but that divergence pre-exists this PR and the fix faithfully preserves the original predicate — not a regression.

## Tests (genuine, non-tautological)
- Verified the X6200 guard FAILS on pre-fix code: temporarily reverted server.py to origin/main and ran the new test — `AssertionError: audio-scope handler received zero FFT frames ... assert 0 >= 1`. After restoring the fix it passes. Real synthetic PCM16 is fed through `_audio_broadcaster._tap_registry.feed()` -> `feed_audio` -> FFT -> dispatch; the `center_freq>0` gate is satisfied via `set_center_freq(14_074_000)` and the fps limit bypassed via `_last_frame_time = 0.0`. Test also asserts the X6200 MAIN `/api/v1/scope` still receives frames.
- IC-7610 sibling test drives `_dispatch_audio_fft_frame` directly and asserts the FFT frame reaches audio-scope but NOT `/api/v1/scope`.
- Smoke test is triple-gated OFF by default: lives under `tests/integration/` (run ignores it), carries `serial_integration` marker, and additionally requires `RIGPLANE_HW_SMOKE=1`.

## Gates (all green, re-run by me)
- `uv run pytest tests/ -q --ignore=tests/integration`: 6347 passed, 7 skipped, 0 failures; smoke not run.
- `uv run ruff check src/ tests/`: All checks passed.
- `uv run lint-imports`: 4 contracts kept, 0 broken.
- `uv run mypy src/`: 6 errors, all the pre-existing baseline in yaesu_cat/radio.py + _civ_rx.py; zero new errors in server.py.

PR is OPEN, not draft, base main. No unplanned changes. Approving for merge once required checks are green.

PR #1696, opened 2026-05-30
fix(MOR-242): arm USB-audio TX for serial Icom backends + RX-only bridge fallback

morozsm:
## Root cause (corrected)

`_IcomSerialRadioBase.start_audio_tx_pcm` opened the correct USB CODEC TX stream via `self._serial_audio_driver.start_tx(...)` but **never set `self._pcm_tx_fmt`**, and the base had **no `push_audio_tx_pcm` override**. So every push fell through to `AudioRuntimeMixin.push_audio_tx_pcm`, whose guard `if self._pcm_tx_fmt is None: raise RuntimeError("PCM TX not started…")` fired on every frame. The auto-bridge enables TX by default, so `_tx_loop` pushed loopback audio and the radio rejected every frame, spamming "PCM TX not started".

The working TX path lived **only** in `Icom7610SerialRadio`, so the bug was systemic to all non-7610 serial Icom backends.

## Part A — real fix (lift the working TX path into the shared base)

`src/rigplane/backends/_icom_serial_base.py`:
- `start_audio_tx_pcm` now clamps `channels → 1` (USB CODEC mic is mono-only; GH#1382, consistent with the MOR-238 RX clamp) and arms `self._pcm_tx_fmt = (sample_rate, 1, frame_ms)`.
- added `push_audio_tx_pcm` (guards `_pcm_tx_fmt`, validates frame size against the armed format, forwards to the USB audio driver).
- added `start/push/stop_audio_tx_opus`.
- `_pcm_tx_fmt` is cleared in `stop_audio_tx_pcm` and `_stop_serial_audio_driver`.

`src/rigplane/backends/icom7610/serial.py`: deleted the now-duplicate TX/teardown overrides so IC-7610-serial **inherits** the base. Behavior is pinned identical by the **unchanged** `tests/test_icom7610_serial_radio.py` (22 tests, green).

**Systemic scope:** fixes IC-705, X6200 (routed through the IC-705 backend), IC-7300, and IC-9700 serial TX.

## Part B2 — bridge safety (RX-only fallback)

`src/rigplane/audio/bridge.py`: `_setup_streams` wraps `start_audio_tx_pcm`, and `_tx_loop` guards the first `push_audio_tx_pcm`, so a `RuntimeError`/`NotImplementedError` downgrades the bridge to RX-only (disable TX, skip/stop `_tx_loop`, log **once** at WARNING) instead of per-frame spam. Insurance so this class of bug can never spam again. LAN, 7610, and yaesu-cat TX paths are unaffected.

## Tests

- Parametrized base-backend TX test (IC-705, X6200 model, IC-7300, IC-9700): arms `_pcm_tx_fmt`, push reaches the fake USB driver. Confirmed it **fails on the unfixed base** and passes after the lift.
- Bridge reproducer: a real `AudioBridge` + `FakeAudioBackend` against a fake-driver-backed serial radio — a captured frame reaches the driver with no RuntimeError (the AsyncMock bridge tests could not catch this).
- Part-B2 test: a TX-rejecting radio → bridge runs RX-only, logs once, no `_tx_loop`.
- Hardware-gated X6200 serial TX smoke test (`tests/hardware`, `RIGPLANE_HW_SMOKE=1`), **skipped by default**: arms the PCM TX path and pushes a few frames to the USB audio OUTPUT only — never engages PTT, never transmits RF.

Fakes reused from `tests/test_icom7610_serial_radio.py` (no one-off mocks).

## Gate results

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **6352 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failed**
- `tests/test_icom7610_serial_radio.py` → **22 passed, file unchanged (zero diff)**
- `uv run ruff check src/ tests/` + `ruff format` → clean
- `uv run mypy src/` → no new errors in touched files (6 pre-existing `yaesu_cat/radio.py` errors are baseline, identical on a clean tree)
- X6200 hardware smoke → skipped by default (verified)

## Compatibility

- IC-7610 LAN path untouched (`_audio_runtime_mixin.py` not modified).
- IC-7610 serial behavior-identical, pinned by the unchanged existing tests.
- On-air TX validation for IC-705/IC-7300/IC-9700 remains pending real-hardware confirmation; the X6200 path is the one with an attached test rig.

Independent review pending.

### DIFF
--- a/src/rigplane/audio/bridge.py
+++ b/src/rigplane/audio/bridge.py
@@ -478,11 +478,25 @@ class AudioBridge:
 
         # --- TX path: virtual device input → radio ---
         if self._tx_enabled:
-            await self._radio.start_audio_tx_pcm(
-                sample_rate=self._sample_rate,
-                channels=self._channels,
-                frame_ms=self._frame_ms,
-            )
+            try:
+                await self._radio.start_audio_tx_pcm(
+                    sample_rate=self._sample_rate,
+                    channels=self._channels,
+                    frame_ms=self._frame_ms,
+                )
+            except (RuntimeError, NotImplementedError) as exc:
+                # The radio cannot arm a PCM TX stream (e.g. a serial backend
+                # that never set up its USB-audio TX path). Degrade to RX-only
+                # instead of spinning a _tx_loop that would reject — and log —
+                # every captured frame (MOR-242).
+                self._tx_enabled = False
+                self._tx_started = False
+                logger.warning(
+                    "%s: radio rejected TX start (%s); running RX-only",
+                    self._label,
+                    exc,
+                )
+                return
             self._tx_started = True
 
             tx_dev_id = dev_id
@@ -829,6 +843,17 @@ class AudioBridge:
 
                 try:
                     await self._radio.push_audio_tx_pcm(pcm_bytes)
+                except (RuntimeError, NotImplementedError) as exc:
+                    # The radio cannot accept PCM TX frames (TX path not armed).
+                    # Degrade to RX-only and stop the loop instead of rejecting
+                    # — and logging — every captured frame (MOR-242).
+                    self._tx_enabled = False
+                    logger.warning(
+                        "%s: radio rejected TX frame (%s); switching to RX-only",
+                        self._label,
+                        exc,
+                    )
+                    return
                 except Exception:
                     if self._tx_frames <= 5:
                         logger.warning("%s: TX push error", self._label, exc_info=True)
--- a/src/rigplane/backends/_icom_serial_base.py
+++ b/src/rigplane/backends/_icom_serial_base.py
@@ -454,7 +454,38 @@ class _IcomSerialRadioBase(CoreRadio):
         await self._serial_audio_driver.stop_rx()
 
     # ------------------------------------------------------------------
-    # Audio TX
+    # Audio TX (Opus)
+    # ------------------------------------------------------------------
+
+    async def start_audio_tx_opus(self) -> None:
+        self._check_connected()
+        await self._serial_audio_driver.start_tx(
+            sample_rate=self.audio_sample_rate,
+            channels=self._serial_audio_channels_for_codec(),
+            frame_ms=20,
+        )
+
+    async def push_audio_tx_opus(self, opus_data: bytes) -> None:
+        self._check_connected()
+        if not self._serial_audio_driver.tx_running:
+            raise RuntimeError("Audio TX not started")
+        payload = bytes(opus_data)
+        if self._serial_codec_is_opus():
+            transcoder = self._get_pcm_transcoder(
+                sample_rate=self.audio_sample_rate,
+                channels=self._serial_audio_channels_for_codec(),
+                frame_ms=20,
+            )
+            payload = transcoder.opus_to_pcm(payload)
+        await self._serial_audio_driver._push_tx_pcm(payload)
+
+    async def stop_audio_tx_opus(self) -> None:
+        await self._serial_audio_driver.stop_tx()
+        self._pcm_tx_fmt = None
+
+    # ------------------------------------------------------------------
+    # Audio TX (PCM) — arms ``_pcm_tx_fmt`` so ``push_audio_tx_pcm`` works
+    # over the serial USB CODEC path (MOR-242).
     # ------------------------------------------------------------------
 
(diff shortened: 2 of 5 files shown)


morozsm:
Agent Review: PASS

Independent adversarial review of PR #1696 (MOR-242 — serial/USB Icom TX). I did not write this code.

## Scope & diff completeness (special-attention check)
The reported `git stash pop` conflict did NOT leave a half-reverted diff. Verified independently:
- `_icom_serial_base.py` CONTAINS the full fix: `push_audio_tx_pcm` override (frame-size validation against the armed `_pcm_tx_fmt`, bytes-like guard, `_serial_audio_driver._push_tx_pcm` sink), the `_pcm_tx_fmt = (sample_rate, tx_channels, frame_ms)` assignment in `start_audio_tx_pcm`, the mono `tx_channels = 1` clamp, the new Opus TX trio, and teardown clearing (`_pcm_tx_fmt = None` in both `stop_audio_tx_opus` and `_stop_serial_audio_driver`).
- `icom7610/serial.py` had its duplicate overrides REMOVED (135 lines gone; the class now only keeps RX-PCM delegation + the CommandError scope guardrail and inherits everything else from the base).
- Lifted base impl is behavior-equivalent to the deleted 7610 impl (compared line-by-line vs git history; only comment wording changed: "IC-7610 ALC" → "radio ALC" / MOR-238 reference). Diff stat: 3 source files + 2 test files, matching the agreed design. `_audio_runtime_mixin.py` and the LAN `IcomRadio` are untouched.

## Part A correctness
- Pre-fix root cause confirmed: original base opened TX with `channels=channels` and never set `_pcm_tx_fmt`, with no `push_audio_tx_pcm` override → mixin guard raised "PCM TX not started" per frame. Exact symptom.
- Guard raises the SAME `RuntimeError` when not started (`_pcm_tx_fmt` is initialized to None in the CoreRadio/runtime `__init__`, so it is RuntimeError, not AttributeError).
- All 4 subclasses inherit the fixed base with NO TX overrides: Ic705SerialRadio (incl. model="X6200", which factory.py routes through Ic705SerialRadio), Ic7300SerialRadio, Ic9700SerialRadio. None re-break it.

## Part B2 non-regression
- RX path (rx_stream, rx_task, subscription) is fully set up BEFORE the TX block, so the early `return` on TX-start failure correctly leaves the bridge RX-only.
- Downgrade triggers ONLY on `(RuntimeError, NotImplementedError)`; the broad `except Exception` is retained after the specific handler for genuine transient errors. LAN / IC-7610 / yaesu-cat whose `start_audio_tx_pcm` succeeds run the normal `_tx_loop` unchanged — no regression.

## Tests
- `tests/test_icom7610_serial_radio.py` is byte-identical to origin/main (empty diff) and green (22 passed) — pins IC-7610-serial behavior after the lift.
- New `tests/test_serial_backend_tx_mor242.py` (7 tests) uses the repo's REAL shared fakes (`_FakeUsbAudioDriver` / `_FakeSerialCivLink`), not one-off MagicMock. Confirmed genuinely non-tautological: reverting source to origin/main makes 6 of them FAIL (4 parametrized at `_pcm_tx_fmt is not None`, bridge reproducer at frame-not-delivered, B2 at no-downgrade); the only test that passes pre-fix is the push-before-start guard, which is expected. Bridge reproducer asserts no RuntimeError + frame delivered; B2 asserts RX-only downgrade + single WARNING + `_tx_task is None` + subscriber present.
- Hardware smoke `tests/hardware/test_x6200_serial_tx_smoke.py` is `pytest.mark.hardware` + env-gated (`RIGPLANE_HW_SMOKE`), skipped in a normal run, and never keys PTT/RF (arms the USB CODEC OUTPUT stream only).

## Gates
- `pytest tests/ --ignore=tests/integration`: 6352 passed, 8 skipped, 0 failures.
- `ruff check src/ tests/`: All checks passed.
- `lint-imports`: 4 contracts kept, 0 broken.
- `mypy src/` (project invocation): 6 errors in 2 files — all pre-existing yaesu_cat/_civ_rx baseline; ZERO in the 3 touched files. (Per-file mypy shows resolution-artifact noise; the project-level run is authoritative.)

No blocking issues. Diff is complete, correct, behavior-preserving for IC-7610, fixes the X6200/705/7300/9700 TX path, and degrades the bridge cleanly when TX cannot arm.

PR #1699, opened 2026-06-01
fix(MOR-250): callback blocksize=0 TX capture + 20 ms framing — restore Windows TX (open-core half)

morozsm:
## Summary

Windows TX delivery fix (open-core half). This PR makes the TX capture stream
produce frames the core PCM TX validator actually accepts, and kills the
~50 Hz read-seam comb that the previous fixed-blocksize capture introduced.

- Linear epic: **MOR-256** — audio-routing-redesign.
- Linear: **MOR-250** — Windows TX splatter / Po=0 W on TX.

This is the **open-core** half of a two-PR delivery. It ships **together** with
the proprietary companion PR (device selection + COM init); see cross-link
below.

## Root cause

On Windows, TX produced **Po = 0 W** (no carrier modulation) even though the
audio path looked connected. Two coupled defects:

1. **Read-seam comb.** Fixed-`blocksize` capture read the input stream in
   discrete chunks, and the boundary between successive reads injected a
   periodic discontinuity — a ~50 Hz comb across the TX audio. Switching to a
   callback-driven `blocksize=0` stream removes the application-side read seam
   entirely.
2. **TX frame-size rejection (the Po=0 W cause).** The callback +
   `blocksize=0` stream emitted **960-byte (10 ms)** frames, but the core PCM
   TX validator requires **1920-byte (20 ms)** frames (48 kHz, mono, S16LE).
   Result: **100 % of TX frames were rejected** at the validator, so nothing
   was ever transmitted. The capture path now re-chunks the callback stream
   into fixed `frame_ms` (1920-byte / 20 ms) frames before handoff, so the
   frames pass the validator unchanged.

## What changed

- `38663671` — **callback-driven `blocksize=0` TX capture.** Replaces the
  fixed-blocksize read loop with a callback stream so there is no
  application-side read seam, killing the ~50 Hz comb.
- `5c5f776` — **re-chunk the callback stream into fixed `frame_ms` frames.**
  Buffers callback output and emits exactly 1920-byte / 20 ms frames, fixing
  the TX frame-size rejection (960-vs-1920) that caused Po=0 W.
- `c88de19` — **regression test: capture ↔ core PCM TX validator seam.** Drives
  the capture output through the actual core validator and asserts acceptance;
  the historical 960-byte regression frame is explicitly rejected by this test.
- `3e7306a4` — **audio-path contract tests (codec/channels/rate SSOT).** Locks
  the TX format single-source-of-truth so codec, sample rate, and channel count
  cannot silently drift apart between capture and validator.

## Tests / verification

- New + seam suite: **23 passed** (`tests/test_audio_backend.py`,
  `tests/test_audio_capture_tx_validator_seam.py`,
  `tests/test_audio_tx_contract_consistency.py`).
- The new tests were **RED-verified without the fix** (they fail on the
  pre-fix capture behaviour), then green after.
- The seam test **rejects the 960-byte regression frame**, proving the
  validator boundary is enforced.

## On-air validation

After this fix, the operator completed **real FT8 QSOs** through the RigPlane
virtual audio driver on Windows. TX carrier modulation is restored end-to-end.

## Boundary

**mixed/split.** This PR is the **open-core** portion: the TX capture stream
and frame re-chunking (generic audio transport, no commercial/device logic).
The companion device-face selection and COM initialization live in the
proprietary `rigplane-pro` PR. **The two PRs must ship together** — locally the
companion consumes this core change via the editable path source; for
release/non-local consumption see the out-of-scope note in the companion PR.

## Out of scope / follow-up

- Release/version coordination (core version bump + companion re-pin) is tracked
  in the companion PR's out-of-scope section.
- **MOR-257** — comb bench harness — remains open.
- **MOR-258** — audio path tracer — remains open.

## Cross-link

Companion (proprietary) PR: https://github.com/rigplane/rigplane-pro/pull/1119

### DIFF
--- a/src/rigplane/audio/backend.py
+++ b/src/rigplane/audio/backend.py
@@ -242,13 +242,19 @@ class _PortAudioRxStream:
     ``blocksize=0`` opens cleanly.
 
     The PortAudio callback runs on the audio thread and must not block. It
-    copies the captured PCM (s16le) and hands it straight to the consumer
-    *callback*. Each callback delivers a device-natural, variable-size block;
-    it is forwarded verbatim so no new fixed-size re-framing seam (which would
-    re-introduce a discontinuity) is added. The consumer contract (see
-    :class:`RxStream`) is that the callback is cheap and thread-safe to invoke
-    from the audio thread; the companion bridge marshals onto its event loop
-    via ``call_soon_threadsafe``.
+    copies the captured PCM (s16le) out of the engine-native, variable-size
+    block and *re-chunks* it into fixed ``frame_ms``-sized frames before
+    handing them to the consumer *callback*. The downstream PCM-TX contract
+    requires fixed-size frames (e.g. 20 ms = 960 samples = 1920 bytes s16le
+    mono @ 48 kHz); a managed core validator rejects any frame whose byte
+    length differs. Re-chunking a *continuous* callback stream is lossless and
+    introduces no scheduling seam (the ~50 Hz comb came from the old
+    blocking-read inter-read gap, not from re-chunking), so this preserves the
+    fixed-frame contract without re-introducing the comb. Sub-frame remainders
+    are buffered across callbacks; a trailing partial frame at stop is dropped.
+    The consumer contract (see :class:`RxStream`) is that the callback is cheap
+    and thread-safe to invoke from the audio thread; the companion bridge
+    marshals onto its event loop via ``call_soon_threadsafe``.
     """
 
     def __init__(
@@ -258,6 +264,7 @@ class _PortAudioRxStream:
         sample_rate: int,
         channels: int,
         blocksize: int,
+        frame_ms: int,
     ) -> None:
         self._sd = sd
         self._device_index = device_index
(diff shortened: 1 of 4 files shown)


morozsm:
Agent Review: PASS for head a44fd72d

Independent review (not the implementation agent). Verified:
- Re-chunk in `_PortAudioRxStream._input_callback` is lossless: every byte enters the `bytearray` accumulator in order via `extend`, whole frames are sliced from the front (`acc[:frame_bytes]` / `del acc[:frame_bytes]`), remainder buffered; no dup/reorder. `_frame_bytes = frame_samples*channels*2` matches `PcmAudioFormat.frame_bytes` exactly (interleaved s16le). Accumulator is audio-thread-local; `stop()` rebinds it and nulls the callback before `stream.stop()` — GIL-atomic, no torn state (verified by the in-suite late-callback no-op assertion). Bridge `_on_tx_capture` marshals via `call_soon_threadsafe` (cheap/thread-safe, immutable `bytes`), so the audio-thread consumer contract holds.
- `blocksize=0` + `callback=` intact; `frame_ms` honored (sizes the emitted frame), no leftover `del frame_ms`; `asyncio` import correctly removed (only the deleted read loop used it).
- Tests hit REAL code, not re-implementations: seam test drives the real `_PortAudioRxStream` (only sounddevice faked) into the real `AudioRuntimeMixin._push_audio_tx_pcm_internal` (only `_check_connected`/`_audio_stream` stubbed). Independently reproduced: validator rejects the 960-byte frame and accepts 1920. `test_verbatim_forwarding_would_be_rejected` proves the guard is non-tautological.
- `a44fd72d` is a benign prose-only Icom-LAN->RigPlane rename in 2 test comment lines (no logic change); grep-gate clean.
- Scope: exactly 1 production file (backend.py) + 3 test files. Ran tests/test_audio_backend.py + the 2 new suites: 63 passed locally on a44fd72d; ruff clean. CI quick/grep-gate green.

PR #1703, opened 2026-06-01
fix(#1702): reset leaked external-CAT session on radio (re)connect

### LINKED ISSUE #1702: control-plane: radio_state goes stale after LAN session restart (rigctld/CAT serve frozen freq; poller paused via external_cat_session)
## Symptom (observed live, IC-7610 over LAN, managed station runtime)
After a managed LAN session restart, the radio's live `radio_state` stopped updating: rigctld (`:4532`) served a **frozen frequency** (`rigctl f` → 14074000) while the actual radio / web GUI were at 14.250 MHz. Control WRITES did not reach the radio (`F 14075000` → `RPRT 0` but no change; desktop banner 'Radio control unavailable — control traffic is not reaching the radio'). WSJT-X CAT did nothing. RX audio, spectrum, and the web state display continued to work.

## Trigger (from companion logs)
At 17:00:52 the core station runtime exited and was restarted:
```
station_runtime.start_retrying ... core station runtime exited before readiness: exit_code_1
rigplane.core.exceptions.TimeoutError: login response timed out   (_control_phase.py _wait_for_packet)
→ lan_session_cooldown → lan_session_data_port_pending → reconnect 17:01:29
```
After reconnect, audio + web-state resynced, but rigctld/CAT state did not.

## Mechanism (code)
- `rigctld/handler.py:390` get_freq reads the shared `self._radio.radio_state.main.freq`, falling back to `_FallbackRigState` / `_pending` when it's invalid. So a stale/invalid live state ⇒ rigctld serves a frozen value.
- `rigctld/poller.py:201-205` and `web/radio_poller.py:638` both **pause polling** while `radio.external_cat_session_active` is True (to avoid colliding on the CI-V wire).
- `external_cat_session_active` (`runtime/radio.py:1326`) = `self._external_cat_session`, set by:
  - `begin_external_cat_session()` / `end_external_cat_session()` — called **only** by `hamlib_bridge.py:226/232` (A1 external CAT, e.g. WSJT-X over Hamlib). Long-lived; **no `finally`** around begin/end at the bridge layer.
  - `_claim_external_cat_session()` per `send_civ_transaction` — released in `finally` (safe).

## Hypotheses to confirm
1. **Leaked external-CAT ownership:** a `begin_external_cat_session()` (hamlib bridge) not matched by `end_external_cat_session()` across the LAN session restart / radio-object swap ⇒ poller paused permanently ⇒ `radio_state` frozen.
2. **No resync on reconnect:** after the managed LAN session restarts, nothing calls `reconcile_state()` / re-fetches initial state, and with the poller paused the live state never refreshes.
3. **Optimistic web state divergence:** the GUI showed 14.250 while the radio/rigctld were elsewhere — set-commands may be acked optimistically without confirming they reached the radio.

## Proposed fixes
- On LAN session restart / radio reconnect: force `reconcile_state()` (re-fetch freq/mode/…), reset `_external_cat_session`/owner, invalidate rigctld `_FallbackRigState`/`_pending`.
- Guard the hamlib-bridge `begin/end_external_cat_session` with try/finally (or tie ownership to the radio-object lifecycle) so it cannot leak.
- Don't serve a stale fallback indefinitely: if `radio_state` is invalid for > N s, surface a rigctld error / 'control unavailable' instead of a frozen value.
- Confirm-or-fail control writes rather than optimistic ack when control traffic isn't reaching the radio.

## Boundary
Public core (`rigplane`) control-plane. Found while validating the Linux companion (rigplane-pro).

morozsm:
Closes #1702

## Root cause (file:line)

After a managed LAN session restart, rigctld served a frozen frequency (`14074000`) and control writes did not refresh live state, while RX audio / spectrum / web-state kept working. Traced to a **leaked external-CAT session** that permanently pauses both pollers:

1. `runtime/radio.py:1330` `begin_external_cat_session()` sets `_external_cat_session = True`. Its only releaser at the bridge layer is `hamlib_bridge.py` `_end_session_and_reconcile()`, called **solely** from `stop()` / `__aexit__`. If `stop()` never runs — managed station runtime crash+restart (the `exit_code_1` / login-timeout in the issue), a dropped bridge object, or `open_transport()` raising **after** `_begin_session()` (was at `hamlib_bridge.py:137`, before the listener/server setup) — the flag leaks.
2. `runtime/radio.py:1054` `connect()` re-fetched initial state on (re)connect but **never reset** `_external_cat_session`, so a leaked flag survived the reconnect.
3. With the flag stuck `True`, `rigctld/poller.py:205` and `web/radio_poller.py:638` pause polling forever, so `radio_state.main.freq` is frozen at the last `_fetch_initial_state` value. `rigctld/handler.py:512-514` then serves that frozen `main.freq` indefinitely (any `freq > 0` wins over the fallback). The `14074000` value is the FT8 default the radio was last on — a stale **live** `main.freq`, not the `_FallbackRigState` default (which is `0`).

So the frozen value is "stale live state, pollers paused", not a fallback artifact.

## Fix (minimal, in-core, behaviour-preserving)

- `runtime/radio.py` — `connect()` calls a new `_reset_external_cat_session()` **before** the control-phase handshake. A fresh connection invalidates any prior external session (its transport is gone), so a leaked flag cannot survive a reconnect and the cooperating pollers resume. This is the load-bearing fix: it recovers regardless of who/how the flag leaked, and the `reconnect_loop` path already routes through `radio.connect()`.
- `hamlib_bridge.py` — `open_transport()` now releases ownership (`_end_session()`) if listener/server setup fails after `_begin_session()`, so a partial start cannot leak.

The legitimate poller-pause during an **active** external CAT session is untouched (the reset only fires on a new connect, when no live session can exist).

## Tests (TDD — added red, then fixed)

- `tests/test_reconnect.py::TestExternalCatSessionResetOnConnect` — `connect()` clears a leaked flag; no-op when already clean.
- `tests/test_hamlib_bridge.py::test_open_transport_failure_does_not_leak_session` — a failure mid-`open_transport` does not leak the session.

## Verification

- `uv run pytest tests/ -q` → 6569 passed, 112 skipped, 11 xfailed, 1 xpassed, 0 failures.
- `uv run ruff check src/ tests/` → clean.
- `uv run lint-imports` → 4 contracts kept, 0 broken.

## Confidence / hardware caveat

High confidence on the leak + no-reset-on-reconnect mechanism and that the fix restores polling after reconnect. The exact in-pro trigger that orphaned the session (managed-runtime crash vs. radio-object swap) lives in `rigplane-pro` and could not be reproduced here; the core fix is defensive and covers all leak sources at the connection boundary. The issue's secondary suggestions (rigctld surfacing "control unavailable" after N s of invalid state; confirm-or-fail control writes) are **not** included — they are larger behavioural changes beyond the minimal root-cause fix and can be tracked separately if desired.

### DIFF
--- a/src/rigplane/hamlib_bridge.py
+++ b/src/rigplane/hamlib_bridge.py
@@ -135,10 +135,23 @@ class HamlibBridge:
         if self._server is not None:
             return self.back_port
         self._begin_session()  # take ownership: pause RigPlane's own pollers
-        self._subscription = self._radio.add_raw_civ_listener(self._on_radio_frame)
-        self._server = await asyncio.start_server(self._handle_back, self.host, 0)
-        self._back_port = self._server.sockets[0].getsockname()[1]
-        await self._server.start_serving()
+        try:
+            self._subscription = self._radio.add_raw_civ_listener(self._on_radio_frame)
+            self._server = await asyncio.start_server(self._handle_back, self.host, 0)
+            self._back_port = self._server.sockets[0].getsockname()[1]
+            await self._server.start_serving()
+        except BaseException:
+            # Never leak external-CAT ownership if setup fails after we claimed
+            # it — a leaked session pauses RigPlane's pollers forever (#1702).
+            if self._subscription is not None:
+                self._subscription.close()
+                self._subscription = None
+            if self._server is not None:
+                self._server.close()
+                self._server = None
+            self._back_port = None
+            self._end_session()
+            raise
         logger.info(
             "hamlib-bridge: back-side CI-V listener up on %s:%d",
             self.host,
@@ -227,11 +240,15 @@ class HamlibBridge:
         if callable(begin):
             begin()
 
-    async def _end_session_and_reconcile(self) -> None:
-        """Release ownership and refresh RigPlane state (both best-effort)."""
+    def _end_session(self) -> None:
+        """Release external-CAT-session ownership if the radio supports it."""
         end = getattr(self._radio, "end_external_cat_session", None)
         if callable(end):
             end()
+
+    async def _end_session_and_reconcile(self) -> None:
+        """Release ownership and refresh RigPlane state (both best-effort)."""
+        self._end_session()
         reconcile = getattr(self._radio, "reconcile_state", None)
         if callable(reconcile):
             try:
--- a/src/rigplane/runtime/radio.py
+++ b/src/rigplane/runtime/radio.py
@@ -1056,7 +1056,15 @@ class CoreRadio(ScopeRuntimeMixin, AudioRuntimeMixin, DualRxRuntimeMixin):
 
         Delegates to the composed ControlPhaseRuntime, then fetches
         initial radio state so RadioState is populated before consumers.
+
+        A fresh (re)connection means any prior external-CAT session is over —
+        its transport is gone — so clear leaked ownership first (#1702). Without
+        this, a ``begin_external_cat_session()`` that was never matched by an
+        ``end_external_cat_session()`` (managed-runtime crash/restart, a dropped
+        Hamlib bridge) would keep cooperating pollers paused forever, freezing
+        ``radio_state`` while rigctld/web serve a stale frequency.
         """
+        self._reset_external_cat_session()
         await self._control_phase.connect()
         await self._fetch_initial_state()
 
@@ -1347,6 +1355,17 @@ class CoreRadio(ScopeRuntimeMixin, AudioRuntimeMixin, DualRxRuntimeMixin):
         """Release legacy external-CAT-session ownership. Idempotent."""
         if self._external_cat_session_owner not in (None, "external"):
             return
+        self._external_cat_session = False
+        self._external_cat_session_owner = None
+
+    def _reset_external_cat_session(self) -> None:
+        """Unconditionally clear external-CAT ownership (#1702).
+
+        Used on (re)connect: a new connection invalidates any prior session
+        regardless of owner, so this drops a leaked ``begin_external_cat_session``
+        that was never released. Unlike :meth:`end_external_cat_session` it does
+        not respect the current owner — the old session can no longer exist.
+        """
         self._external_cat_session = False
         self._external_cat_session_owner = None
 
--- a/tests/test_hamlib_bridge.py
+++ b/tests/test_hamlib_bridge.py
@@ -281,3 +281,18 @@ async def test_bridge_runs_against_icom_serial_backend() -> None:
     writer.close()
     await bridge.stop()  # reconcile is best-effort (radio not connected) — tolerated
     assert radio.external_cat_session_active is False
+
+
+async def test_open_transport_failure_does_not_leak_session(
+    radio: FakeRawPipeRadio,
+) -> None:
+    """Regression for #1702: if open_transport() raises *after* taking session
+    ownership, the external-CAT flag must not leak — otherwise cooperating
+    pollers pause forever and radio_state freezes."""
+    bridge = HamlibBridge(radio, model="3078")
+    boom = RuntimeError("listener setup failed")
+    with patch.object(radio, "add_raw_civ_listener", side_effect=boom):
+        with pytest.raises(RuntimeError, match="listener setup failed"):
+            await bridge.open_transport()
+    # Ownership was taken (_begin_session) then released on failure — no leak.
+    assert radio.session_active is False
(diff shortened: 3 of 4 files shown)


morozsm:
Agent Review: PASS

Independent adversarial review (separate worktree at `d3a160e2`, Python 3.14). I did not write this PR and re-verified everything myself.

## Gate (ran myself, actual results)
- `uv run pytest tests/ -q` → **6619 passed, 62 skipped, 11 xfailed, 1 xpassed, 0 failed** (full suite incl. integration/validation, 178s). Slightly different counts than the PR body (6569/112) — Python-version/env difference, no failures either way.
- `uv run ruff check src/ tests/` → **All checks passed**.
- `uv run lint-imports` → **4 contracts kept, 0 broken** (252 files, 622 deps). Layer boundaries respected.

## Concerns cleared
- **Red→green (real regression guards):** swapped the two source files back to `main` (keeping the new tests) and re-ran — `test_connect_resets_leaked_external_cat_session` and `test_open_transport_failure_does_not_leak_session` both **FAIL** pre-fix (`external_cat_session_active`/`session_active` stuck `True`), pass post-fix. The clean-connect no-op test passes in both. Tests use a real `IcomRadio` fixture (not a mock) so the owner field is genuinely exercised. Pollers (`rigctld/poller.py:205`, `web/radio_poller.py:638`) re-check the flag every cycle with no latching, so clearing it un-pauses them — i.e. the fix actually resumes polling and un-freezes state.
- **No mid-session regression:** `_reset_external_cat_session()` is called *only* from `CoreRadio.connect()`. During an active Hamlib session the bridge serves on an existing connection and nobody calls `connect()`, so the reset cannot pollute a live external master's CI-V stream. `reconnect_loop` (`radio_reconnect.py:145-146`) recreates the transport and calls `radio.connect()` — at that point the old session's transport is genuinely gone, so unconditional clear is correct.
- **Covers the #1702 trigger:** the documented trigger (managed runtime crash+restart → login timeout → `lan_session_cooldown` → reconnect) routes through `reconnect_loop` → `CoreRadio.connect()`, which now resets. Both CI-V-watchdog full-reconnect paths (`_civ_rx.py:588,611`) also go through `_host.connect()`. Covered.
- **Bridge release-on-failure:** `open_transport()` now wraps listener/server setup in try/except-`BaseException`; on failure it closes the subscription + server (idempotent `is not None` guards), nulls `_back_port`, and `_end_session()` before re-raising. `_end_session`→`end_external_cat_session` is owner-scoped (only releases `None`/`"external"`) and idempotent, so a later `stop()` can't double-release or stomp another owner. Correct.
- **Owner-scoping / scoped claims:** `_reset_external_cat_session` clears unconditionally (by design, on connect), but scoped `send_civ_transaction` claims run on a live connection inside a `finally` release; a connect can only overlap a transaction whose transport is already torn down, where the idempotent `finally` release is harmless. No corruption of `_claim/_release_external_cat_session` semantics.

## Non-blocking nit
- One narrow leak-recovery path is not covered: `_control_phase.soft_reconnect()` falling back to `self.connect()` (control transport gone, `_control_phase.py:587`) rebuilds the control session via `ControlPhaseRuntime.connect`, which does **not** call `_reset_external_cat_session`. The documented #1702 trigger does not use this path (it goes through `reconnect_loop`/`CoreRadio.connect`), and the bridge release-on-failure closes the other main leak source — so not a blocker. For defense-in-depth, consider resetting in that soft-reconnect full-connect fallback too.

No blockers found. PASS.

morozsm:
Agent Review: PASS

Re-confirmation review of the new PR head `517407d2` (Closes #1702). This re-confirms the prior `Agent Review: PASS` (which was posted against `d3a160e2`); the gate needed a fresh PASS because a follow-up commit changed the head.

**Delta since the previously-reviewed commit** (`git diff d3a160e2..517407d2`): formatting-only. The sole change is `ruff format` whitespace reformatting of `tests/test_reconnect.py` (1 file, 2 insertions / 6 deletions) — two multi-line `patch.object(radio._control_phase, "connect", new_callable=AsyncMock)` calls collapsed onto single lines. No logic, source, or behavioral change.

**Gate results on `517407d2`:**
- `uv run ruff check src/ tests/` — PASS (All checks passed!)
- `uv run ruff format --check src/ tests/` — PASS (519 files already formatted; this was the previously-failing gate and is now green)
- `uv run lint-imports` — PASS (4 contracts kept, 0 broken)
- `uv run pytest tests/test_reconnect.py tests/test_hamlib_bridge.py -q` — PASS (28 passed)
- CI `quick` check — green (pass)

Safe to merge.

PR #1720, opened 2026-06-07
fix: derive discovery legacy url/urls from primary fleet member (MOR-380)

morozsm:
## Problem (MOR-380)

`DiscoveryResponder.build_response()` (`src/rigplane/web/discovery.py`) built the legacy
single-radio top-level `url` and all `urls.*` entries from the static `self._web_port`
(default 8080), even when a `fleet_provider` is configured. In a multi-radio station
front door each fleet member carries its own `web_port`, and the legacy `radio` block is
already derived from `fleet[0]` — but the legacy `url`/`urls.*` were not, so legacy
single-radio consumers were pointed at a dead port (e.g. :8080) instead of the
selected/primary member's live `:808x`.

## Fix

When a fleet is present, derive the legacy `url` / `urls.*` (and scheme) from `fleet[0]`
(`web_port`, and `tls` with fallback to the responder's own `tls` when the member's is
`None`) — consistent with how the legacy `radio` block already uses `fleet[0]`, and with
how the per-member `radios[]` URLs already use `member.web_port`/`member.tls`. The
no-fleet single-radio path is unchanged (still `self._web_port`/`self._tls`); empty fleet
`[]` is treated as no-fleet. The `radios[]` block and the discovery JSON schema are
unchanged — only the legacy block's port/scheme values change.

## Tests

- Updated `test_single_radio_fields_preserved_with_fleet` (its `:8080` assertion encoded
  the bug → now expects `fleet[0]`'s port).
- Added `test_legacy_url_uses_fleet_member_port_and_tls` (fleet `web_port=8443, tls=True`
  → legacy `url`+`urls.*` are `https://…:8443`, `radio` still from `fleet[0]`).
- Locked the no-fleet path (empty fleet → legacy `url` uses static `:8080`).

## Gate (full, in an isolated worktree)

- `pytest tests/ --ignore=tests/integration`: 7205 passed, 0 failures
- `mypy src/`: 20 = unchanged baseline, 0 new (0 in `discovery.py`)
- `lint-imports`: 4 contracts kept, 0 broken

Boundary: `core`/`web` only; no protocol/transport/upper-layer change.

Linear: MOR-380

### DIFF
--- a/src/rigplane/web/discovery.py
+++ b/src/rigplane/web/discovery.py
@@ -178,21 +178,32 @@ class DiscoveryResponder:
     def build_response(self, remote_addr: str) -> bytes:
         """Build the JSON discovery response payload."""
         local_ip = _get_local_ip_for(remote_addr)
-        scheme = "https" if self._tls else "http"
 
         radio_info = self._radio_provider() if self._radio_provider else None
         fleet = self._fleet_provider() if self._fleet_provider else None
 
-        # Back-compat: keep the single-radio top-level block populated from the
-        # first/selected fleet member when no dedicated radio_provider is set,
-        # so legacy parsers still resolve a valid station_server.
-        if radio_info is None and fleet:
+        # Derive the effective web_port/tls for the legacy single-radio url/urls
+        # block. When a fleet is present, the primary member (fleet[0]) owns the
+        # legacy block, so its port/tls must drive these URLs — otherwise the
+        # top-level url stays stuck on the responder's static web_port even
+        # though the radio it describes lives elsewhere.
+        web_port = self._web_port
+        tls = self._tls
+        if fleet:
             first = fleet[0]
-            radio_info = RadioInfo(
-                model=first.model,
-                connected=first.connected,
-                radio_ready=first.connected,
-            )
+            web_port = first.web_port
+            tls = first.tls if first.tls is not None else self._tls
+            # Back-compat: keep the single-radio top-level block populated from
+            # the primary fleet member when no dedicated radio_provider is set,
+            # so legacy parsers still resolve a valid station_server.
+            if radio_info is None:
+                radio_info = RadioInfo(
+                    model=first.model,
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

**Head:** d971fdf9 (`codex/mor-380-discovery-fleet-url` → `main`, MOR-380)

Independent review (reviewer ≠ implementer) of the discovery legacy-URL fleet-port fix.

**Scope:** exactly 2 files — `src/rigplane/web/discovery.py` (+25/−17) and `tests/test_discovery_responder.py` (+35/−1). No stray debug/logging, no commented-out code, no scope creep, no JSON schema/key change; `radios[]` per-member URL construction untouched.

**Correctness (all acceptance bullets MET, verified by file:line):**
- fleet present → `web_port = fleet[0].web_port`, `tls = fleet[0].tls if not None else self._tls`, `scheme` derived AFTER the effective tls; `fleet[0]` bound once (`discovery.py:190-195`).
- legacy `url` + all 5 `urls.*` (base/health/readiness/runtime/station) use the local `web_port`/`scheme` (`discovery.py:220-227`).
- no-fleet (None **or** empty `[]`) → byte-identical old behavior (`self._web_port`/`self._tls`); `[]` is falsy so it stays static.
- tls fallback uses `is not None` (member `tls=None` → falls back to `self._tls`, not hardcoded `False`).

**Falsification:** re-introducing the bug (legacy block back to `self._web_port`) makes both guard tests fail (`':8081' not in :8080`, `:8080` not `https://…:8443`); restored via `git checkout --` (no stash), worktree clean at d971fdf9.

**Gate (independently re-run, verbatim):**
- `pytest tests/ --ignore=tests/integration`: 7205 passed, 0 failures
- `mypy src/`: 20 = unchanged baseline, 0 new (0 in discovery.py)
- `lint-imports`: 4 contracts kept, 0 broken

**Blocking issues:** none.

PR #1729, opened 2026-06-07
fix: restart RX through the real consumer after TX, not a no-op (MOR-506)

morozsm:
## Summary
Fixes a latent **Icom-only** bug: on IC-7610 (and other CI-V backends routed through the web `RadioPoller`), browser RX audio dies after the first transmit.

### Root cause
The radio can't full-duplex, so `RadioPoller` restarts RX on the TX→RX transition (`case PttOff()`). The old code restarted it with a throwaway `async def _noop_rx(_pkt): pass` passed to `radio.start_audio_rx_opus(...)`, which stores into a **single-slot** `_opus_rx_user_callback` — clobbering the real consumer (`AudioBus._on_opus_packet`) registered by `AudioBus._start_rx`. After that, frames route into the no-op → silence (and on no-subscriber cycles it even spun a phantom RX stream while `_rx_active` stayed `False`).

### Fix (3 files, +174/-14)
- `audio/bus.py`: new `AudioBus.restart_rx()` — reinstalls the bus's own `_on_opus_packet` via `start_audio_rx_opus`, under the same `_lock` as the rest of the RX lifecycle; **no-op when there are no subscribers** (nothing is listening, so re-issuing the protocol RX-start would only feed a phantom loop).
- `web/radio_poller.py`: `case PttOff()` now `await radio.audio_bus.restart_rx()` instead of installing `_noop_rx`. Keeps the web layer out of radio-private `_opus_rx_user_callback`.
- `tests/test_audio_rx_tx_transition.py`: end-to-end real-`AudioBus` clobber regression test + 3 stale mock assertions corrected (they had encoded the buggy no-op path).

### Tests
The regression test drives a real `AudioBus` + a fake radio with faithful single-slot RX semantics through `RadioPoller._execute(PttOn())` → `_execute(PttOff())` and asserts the bus subscriber still receives the post-TX frame. Verified it **FAILS on `origin/main`** (`QueueEmpty` + `_noop_rx was never awaited`) and **passes on the fix**.

### Gate
- `pytest --ignore=tests/integration` → 7257 passed, 0 failures
- `mypy src/` → 20 baseline errors, 0 new (none in changed files)
- `lint-imports` → 4 kept, 0 broken (web→runtime→audio clean)

### Independent review
Implementer ≠ reviewer. Independent reviewer ran old-code falsification (4 tests fail on `origin/main`, all 7 pass on fix), confirmed the no-subscribers edge is SAFE (strictly better than the old phantom-stream behavior), no double-start/lock-reentrancy/layering issues, Yaesu path inert → **Agent Review: PASS**.

### Scope / deferred
- Yaesu (FTX-1) is unaffected — it uses its own poller, never `RadioPoller`.
- Live IC-7610 confirmation (transmit, verify browser RX resumes after PTT-off) is deferred until hardware.
- Known cosmetic follow-up: on the IC-7610 LAN stream the re-arm logs a swallowed "failed to re-arm RX" (the LAN rx_loop isn't torn down during TX, so `start_rx` raises `RuntimeError(RECEIVING)` — harmless; the callback is reinstated before the raise, so audio resumes). Could be demoted to debug-level later.

### DIFF
--- a/src/rigplane/audio/bus.py
+++ b/src/rigplane/audio/bus.py
@@ -318,6 +318,27 @@ class AudioBus:
         except Exception:
             logger.exception("audio-bus: failed to start RX")
 
+    async def restart_rx(self) -> None:
+        """Re-establish RX on the radio using the bus's own callback.
+
+        Used after a half-duplex TX cycle (e.g. the web poller's PTT-off
+        transition on Icom CI-V backends): the radio's single-slot RX
+        callback must be restored to :meth:`_on_opus_packet` so subscribers
+        keep receiving frames. No-op when the bus has no active subscribers.
+        """
+        async with self._lock:
+            if not self._subscribers:
+                return
+            try:
+                await self._radio.start_audio_rx_opus(
+                    self._on_opus_packet,
+                    jitter_depth=self._jitter_depth,
+                )
+                self._rx_active = True
+                logger.info("audio-bus: RX re-armed after TX")
+            except Exception:
+                logger.exception("audio-bus: failed to re-arm RX")
+
     async def _stop_rx(self) -> None:
         """Stop receiving audio from the radio."""
         if not self._rx_active:
--- a/src/rigplane/web/radio_poller.py
+++ b/src/rigplane/web/radio_poller.py
@@ -1192,11 +1192,12 @@ class RadioPoller:
                             await radio.stop_audio_tx_opus()
                         logger.info("poller: TX audio stream stopped")
 
-                        # Restart RX audio after TX (IC-7610 doesn't support full duplex)
-                        async def _noop_rx(_pkt: Any) -> None:
-                            pass
-
-                        await radio.start_audio_rx_opus(_noop_rx)
+                        # Restart RX audio after TX (IC-7610 doesn't support
+                        # full duplex). Re-arm through the AudioBus so the
+                        # real subscriber callback is reinstated rather than a
+                        # throwaway no-op clobbering the single-slot RX callback
+                        # (MOR-506).
+                        await radio.audio_bus.restart_rx()
                         logger.info("poller: RX audio stream restarted")
                     except Exception as e:
                         logger.debug("poller: audio stream transition failed: %s", e)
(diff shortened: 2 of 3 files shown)


morozsm:
**Agent Review: PASS**

**Old-code falsification (key evidence tests are real):** reverted ONLY `audio/bus.py` + `web/radio_poller.py` to `origin/main` (kept new tests), ran `tests/test_audio_rx_tx_transition.py` → **4 failed, 3 passed**:
- `test_ptt_off_does_not_clobber_audio_bus_subscriber` (the real end-to-end regression) → `QueueEmpty` at the post-TX frame assertion + `RuntimeWarning: coroutine '_noop_rx' was never awaited`
- `test_ptt_off_restarts_rx_audio`, `test_ptt_cycle_full_sequence`, `test_multiple_ptt_cycles` → old code never calls `audio_bus.restart_rx`
After restore → all 7 pass.

**NO-SUBSCRIBERS edge → SAFE.** Traced every RX-stream consumer reachable via `RadioPoller.PttOff`: browser RX (`handlers/audio.py:451`) and BlackHole bridge (`audio/bridge.py:475`) are both bus subscribers; the only non-bus `start_audio_rx_opus` callers (`cli/__init__.py`, `runtime/sync.py`) don't route through `RadioPoller` (web-only). So when `_subscribers` is empty nothing is listening, and the old `_noop_rx` path was strictly worse (started a phantom RX stream into a no-op while `_rx_active` stayed `False`). New skip is safer, not a regression.

**Other angles clear:** `_rx_active` set True only on success under `_lock` (no orphaned loop); no lock re-entrancy (PttOff isn't inside a bus lock); Yaesu inert (own poller); diff exactly 3 files, +174/-14, no scope creep.

**Full gate (verbatim):**
- `pytest --ignore=tests/integration` → 7257 passed, 9 skipped, 11 xfailed, 1 xpassed, 0 failed
- `ruff check` → All checks passed! · `ruff format --check` → 545 files already formatted
- `mypy src/` → 20 baseline, 0 new (none in bus.py)

Non-blocking: cosmetic "failed to re-arm RX" log on the IC-7610 LAN PTT cycle (callback reinstated before the swallowed RuntimeError → audio resumes). Live IC-7610 re-validation deferred.

PR #1730, opened 2026-06-07
fix: convert FTX-1 filter_width readback SH code → Hz via current mode (MOR-507)

morozsm:
## Summary
Fixes the FTX-1 v2 web filter-width slider snapping back to ~500 Hz after the operator drags it (the rig itself held the correct width).

### Root cause
`YaesuCatRadio.read_filter_width` resolved the mode-specific width table from the legacy `self._state` mirror — which the **observation adapter never updates**. During polling that mode was stale/default, so `_filter_width_table()` picked the wrong table (or `None`) and the method fell back to returning the **raw SH code** (e.g. `20`) as if it were Hz. `/api/v1/state main.filterWidth` then carried the raw code, and the frontend `hzToTableIndex()` mis-mapped it (~500) → the slider snapped back every poll. SET was already correct (mode-aware Hz→code).

### Fix (READ-only; SET and frontend untouched)
- `read_filter_width(receiver, mode=None)` resolves the table from the radio's **current** mode; `_filter_width_table` gained an optional `mode` param so SET and READ share one resolution path.
- `observations.py poll_medium` already reads `read_mode(0)` immediately before `read_filter_width` — it now threads that result as `read_filter_width(0, mode=main_mode)`, so **zero extra CAT round-trips on the hot poll path**.
- When `mode is None` (non-poll call sites, e.g. public `get_filter_width`), it reads the mode fresh via `read_mode(receiver)`. Degrades safely under `_safe_read` (skips the field rather than raising).
- `rule.fixed → defaults[0]` (AM/FM) and the raw-index compat fallback are preserved; SUB resolves its own fresh mode.

### Tests
New `test_read_filter_width_uses_current_mode_not_stale_state`: stale mirror `USB` (SH code 20 = 3200 Hz) while the rig's actual mode read via CAT is `CW` (code 20 = 4000 Hz) → asserts the readback returns 4000, not 3200 and not the raw code 20. Verified it **fails on `origin/main`** (behaviorally: old returns 3200) and passes here. The 3 pre-existing tests touched are mechanical updates for the now-explicit `read_mode` call / honest two-query (`MD0;`+`SH0;`) assertions — no assertion weakened (independently reviewed).

### Gate
- `pytest --ignore=tests/integration` → 7258 passed, 0 failures
- `mypy src/` → 20 baseline errors, 0 new

### Independent review
Implementer ≠ reviewer. Independent reviewer ran old-code falsification (new test fails on `origin/main`), verified each pre-existing test edit is legitimate (not coverage-weakening) with before/after quotes, confirmed no hot-path CAT traffic, safe `mode=None` fallback, correct per-mode table resolution, SUB handled → **Agent Review: PASS**.

### Deferred
Live FTX-1 re-validation (drag the web filter-width slider, confirm no snap-back, `/state main.filterWidth` in Hz) — radio is connected; orchestrator follow-up.

### DIFF
--- a/src/rigplane/backends/yaesu_cat/observations.py
+++ b/src/rigplane/backends/yaesu_cat/observations.py
@@ -192,7 +192,9 @@ class YaesuObservationRadio(Protocol):
 
     async def read_agc(self, receiver: int = 0) -> int: ...
 
-    async def read_filter_width(self, receiver: int = 0) -> int: ...
+    async def read_filter_width(
+        self, receiver: int = 0, mode: str | None = None
+    ) -> int: ...
 
     async def read_if_shift(self, receiver: int = 0) -> int: ...
 
@@ -286,9 +288,14 @@ class YaesuObservationAdapter:
                 observations.append(
                     adapter.observation(_MAIN_FREQ, value, native_id="read_freq")
                 )
+        # Capture the MAIN mode so filter_width (below) can resolve its
+        # mode-specific width table from the freshly-read mode rather than
+        # issuing a redundant CAT mode query on the hot poll path (MOR-507).
+        main_mode: str | None = None
         if self._can_poll(_MAIN_MODE):
             ok, result = await self._safe_read("main.mode", self.radio.read_mode(0))
             if ok and result is not None:
+                main_mode = result[0]
                 observations.append(
                     adapter.observation(_MAIN_MODE, result[0], native_id="read_mode")
                 )
@@ -318,7 +325,7 @@ class YaesuObservationAdapter:
             _MAIN_FILTER_WIDTH
         ):
             ok, value = await self._safe_read(
-                "main.filter_width", self.radio.read_filter_width(0)
+                "main.filter_width", self.radio.read_filter_width(0, mode=main_mode)
             )
             if ok:
                 observations.append(
--- a/src/rigplane/backends/yaesu_cat/radio.py
+++ b/src/rigplane/backends/yaesu_cat/radio.py
@@ -1257,38 +1257,56 @@ class YaesuCatRadio:
 
     # -- D4: Filters --------------------------------------------------------
 
-    def _filter_width_table(self, receiver: int = 0) -> tuple[int, ...] | None:
-        """Return the filter-width table for the current mode, or None."""
+    def _filter_width_table(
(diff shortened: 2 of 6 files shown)


morozsm:
**Agent Review: PASS**

**Old-code falsification:** reverted ONLY the two production files (`radio.py`, `observations.py`) to `origin/main`, ran `tests/test_yaesu_cat_observation_adapter.py` → **1 failed, 35 passed**. `test_read_filter_width_uses_current_mode_not_stale_state` failed on old code; behaviorally the no-kwarg path returns **3200** (stale USB/SSB table) where correct is **4000** (CW table, code 20) — genuine discrimination, verified against `rigs/ftx1.toml`. Restored after.

**Pre-existing test edits — all LEGITIMATE, none weakened (per-test before/after reviewed):**
- `test_ftx1_radio::test_get_filter_width` — `assert_called_once_with("SH0;")` → `assert_any_call("SH0;")`+`assert_any_call("MD0;")`; value assertion `==2100` unchanged. Honest update for the new two-query path.
- `test_dsp_filter_family` (2 tests) — only added `read_mode` stub; value assertions `==2400` untouched.
- `observation_adapter` renamed test — all 4 non-mutation assertions preserved + `read_mode` stub.
- poller/adapter stub classes — mandatory `mode: str|None=None` param additions; bodies/assertions unchanged.

**Other angles — all PASS:** zero extra CAT round-trips on the hot poll path (threads the already-read mode); `mode=None` fallback fires only off-poll and degrades safely under `_safe_read`; per-mode table resolution correct (USB/LSB→SSB, CW→CW, AM/FM `fixed`→defaults[0]); SET byte-for-byte unchanged; `read_mode` shape `tuple[str,int|None]` matches; SUB resolves its own fresh mode.

**Full gate (verbatim):**
- `pytest --ignore=tests/integration` → 7258 passed, 9 skipped, 11 xfailed, 1 xpassed, 0 failed
- target 4 test files → 372 passed
- `ruff check` → All checks passed! · `ruff format --check` → 545 files already formatted
- `mypy src/` → 20 errors (= origin/main baseline, 0 new)

Non-blocking: theoretical extra `read_mode` only if a future config polls filter_width but not mode (not the case for the current FTX-1 profile). Live FTX-1 re-validation pending (orchestrator).

PR #1732, opened 2026-06-07
fix: adaptive auto-range for FFT audio scope dB window (MOR-512)

morozsm:
## Summary
The v2 AUDIO SCOPE FFT rendered nearly empty for low-level radio audio (FTX-1): the fixed dB window `_DB_FLOOR=-70 / _DB_CEIL=-15` sits ~16–20 dB above the FTX-1's actual audio (measured in-band per-bin ≈ −86 dB over a −126 dB noise floor → 0/124 bins clear the −70 floor → invisible).

## Fix — per-stream adaptive auto-range
New `_update_db_window(avg_db)` (called each frame) replaces the fixed window:
- **Floor** = 10th percentile of per-bin dB → slow EMA (`α=0.05`), placed 5 dB below.
- **Ceil** = 99th percentile → attack/decay EMA (`0.30`/`0.05`), 3 dB above.
- Seeded from the first frame; **min-span guard** `_MIN_SPAN_DB=45` (so silence/noise can't be stretched to fill the screen); absolute clamps floor `[-160,-20]`, ceil `≤0`. EMA state reset in `stop()`.

Per-frame dB compute, Hann window, averaging, frame-rate limit, center-freq/bandwidth overlay, and the `_center_freq<=0` guard (MOR-505) are byte-unchanged. Frontend renderer plots `pixels/160` linearly → **backend-only**.

## Why it won't pump or amplify silence
Floor EMA is very slow (stable baseline); ceil decays slowly so a vanishing signal doesn't snap the window down and re-amplify noise; the 45 dB min-span keeps a flat noise field pinned to the bottom. Reviewer's adversarial harness: 60 frames of pure low-level noise → max 30%/median 14% (not a fake spectrum); steady weak tone → converges to 152–153/160 with 1px frame-to-frame stability; loud transient recovers via ceil decay (no permanent saturation).

## Tests
New `TestAudioFftScopeAdaptiveAutoRange` (4 tests): weak band (~−86 dB/bin) renders well above baseline; converges + stable; loud not permanently saturated; low-level noise stays low. Verified the weak-band tests **fail on `origin/main`** (0/160 px). Test diff is purely additive — no pre-existing test weakened.

## Gate
- `pytest --ignore=tests/integration` → 7262 passed, 0 failures
- `ruff check` / `ruff format --check` → clean
- `mypy src/` → 20 baseline errors, 0 new (none in fft_scope.py)
- `lint-imports` → 4 kept, 0 broken

## Independent review
[...]

PR #1733, opened 2026-06-07
fix: estimate FFT scope auto-range floor from the in-band audio region (MOR-512)

morozsm:
## Summary
Follow-up to the MOR-512 adaptive auto-range (#1732). On the live FTX-1 the scope showed a **high noise floor regardless of signal**: the adaptive floor/ceil percentiles ran over the FULL spectrum, but radio audio is bimodal — an audio baseband (~0–3.2 kHz, always carrying receiver band-noise) plus a quiet out-of-band region (>~4 kHz). The quiet out-of-band (majority of bins) dragged the floor to ~−128 while the ceil latched to the in-band band-noise (~−98), so the ever-present band-noise rendered at ~68% of screen height with no signal.

## Fix
`_update_db_window` now estimates floor/ceil from the **in-band audio region** via a new `_inband_db()`:
- In-band = `50 Hz` ≤ freq ≤ upper edge; upper edge = `_crop_max_hz/2` when the mode bandwidth is known, else fallback `_INBAND_FALLBACK_HI_HZ = 3500`. (FTX-1 reports no `max_hz` → fallback — the live regression case.)
- `_FLOOR_PCT` 10 → 30 (in-band low percentile = band-noise floor); `_CEIL_PCT` kept 99 (signal peak).
- Empty-mask guard returns the full array (degenerate fft_size/sr).
- All smoothing (slow-EMA floor / attack-decay ceil), `_MIN_SPAN_DB=45`, margins, clamps, first-frame seeding, `stop()` reset, the `_center_freq<=0` guard, and the bandwidth-crop overlay are **unchanged**.

Measured effect (live numbers): in-band band-noise 68% → **~15%** (bottom); a +25 dB tone renders at ~70% (55% separation).

## Tests
New `test_bimodal_no_signal_inband_noise_maps_low` (in-band noise <30%, was ~65% on the merged full-spectrum code) and `test_bimodal_with_signal_tone_rises_above_band_noise`. Calibrated to the live capture (in-band −102 dB / out-of-band −126 dB). Both **fail on the merged code**, pass here. The 4 pre-existing adaptive guards (anti-pump, silence-low, loud-not-saturated, converges-stable) are untouched and pass.

## Gate
- `pytest --ignore=tests/integration` → 7270 passed, 0 failures
- `mypy src/` → 20 baseline, 0 new (none in fft_scope.py)

## Independent review
Implementer ≠ reviewer. Reviewer ran old-code falsification, an independent bimodal harness (noise pins ~15%, +25 dB tone 70%, +8 dB tone 32%/visible; steady tone 0.000 dB pumping; loud not saturated), confirmed in-band edge math (`_crop_max_hz/2` matches the displayed one-sided passband), fallback robustness, IC-7610 SSB path solid, all guards/state unchanged → **Agent Review: PASS**.

## Follow-up (non-blocking, filed separately)
Reviewer noted IC-7610 **AM** (`max_hz=10000` → in-band edge 5000 Hz) can re-include the quiet tail if the actual AM audio is narrow → cap the in-band upper edge at ~3500 Hz even when a wider crop is known (`min(_crop_max_hz/2, _INBAND_FALLBACK_HI_HZ)`). Narrow AM-only edge; SSB/CW/data/RTTY and the FTX-1 fallback unaffected. Tracked as a hardening follow-up.

## Deferred
Live FTX-1 re-validation (noise floor low, signal above it) — radio connected.

### DIFF
--- a/src/rigplane/audio/fft_scope.py
+++ b/src/rigplane/audio/fft_scope.py
@@ -40,11 +40,35 @@ _SCOPE_MODE_CENTER = 0
 # and slide the window to follow them, so any radio/level stays visible.
 _PIXEL_MAX = 160
 
-# Robust per-frame floor/ceil estimators (percentiles of the dB array).
-# A low percentile rejects the few near-silent bins; a high percentile rejects
-# single-bin spikes (DC leakage etc.) that would otherwise inflate the ceiling.
-_FLOOR_PCT = 10.0
+# Robust per-frame floor/ceil estimators (percentiles of the IN-BAND dB array).
+#
+# Radio RX audio is BIMODAL (MOR-512): the demodulated audio baseband
+# (~0-3.2 kHz) always carries receiver band-noise, while the out-of-band region
+# (>~4-5 kHz) sits far lower, near the true noise floor. Estimating the floor
+# over the FULL spectrum latches it onto the quiet out-of-band majority, which
+# drags the floor way down and pushes the ever-present in-band band-noise up to
+# ~68% of the screen with NO signal present ("high noise floor regardless of
+# signal"). We therefore estimate floor and ceil over the IN-BAND region only —
+# the bins that actually carry radio audio (and that the display shows) — so the
+# band-noise maps LOW and a real signal stands out above it.
+#
+# Floor: a low-ish percentile of the in-band bins (rejects the quietest bins
+# without latching onto out-of-band). Ceil: a high percentile (rejects single-bin
+# DC-leakage spikes that would otherwise inflate the ceiling).
+_FLOOR_PCT = 30.0
 _CEIL_PCT = 99.0
+
+# In-band (audio baseband) region used for the floor/ceil estimate.
+#
+# Lower edge: skip DC / sub-sonic bins (hum, DC leakage) that do not represent
+# band-noise. Upper edge: when the mode bandwidth is known the in-band region is
+# the displayed passband (``_crop_max_hz / 2`` of audio, since the crop keeps
+# ±half on each side of center); when it is unknown (e.g. FTX-1, which reports no
+# filter ``max_hz`` and shows the full spectrum) fall back to a fixed audio
+# baseband limit that covers a typical voice passband without reaching into the
+# quiet out-of-band tail that caused the regression.
+_INBAND_LO_HZ = 50.0
+_INBAND_FALLBACK_HI_HZ = 3500.0
 
 # EMA / attack-decay smoothing of the tracked floor & ceil. The floor adapts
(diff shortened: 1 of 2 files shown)


morozsm:
**Agent Review: PASS**

**Old-code falsification:** reverted ONLY `fft_scope.py` to `origin/main` → `test_bimodal_no_signal_inband_noise_maps_low` FAILS (`assert 0.6459 < 0.3` — band-noise at 65%, the live regression); `test_bimodal_with_signal_tone_rises_above_band_noise` FAILS (64%). Restored after.

**Noise-low + signal-visible (make-or-break) — independent bimodal harness, 60-frame settle, FTX-1 fallback path:**
| Scenario | floor | ceil | band-noise | signal | sep |
|---|---|---|---|---|---|
| no-signal | −108.6 | −63.6 | **14.7%** (oob 0%) | — | — |
| +25 dB tone | −108.6 | −63.6 | 14.8% | 69.9% | 55% |
| +8 dB weak tone | −108.6 | −63.6 | 14.8% | 32.1% | 17% (visible) |
| steady 120-frame | — | — | — | — | floor/ceil span **0.000 dB** (no pump) |
| very loud broadband | −26.9 | +18.1 | 15% | — | not saturated |

**In-band region** correct: `50 Hz ≤ freq ≤ _crop_max_hz/2` (matches the displayed one-sided passband) else 3500 fallback; empty-mask guard returns full array (no crash/RuntimeWarning). **Fallback robust** (in-band-mostly-filled doesn't collapse the floor; signal near the 3500 edge fine). **IC-7610 SSB solid** (`max_hz=3600` → edge 1800 Hz; no-signal 13.8%, a 2200 Hz tone still renders 68.9% since the slice only affects floor/ceil, not the pixel map). **Guards/state** all unchanged; test diff purely additive; 4 pre-existing adaptive tests untouched & green.

**Full gate (verbatim):**
- `pytest --ignore=tests/integration` → 7270 passed, 9 skipped, 11 xfailed, 1 xpassed, 0 failed
- `pytest tests/test_audio_fft_scope.py` → 35 passed
- `ruff check` → All checks passed! · `ruff format --check` → 546 files already formatted
- `mypy src/` → 20 baseline, 0 new (none in fft_scope.py)

**Non-blocking follow-up:** IC-7610 AM (`max_hz=10000` → edge 5000 Hz) can re-trigger the regression shape if AM audio is narrow → cap edge at `min(_crop_max_hz/2, ~3500)`. AM-only, narrow-audio edge; SSB/CW/data/RTTY + FTX-1 fallback unaffected. Filed as a hardening follow-up. Live FTX-1 re-validation pending.

PR #1738, opened 2026-06-09
feat: USB duplex-policy resolver on UsbAudioDriver (MOR-534, AudioTransport 1/12)

morozsm:
## Summary

Step 1/12 of the MOR-532 AudioTransport epic (Linear MOR-534).

Adds a pure read-only USB duplex policy to `src/rigplane/audio/usb_driver.py`:

- Module-level `resolve_usb_duplex_mode(rx_dev, tx_dev) -> Literal["full", "exclusive"]` — returns `"exclusive"` iff the platform is macOS AND RX and TX resolve to the same device index AND the device is a real CODEC (not a virtual loopback); `"full"` otherwise (separate devices, virtual loopbacks like BlackHole/VB-Cable, non-macOS).
- Lazy `UsbAudioDriver.duplex_mode` property — resolves devices via the normal `_ensure_selected_devices` selection path on first access.

Rationale: single source of truth for the macOS CoreAudio AUHAL -50 same-device reality (MOR-531 context). Pure policy — nothing consumes it yet; later epic steps route stream setup through it. Zero behavior change anywhere.

The virtual-loopback predicate `_is_virtual_loopback_device` stays in `audio/bridge.py` (shared via late module-attribute lookup, not moved), so existing monkeypatch targets in bridge tests keep working — verified by a dedicated test.

## Tests (TDD, falsification-first)

New `tests/test_usb_duplex_mode.py` (7 tests, confirmed FAILING before implementation, PASSING after):

- same CODEC device on macOS → `"exclusive"`
- separate devices on macOS → `"full"`
- virtual loopback (BlackHole) same-device on macOS → `"full"`
- same device on non-Darwin (monkeypatched platform) → `"full"`
- monkeypatching `bridge._is_virtual_loopback_device` steers the resolver (shared predicate, late binding)
- `UsbAudioDriver.duplex_mode` lazy property: CODEC → `"exclusive"`, loopback → `"full"` (using `FakeAudioBackend`, no one-off mocks)

## Gate results

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — 7295 passed, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures (incl. `tests/test_audio_duplex.py` and bridge tests unchanged and green)
- `uv run ruff check src/ tests/` — all checks passed; `uv run ruff format --check` — 548 files already formatted
- `uv run mypy src/` — 20 errors in 7 files (baseline, zero new)

## Scope

2 files, +173/-1 (52 LOC in `usb_driver.py`, rest is the test module). Within guardrails.

### DIFF
--- a/src/rigplane/audio/usb_driver.py
+++ b/src/rigplane/audio/usb_driver.py
@@ -15,7 +15,7 @@ import asyncio
 import logging
 import sys
 from dataclasses import dataclass
-from typing import Any, Callable
+from typing import Any, Callable, Literal
 
 from .backend import (
     AudioBackend,
@@ -446,6 +446,44 @@ def _extract_sounddevice_module(backend: AudioBackend) -> Any | None:
     return None
 
 
+def resolve_usb_duplex_mode(
+    rx_dev: UsbAudioDevice,
+    tx_dev: UsbAudioDevice,
+) -> Literal["full", "exclusive"]:
+    """Resolve the USB duplex policy for a selected RX/TX device pair.
+
+    Returns ``"exclusive"`` iff the platform is macOS AND RX and TX resolve
+    to the same device index AND the device is a real CODEC (not a virtual
+    loopback such as BlackHole/VB-Cable). On macOS CoreAudio, two separate
+    streams on one real C-Media USB CODEC fail with AUHAL -50 (MOR-531), so
+    such a device must be owned exclusively by ONE full-duplex stream.
+    Everything else — separate devices, virtual loopbacks, non-macOS
+    platforms — supports the two-stream path: ``"full"``.
+
+    Pure read-only policy (MOR-534, AudioTransport 1/12): nothing consumes
+    it yet; later epic steps route stream setup through it.
+    """
+    if sys.platform != "darwin":
+        return "full"
+    if rx_dev.index != tx_dev.index:
+        return "full"
+    # Single source of truth with the bridge path: reuse (not move) the
+    # virtual-loopback predicate. Imported lazily and called via the module
+    # attribute so existing monkeypatch targets on ``rigplane.audio.bridge``
+    # keep steering both consumers.
+    from rigplane.audio import bridge
+
+    info = AudioDeviceInfo(
+        id=AudioDeviceId(rx_dev.index),
+        name=rx_dev.name,
+        input_channels=rx_dev.input_channels,
+        output_channels=rx_dev.output_channels,
+    )
+    if bridge._is_virtual_loopback_device(info):
+        return "full"
+    return "exclusive"
+
+
 class UsbAudioDriver:
     """Stateful USB audio driver with deterministic device selection.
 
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of head c57eeef3 in a throwaway worktree (origin/codex/mor-534-usb-duplex-policy).

**Scope verified**
- Diff is exactly 2 files (+173/−1): `src/rigplane/audio/usb_driver.py`, new `tests/test_usb_duplex_mode.py`. `bridge.py` is untouched (0-line diff vs main).
- usb_driver changes are purely additive: `Literal` import, module-level `resolve_usb_duplex_mode`, lazy `UsbAudioDriver.duplex_mode` property, `__all__` entry. `_ensure_selected_devices` itself is unmodified.
- No consumer of `duplex_mode` anywhere in `src/` and the driver defines no `__repr__`/asdict/vars iteration, so the property cannot be accidentally triggered by logging or introspection — zero behavior change confirmed.

**Policy correctness**
- Platform gate uses `sys.platform != "darwin"`, consistent with the module's two existing checks (usb_driver.py:361, :466).
- Same-device check compares `UsbAudioDevice.index` from a single enumeration — semantically equivalent to bridge's own `int(tx_dev_id) == int(dev_id)` duplex decision.
- `AudioDeviceInfo` constructed from `UsbAudioDevice` maps id/name/channels faithfully; the bridge predicate `_is_virtual_loopback_device` reads only `dev.name`, so no semantic distortion.
- Lazy `from rigplane.audio import bridge` + call-time attribute lookup: bridge.py does not import usb_driver at module level (no hard cycle either way), and the call-time module-attribute access preserves existing monkeypatch targets — proven by `test_resolver_shares_bridge_loopback_predicate`, which asserts the patched predicate is actually invoked.

**Tests (falsified)**
- 7 tests cover all 4 spec scenarios (same-device-macOS→exclusive, separate→full, loopback→full, non-darwin→full) plus late-binding monkeypatch and the lazy property via `FakeAudioBackend` (both exclusive and loopback paths); the property test also asserts device selection occurred.
- Falsification: with main's usb_driver swapped in, the new test file fails collection (ImportError on `resolve_usb_duplex_mode`); restored to PR head, 7/7 pass.

**Full gate (run locally)**
- `pytest tests/ -q --ignore=tests/integration`: 7295 passed, 0 failed (9 skipped, 11 xfailed, 1 xpassed)
- `tests/test_audio_duplex.py` + `tests/test_audio_bridge.py`: 53 passed, files unchanged
- `ruff check`: clean · `ruff format --check`: clean (548 files)
- `mypy src/`: 20 errors in 7 files — exact baseline, none in usb_driver.py
- `lint-imports`: 4 contracts kept, 0 broken

**Non-blocking note for later epic steps**: bridge's `use_duplex` has no platform gate (it opens a single duplex stream for same-device on any OS), while this policy returns "full" for non-Darwin same-device. Intentional per spec (AUHAL -50 is macOS-only), but when a future step routes bridge stream setup through this policy, Linux same-device behavior would shift from single-duplex to two-stream — worth a conscious decision at that point.

PR #1741, opened 2026-06-09
feat: audio descriptors on the Icom serial base (MOR-536, AudioTransport 3/12)

morozsm:
## Summary

Step 3/12 of the MOR-532 AudioTransport epic (Linear MOR-536): add the audio descriptor surface to `_IcomSerialRadioBase`, covering X6200 and the other Icom serial (USB CI-V) backends.

- `audio_tx_codec` → always `AudioCodec.PCM_1CH_16BIT`. The serial USB CODEC mic input is mono-only by hardware; `start_audio_tx_pcm` clamps the PortAudio output stream to one channel (GH#1382), so the TX payload is mono 16-bit PCM regardless of the configured RX codec.
- `audio_duplex_mode` → delegates to the USB audio driver's `duplex_mode` policy (MOR-534, merged in aa1c4852) via `getattr`, so the internal `_SerialAudioDriver` protocol stays narrow. Falls back to `"full"` when the driver does not expose the property or when resolving it raises (device enumeration may fail on offline hosts and test doubles).

Docstrings mirror the MOR-535 LAN descriptor wording in `runtime/_audio_runtime_mixin.py`. No `AudioCapable`/protocol changes, no consumers wired yet.

## Tests (TDD, falsification-first)

New tests in `tests/test_icom7610_serial_radio.py`, using the existing `_FakeUsbAudioDriver` stub pattern (no one-off MagicMocks):

- both descriptors present on `Icom7610SerialRadio` and `Ic705SerialRadio`
- `audio_tx_codec == AudioCodec.PCM_1CH_16BIT`, including with a stereo RX codec configured
- `audio_duplex_mode` returns the driver's value (`"exclusive"` / `"full"`) when the driver exposes `duplex_mode`
- `"full"` fallback for stubs without the property and for a driver whose property raises

Red→green: the delegation test failed before implementation (`assert 'full' == 'exclusive'` — the inherited LAN mixin descriptor returned `"full"` unconditionally); all 26 tests in the file pass after.

## Gate results (worktree)

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7304 passed**, 9 skipped, 11 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` — All checks passed; `ruff format --check` — 548 files already formatted
- `uv run mypy src/` — 20 errors in 7 files (= baseline, zero new)

## Scope

2 files, +100 LOC (29 src / 71 tests).

### DIFF
--- a/src/rigplane/backends/_icom_serial_base.py
+++ b/src/rigplane/backends/_icom_serial_base.py
@@ -178,6 +178,35 @@ class _IcomSerialRadioBase(CoreRadio):
 
         return getattr(self._serial_audio_driver, "usb_audio_contract", None)
 
+    @property
+    def audio_tx_codec(self) -> AudioCodec:
+        """Effective TX codec (MOR-532 codec descriptor surface).
+
+        The serial USB CODEC mic input is mono-only by hardware:
+        ``start_audio_tx_pcm`` clamps the PortAudio output stream to one
+        channel (GH#1382), so the TX payload is always mono 16-bit PCM
+        regardless of the configured RX codec. Consumed by later MOR-532
+        epic steps.
+        """
+        return AudioCodec.PCM_1CH_16BIT
+
+    @property
+    def audio_duplex_mode(self) -> str:
+        """Duplex capability (MOR-532 duplex descriptor surface).
+
+        Delegates to the USB audio driver's ``duplex_mode`` policy (MOR-534)
+        via ``getattr`` so the internal ``_SerialAudioDriver`` protocol stays
+        narrow. Falls back to ``"full"`` when the driver does not expose the
+        property or when resolving it raises (device enumeration may fail on
+        offline hosts and test doubles). Single source of duplex policy for
+        later MOR-532 epic steps.
+        """
+        try:
+            mode = getattr(self._serial_audio_driver, "duplex_mode", None)
+        except Exception:
+            return "full"
+        return mode if isinstance(mode, str) else "full"
+
     # ------------------------------------------------------------------
     # Connection properties
     # ------------------------------------------------------------------
--- a/tests/test_icom7610_serial_radio.py
+++ b/tests/test_icom7610_serial_radio.py
@@ -7,6 +7,7 @@ import asyncio
 import pytest
 
 from rigplane import IcomRadio, RadioConnectionState
+from rigplane.backends.ic705 import Ic705SerialRadio
 from rigplane.backends.icom7610 import Icom7610SerialRadio
 from rigplane import IC_7610_ADDR
 from rigplane.commands import (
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of head `2c6caf8b` (MOR-536, AudioTransport 3/12).

**Scope** — `gh pr diff` confirms exactly 2 files, purely additive: `src/rigplane/backends/_icom_serial_base.py` (+29, two properties) and `tests/test_icom7610_serial_radio.py` (+71, 4 tests + 2 fake-driver subclasses). `_SerialAudioDriver` protocol untouched, no behavior removed.

**MRO verified empirically** — for all four serial subclasses (Icom7610SerialRadio, Ic705SerialRadio, Ic7300SerialRadio, Ic9700SerialRadio) the MRO is `[Subclass, _IcomSerialRadioBase, CoreRadio, ScopeRuntimeMixin, AudioRuntimeMixin, DualRxRuntimeMixin]`, so the new base-class properties now shadow the MOR-535 `AudioRuntimeMixin` descriptors (f2c0ca19) on every serial backend. X6200 routes through `Ic705SerialRadio` per `backends/factory.py` (`model in ("IC-705", "X6200")`), so it is covered too.

**Override vs inheritance** — the override is justified, not a restatement:
- `AudioRuntimeMixin.audio_tx_codec` reads `_audio_stream_contract.tx_codec`, which `resolve_lan_audio_stream_request()` (audio/route.py) takes from `profile.tx_codec` when set. Today every shipped profile pins `PCM_1CH_16BIT` (so values coincide), but the schema permits any codec — a future profile with a non-mono LAN `tx_codec` would have silently leaked onto the serial path, contradicting the GH#1382 hardware mono clamp in `start_audio_tx_pcm`. The override pins the correct serial invariant.
- `AudioRuntimeMixin.audio_duplex_mode` is a hardcoded `"full"` (correct for Icom LAN UDP); the serial override delegates to the driver's MOR-534 `duplex_mode` — genuinely different behavior. Real `UsbAudioDriver.duplex_mode` lazily enumerates devices and can raise on offline hosts, so the `try/except → "full"` fallback is warranted. `_serial_audio_driver` is never `None` (always constructed in `__init__`), and `getattr(..., None)` handles drivers without the property.

**Falsification (run myself)** — with `origin/main`'s `_icom_serial_base.py` + the PR's tests: exactly 1 failure, `test_serial_audio_duplex_mode_delegates_to_driver` (`assert 'full' == 'exclusive'`); the other 3 pass via the inherited mixin (coinciding fallbacks), matching the author's self-flag. Restored PR file: 26/26 pass. One falsifying test + three pinning tests is adequate for a descriptor-surface step; the pinning tests guard against future regressions in the contract-resolution path.

**Full gate (all run in a clean worktree of the PR head):**
- `pytest tests/ -q --ignore=tests/integration` → 7304 passed, 9 skipped, 11 xfailed, 1 xpassed, 0 failures
- `ruff check src/ tests/` → clean; `ruff format --check` → 548 files already formatted
- `mypy src/` → 20 errors in 7 files (exact baseline, zero new, none in touched files)
- `lint-imports` → 4 kept / 0 broken

**Non-blocking notes:** Ic7300/Ic9700 serial backends inherit the new properties but have no direct test (base-class coverage via 7610/705 is acceptable); the broad `except Exception → "full"` could report full duplex for an exclusive-only device when enumeration fails, but that matches the documented descriptor-fallback semantics and the actual stream open still enforces reality.

Mergeable once required checks are green.

PR #1744, opened 2026-06-09
feat: AudioTransport neutral methods on the Icom serial base (MOR-540, AudioTransport 7/12)

morozsm:
## Summary

Step 7/12 of the MOR-532 AudioTransport epic (Linear MOR-540): the Icom serial base gains the codec/transport-neutral `AudioTransport` methods, and the synthetic packet ident gets a documented name.

- **`src/rigplane/audio/lan_stream.py`** — new module constant `SYNTHETIC_RX_IDENT = 0x9781` next to the existing LAN RX ident constants, documenting that it marks locally-framed (non-LAN-wire) AudioPackets produced by USB-codec backends. Purely additive; intentionally NOT exported from `rigplane.audio.__init__` (the step-12 conformance/export step handles that — `tests/test_exports.py` untouched).
- **`src/rigplane/backends/_icom_serial_base.py`** — five neutral methods over `_serial_audio_driver`:
  - `start_rx(cb)` = body of the former `start_audio_rx_opus` (AudioPacket framing now via `SYNTHETIC_RX_IDENT`, PCM→Opus transcode branch kept inside, keyed on the existing codec check)
  - `stop_rx()`, `start_tx()` (no-arg TX arming), `push_tx(data)` (incl. Opus→PCM transcode + "Audio TX not started" guard), `stop_tx()` (clears `_pcm_tx_fmt`)
  - Legacy opus-family methods re-pointed to one-line delegates onto the neutral ones (`start_audio_rx_opus` keeps its `jitter_depth` validation/assignment — the neutral surface has no jitter knob). PCM-family methods unchanged.
- Packet bytes are **identical** to before: ident value unchanged (0x9781), wrapping uint16 `send_seq` unchanged, payload unchanged. MOR-242 TX clamp and MOR-238/MOR-236 RX channel tests pass unmodified.

## Tests (TDD)

New `tests/test_audio_transport_serial.py` (reuses the existing serial fakes from `test_icom7610_serial_radio.py`), confirmed red before implementation (5 failed), green after:
- `isinstance(Icom7610SerialRadio/Ic705SerialRadio, AudioTransport)` is True
- legacy↔neutral RX equivalence: identical `AudioPacket` sequences + identical driver start counts
- legacy↔neutral TX equivalence: identical `start_tx` kwargs + identical pushed frames
- byte-compat lock: `start_rx` packets carry `ident == 0x9781 == SYNTHETIC_RX_IDENT`, seq 0,1
[...]

PR #1745, opened 2026-06-09
feat: AudioTransport on YaesuCatRadio — un-impersonate FTX-1 (MOR-541, AudioTransport 8/12)

morozsm:
## Summary

Step 8/12 of the MOR-532 AudioTransport epic (Linear MOR-541). `YaesuCatRadio` now implements the codec/transport-neutral `AudioTransport` surface (`start_rx` / `stop_rx` / `start_tx` / `push_tx` / `stop_tx`) over the USB audio driver, and the legacy `*_opus`/`*_pcm` family becomes back-compat delegates — mirroring the merged LAN (MOR-539) and Icom-serial (MOR-540) implementations.

## Spec mapping

- `start_rx(cb)` carries the body of the old `start_audio_rx_opus`: the `_on_pcm_frame` wrapper fabricating synthetic `AudioPacket`s, with the ident sourced from `rigplane.audio.lan_stream.SYNTHETIC_RX_IDENT` instead of the hardcoded `0x9781` literal. Packet bytes are **identical** to before: same ident value (0x9781), same wrapping uint16 `send_seq`, same PCM payload. Docstring states frames carry PCM per `audio_codec` — no Opus claim.
- `start_tx()` = legacy `start_audio_tx_pcm()` defaults (mono PCM at `audio_sample_rate`, 20 ms frames); `push_tx(data)` = the `_push_pcm_tx` path; `stop_rx()`/`stop_tx()` close the driver streams.
- Recursion safety (MOR-539 case): `_push_pcm_tx` stays the private terminal push helper — `push_tx` and both legacy push shims funnel into it; `start_tx`/`stop_tx` call the driver directly. Verified against the real call graph: no `pcm↔opus` cycle exists.

## Behavior change (deliberate)

`start_audio_tx_opus` changes from a **silent no-op** (`pass`) to delegating to `start_tx()`. This fixes the bug class where the web handler's opus TX branch (`web/handlers/audio.py` `_handle_control`) silently did nothing on Yaesu backends, leaving the driver TX path unarmed. On double start it now raises `AudioDriverLifecycleError` ("TX stream already started.", a `RuntimeError` subclass) — the exact same exception `start_audio_tx_pcm` raises today, so the handler's existing `except RuntimeError` double-start handling sees no new exception type.

`push_audio_tx_opus`'s push-opus-as-pcm oddity is kept byte-for-byte unchanged and documented as legacy shim behavior.

## Tests (TDD, `tests/test_yaesu_audio.py`)

All five new tests confirmed failing before implementation, passing after:

1. `isinstance(YaesuCatRadio(...), AudioTransport)` is True.
2. `start_rx` packets carry `ident == SYNTHETIC_RX_IDENT == 0x9781`, incrementing seq, payload passthrough.
3. Full `AudioPacket` equality + identical driver-call traces, legacy vs neutral RX session (mirrors the MOR-540 test approach).
4. Identical driver-call traces legacy-opus vs neutral TX session (delegates only, no divergent framing logic).
5. Regression: `start_audio_tx_opus` now arms TX (driver `start_tx` called, `tx_running` True) where before it was a no-op.

Existing FTX-1/Yaesu audio tests pass unmodified.

## Gate results

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7335 passed**, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` — All checks passed; `ruff format --check` — 551 files already formatted
- `uv run mypy src/` — 21 errors in 8 files, **identical to the 582e4d7d baseline** (diff is line-number shifts only; zero new errors)

### DIFF
--- a/src/rigplane/backends/yaesu_cat/radio.py
+++ b/src/rigplane/backends/yaesu_cat/radio.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Literal, Sequence, cast
 
 from ...audio import AudioPacket
+from ...audio.lan_stream import SYNTHETIC_RX_IDENT
 from ...command_spec import CatCommandSpec
 from ...commands import hz_to_table_index, table_index_to_hz
 from ...types import AudioCodec, BreakInMode
@@ -369,29 +370,30 @@ class YaesuCatRadio:
             self._audio_bus = AudioBus(self)
         return self._audio_bus
 
-    # -- AudioCapable methods -----------------------------------------------
+    # -- Neutral AudioTransport surface (MOR-532 epic, MOR-541) --------------
 
-    async def start_audio_rx_opus(
+    async def start_rx(
         self,
         callback: Callable[[AudioPacket | None], None],
-        *,
-        jitter_depth: int = 5,
     ) -> None:
+        """Start RX capture from the USB audio device (``AudioTransport.start_rx``).
+
+        Frames carry raw PCM encoded per :attr:`audio_codec`
+        (``PCM_1CH_16BIT``) — the FTX-1 USB path has no Opus framing.
+        Each captured frame is wrapped in a synthetic
+        :class:`AudioPacket` marked with ``SYNTHETIC_RX_IDENT`` and a
+        locally counted wrapping uint16 ``send_seq`` — there is no LAN
+        wire header on this path.
+        """
         if not callable(callback):
             raise TypeError("callback must be callable and accept AudioPacket | None.")
-        if isinstance(jitter_depth, bool) or not isinstance(jitter_depth, int):
-            raise TypeError(
-                f"jitter_depth must be an int, got {type(jitter_depth).__name__}."
-            )
-        if jitter_depth < 0:
-            raise ValueError(f"jitter_depth must be >= 0, got {jitter_depth}.")
         self._require_connected()
 
         self._opus_rx_user_callback = callback
 
         def _on_pcm_frame(pcm_frame: bytes) -> None:
             packet = AudioPacket(
-                ident=0x9781,
+                ident=SYNTHETIC_RX_IDENT,
                 send_seq=self._audio_seq,
                 data=pcm_frame,
             )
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of head `bce45a40` (MOR-541, AudioTransport epic 8/12). Verified in a throwaway worktree at the PR head.

**Scope** — exactly 2 files as claimed: `src/rigplane/backends/yaesu_cat/radio.py` (+117/−24), `tests/test_yaesu_audio.py` (+190). No unplanned changes.

**The deliberate behavior change (`start_audio_tx_opus`: no-op → `start_tx()`)**
- Regression test `test_start_audio_tx_opus_arms_tx` is real: with `radio.py` reverted to origin/main, all 5 new tests fail (the regression test fails precisely on `driver.tx_running is False`); restored, 40/40 pass.
- Production reachability audited: the only caller that hits the opus branch for a YaesuCatRadio is `web/radio_poller.py:1173` (PttOn) — `_audio_tx_codec_and_rate()` consults only `audio_stream_contract` (absent on Yaesu) → `tx_codec=None` → opus branch. So PttOn now genuinely arms USB TX on FTX-1, matching the already-merged serial-Icom precedent (MOR-242 #1696, MOR-540 #1744). The web handler (`web/handlers/audio.py:706`) cannot reach the opus branch for Yaesu: `_tx_codec()` falls back to `audio_codec` = `PCM_1CH_16BIT` → always the PCM branch. `_handle_tx_audio` likewise selects the PCM push branch. No code path relied on the no-op.
- Double-start: `AudioDriverLifecycleError` is a `RuntimeError` subclass (`exceptions.py:58`), so the handler's `except RuntimeError` at `audio.py:712` catches it. Caveat (pre-existing, not introduced here): the USB driver's message is "TX stream already started.", which does NOT match the handler's `"Already transmitting"` tolerance (that string comes from `lan_stream.py:602` only), so a poller-first double start re-raises. Mitigations verified: normal ordering is handler-first (frontend sends `audio_start tx` at mic-enable, before PTT); the poller side swallows lifecycle errors in its broad except; worst case the reader task exits, `run()` logs and closes the audio WS, and the frontend auto-reconnects — no crash. This exact exposure already exists on main for serial Icom backends since #1696/#1744. Recommended follow-up (separate PR, would exceed this PR's 2-file scope): broaden the handler matcher to also tolerate `AudioDriverLifecycleError`/"already started".

**Packet byte-compat** — tests exercise the REAL code path: `TracingAudioDriver` extends `FakeAudioDriver`, frames are injected through the actual driver RX callback closure, and full frozen `AudioPacket` equality is asserted legacy-vs-neutral (ident `0x9781 == SYNTHETIC_RX_IDENT`, wrapping uint16 seq `[0, 1]`, payload identity). TX side compares driver `start_tx` kwargs and pushed frames legacy-vs-neutral. No hand-built expectations.

**Call graph** — acyclic: `push_tx → _push_pcm_tx` (private terminal helper, MOR-539 pattern); `push_audio_tx_{opus,pcm} → push_tx`; `start_audio_rx_opus → start_rx`; `start/stop_audio_tx_opus → start/stop_tx`. No `*_opus` method contains framing or transcode logic.

**mypy baseline verified myself** — reverted `radio.py` to origin/main in the same env and ran `uv run mypy src/` both ways: main baseline is **21 errors in 8 files** (the historical "20 in 7" is stale), PR is 21 errors in 8 files, and the error sets are **identical modulo line-number shifts** (+69 from added code in `radio.py`). Zero new errors attributable to this PR.

**Full gate** (in the PR worktree, `uv sync --all-extras`):
- `pytest tests/ -q --ignore=tests/integration`: 7335 passed, 9 skipped, 11 xfailed, 1 xpassed, 0 failures
- `ruff check src/ tests/`: clean; `ruff format --check`: 551 files already formatted
- `mypy src/`: 21 errors / 8 files == verified main baseline (zero new)

PR #1747, opened 2026-06-09
feat: radio_poller PTT path reads audio_duplex_mode + neutral TX (MOR-543, AudioTransport 10/12)

morozsm:
## Summary

Step 10/12 of the MOR-532 AudioTransport epic (Linear MOR-543). The radio_poller PTT audio path now targets the codec/transport-neutral `AudioTransport` surface and reads the `audio_duplex_mode` descriptor — with **NO behavior flip**: RX re-arm after TX still fires unconditionally for every radio, exactly as in MOR-506.

### Changes (`src/rigplane/web/radio_poller.py`)

- **PttOn**: when the radio exposes the neutral surface, `await radio.start_tx()` (backend resolves the TX format from its negotiated contract). `getattr` fallback keeps the current `_audio_tx_codec_and_rate()` codec-branched path for legacy radios.
- **PttOff**: `await radio.stop_tx()` when neutral (same fallback), then the RX re-arm is gated through a new explicit helper `_should_restart_rx(mode: str) -> bool` that takes `getattr(radio, "audio_duplex_mode", "half")`, logs the mode at debug level, and **returns True for ALL modes** — preserving MOR-506 unconditional re-arm semantics exactly. The future skip-on-"full" flip becomes a one-line change with the decision point visible.
- Deleted the misleading "IC-7610 doesn't support full duplex" comment — the LAN stream IS full-duplex; the re-arm exists to re-install the single-slot RX callback (MOR-506). Replaced with the duplex-mode rationale + the pending hardware-gated flip note.
- `_audio_tx_codec_and_rate` retained, now only reachable from the legacy fallback (docstring states this).

**Behavior is identical to today for every radio.** The IC-7610 LAN path goes through neutral `start_tx`/`stop_tx`, which MOR-539's review proved byte-equivalent to the per-codec calls; the re-arm still always fires. The operator decision to defer the skip-restart_rx-on-"full" flip to a separate hardware-gated ticket (after live IC-7610 verification) is intentional and out of scope here.

### Tests (`tests/test_audio_rx_tx_transition.py`, TDD red → green)

New tests (failed before implementation, pass after):
- `test_ptt_on_neutral_radio_uses_start_tx` — neutral radio: `start_tx()` awaited, NO codec branching (legacy mocks not awaited).
- `test_ptt_off_neutral_radio_stops_tx_then_restarts_rx[full|exclusive|None]` — `stop_tx()` then `audio_bus.restart_rx()` in order, for mode `"full"`, `"exclusive"`, AND missing-attr — all re-arm (behavior preserved).
- `test_legacy_radio_without_neutral_methods_uses_codec_fallback` — legacy radio still routes through `_audio_tx_codec_and_rate` branches.
- `test_should_restart_rx_returns_true_for_all_modes` — pins the no-flip decision for `"full"`/`"half"`/`"exclusive"`.

All 7 existing MOR-506 tests pass **unmodified** (incl. the real-AudioBus clobber regression and the PCM-contract test).

### Gate results (worktree)

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7341 passed**, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` → All checks passed; `ruff format --check` → 551 files already formatted
- `uv run mypy src/` → 21 errors in 8 files — **identical to the main baseline** (the 4 radio_poller entries are pre-existing `SetData*ModInput` capture-pattern errors, shifted by exactly the +44-line diff offset; zero new)
- `uv run lint-imports` → Contracts: 4 kept, 0 broken

Guardrails: 2 files, +148/−20.

### DIFF
--- a/src/rigplane/web/radio_poller.py
+++ b/src/rigplane/web/radio_poller.py
@@ -187,12 +187,32 @@ _SLOW_INTERVAL: float = 0.25  # levels/settings — rarely change
 
 
 def _audio_tx_codec_and_rate(radio: Any) -> tuple[AudioCodec | None, int]:
+    """Legacy TX-format resolution for radios WITHOUT the neutral
+    ``AudioTransport`` surface; only reachable from the PTT fallback path
+    (MOR-543). Radios exposing ``start_tx``/``stop_tx`` resolve the format
+    themselves from their negotiated contract.
+    """
     contract = getattr(radio, "audio_stream_contract", None)
     tx_codec = getattr(contract, "tx_codec", None)
     tx_sr = getattr(contract, "tx_sample_rate_hz", None)
     if not isinstance(tx_sr, int) or isinstance(tx_sr, bool) or tx_sr <= 0:
         tx_sr = 48000
     return tx_codec, tx_sr
+
+
+def _should_restart_rx(mode: str) -> bool:
+    """Whether the RX path must be re-armed after a PTT TX cycle.
+
+    *mode* is the radio's ``audio_duplex_mode`` descriptor. Returns True
+    for ALL modes for now, preserving the MOR-506 unconditional re-arm
+    semantics exactly: the re-arm reinstates the single-slot RX callback
+    through the AudioBus, and skipping it for ``"full"`` transports is
+    deferred to a hardware-gated follow-up (verify on a real IC-7610
+    first). When that flip lands, this helper is the one-line decision
+    point (MOR-543).
+    """
+    logger.debug("poller: audio_duplex_mode=%s -> restart_rx=True", mode)
+    return True
 
 
 # ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of head e65dbf6a (codex/mor-543-poller-duplex-mode), reviewed in a throwaway worktree.

**Scope**: exactly 2 files, +148/−20, matching the claim — `src/rigplane/web/radio_poller.py`, `tests/test_audio_rx_tx_transition.py`. Test file has 0 deletions: all 7 pre-existing MOR-506 tests are byte-identical and pass unmodified (13/13 in the file).

**Point 1 — IC-7610 LAN path parity (verified against merged `runtime/_audio_runtime_mixin.py`)**:
- IcomRadio always builds `_audio_stream_contract` in `__init__` (radio.py:838) and mirrors `_audio_tx_codec = contract.tx_codec` (:842). The neutral `start_tx()` (mixin:89-103) branches on the `audio_tx_codec` property (:520-532), which reads `contract.tx_codec` — the SAME source the old poller helper `_audio_tx_codec_and_rate` read via the public `audio_stream_contract` property.
- PCM-negotiated contract: old poller called `start_audio_tx_pcm(sample_rate=contract.tx_sample_rate_hz)`; neutral `start_tx()` calls `start_audio_tx_pcm()` whose None-default resolves `contract.tx_sample_rate_hz` (mixin:281-285) — same value, same `_pcm_tx_fmt` tuple, same downstream stream start.
- Opus contract: old `start_audio_tx_opus()` is now a back-compat delegate to `_start_tx_stream()` (mixin:247-254), which is exactly what neutral `start_tx()`'s else-branch calls. Identical body.
- Stop side: old poller's `stop_audio_tx_pcm()`/`stop_audio_tx_opus()` BOTH delegate to neutral `stop_tx()` (mixin:361-370), so the PR's direct `stop_tx()` call is wire-identical for both codecs.
- Error parity: PttOn keeps the same try/except → `logger.warning`, PttOff the same try/except → `logger.debug`; `restart_rx` stays inside the same try block as on main. No new crash path in the poller loop.
- Only theoretical divergence: a contract-LESS radio that nevertheless exposes `start_tx` would default to the PCM branch (`_audio_tx_codec` fallback) where the old poller defaulted to opus on `tx_codec=None` — unreachable for IcomRadio (contract always set) and the PCM branch still lands on the same `_start_tx_stream()`. Noting for the record, not a blocker.

**Point 2 — re-arm unconditionality**: `_should_restart_rx` (radio_poller.py:203-215) literally `return True` for every input, logs the mode at debug; single call site (radio_poller.py:1243, PttOff only). MOR-506 unconditional re-arm semantics preserved exactly; the "full"-skip is correctly deferred as a hardware-gated follow-up per the no-flip mandate.

**Point 3 — legacy fallback**: the fallback block is character-for-character the old code: `_audio_tx_codec_and_rate(radio)` → `start_audio_tx_pcm(sample_rate=tx_sr)` / `start_audio_tx_opus()`, and `stop_audio_tx_pcm()` / `stop_audio_tx_opus()`, same kwargs, same log lines. Confirmed by test `test_legacy_radio_without_neutral_methods_uses_codec_fallback`.

**Point 4 — tests**: `test_ptt_off_neutral_radio_stops_tx_then_restarts_rx` pins ordering via `assert order == ["stop_tx", "restart_rx"]` and parametrizes `full`/`exclusive`/missing-attr; `test_should_restart_rx_returns_true_for_all_modes` pins the no-flip. Falsification check: with `radio_poller.py` reverted to origin/main, the file fails collection (ImportError on `_should_restart_rx`); restored to PR head → 13/13 pass.

**Full gate** (worktree, uv sync --all-extras):
- `pytest tests/ -q --ignore=tests/integration`: **7341 passed**, 9 skipped, 11 xfailed, 1 xpassed, 0 failures
- `ruff check src/ tests/`: clean; `ruff format --check`: 551 files already formatted
- `mypy src/`: 21 errors in 8 files — error set diffed against main with line numbers normalized: **IDENTICAL, zero new**

Verdict: merge-ready once required checks are green.

PR #1735, opened 2026-06-08
feat: full-duplex USB audio path (single sd.Stream) for same-device RX+TX (MOR-531)

morozsm:
## Summary
Adds a **single-`sd.Stream` full-duplex** USB audio path so a USB-CODEC radio (FTX-1, X6200) can do digital-mode TX (computer audio → radio via USB MOD) **while** the browser RX stream + FFT scope keep running — without the macOS CoreAudio `AUHAL -50` that two separate streams (`InputStream`+`OutputStream`) cause on one C-Media device. Replaces the `--bridge-rx-only` workaround for the same-device case. Do-now resolution of **MOR-505 layer-2**, scoped from the 2026-06-07 unified-audio-architecture analysis; live de-risk confirmed a single duplex stream starts cleanly on the C-Media (no `-50`).

## Design (additive)
- **`_PortAudioDuplexStream`** + `AudioBackend.open_duplex` (`audio/backend.py`): wraps ONE `sd.Stream(device=(idx,idx), channels=(rx_open,tx), …)`. Its single callback does **RX-fan** (via a new shared `_RxFramer` — the SAME `rx_audio_channel` downmix + lossless fixed-`frame_ms` re-chunk that `_PortAudioRxStream` used, now extracted and used by both) and **TX-pull** (by embedding a `_PortAudioTxStream` and driving its `_output_callback`/ring; its own `sd` stream is never opened; silence on empty).
- **`UsbAudioDriver.start_duplex`** (`audio/usb_driver.py`): opens duplex when `rx.index == tx.index`, else raises (caller uses the two-stream path); forces a shared sample rate.
- **Bridge** (`audio/bridge.py`): `use_duplex = tx_armed AND tx_dev == rx_dev AND not virtual_loopback` → one duplex stream serves both legs. Virtual loopback (BlackHole/VB-Cable/PipeWire) stays two-stream; **MOR-242 RX-only degrade preserved** (and cleaner).

`open_rx`/`open_tx` signatures **unchanged**; `_PortAudioTxStream` **byte-identical**; nothing renamed/removed (rigplane-pro-stable `audio/__init__` contract intact). All within the `audio` layer.

## Tests
[...]

PR #1750, opened 2026-06-10
fix: subscribe bridge RX before arming radio TX — LAN RX dead + packet-queue flood (MOR-556)

morozsm:
## Root cause

Regression from #1735 (MOR-531; found in live IC-7610 validation). Commit `d0a15d27` reordered `AudioBridge._setup_streams` so the radio TX arm (which feeds the `use_duplex` device-stream topology decision) runs BEFORE the AudioBus subscribe. On LAN backends this kills RX entirely:

- `LanAudioStream.start_tx()` flips the single state field to TRANSMITTING (`audio/lan_stream.py:603`).
- The later first-subscriber `AudioBus._begin_rx` → LAN mixin `start_rx` → `lan_stream.start_rx` raises `RuntimeError("Cannot start RX in state transmitting")` (`lan_stream.py:522-523`).
- The LAN state machine supports RX-then-TX only (`_rx_loop` keeps running through TRANSMITTING; `stop_tx` reverts to RECEIVING when `_rx_callback` is set). TX-then-RX is impossible.
- With no `_rx_loop`, nothing drains the audio transport's packet queue → `core/transport.py:644` overflow warning floods at ~150/s.

**Live evidence (IC-7610 LAN, 2026-06-09):** "Audio TX started" → "audio-bus: +subscriber 'audio-bridge'" → `ERROR audio-bus: failed to start RX` (exact traceback), repeated for 'web-audio'; WS RX 0 frames / analyzer rms −96; 21k+ packet-queue overflow warnings in 3 minutes.

## Fix (MOR-556 option 1 — minimal)

Move the AudioBus subscribe + subscription start in `_setup_streams` BEFORE the radio-TX-arm block. The `use_duplex` decision depends only on `tx_armed` and device identity (RX/TX device IDs + virtual-loopback check), not on bus state, so the MOR-531 duplex topology is unaffected. `rx_only` semantics are unchanged — the TX-arm block is still skipped entirely when TX is disabled.

## Falsification evidence (TDD)

The regression test was written FIRST, against a LAN-like radio stub that mirrors the real `LanAudioStream` state machine (`start_rx` requires IDLE, `start_tx` flips to TRANSMITTING) — order-insensitive `FakeAudioBackend` stubs are why CI missed this.

- **Red on unmodified main (`a52263d9`):** `test_bridge_subscribes_rx_before_arming_tx` fails with `RuntimeError: Cannot start RX in state transmitting` + `ERROR audio-bus: failed to start RX` — the exact live traceback.
- **Green with the fix.** Also asserts TX is still armed AFTER RX for non-rx_only configs (`calls == ["start_rx", "start_tx"]`, state ends TRANSMITTING) and that rx_only never arms TX.

## Gate results (worktree @ fix)

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7369 passed, 0 failed**, 9 skipped, 11 xfailed, 1 xpassed (one flaky perf-timing test `test_frame_creation_latency_distribution` failed once under full-suite load on the first run, passed in isolation and on full-suite rerun — unrelated to this change)
- `uv run ruff check src/ tests/` — All checks passed; `uv run ruff format --check` — 552 files already formatted
- `uv run mypy src/` — 21 errors in 8 files, error set **byte-identical** to main baseline (zero new)
- `uv run lint-imports` — 4 kept / 0 broken

Diff: 2 files, +105/−5 (within guardrails).

### DIFF
--- a/src/rigplane/audio/bridge.py
+++ b/src/rigplane/audio/bridge.py
@@ -501,6 +501,15 @@ class AudioBridge:
         else:
             self._decoder = None
 
+        # --- Subscribe to AudioBus BEFORE arming radio TX (MOR-556). The
+        # first subscriber triggers radio RX start, and the LAN stream state
+        # machine supports RX-then-TX only: arming TX first puts it in
+        # TRANSMITTING and the later RX start raises, leaving RX dead and the
+        # transport packet queue undrained (regression from #1735).
+        bus = self._radio.audio_bus
+        self._subscription = bus.subscribe(name="audio-bridge")
+        await self._subscription.start()
+
         # --- Arm the radio TX path BEFORE choosing the device-stream topology.
         # The radio may reject TX (serial backend with no USB-audio TX path):
         # degrade to RX-only and use a plain output stream, never a duplex one
@@ -583,11 +592,6 @@ class AudioBridge:
                 frame_ms=self._frame_ms,
             )
             await self._rx_stream.start()
-
-        # Subscribe to AudioBus
-        bus = self._radio.audio_bus
-        self._subscription = bus.subscribe(name="audio-bridge")
-        await self._subscription.start()
 
         # --- TX path: device input → radio (capture) ---
         if tx_armed:
--- a/tests/test_audio_bridge.py
+++ b/tests/test_audio_bridge.py
@@ -300,6 +300,102 @@ async def test_bridge_stop_when_not_running():
 
 
 # ---------------------------------------------------------------------------
+# RX-before-TX start order against a LAN-like state machine (MOR-556)
+# ---------------------------------------------------------------------------
+
+
+class _LanLikeRadio:
+    """Radio stub mirroring ``LanAudioStream``'s RX/TX state machine.
+
+    The real LAN stream supports RX-then-TX only: ``start_rx`` requires
+    IDLE state, while ``start_tx`` flips the single state field to
+    "transmitting". Arming radio TX before the AudioBus RX subscribe
+    therefore kills RX entirely (regression from #1735, MOR-556).
+    ``FakeAudioBackend``-style order-insensitive stubs cannot catch this.
+    """
+
+    def __init__(self) -> None:
+        from rigplane.audio_bus import AudioBus
+        from rigplane.types import AudioCodec
+
+        self.state = "idle"
+        self.calls: list[str] = []
+        self.rx_callback: object | None = None
+        self.audio_tx_codec = AudioCodec.PCM_1CH_16BIT
+        self.audio_bus = AudioBus(self)
+
+    async def start_rx(
+        self, callback: object, *, jitter_depth: int | None = None
+    ) -> None:
+        self.calls.append("start_rx")
+        if self.state != "idle":
+            raise RuntimeError(f"Cannot start RX in state {self.state}")
+        self.rx_callback = callback
+        self.state = "receiving"
+
+    async def stop_rx(self) -> None:
+        if self.state == "receiving":
+            self.state = "idle"
+        self.rx_callback = None
+
+    async def start_tx(self) -> None:
+        self.calls.append("start_tx")
+        if self.state == "transmitting":
+            raise RuntimeError("Already transmitting")
+        self.state = "transmitting"
+
+    async def stop_tx(self) -> None:
+        if self.state != "transmitting":
+            return
+        self.state = "receiving" if self.rx_callback is not None else "idle"
+
+    async def push_tx(self, audio_data: bytes) -> None:
+        if self.state != "transmitting":
+            raise RuntimeError(f"Cannot push TX in state {self.state}")
+
+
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of PR #1750 at head `dbaa49e7` (MOR-556 urgent fix: subscribe bridge RX before arming radio TX). Reviewed in a throwaway worktree; reviewer did not author this code.

## 1. Reorder side effects (traced `_setup_streams` head vs origin/main)

- The moved block (`src/rigplane/audio/bridge.py:504-511`) only registers a bus subscriber and starts radio RX (`bus.py` `_add_subscriber` → `radio.start_rx`); it opens no device streams and reads no TX state. Nothing between the old position and the new one depends on bus/RX state: the TX-arm block reads only `start_tx`/`audio_tx_codec`, and `use_duplex` (bridge.py:567-571) = `tx_armed` + device-id equality + virtual-loopback check — all bus-independent. MOR-531 same-device duplex topology is intact: `tests/test_audio_duplex.py::TestBridgeDuplexSelection` (duplex for same device, two-stream for separate devices, RX-only degrade) all pass on the PR head; `open_duplex` + `duplex_stream.start` still happen after TX arm, same relative order as main.
- Radio-side RX-then-TX order is safe on every backend: LAN `LanAudioStream.start_tx` explicitly allows the RECEIVING→TRANSMITTING transition ("Can be called while already receiving", `lan_stream.py:593-606`); USB-driver radios (serial Icom `_icom_serial_base.py:441`, Yaesu CAT/FTX-1 `_audio_runtime_mixin.py:89`) have independent RX/TX legs under separate locks — order-insensitive.
- Error path: TX-arm rejection (`RuntimeError`/`NotImplementedError`) is still caught → RX-only degrade with the subscription live and a plain output stream (covered by `test_bridge_rx_only_degrade_preserved_on_duplex`). Non-blocking observation: an UNCAUGHT failure after the subscribe (e.g. `ConnectionError` from `_require_connected` inside `start_tx`, or backend `open_tx`/`open_duplex` raising) propagates out of `start()` (bridge.py:757-761), which only sets IDLE and re-raises — the subscription is never closed, leaving radio RX running, and `stop()` early-returns on IDLE. This leak window pre-exists on main (subscribe sat before the TX-capture open there too); the reorder widens it to include TX-arm + device-open. The reconnect path self-heals via `_teardown_streams`. Suggested follow-up (not required here): call `_teardown_streams()` in `start()`'s except clause.

## 2. Stub fidelity (`_LanLikeRadio` vs real `LanAudioStream`)

- Faithful on every semantics the fix depends on: `start_rx` raises unless IDLE (`lan_stream.py:522-523`); accepts the `jitter_depth` kwarg so the bus's signature inspection exercises the real call shape; `start_tx` flips the single state field, raising only when already TRANSMITTING (`lan_stream.py:603-606`); `stop_tx` reverts to RECEIVING when the RX leg is alive (`lan_stream.py:644-650`; the stub proxies rx_task-alive via `rx_callback`); `push_tx` requires TRANSMITTING. The strict `calls == ["start_rx", "start_tx"]` assertion pins the order — a loose stub could not pass a wrong fix.
- One minor divergence: real `stop_rx` no-ops unless state == RECEIVING (`lan_stream.py:535-536`); the stub clears `rx_callback` unconditionally. Consequence: the test's trailing `radio.state == "idle"` would be "receiving" on real LAN after `bridge.stop()` (subscription closes while TRANSMITTING → stop_rx no-op → stop_tx reverts to RECEIVING). This is pre-existing teardown semantics identical to pre-#1735 production behaviour, not introduced by this PR — noting as a potential follow-up only.

## 3. Falsification

With `git checkout origin/main -- src/rigplane/audio/bridge.py`: `test_bridge_subscribes_rx_before_arming_tx` FAILS exactly as on live hardware — bus logs `RuntimeError: Cannot start RX in state transmitting` and `rx_active == False` (the bus swallows the error, matching the silent-dead-RX + transport queue-flood symptom). Restored to PR head: both new tests pass.

## 4. Gates (worktree at dbaa49e7)

- Full suite `--ignore=tests/integration`: **7369 passed, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed — 0 failures** (4:57)
- `tests/test_civ_command_profiling.py`: 8 passed in the full parallel run AND in isolation — the previously flagged p99 flake did not reproduce; it is a latency-distribution perf test, plausibly load-sensitive and unrelated to this change
- `ruff check`: clean · `ruff format --check`: 552 files already formatted
- `mypy src/`: 21 errors in 8 files — identical to baseline, none in `audio/bridge.py`, zero new
- Bridge/duplex files (test_audio_bridge.py, test_audio_bridge_stereo.py, test_audio_duplex.py, test_audio_fullduplex.py, test_hamlib_bridge.py, test_usb_duplex_mode.py): 94 passed, 2 skipped

Scope: 2 files, +105/−5, within guardrails; the bridge.py change is a pure reorder with an explanatory comment.

PR #1751, opened 2026-06-10
fix: project 0x27 scope-control answers into the StateStore — v2 scope toolbar (MOR-557)

morozsm:
## Root cause (URGENT — v2 scope toolbar dead on IC-7610)

Regression from the MOR-334 squash-merge (`e19d8988`), verified live on the IC-7610 on 2026-06-09:

- The RadioPoller's 0x27 scope-control fetch works and `_handle_27` (`src/rigplane/runtime/_civ_rx.py`) decodes the answers — but writes them **only** to the legacy `RadioState` mirror.
- The StateStore observation decoder `_observations_from_frame` had **no 0x27 branch** (`_CIV_STATE_FIELD_FAMILIES[0x27]` is diagnostics labeling only).
- Post-MOR-334, `build_public_state` (`web/server.py`) is snapshot-only, so all 9 `scopeControls.*` web fields stayed `availability: missing, observed: false`.
- The v2 toolbar's `disabled={!isFieldAvailable(...)}` gating therefore disabled every control: SPAN/SPEED/REF rendered "—" and clicks never sent commands.

## Fix

1. **`src/rigplane/runtime/_civ_rx.py`** — new `_scope_control_observations()` mirrors the `_handle_27` sub-command dispatch (subs `0x12/0x13/0x14/0x15/0x16/0x17/0x19/0x1A`) using the already-imported `parse_scope_*_response` helpers, emitting `Observation`s at `FieldPath.scope_control("display", <name>)` → store paths `scope_controls.global.display.{receiver,dual,mode,span,edge,hold,ref_db,speed}` (exactly the storePaths the web fieldStatus seeds reference). Wired as an `elif frame.command == 0x27 and frame.sub` branch in `_observations_from_frame`; sub `0x00` waterfall data never reaches it (`_route_civ_frame` short-circuits it) and echoes are dropped by the `from_addr` filter.
2. **`src/rigplane/web/runtime_helpers.py`** — `_snapshot_field_public_paths` and `_apply_snapshot_field` now cover **all eight** scope-control leaves (previously only `span` + receiver-scoped `receiver`), via an inverse of `_SCOPE_CONTROL_PUBLIC_FIELDS`, projecting to the public fields the frontend reads: `scopeControls.{span,speed,refDb,mode,hold,edge,dual,receiver}`. No frontend changes.
[...]

PR #1752, opened 2026-06-10
fix: drop never-observed scope_controls group fieldStatus (MOR-557 fix 2 — parent veto)

morozsm:
## Root cause

The first MOR-557 fix (#1751) made the eight `scopeControls.<leaf>` fieldStatus entries genuinely observable, but the v2 scope toolbar remained dead. `_default_snapshot_field_status` (`src/rigplane/web/runtime_helpers.py`) seeded a **group-level** `fieldStatus["scopeControls"]` entry — because `"scope_controls"` is a member of `_GLOBAL_SLOW_STATE_FIELDS` — keyed to the store path `global.slow_state.scope_controls`. No observation ever writes that path (the #1751 observations land under `scope_controls.global.display.*`), so the group entry was permanently `availability: missing, observed: false`. The frontend's MOR-429 parent-veto rule (`frontend/src/lib/state/field-status.ts`: an own `available` entry is overridden by a `missing`/`stale` ancestor) therefore resolved `isFieldAvailable("scopeControls.<leaf>")` to false for ALL 8 leaves, keeping every toolbar control disabled despite the leaves being observed and available. Verified live on IC-7610.

## Fix (backend-only, no dist rebuild)

Skip seeding the group-level entry for `"scope_controls"` in the `_GLOBAL_SLOW_STATE_FIELDS` loop. The eight per-leaf `scopeControls.*` entries seeded immediately after remain the real gate. Every other slow-state field is seeded byte-identically.

Consumer audit: no code in `src/`, `tests/`, or `frontend/src` gates on the bare `scopeControls` fieldStatus key — the only reader was the generic `parentAvailability` ancestor walk (the veto itself). Frontend synthetic tests that construct a group entry by hand still exercise the (intentionally kept) parent-veto semantics. The committed golden fixture `frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json` regenerated via its own test (group key removed); the frontend vitest suite consuming it passes (8/8).

## Falsification (red → green)

Two new tests in `tests/test_web_runtime_helpers.py`, both RED on `c95bf377` (main):

1. `test_default_field_status_has_no_scope_controls_group_entry` — default snapshot contains no bare `scopeControls` entry while all 8 leaves remain seeded `missing`. RED: `assert 'scopeControls' not in {...}` failed.
2. `test_observed_scope_control_leaves_survive_frontend_parent_veto` — replicates the frontend MOR-429 parent-availability walk over the real backend payload; after observations land, every leaf must resolve `available`. RED: `mode` resolved `missing` (parent veto), exactly the live symptom.

Both GREEN after the fix; no existing test needed modification.

## Gate results (worktree, after final formatting)

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7383 passed**, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` — All checks passed; `ruff format --check` — 552 files already formatted
- `uv run mypy src/` — 21 errors in 8 files (baseline, zero new)
- frontend `vitest run src/lib/state/__tests__/field-status.test.ts` — 8 passed

### DIFF
--- a/frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json
+++ b/frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json
@@ -593,12 +593,6 @@
     "observed": false,
     "storePath": "global.slow_state.scanning"
   },
-  "scopeControls": {
-    "availability": "missing",
-    "freshness": "unknown",
-    "observed": false,
-    "storePath": "global.slow_state.scope_controls"
-  },
   "scopeControls.dual": {
     "availability": "missing",
     "freshness": "unknown",
--- a/src/rigplane/web/runtime_helpers.py
+++ b/src/rigplane/web/runtime_helpers.py
@@ -407,6 +407,15 @@ def _default_snapshot_field_status(receiver_count: int) -> dict[str, dict[str, A
             FieldPath.global_("meters", name),
         )
     for name in _GLOBAL_SLOW_STATE_FIELDS:
+        if name == "scope_controls":
+            # No observation ever writes ``global.slow_state.scope_controls``
+            # (scope-control observations land under
+            # ``scope_controls.global.display.*``), so a group-level
+            # ``scopeControls`` entry would stay ``missing`` forever and the
+            # frontend MOR-429 parent-veto rule would disable every observed
+            # ``scopeControls.<leaf>`` (MOR-557). The eight per-leaf entries
+            # seeded below are the real gate.
+            continue
         _set_missing_field_status(
             statuses,
             _to_camel(name),
--- a/tests/test_web_runtime_helpers.py
+++ b/tests/test_web_runtime_helpers.py
@@ -19,7 +19,7 @@ from rigplane.core.state_pipeline_contracts import (
     Observation,
     SourceMetadata,
 )
-from rigplane.core.state_store import FreshnessClock, StateStore
+from rigplane.core.state_store import FreshnessClock, StateSnapshot, StateStore
 
 
 class _FakeWriter:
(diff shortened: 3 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of b34156b9 (base c95bf377, exactly 3 files, +79/−7).

**Falsification** — reverted `src/rigplane/web/runtime_helpers.py` to origin/main: exactly the 2 new tests fail, and `test_observed_scope_control_leaves_survive_frontend_parent_veto` fails with precisely the diagnosed symptom (observed leaf resolves `missing` via the dead `scopeControls` ancestor veto). Restored → 19/19 pass.

**Point 1 — MOR-429 replica fidelity** (vs `frontend/src/lib/state/field-status.ts`): the test's `_frontend_availability` faithfully reproduces the availability semantics — most-specific-first ancestor walk with first-entry-wins (matches `parentAvailability` lines 27-37), missing/non-available parent vetoes an available own entry (lines 54-58), explicit missing own entry is NOT rescued by an available parent, and no-own-entry inherits parent ?? 'available' (lines 47-51). One minor fidelity gap: the replica omits the freshness→'stale' mapping (real code lines 33 and 53). Irrelevant to the scenario under test (all entries fresh or unknown-missing), so not a blocker — just don't reuse the replica for stale-path assertions.

**Point 2 — absent-parent collateral**: in `parentAvailability`, an ancestor with NO fieldStatus entry is simply skipped (`if (status)` line 32); the walk exhausting returns `undefined` = no opinion, so the leaf's own `available` entry wins. Dropping the group entry is therefore the correct mechanism, and no frontend consumer reads the bare `scopeControls` key for availability — SpectrumToolbar gates only on the 8 `scopeControls.<leaf>` paths (SpectrumToolbar.svelte:113-120). Frontend vitest: targeted `src/lib/state` + `src/components/spectrum` 200/200, full suite 119 files / 2051 tests, all green with the regenerated fixture.

**Point 3 — skip scope**: exactly one member (`scope_controls`) skipped; all other `_GLOBAL_SLOW_STATE_FIELDS` seeded byte-identically (fixture diff is solely the 6-line bare-key removal). Note on the code comment's "no observation ever writes" claim: the legacy `sync_state_store_from_radio_state` bridge (server.py:1513, `_LEGACY_GLOBAL_SLOW_STATE_FIELDS` includes `scope_controls`) COULD write that path, but no production backend exercises the legacy `StatePollable` branch (yaesu_cat implements `ObservationPollable`, which takes precedence in web_startup.py:96), and even if one did, the observed entry is added dynamically — the skip only removes the permanently-missing default seed, so legacy behavior is unchanged. CI-V 0x27 answers land at `FieldPath.scope_control("display", ...)` (_civ_rx.py:1881), confirming the diagnosis for IC-7610.

**Lurking dead-parent family (follow-up, not a blocker)**: `main.vfoA` / `main.vfoB` / `sub.vfoA` / `sub.vfoB` group entries are seeded keyed to `receiver.<rx>.slow_state.vfo_a/vfo_b`, which no observation ever writes (the legacy receiver slow_state bridge only carries `contour`; VFO slot data lands at `FieldPath.vfo_slot(...)` → `main.vfoA.freqHz` leaves). Any future `isFieldAvailable` gate on a `*.vfoA.*` leaf would be vetoed identically. No frontend consumer gates on those paths today, so it is latent — recommend a follow-up issue.

**Point 4 — fixture legitimacy**: `tests/test_field_status_frontend_fixture.py` genuinely auto-regenerates (rewrites on mismatch, then asserts equality); full pytest passed with a clean tree, so the committed fixture equals the live projection.

**Full gate** (worktree at b34156b9): pytest 7383 passed / 9 skipped / 11 xfailed / 1 xpassed, zero failures · ruff check clean · ruff format 552 files clean · mypy 21 errors in 8 files (= baseline, zero new) · lint-imports 4 kept / 0 broken · frontend vitest 2051/2051.

OK to merge once required checks are green.

PR #1753, opened 2026-06-10
fix: branch bridge stream order on audio_duplex_mode — restore FTX-1 exclusive duplex (MOR-559)

morozsm:
## Root cause

[MOR-559](https://linear.app/morozsm/issue/MOR-559) (URGENT). The MOR-556 reorder (#1750) made `AudioBridge._setup_streams` subscribe the AudioBus — whose first subscriber triggers radio `start_rx` — BEFORE arming radio TX. Two backends have **conflicting order requirements**:

- **LAN (IC-7610)**: `LanAudioStream` supports RX-then-TX only — `start_tx` flips the single state field to TRANSMITTING and a later `start_rx` raises `"Cannot start RX in state transmitting"`. RX must be subscribed first (MOR-556, correct for LAN).
- **FTX-1 (same-device macOS C-Media USB CODEC, `audio_duplex_mode == "exclusive"`)**: the radio side uses two separate `sd` streams (`UsbAudioDriver.start_rx`/`start_tx`; `start_duplex` is not yet wired into the radio path). Per the MOR-531 live de-risk, the AUHAL `-50` trigger is specifically **"add TX playback to an already-running RX capture"** on one C-Media device — and it kills the capture **silently** (no Python exception), so the bridge still logged `started (RX+TX)` while RX was dead. The reverse direction (RX onto a running TX leg) is the MOR-531 live-validated path.

Live evidence (FTX-1, 2026-06-09, main @ `7f271fcf`): `bridge codec=... duplex_mode=exclusive` → `audio-bus: RX started` → `||PaMacCore (AUHAL)|| Error on line 2523: err='-50'` → `started (RX+TX, 48000Hz, 1ch, 20ms frames)` — but WS audio probe 0 frames, `/api/v1/audio/analysis` rms −96, scope 0 frames.

## Fix

Branch `_setup_streams` on `getattr(radio, "audio_duplex_mode", "full")` — the MOR-534/537 descriptor built exactly for this arbitration, consulted on the **radio** (FakeAudioBackend-compatible; the FTX-1/X6200 properties delegate to the driver policy and fall back to `"full"`):

- `"exclusive"` → pre-MOR-556 order restored for this case: radio TX-arm + `use_duplex` topology decision first, bus subscribe (radio RX start) after the device streams are up.
- anything else (`"full"`/missing — LAN, virtual loopbacks, separate devices, non-macOS) → MOR-556 RX-first order kept.

Capture-health surface (cheap, in-scope): `AudioBus._start_rx` swallows radio RX-start failures (`logger.exception`, `rx_active` stays False). The new `_subscribe_bus` helper raises when `bus.rx_active` is False after subscribe, so the bridge fails loudly instead of claiming "started" with dead capture. Note: the CoreAudio-level **silent** capture death (tonight's -50 mode) is not Python-visible at open time — runtime capture-health probing (e.g. zero-RX-frame watchdog) is noted as a follow-up.

## Falsification

- NEW `test_bridge_arms_tx_before_rx_on_exclusive_duplex` + `_ExclusiveUsbRadio` stub mirroring the real semantics (`audio_duplex_mode = "exclusive"`; TX open while RX capture runs silently kills the capture — the live -50 behavior, not a raise): **RED on main** (`AssertionError: assert ['start_rx', 'start_tx'] == ['start_tx', 'start_rx']`, capture dead) → **GREEN with the fix**.
- MOR-556 LAN regression test `test_bridge_subscribes_rx_before_arming_tx` (`_LanLikeRadio` state machine): **green before and after** — LAN keeps RX-first.
- NEW `test_bridge_rx_only_on_exclusive_duplex_still_starts_rx` guards the `--bridge-rx-only` path under the exclusive branch.

## Gate (all in worktree)

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7385 passed, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed**, zero failures
- `uv run ruff check src/ tests/` — All checks passed; `uv run ruff format --check` — 552 files already formatted
- `uv run mypy src/` — 21 errors in 8 files (exact baseline, **zero new**)

Diff: 2 files, +131/−8.

## Acceptance still pending (live)

Live FTX-1 full-bridge validation (no `--bridge-rx-only`: no -50, WS RX frames non-silent, scope flowing) and IC-7610 LAN re-check per MOR-559 acceptance — to be run after merge on the live server.

### DIFF
--- a/src/rigplane/audio/bridge.py
+++ b/src/rigplane/audio/bridge.py
@@ -501,14 +501,22 @@ class AudioBridge:
         else:
             self._decoder = None
 
-        # --- Subscribe to AudioBus BEFORE arming radio TX (MOR-556). The
-        # first subscriber triggers radio RX start, and the LAN stream state
-        # machine supports RX-then-TX only: arming TX first puts it in
-        # TRANSMITTING and the later RX start raises, leaving RX dead and the
-        # transport packet queue undrained (regression from #1735).
-        bus = self._radio.audio_bus
-        self._subscription = bus.subscribe(name="audio-bridge")
-        await self._subscription.start()
+        # --- Bus-subscribe vs radio-TX-arm order is backend-dependent
+        # (MOR-559); the first bus subscriber triggers radio RX start:
+        #
+        # - "full"/missing (LAN, MOR-556): subscribe BEFORE arming TX. The
+        #   LAN stream state machine supports RX-then-TX only — arming TX
+        #   first puts it in TRANSMITTING and the later RX start raises,
+        #   leaving RX dead and the packet queue undrained (#1735 regression).
+        # - "exclusive" (same-device macOS USB CODEC, MOR-531/534): arm TX
+        #   FIRST and subscribe only after the device streams are up. Adding
+        #   the TX leg to an already-running RX capture on one C-Media device
+        #   triggers CoreAudio AUHAL paramErr -50 and silently kills the
+        #   capture; the reverse order (RX onto a running TX leg) is the
+        #   MOR-531 live-validated path.
+        rx_first = getattr(self._radio, "audio_duplex_mode", "full") != "exclusive"
+        if rx_first:
+            await self._subscribe_bus()
 
         # --- Arm the radio TX path BEFORE choosing the device-stream topology.
         # The radio may reject TX (serial backend with no USB-audio TX path):
@@ -593,6 +601,11 @@ class AudioBridge:
             )
             await self._rx_stream.start()
 
+        # Exclusive same-device duplex: the radio TX leg is armed and the
+        # device streams are open — radio RX may start now (MOR-559).
+        if not rx_first:
+            await self._subscribe_bus()
+
         # --- TX path: device input → radio (capture) ---
         if tx_armed:
             self._tx_queue = asyncio.Queue(maxsize=64)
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS — Falsification confirmed: `test_bridge_arms_tx_before_rx_on_exclusive_duplex` is red on main's bridge.py (fails with `['start_rx', 'start_tx']` — the exact live bug signature) and green with the PR; the MOR-556 LAN guard `test_bridge_subscribes_rx_before_arming_tx` (stub without `audio_duplex_mode`, exercising the getattr-default "full" path) and the rx_only exclusive test both pass. The exclusive branch reproduces the pre-MOR-556 order exactly (verified against 0acfea3f^: TX arm → device streams → bus subscribe → duplex start), the IC-7610 LAN mixin hard-returns "full" so it can never take the exclusive branch, and all three `audio_duplex_mode` implementations are exception-safe with "full" fallbacks. The `rx_active` guard cannot false-positive: AudioBus sets `_rx_active` synchronously inside the awaited `subscription.start()` under the bus lock, so False-after-subscribe deterministically means a swallowed `_start_rx` failure; the reconnect loop tears down the subscription before retrying and the CLI catches the new RuntimeError. Full gate in a clean worktree: pytest 7385 passed / 0 failed, ruff check + format clean, mypy exactly at the 21-errors/8-files baseline (none in bridge.py), lint-imports 4 contracts kept / 0 broken.

PR #1755, opened 2026-06-10
fix(MOR-560): teardown bridge bus subscription on failed initial start

morozsm:
## Problem (MOR-560, step 1 of epic MOR-562)

`AudioBridge.start()` caught `_setup_streams` failures, set state IDLE, and re-raised **without teardown**. The rx-first path (MOR-556) subscribes to the AudioBus early in `_setup_streams`, so any later setup failure — device open, TX arm, `open_duplex`, or the `_subscribe_bus` `rx_active=False` RuntimeError itself — left the just-registered subscription orphaned on the bus. `stop()` early-returns on IDLE, so nothing ever removed it: the leaked subscriber kept the bus subscriber count > 0 and **radio RX running with no consumer draining the queue**.

The reconnect loop is NOT affected (it calls `_teardown_streams` before each retry); only the initial-start failure path leaked.

## Fix

In `start()`'s failure path, run the existing `_teardown_streams()` (which closes the early bus subscription via `aclose()`, triggering last-subscriber RX stop on the bus) before re-raising. The teardown call is guarded so a teardown error cannot mask the original exception (callers distinguish e.g. `LoopbackNotFoundError` in auto-bridge mode). No changes to the reconnect loop or the success path; no new abstractions.

## Red → green (falsification-first TDD)

Two regression tests written and confirmed **red** on the pre-fix code, both failing on the leak assertion:

```
assert radio.audio_bus.subscriber_count == 0
E   assert 1 == 0
```

- `test_bridge_failed_rx_start_releases_bus_subscription` — radio stub whose `start_rx` raises; the bus swallows it (`rx_active` stays False) and `_subscribe_bus` raises after the subscription is registered.
- `test_bridge_failed_stream_open_releases_bus_subscription` — `FakeAudioBackend` subclass whose `open_tx` raises after a successful subscribe; additionally asserts `rx_active is False` and `stop_audio_rx_opus` awaited once (last-subscriber removal stopped radio RX).

Both green after the fix. The MOR-556 (`test_bridge_subscribes_rx_before_arming_tx`) and MOR-559 (`test_bridge_arms_tx_before_rx_on_exclusive_duplex`, `test_bridge_rx_only_on_exclusive_duplex_still_starts_rx`) regressions still pass.

## Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7387 passed**, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` → clean; `ruff format --check` → 552 files already formatted
- `uv run mypy src/` → 21 errors in 8 files (exact baseline, no new errors)

Diff: 2 files, +94/-0 (`src/rigplane/audio/bridge.py` +13, `tests/test_audio_bridge.py` +81).

### DIFF
--- a/src/rigplane/audio/bridge.py
+++ b/src/rigplane/audio/bridge.py
@@ -787,6 +787,19 @@ class AudioBridge:
         try:
             await self._setup_streams()
         except Exception:
+            # _setup_streams may fail AFTER the bus subscription is registered
+            # (the rx-first path subscribes early, MOR-556). Tear down before
+            # re-raising — stop() early-returns on IDLE, so a leaked
+            # subscriber would keep radio RX running with no consumer
+            # draining the queue (MOR-560).
+            try:
+                await self._teardown_streams()
+            except Exception:
+                logger.debug(
+                    "%s: teardown after failed start error",
+                    self._label,
+                    exc_info=True,
+                )
             self._set_state(BridgeState.IDLE, "start_failed")
             raise
 
--- a/tests/test_audio_bridge.py
+++ b/tests/test_audio_bridge.py
@@ -489,6 +489,87 @@ async def test_bridge_rx_only_on_exclusive_duplex_still_starts_rx():
 
 
 # ---------------------------------------------------------------------------
+# Failed initial start must release the bus subscription (MOR-560)
+# ---------------------------------------------------------------------------
+
+
+class _RxStartFailRadio:
+    """Radio stub whose RX start fails.
+
+    The AudioBus swallows the ``start_rx`` error (logs "audio-bus: failed to
+    start RX") and leaves ``rx_active`` False — ``_subscribe_bus`` then raises
+    AFTER the subscription is already registered with the bus.
+    """
+
+    def __init__(self) -> None:
+        from rigplane.audio_bus import AudioBus
+
+        self.audio_bus = AudioBus(self)
+
+    async def start_rx(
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: PASS — Falsification confirmed: with origin/main's bridge.py both new tests fail exactly at the leaked subscriber (`assert 1 == 0` in both `test_bridge_failed_rx_start_releases_bus_subscription` and `test_bridge_failed_stream_open_releases_bus_subscription`); restoring the PR bridge.py turns the file green (47/47). Full gate on pr-1755 (7cc001f8): pytest 7387 passed / 0 failed (9 skipped, 11 xfailed, 1 xpassed), ruff check + format clean, mypy exactly at baseline (21 errors / 8 files), lint-imports 4 kept / 0 broken. Code review: teardown in the except path is guarded by its own try/except (debug-logged, original exception re-raised via bare raise; CancelledError not swallowed), `_teardown_streams` is None-guarded per resource and idempotent for partial setup (LoopbackNotFoundError pre-subscribe path is a full no-op), success path and `_reconnect_loop` are byte-identical to main. Scope is 2 files / +94 LOC as claimed. Non-blocking note: a CancelledError raised inside `_setup_streams` still bypasses the new teardown (outer except is `Exception`), but state then remains CONNECTING, so `stop()` does not early-return and recovers the subscription — pre-existing behavior, out of MOR-560 scope.

PR #1756, opened 2026-06-10
feat(MOR-564): AudioBus RX heartbeat exposed in runtime payload

morozsm:
## Goal

MOR-564 (step 3 of epic MOR-562, RX health model): make RX liveness observable. Stamp a monotonic timestamp on every RX frame fan-out in `AudioBus` and expose it — plus the already-existing per-subscriber stats — in the web runtime payload (`GET /api/v1/runtime`).

**Additive observability only.** No watchdog, no behavior change, no new fields renamed or removed. This is the first brick of the health model; the watchdog comes in a later step.

## Changes

- `src/rigplane/audio/bus.py`
  - `AudioBus._last_rx_frame_monotonic: float | None`, stamped via a single `time.monotonic()` assignment at the top of `_on_opus_packet` (RX hot path — no locking, no logging).
  - Read-only property `last_rx_frame_monotonic`; also included in `AudioBus.stats` as `last_rx_frame_monotonic`.
- `src/rigplane/web/server.py`
  - New `_runtime_audio_bus_payload()` next to `_runtime_bridge_payload()`, wired into `_serve_runtime` as a new top-level `"audioBus"` key next to `"bridge"`:
    ```json
    "audioBus": {
      "enabled": true,
      "lastRxFrameMonotonic": 12345.678,
      "stats": {"rx_active": true, "subscriber_count": 1, "last_rx_frame_monotonic": 12345.678, "subscribers": [{"name": "...", "active": true, "received": N, "dropped": N, "queued": N}]}
    }
    ```
    `{"enabled": false}` when no bus exists. The helper reads the private `_audio_bus` slot (same pattern as the `_acquisition_scheduler` diagnostics read) so a runtime poll never lazily instantiates the bus as a side effect.
- `tests/test_audio_bus_heartbeat.py` — new falsification tests.

## Red → Green (falsification-first)

Tests written first; confirmed RED on the unmodified tree:
- `AttributeError: 'AudioBus' object has no attribute 'last_rx_frame_monotonic'` (x2)
- `KeyError: 'audioBus'` in the runtime payload (x2)

After implementation, all 4 GREEN: stamp is `None` before any packet; stamps within a `[before, after]` monotonic window and advances on a second packet; runtime payload exposes `audioBus.lastRxFrameMonotonic` + per-subscriber `received`/`dropped`; `{"enabled": false}` without a bus.

## Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → 7389 passed, 9 skipped, 3 deselected, 11 xfailed, 1 xpassed, **0 failures**
- `uv run ruff check src/ tests/` → clean; `ruff format --check` → 553 files already formatted
- `uv run mypy src/` → exactly baseline: 21 errors in 8 files (no new errors)
- `uv run lint-imports` → 4 kept, 0 broken

### DIFF
--- a/src/rigplane/audio/bus.py
+++ b/src/rigplane/audio/bus.py
@@ -41,6 +41,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import time
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
@@ -264,6 +265,7 @@ class AudioBus:
         self._subscribers: list[AudioSubscription] = []
         self._rx_active = False
         self._lock = asyncio.Lock()
+        self._last_rx_frame_monotonic: float | None = None
 
     @property
     def subscriber_count(self) -> int:
@@ -274,10 +276,19 @@ class AudioBus:
         return self._rx_active
 
     @property
+    def last_rx_frame_monotonic(self) -> float | None:
+        """``time.monotonic()`` of the most recent RX fan-out, or None.
+
+        RX liveness heartbeat (MOR-564): observability only, no watchdog.
+        """
+        return self._last_rx_frame_monotonic
+
+    @property
     def stats(self) -> dict[str, Any]:
         return {
             "rx_active": self._rx_active,
             "subscriber_count": len(self._subscribers),
+            "last_rx_frame_monotonic": self._last_rx_frame_monotonic,
             "subscribers": [s.stats for s in self._subscribers],
         }
 
@@ -291,6 +302,7 @@ class AudioBus:
 
     def _on_opus_packet(self, packet: "AudioPacket | None") -> None:
         """Internal callback — distributes packet to all active subscribers."""
+        self._last_rx_frame_monotonic = time.monotonic()
         for sub in self._subscribers:
             sub.deliver(packet)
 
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS — Falsification confirmed: with main's bus.py/server.py under the PR's new test file, all 4 tests fail exactly as expected (AttributeError: last_rx_frame_monotonic, KeyError: 'audioBus'); restoring PR sources turns them green. Full gate on head bbc2bfc4: pytest 7389 passed / 0 failed (9 skipped, 11 xfailed, 1 xpassed), ruff check + format clean, mypy exactly at baseline (21 errors in 8 files), lint-imports 4 kept / 0 broken. Key risks verified: the hot-path addition in _on_opus_packet is a single time.monotonic() float assignment (no lock, no logging, no allocation, no exception path); the runtime payload reads the private _audio_bus slot via getattr (same pattern as _acquisition_scheduler) so GET /runtime never lazily constructs the bus — and both IcomRadio and the yaesu_cat backend use that slot name; JSON is additive-only (new top-level audioBus key, {"enabled": false} when no bus, no existing fields renamed/removed); the property is read-only and the value stays in the monotonic domain with no consumer treating it as epoch time. Non-blocking note: a None flush packet also stamps the heartbeat — acceptable for an observability-only liveness signal.

Issue #295, opened 2026-03-17
epic: Standardize command naming — TOML as single source of truth


morozsm:
## Summary

Standardize command naming across all layers — TOML profiles become the **single source of truth** for command names. Every layer (commands.py, radio.py, web handlers) must use the same canonical names defined in TOML.

Currently there are **115 naming mismatches** across 5 categories:
- 32 TOML overrides missing `get_`/`set_` prefix
- 21 commands.py functions with different names than TOML
- 37 radio.py methods with different names than TOML
- 23 web handler commands with different names than TOML
- 2 asymmetric get/set pairs (missing `get_squelch`)

## Motivation

Today, `grep "get_rf_power"` only finds the TOML definition — commands.py calls it `get_power`, radio.py calls it `get_power`, web UI calls it `set_power`. This makes cross-layer debugging painful, and new contributors must learn an implicit mapping between layers.

**After this work:**
- `grep "get_rf_power"` finds TOML → commands.py → radio.py → web handlers → tests
- Adding a new command = adding it to TOML + implementing with the **same name** everywhere
- A CI validation test enforces that every TOML command name exists in commands.py and radio.py

## Design Principle

**TOML = canonical name.** All other layers conform to TOML.

Exception: **compound commands** where radio.py wraps multiple CI-V commands into a single higher-level method. These are documented in an explicit `_COMPOUND_COMMANDS` registry and excluded from the naming parity test:

| TOML names | radio.py method | Reason |
|---|---|---|
| `ptt_on`, `ptt_off` | `set_ptt(on: bool)` | bool toggle |
| `power_on`, `power_off` | `set_powerstat(on: bool)` | bool toggle |
| `scope_on`, `scope_off`, `scope_data_output` | `enable_scope()` / `disable_scope()` | orchestrated |
| `set_dual_watch_on`, `set_dual_watch_off` | `set_dual_watch(on: bool)` | bool toggle |

### Phase 1: TOML Standardization → sub-issue
- Add `get_`/`set_` prefixes to all `[commands.overrides]` entries
- Affects: ic7610.toml (18 entries), ic7300.toml (14 entries)
- No code changes, only TOML files

### Phase 2: commands.py Renames → sub-issue
- Rename ~15 functions to match TOML names
- Add missing functions (`get_squelch`, `get_manual_notch_width`)
- Update all internal callers

### Phase 3: radio.py Renames → sub-issue
- Rename ~20 methods to match TOML names
[...]

PR #1760, opened 2026-06-10
test(MOR-567): audio lifecycle conformance suite across all backends

morozsm:
## Summary

Capstone of the MOR-562 de-risking tranche (ADR §3.9.2). Adds `tests/contracts/test_audio_lifecycle_conformance.py` — a lifecycle scenario matrix run for **every shipping backend against its fake driver/link**, built so a reintroduced MOR-556/MOR-559-style stream start-order bug fails the suite on the affected transport. Tests-only: **zero `src/` changes.**

## Scenario × backend matrix

| Scenario | lan-icom (real `LanAudioStream`) | icom7610 / ic705¹ / ic7300 / ic9700 serial (real `UsbAudioDriver` + strict `FakeAudioBackend`) | yaesu-ftx1 (same driver pair) | lan-graph-stub | exclusive-graph-stub |
|---|---|---|---|---|---|
| `rx_then_tx_then_ptt_cycle_then_stop` (2 PTT cycles, ends clean: 0 bus subscribers, 0 open streams) | ✅ | ✅ ×4 | ✅ | — | — |
| `tx_while_rx` (TX arm must NOT kill running RX) | ✅ | ✅ ×4 | ✅ | — | — |
| `rx_while_tx` (LAN: typed `AudioAlreadyStartedError` from TRANSMITTING; independent-leg transports: clean late RX) | ✅ raises | ✅ clean ×4 | ✅ clean | — | — |
| `double_start` (RX and TX: typed `AudioAlreadyStartedError`, never bare `RuntimeError`) | ✅ | ✅ ×4 | ✅ | — | — |
| `start_failure_cleanup` (MOR-560 generalized: failed RX start leaks no bus subscription/streams) | ✅ | ✅ ×4 | ✅ | — | — |
| bridge round-trip honoring `audio_duplex_mode` (RX alive + TX armed after `bridge.start()`, clean after `stop()`) | ⚠️ xfail² | ✅ ×4 | ✅ | ✅ | ✅ |
| demand-arrival permutation convergence (graph encoding: both orders converge or raise consistently) | — | — | — | ✅ rx/tx | ✅ rx/tx |

¹ X6200 ships through the IC-705 serial class.
² Known shipping bug surfaced by this suite — see below.

**Covered with real code, no hardware:** LAN `IcomRadio` drives the REAL `LanAudioStream` state machine (the actual MOR-556 culprit) over a fake UDP transport; the serial family and FTX-1 drive the REAL `UsbAudioDriver` over `FakeAudioBackend(strict_device_exclusive=True)` (MOR-566) with separate RX/TX devices (deterministic `"full"` duplex on every host platform).

**Skipped-with-reason:** the FTX-1 *same-device* exclusive path cannot run against the strict fake — live CoreAudio kills asymmetrically (RX onto running TX is clean; TX onto running RX dies with AUHAL −50) while `strict_device_exclusive` rejects any second stream symmetrically. That graph is exercised via the shared `ExclusiveUsbRadio` stub (declared from the MOR-531 live de-risk), including through the real `AudioBridge`.

## Proof the suite bites (experiments run locally, then reverted — committed tree is green)

1. **MOR-556 revert** — `bridge.py`: replaced `rx_first = getattr(radio, "audio_duplex_mode", "full") != "exclusive"` with `rx_first = False` (pre-#1735-fix order: TX armed before bus RX subscribe) → **2 failures**: `test_bridge_roundtrip_respects_declared_setup_order[lan-icom]` and `..._on_declared_transition_graphs[LanLikeRadio]`, both with the live symptom (`RuntimeError: radio RX failed to start for the bridge subscription` ← `AudioAlreadyStartedError: Cannot start RX in state transmitting`).
2. **MOR-559 revert** — same line set to `rx_first = True` (exclusive branch removed) → **1 failure**: `..._on_declared_transition_graphs[ExclusiveUsbRadio]` with `bridge start left radio RX capture dead (MOR-556/559)` — the silent −50 RX kill exactly as observed live.
3. **MOR-560 revert** — removed the `_teardown_streams()` call from `AudioBridge.start()`'s failure path → **6 failures**: `test_start_failure_leaves_no_leaked_subscription` on ALL six shipping-backend rows (`subscriber_count == 1`, the leaked bus subscription).

Each experiment was applied, the suite run, the failure observed, and `git checkout src/rigplane/audio/bridge.py` restored the tree.

## Real bug surfaced (encoded as xfail, NOT fixed here)

`AudioBridge.stop()` tears down the bus subscription (RX demand) **before** stopping radio TX, and `LanAudioStream.stop_rx` early-returns while state is TRANSMITTING — so on the LAN backend a bridge stop with TX armed leaks the running RX stream (final state RECEIVING, live `_rx_task`, registered callback, zero bus subscribers). Stop-side instance of the MOR-556 ordering class; the existing `LanLikeRadio` stub masks it because its `stop_rx` clears the callback unconditionally, unlike the real stream. The LAN bridge round-trip row asserts the leak explicitly and `pytest.xfail`s with the reason; needs a follow-up issue.

## Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7451 passed**, 9 skipped, 3 deselected, 12 xfailed, 1 xpassed (xpass pre-existing/platform-specific; the new xfail is imperative and cannot xpass)
- `uv run ruff check src/ tests/` → clean; `ruff format --check` → 556 files already formatted
- `uv run mypy src/` → baseline **21 errors in 8 files** (unchanged)
- `uv run lint-imports` → **4 kept, 0 broken**

New suite alone: 41 passed, 1 xfailed. 1 file, +396 lines (≈260 non-docstring logic lines, of which ~17 imports and ~95 backend-row/harness table).

### DIFF
--- a/tests/contracts/test_audio_lifecycle_conformance.py
+++ b/tests/contracts/test_audio_lifecycle_conformance.py
@@ -1,0 +1,396 @@
+"""Audio lifecycle conformance suite across all shipping backends (MOR-567).
+
+ADR §3.9.2, capstone of the MOR-562 de-risking tranche: one scenario
+matrix, run for every shipping transport against its fake driver/link,
+that CATCHES the MOR-556/MOR-559 start-order bug class instead of
+pinning a single happy path.
+
+Backend rows (all hardware-free):
+
+- ``lan-icom``        — LAN :class:`IcomRadio` driving the REAL
+  :class:`~rigplane.audio.lan_stream.AudioStream` state machine (the
+  actual MOR-556 culprit) over a fake UDP transport.
+- ``icom7610/ic705/ic7300/ic9700-serial`` — the real serial classes
+  (X6200 ships via the IC-705 class) over ``_FakeSerialCivLink`` with
+  the REAL :class:`UsbAudioDriver` on a strict-exclusive
+  :class:`FakeAudioBackend` (separate RX/TX devices → ``"full"`` duplex
+  on every host platform, deterministic).
+- ``yaesu-ftx1``      — :class:`YaesuCatRadio` with the same real
+  driver/backend pair and a patched CAT transport.
+- ``lan-graph-stub`` / ``exclusive-graph-stub`` — the shared
+  order-sensitive stubs (MOR-566) with declared transition graphs;
+  bridge round-trip rows only.
+
+Documented gap: the FTX-1 *same-device* exclusive path cannot run
+against the strict fake — live CoreAudio kills asymmetrically (RX onto
+running TX is clean, TX onto running RX dies with AUHAL -50) while
+``strict_device_exclusive`` rejects any second stream symmetrically.
+That graph is therefore exercised via ``ExclusiveUsbRadio`` (declared
+from the MOR-531 live de-risk), not a real backend row.
+
+Known shipping bug surfaced by this suite (kept as a strict xfail, not
+fixed here — MOR-567 is tests-only): ``AudioBridge.stop()`` drops the
+RX demand BEFORE stopping radio TX, and ``LanAudioStream.stop_rx``
+early-returns while TRANSMITTING — so on the LAN backend a bridge stop
+with TX armed leaks the running RX stream (state ends RECEIVING with a
+live ``_rx_task`` and zero bus subscribers). Stop-side instance of the
+MOR-556 ordering class.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from dataclasses import dataclass, field
+from typing import Any, Awaitable, Callable
+
+import pytest
+from _order_sensitive_radios import ExclusiveUsbRadio, LanLikeRadio
+from test_icom7610_serial_radio import _FakeSerialCivLink
+
+from rigplane.audio.backend import AudioDeviceId, AudioDeviceInfo, FakeAudioBackend
+from rigplane.audio.bridge import AudioBridge, BridgeState
+from rigplane.audio.lan_stream import AudioState, AudioStream
+from rigplane.audio.usb_driver import AudioAlreadyStartedError, UsbAudioDriver
+from rigplane.backends.ic705 import Ic705SerialRadio
+from rigplane.backends.ic7300 import Ic7300SerialRadio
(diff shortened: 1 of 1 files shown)


morozsm:
Agent Review: PASS — Independent falsification-first review on a throwaway worktree (pr-1760 @ c99ea3f1, parent = main f89aa4d1). Suite-is-not-vacuous independently reproduced via all three bite experiments: MOR-560 revert (removed `_teardown_streams()` from `AudioBridge.start()` failure path) → exactly 6 failures (subscriber leak on every backend row); MOR-559 revert (`rx_first = True` always) → exactly 1 failure (ExclusiveUsbRadio silent RX kill); MOR-556 revert (tx-first always) → exactly 2 failures (real lan-icom row + LanLikeRadio stub) — all matching the implementer's claimed numbers, experiments reverted, tree pristine. The single xfail documents a REAL shipping bug verified by code reading: `AudioBridge.stop()` closes the bus subscription (dropping RX demand) before `stop_tx`, `LanAudioStream.stop_rx` early-returns while TRANSMITTING (lan_stream.py:536), and `stop_tx` then reverts to RECEIVING with a live `_rx_task` and zero subscribers; the xfail is imperative (`pytest.xfail` behind an assert of the buggy state, bridge round-trip lan row) so it can never silently xpass and will fail loudly once the bug is fixed — correct tests-only disposition. Scope confirmed tests-only (one file, +396/-0, zero src changes); over the 200-LOC soft cap but justified as parametrization across 6 real-backend rows (real LanAudioStream over fake UDP, real UsbAudioDriver over strict-exclusive FakeAudioBackend) + 2 declared-graph stubs, with the FTX-1 same-device gap clearly documented. Full gate: pytest 7451 passed / 0 failed / 12 xfailed / 1 xpassed (xpass is pre-existing test_naming_parity epic-#295, not this file; new file standalone = 41 passed + 1 xfailed + 0 xpass); ruff check + format clean; mypy exactly 21 errors in 8 files (baseline, no src drift); lint-imports 4 kept / 0 broken.

PR #1761, opened 2026-06-10
feat(MOR-575): audio_setup_order descriptor derived from audio_duplex_mode

morozsm:
## Summary

MOR-562 epic step 7 (ADR §3.3): add the **additive** `audio_setup_order -> Literal["rx_first", "tx_first", "atomic"]` property to the three AudioTransport-shaped backend bases, derived from each backend's own `audio_duplex_mode` so the two descriptors share a single source of truth and can never drift. Behavior is byte-identical — **nothing consumes the descriptor yet** (the bridge keeps its own `rx_first` branch until steps 8/9).

## Derivation (reproduces current semantics exactly)

| `audio_duplex_mode` | `audio_setup_order` | Why |
|---|---|---|
| `"full"` | `"rx_first"` | LAN UDP stream reverts RX from TRANSMITTING — RX must be armed before TX; USB separate-devices "full" is order-indifferent |
| `"exclusive"` | `"atomic"` | Same-device macOS: one duplex stream — setup does not decompose into rx/tx-first |
| `"half"` (or anything unexpected/raising) | `"rx_first"` | Safe default, same robustness `audio_duplex_mode` has toward the driver |

## Where

- `src/rigplane/runtime/_audio_runtime_mixin.py` — LAN `IcomRadio` (duplex hard-`"full"` → `"rx_first"`)
- `src/rigplane/backends/_icom_serial_base.py` — IC-705/7300/9700/7610-serial + X6200 family
- `src/rigplane/backends/yaesu_cat/radio.py` — FTX-1 (the `"exclusive"`-capable backend)

Each property mirrors the exact style/placement of the adjacent `audio_duplex_mode` property, including the try/except-with-default degrade.

## Deliberately NOT on the Protocol

`audio_setup_order` is duck-typed — it is NOT added to the `AudioTransport` Protocol in `core/radio_protocol.py`, so the frozen 10-member conformance pin is untouched. Consumers will read it via `getattr(radio, "audio_setup_order", "rx_first")` — exactly how the bridge reads `audio_duplex_mode` today. A new pin (`test_setup_order_not_on_audio_transport_protocol`) guards that the Protocol member set did not grow.

## TDD (red → green)

Pins written first in `tests/contracts/test_audio_transport_conformance.py` (section 6): 14 new tests failed with `AttributeError: ... has no attribute 'audio_setup_order'` (16 pre-existing green), then the three properties were added → all 30 pass. Covers: presence on all 6 shipping backend classes, per-backend duplex→setup-order consistency (LAN, serial, Yaesu × full/exclusive/half), raising-driver degrade to `"rx_first"`, and the Protocol-pin-not-grown guard.

(Pins live in `test_audio_transport_conformance.py`, not `test_audio_lifecycle_conformance.py`, to stay clear of the parallel MOR-574 PR.)

## Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7466 passed**, 9 skipped, 3 deselected, 12 xfailed, 1 xpassed, 0 failures
- `uv run ruff check src/ tests/` — clean; `ruff format --check` — 556 files already formatted
- `uv run mypy src/` — baseline preserved: exactly **21 errors in 8 files**
- `uv run lint-imports` — **4 kept, 0 broken**

No bridge changes, no new abstractions. 3 src files + 1 test file.

### DIFF
--- a/src/rigplane/backends/_icom_serial_base.py
+++ b/src/rigplane/backends/_icom_serial_base.py
@@ -11,7 +11,7 @@ import asyncio
 import logging
 import os
 import time
-from typing import TYPE_CHECKING, Callable, Protocol
+from typing import TYPE_CHECKING, Callable, Literal, Protocol
 
 from .._connection_state import RadioConnectionState
 from ..audio import AudioPacket
@@ -207,6 +207,29 @@ class _IcomSerialRadioBase(CoreRadio):
         except Exception:
             return "full"
         return mode if isinstance(mode, str) else "full"
+
+    @property
+    def audio_setup_order(self) -> Literal["rx_first", "tx_first", "atomic"]:
+        """Setup ordering descriptor (MOR-575, ADR §3.3).
+
+        Derived from :attr:`audio_duplex_mode` — single source of truth,
+        so the two descriptors never drift: ``"exclusive"`` (same-device
+        macOS: one duplex stream, setup does not decompose into
+        rx/tx-first) → ``"atomic"``; ``"full"`` (separate RX/TX devices,
+        order-indifferent) → ``"rx_first"``; ``"half"`` or any
+        unexpected/raising duplex mode degrades to the ``"rx_first"``
+        safe default — the same robustness ``audio_duplex_mode`` has
+        toward the driver. Nothing consumes this yet — the AudioSession
+        (MOR-562 step 8) and bridge (step 9) will read it.
+        """
+        try:
+            mode = self.audio_duplex_mode
+        except Exception:
+            return "rx_first"
+        if mode == "exclusive":
+            return "atomic"
+        # "full", "half", and anything unexpected → rx_first (safe default).
+        return "rx_first"
 
     # ------------------------------------------------------------------
     # Connection properties
--- a/src/rigplane/backends/yaesu_cat/radio.py
+++ b/src/rigplane/backends/yaesu_cat/radio.py
@@ -351,6 +351,29 @@ class YaesuCatRadio:
         return "full"
 
     @property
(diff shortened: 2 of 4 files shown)


morozsm:
Agent Review: PASS — Independent falsification-first review of 34f06a01 (based on current main 7bfd2390). Falsification: with main's 3 backend files restored over the PR, the 14 new section-6 pins fail with AttributeError (no audio_setup_order); restoring the PR files turns all 15 green. Derivation is correct and single-sourced: each property reads self.audio_duplex_mode (the exact attribute bridge.py:517 reads today) with the same try/except-degrade robustness; mapping full/half/raising→rx_first, exclusive→atomic matches ADR §3.3 verbatim (exclusive same-device = single duplex stream via UsbAudioDriver.ensure — setup does not decompose into rx/tx-first, and ADR step 12 treats tx_first/atomic identically for post-TX re-arm), so 'atomic' is the right name for the exclusive case, not a behavior change in waiting. Scope verified: exactly 3 backend files + transport conformance test (+160/−4); bridge.py untouched (rx_first branch intact), core/radio_protocol.py untouched (10-member AUDIO_TRANSPORT_MEMBERS pin frozen, guarded by test_setup_order_not_on_audio_transport_protocol), test_audio_lifecycle_conformance.py untouched (MOR-574's file); grep confirms nothing in src consumes audio_setup_order — only the 3 property definitions. Full gate in a clean worktree: pytest 7466 passed / 0 failed (9 skipped, 12 xfailed, 1 xpassed), ruff check + format clean, mypy exactly 21 errors in 8 files (baseline, no new), lint-imports 4 kept / 0 broken.

Issue #765, opened 2026-04-17
audio: report real frame_ms + dead-code removal (AudioBufferPool + PCM-8 mapping) + docstring


morozsm:
Consolidated replacement for #758, #759, #760, #761, #763 after re-review (\`.claude/workflow/audio-arch-re-review.md\`). Part of epic #764.

## Goal

Single PR that lands the **Option B** fix for the 2026-04-16 companion crash, adds a regression-guard test, and removes the speculative dead code that gave cover to the original mis-diagnosis. Net line **reduction**.

### Production (3 files, ~net −40 LOC)

1. **\`src/icom_lan/web/handlers/audio.py\`**:
   - Line ~305: replace hardcoded \`frame_ms=20\` in \`encode_audio_frame(...)\` with a payload-derived value:
     \`\`\`python
     bytes_per_sample = 2
     frame_ms_hdr = (len(audio_data) * 1000) // (self._sample_rate * self._channels * bytes_per_sample)
     frame_ms_hdr = max(1, min(frame_ms_hdr, 255))  # uint8 clamp
     \`\`\`
   - Lines ~89–93: remove the \`AudioBufferPool\` construction (dead code; never acquired on the hot path).
   - Lines ~213–214: remove the \`PCM_1CH_8BIT / PCM_2CH_8BIT → AUDIO_CODEC_PCM16\` mapping (no backend negotiates 8-bit; \"upcast in future\" never happens; loud failure on future PCM-8 backend beats silent wrong-label).
   - Drop the \`AudioBufferPool\` import + any now-unused references.

2. **\`src/icom_lan/_audio_buffer_pool.py\`**: delete the file (never referenced outside the unused instantiation).

3. **\`src/icom_lan/env_config.py\`**: remove \`get_audio_buffer_pool_size\` + its export.

### Docstring (1 file, ~4 LOC)

4. **\`src/icom_lan/web/protocol.py\`** line ~99: update the \`frame_ms\` docstring to mark it advisory — \"Frame duration in ms; derived from payload size on emit and NOT authoritative. Consumers should compute \`len(payload) * 1000 / (sample_rate * channels * bytes_per_sample)\` directly for any buffer-size decisions.\"

### Tests (1 file, ~+30 LOC)

5. **\`tests/test_web_audio_streaming_profile.py\`** (or similar): add a round-trip invariant test
   \`\`\`python
[...]

Issue #762, opened 2026-04-17
audio: decode Opus before DSP + tap dispatch (or document the gate)


morozsm:
Parent epic: #764

---

## Problem

The DSP pipeline and the FFT-scope tap registry are both gated on `web_codec == AUDIO_CODEC_PCM16`:

- `src/icom_lan/web/handlers/audio.py:284-291` — DSP pipeline skipped when `web_codec` is Opus.
- `src/icom_lan/web/handlers/audio.py:296` — tap fan-out (FFT scope, analyzers) skipped when `web_codec` is Opus.

For radios whose native codec maps to `AUDIO_CODEC_OPUS` (currently IC-705 and anything using `AudioCodec.OPUS_1CH` / `AudioCodec.OPUS_2CH` per `_CODEC_MAP` at `audio.py:209-210`), this silently means:

- Noise gate, limiter, and any configured DSP nodes don't run on RX.
- The waterfall / audio FFT scope panels in the web frontend have no data.
- Audio analyzer signals (SNR, peak, RMS) are not computed.

This is a feature gap, not a bug — nothing observably breaks — but it is undocumented and asymmetric across radios.

## Proposed fix

Pick one — **option A preferred if an Opus decoder is already instantiated for transcoding**:

### Option A: decode Opus once in the relay, run full pipeline

In `_relay_loop`, when `web_codec == AUDIO_CODEC_OPUS`:
1. Decode the Opus payload to PCM16 once.
2. Run DSP, feed tap registry (exactly as the PCM branch already does).
3. Re-encode to Opus before `encode_audio_frame` — OR emit the decoded PCM on the wire instead (changes wire behavior, so no).
4. Ship the (re-)encoded Opus frame as before.

Cost: one decode + one re-encode per 20 ms frame. `_audio_transcoder.PcmOpusTranscoder` is already the class for this and may be reusable.

Trade-off: re-encoding loses some quality vs. pass-through. Negligible for AM/FM ham audio but worth noting.

### Option B: decode Opus to PCM for DSP/tap only, still ship original Opus

Decode to a throwaway PCM buffer just for DSP (to apply, then discard) and tap feed. Ship the **original** Opus frame on the wire. Zero quality loss, but DSP is effectively read-only — noise gate / limiter decisions do not affect the audible output, only the scope/analyzer data.

Semantically weaker: users enabling DSP would expect it to affect their audio.

### Option C: document the gate and deprecate DSP/scope for Opus radios

[...]

PR #1771, opened 2026-06-10
feat(MOR-584): per-connection web egress codecs (per-client encoder pool)

morozsm:
## Summary

Implements Linear **MOR-584** (step 17 of epic MOR-562, ADR §3.6 problem P11): web egress encoding is now **per-connection** instead of one aggregate codec for all browser clients.

### Problem

The relay built ONE shared frame for all WS clients from an aggregate `_web_codec` — `_resolve_web_rx_codec` contained `if AUDIO_CODEC_PCM16 in self._client_rx_codec.values(): return AUDIO_CODEC_PCM16`, so a single PCM16-preferring client forced PCM16 for everyone, and an explicit Opus request was never honored. This coupled codec choice to the client population and blocked the WAN/Opus story.

### Per-client encoder pool design

Position in the pipeline (unchanged spine, new fan-out):

```
ulaw decode → DSP → tap fan-out (FFT scope) → frame_ms derive
                                   │  (PCM spine — untouched)
                                   ▼
              per-client egress encode + per-client queue fan-out   ← MOR-584
```

- The post-DSP/tap **PCM frame is produced once**; at fan-out each client's codec is resolved via `_client_egress_codec`: **client-requested (`preferred_rx_codec`) → profile policy default (`browser_rx_transport`, cached `_web_codec`) → PCM16**.
- Each **Opus client owns a dedicated `PcmOpusTranscoder`** (Opus encoders are stateful), keyed by client id in `_client_opus_transcoders`, format-keyed `(sample_rate, channels, frame_ms)` for mid-stream format changes. Reuses the existing transcoder — no new encoder code.
- **Teardown on every disconnect path**: `unsubscribe`, `reap_dead_clients`, and the relay-loop dead-WS sweep (also now drops the stale `_client_rx_codec` entry — pre-existing leak in the reap path).
- **PCM16 clients share one cached zero-copy frame** per packet; Opus-native passthrough likewise. All frames for one packet carry the same sequence number.
- Per-client Opus encode failure (libopus missing) falls back to PCM16 for that client with the existing one-shot warning.

### Negotiation ack + back-compat ladder

After `audio_start direction=rx` the server sends an advisory text ack:

```json
{"type": "audio_format", "codec": "pcm16|opus", "sample_rate": 48000, "channels": 1, "frame_ms": 20}
```

- Fire-and-forget through `_send_json` (failure tolerated — tested).
- **Absence of ack handling = legacy**: a client that ignores the ack, or sends no `preferred_rx_codec`, gets exactly today's default (profile policy → PCM16). The single-default-client PCM16 frame is **byte-identical** to pre-change output (pinned by `test_single_default_client_pcm16_frame_byte_identical` against a literal `encode_audio_frame(...)` expectation).
- `frame_ms` in the ack is the 20 ms nominal — advisory like the wire header (#765 rule: consumers size buffers from payloads).
- Static per-connection choice only; adaptive switching is explicitly **out of scope** (step 19, MOR-586).

### T1a / spine-stays-PCM boundary

- The encoder pool lives **only** in the broadcaster's WS fan-out. The AudioBridge consumes RX via `session.subscribe_rx` → AudioBus, **upstream** of the broadcaster; FFT scope/taps are fed raw PCM before the egress encode.
- `test_pcm_spine_and_bridge_path_never_get_an_encoder` proves: the post-DSP tap and a bridge-like bus subscriber both receive raw PCM, and the pool is keyed exclusively by the Opus WS client's queue id.
- Opus-native radios (IC-705) keep un-decoded passthrough for every client regardless of preference (no PCM exists to re-encode, issue #762 unchanged).

### Frontend

**Untouched.** Verified: `rx-player.ts` dispatches per-frame on the binary header codec byte (PCM16 direct / Opus via WebCodecs), and `audio-manager.ts`'s `ws.onmessage` only processes `ArrayBuffer` data — the JSON ack is silently ignored by the existing client, which is exactly the legacy-tolerant behavior required.

### Falsification (RED → GREEN)

New suite `tests/test_web_audio_per_client_codec.py` — **8/9 failed on main**, all green after:

- (a) **mixed clients**: PCM16 + Opus concurrently → each receives its own codec; on main the Opus client was forced to PCM16. Same seq across both frames; one encoder, fed the shared PCM.
- (b) **default single client**: PCM16 frame byte-identical, empty encoder pool.
- (c) **T1a boundary**: tap + bridge-like bus subscriber get raw PCM; pool keyed by WS client only.
- (d) **teardown**: encoder destroyed on `unsubscribe` and on `reap_dead_clients`.
- (e) **ack**: sent with negotiated format (opus and pcm16 variants); ack send failure does not break frame delivery.
- Opus-native passthrough for all clients pinned.

Adapted: 2 `SimpleNamespace` broadcaster doubles in `tests/test_handlers_coverage.py` gained the new `negotiated_rx_format` method (no assertions weakened).

### Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7532 passed**, 0 failed (9 skipped, 3 deselected, 11 xfailed, 1 xpassed)
- `uv run ruff check src/ tests/` → clean; `ruff format --check` → 563 files already formatted
- `uv run mypy src/` → baseline exactly **21 errors in 8 files** (no new)
- `uv run lint-imports` → **4 kept, 0 broken**
- Frontend not touched → no frontend build required

Files: `src/rigplane/web/handlers/audio.py` (+104/−45), `tests/test_web_audio_per_client_codec.py` (new, 260), `tests/test_handlers_coverage.py` (+2).

### DIFF
--- a/src/rigplane/web/handlers/audio.py
+++ b/src/rigplane/web/handlers/audio.py
@@ -115,8 +115,13 @@ class AudioBroadcaster:
         self._radio_codec: AudioCodec | None = None
         self._sample_rate: int = 48000
         self._channels: int = 1
-        self._browser_opus_transcoder: PcmOpusTranscoder | None = None
-        self._browser_opus_transcoder_key: tuple[int, int, int] | None = None
+        # Per-client egress encoder pool (MOR-584, ADR §3.6): each Opus
+        # WS client owns a dedicated transcoder (Opus encoders are
+        # stateful), keyed by client id and torn down on disconnect.
+        # Value: (transcoder, (sample_rate, channels, frame_ms)).
+        self._client_opus_transcoders: dict[
+            int, tuple[PcmOpusTranscoder, tuple[int, int, int]]
+        ] = {}
         self._browser_opus_warned: bool = False
         self._lock = asyncio.Lock()
         # Optional DSP pipeline (inserted between codec decode and tap/distribute).
@@ -173,8 +178,6 @@ class AudioBroadcaster:
                     self._subscription = None
                 if self._radio:
                     await self._start_relay()
-            elif preferred_rx_codec is not None:
-                self.invalidate_codec_state()
         self._notify_client_count_change()
         logger.info("audio-broadcaster: client added (total=%d)", len(self._clients))
         return queue
@@ -186,9 +189,8 @@ class AudioBroadcaster:
         async with self._lock:
             removed = self._clients.pop(client_id, None) is not None
             self._client_ws.pop(client_id, None)
-            removed_codec = self._client_rx_codec.pop(client_id, None)
-            if removed_codec is not None:
-                self.invalidate_codec_state()
+            self._client_rx_codec.pop(client_id, None)
+            self._client_opus_transcoders.pop(client_id, None)
             if (
                 not self._clients
                 and self._subscription is not None
@@ -295,6 +297,8 @@ class AudioBroadcaster:
             for cid in dead_ids:
                 self._clients.pop(cid, None)
                 self._client_ws.pop(cid, None)
+                self._client_rx_codec.pop(cid, None)
+                self._client_opus_transcoders.pop(cid, None)
             # Stop relay if no clients remain (and no PCM tap active)
             if (
                 not self._clients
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS — Falsification confirmed: with main's handlers/audio.py restored over the PR, the new suite goes 8/9 RED and the crux mixed-client test fails with exactly the targeted regression (the Opus client's frame carries codec byte 2/PCM16 — force-downgraded by the PCM16 client via the aggregate _web_codec); PR restored → 9/9 green. Spine-stays-PCM / T1a verified by code read and test: ulaw-decode → DSP → tap fan-out all run on the shared raw PCM buffer before the per-client encode, which sits strictly inside the WS-client fan-out (encoder pool keyed by WS client id only; the post-DSP tap and a bridge-like bus subscriber both receive raw PCM); the single-default-client case is pinned byte-identical against a literal encode_audio_frame expectation with an empty encoder pool, all per-client frames for one source packet share the same seq (increment moved after the loop), and Opus-native radio passthrough is preserved for every client with no double-encode. Encoder + _client_rx_codec teardown verified on all three disconnect paths (unsubscribe, reap_dead_clients, relay dead-WS sweep — the sweep also stops leaking _client_rx_codec), and the audio_format ack is fire-and-forget via the exception-swallowing _send_json with the requested → profile-policy → PCM16 ladder intact (no-preference and ack-ignoring legacy clients keep today's behavior, ack-send failure does not break frame delivery). Gate on the PR worktree: pytest 7532 passed / 0 failed (9 skipped, 11 xfailed, 1 xpassed; no flake reruns needed), ruff check clean + 563 files already formatted, mypy exactly 21 errors in 8 files (baseline, no new), lint-imports 4 kept / 0 broken; scope is exactly 3 files with no frontend or PCM-spine (bus/fft_scope/bridge/dsp) changes.

PR #1772, opened 2026-06-10
refactor(MOR-586): session-demand re-establishment replaces legacy audio reconnect replay

morozsm:
## Summary

Epic MOR-562 step 20 — recovery unification (ADR §3.4 rule 4, "one recovery loop"). Linear: MOR-586.

### The latent bypass bug

`runtime/_audio_recovery.py` predates the AudioSession: on transport reconnect (LAN reconnect loop + EPIPE-storm audio watchdog in `runtime/radio_reconnect.py`) it replayed the legacy `*_opus`/`*_pcm` start methods from a pre-disconnect snapshot. After the MOR-562 migration the session owns the radio RX/TX legs, so the replay re-armed the radio **behind the session**:

- demand dropped during the outage was **resurrected** (a re-armed radio leg with zero consumers draining it);
- a stale snapshot TX leg left the transport `TRANSMITTING` while the session's demand-derived state said `RX_ONLY` — **session vs radio desync**;
- the replay order was hardcoded rx-then-tx, ignoring the MOR-575 `audio_setup_order` descriptor.

### The mechanism

- **`AudioSession.reestablish()`** (new, `audio/session.py`): re-arms the radio legs from the **live demand counters** — never the snapshot — under the session lock, in the transport-declared order (`tx_first`/`atomic`: TX before RX joins the device, the MOR-559 order). RX re-arm goes through `bus.restart_rx()` (the bus's own callback — single fan-out preserved). Idempotent (typed already-started errors are benign). A dead RX leg raises so the caller surfaces FAILED (demand preserved; the next reconnect or demand edge retries); a TX re-arm failure settles at RX_ONLY, mirroring the deferred-arm policy, and never leaves TX without RX. Resets the watchdog liveness reference so the fresh leg gets its grace window; clears `RECOVERING` shadowing.
- **`AudioRecoveryRuntime`** (`runtime/_audio_recovery.py`): `recover()` routes through `session.reestablish()` when the session holds live demand; otherwise falls back to the legacy replay (extracted unchanged into `_replay_legacy`). The probe reads the **private `_audio_session` slot** (web-server runtime-payload pattern) rather than the lazy `audio_session` property, so probing never instantiates a session on a legacy-only radio. `capture_snapshot()` returns a marker snapshot when session demand is live, so the reconnect hooks' `snapshot is not None` gating becomes demand-driven (covers RX dying before the reconnect fires). The `RECOVERING`/`RECOVERED`/`FAILED` callback surface and the reconnect hooks themselves are unchanged.

### Moved to the session vs kept legacy

- **Moved**: all session-managed audio (bus subscribers — web relay, bridge, recorder — and TX leases) now recovers exclusively through the session's live demand.
- **Kept legacy**: direct `*_opus`/`*_pcm` consumers with no session demand (public API users; pinned by `tests/test_audio_recovery.py`, untouched and green, plus a new fallback test).
- **Bridge radio-side fold**: no `bridge.py` change needed — landed **by construction**. The bridge's reconnect state machine is device-side only (loopback stream errors); its radio-side demand already lives in session handles (MOR-577) that survive the outage and are re-armed by `reestablish()`. Nothing remained to migrate or delete.

### Desync falsification (RED → GREEN)

On the pre-fix code, `tests/test_audio_recovery_session.py` fails exactly on the inconsistency:

- `test_recover_reestablishes_live_demand_not_stale_snapshot` — RED: `AudioState.TRANSMITTING is not RECEIVING` (legacy replay re-armed the released TX lease; session said RX_ONLY).
- `test_recover_with_all_demand_dropped_resurrects_nothing` — RED: `start_rx_count 1 != 0` (replay re-armed RX with no consumer).
- Plus: declared-order re-arm (rx_first vs atomic), RX-failure → `RuntimeError` + FAILED, TX-failure → RX_ONLY with leases kept, recovery-callback state sequences, legacy fallback without session instantiation, and no leaked watchdog task / bus subscription across a reconnect cycle (MOR-567/MOR-581 hygiene).

### Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **7551 passed**, 1 failed: `tests/smoke/test_audio_path_smoke.py::test_web_audio_ws_relays_nonsilent_rx_frames` — **pre-existing on origin/main @ ddd8c232** (verified failing identically on a clean baseline worktree): the MOR-584 per-connection egress merge sends a text WS frame before binary audio, breaking the MOR-583 smoke test from the previous commit. The fix belongs to the parallel PR touching `web/handlers/audio.py` + frontend (out of bounds for this PR).
- `uv run ruff check src/ tests/` — clean; `ruff format --check` — clean.
- `uv run mypy src/` — baseline exactly **21 errors in 8 files** (no new).
- `uv run lint-imports` — **4 kept, 0 broken**.
- MOR-567 conformance, MOR-576/581 session/health, `tests/test_audio_recovery.py`, smoke (except the pre-existing failure above) — green.

### Deferred

- **Live reconnect re-validation is operator-gated** (radio offline): IC-7610 LAN EPIPE / network-blip recovery and FTX-1 USB replug must be re-validated on hardware before this is considered live-proven. Gate here is fakes + conformance + smoke.

### DIFF
--- a/src/rigplane/audio/session.py
+++ b/src/rigplane/audio/session.py
@@ -23,7 +23,9 @@ RECOVERING (MOR-581, ADR §3.4, tenet T3 "no silent audio death"): a ~1 s
 watchdog task — decoupled from the keep-alives — reads the bus RX heartbeat
 (MOR-564); ~3 s of silence ⇒ RECOVERING + a local
 :class:`AudioSessionEvent`; frames resuming return the demand-derived
-state. Step 14 only SURFACES the death — the recovery loop is step 20.
+state. Step 14 only SURFACES the death; step 20 (MOR-586) adds
+:meth:`AudioSession.reestablish` — the radio-side reconnect re-arms the
+session's LIVE demand instead of replaying the legacy snapshot.
 
 Arming order is transport-owned via the MOR-575 descriptor
 ``audio_setup_order`` (read with ``getattr(radio, ..., "rx_first")``):
@@ -57,6 +59,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Any, AsyncIterator, Callable, Literal, cast
 
 from rigplane.audio.bus import AudioBus, AudioSubscription
+from rigplane.audio.usb_driver import AudioAlreadyStartedError
 
 if TYPE_CHECKING:
     from rigplane.audio import AudioPacket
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS — Desync falsification confirmed on the legacy path (restored origin/main `_audio_recovery.py`): `test_recover_reestablishes_live_demand_not_stale_snapshot` failed with the transport TRANSMITTING on a released TX lease while the session said RX_ONLY, and `test_recover_with_all_demand_dropped_resurrects_nothing` failed with start_rx_count 1 (consumer-less RX resurrected); both pass with `reestablish()`, which re-arms from live demand counters under the session lock in the MOR-575 transport order (tx_first/atomic arms TX before RX joins, TX disarmed if RX fails — never TX-without-RX), resets `_rx_armed_at` for fresh watchdog grace and clears the RECOVERING shadow with no flap (max(last, armed_at) reference). Bridge-fold-empty-by-construction verified true: bridge radio-side demand is session handles (rx_sub always, TX lease whenever the radio has neutral `start_tx` — which IcomRadio, the only AudioRecoveryRuntime owner, always has via the mixin), `_rx_loop` silence-fills on radio-side gaps and only device OSError triggers bridge reconnect, so handles survive the outage and nothing needed migrating; legacy fallback correct — `_session_with_demand()` reads the private `_audio_session` slot (lazy property at radio.py:1027 untouched, pinned by `radio._audio_session is None` after probe), and the inert marker snapshot makes `_replay_legacy` a no-op if demand fully drops mid-outage. Reviewed on PR + clean merge of current main (smoke fix #1773 in): scope exactly 3 files (handlers/audio.py, frontend, bridge.py, radio_reconnect.py untouched; RECOVERING/RECOVERED/FAILED callbacks and keep-alive/EPIPE timings unchanged); full gate — pytest 7559 passed / 0 failed (incl. smoke, MOR-567 conformance, MOR-576/581 session/health), ruff check + format clean, mypy exactly 21 errors in 8 files (baseline), lint-imports 4 kept / 0 broken. Live radio-reconnect re-validation deferred to the next hardware session.

PR #1775, opened 2026-06-10
feat(MOR-588): adaptive per-client egress codec controller (PCM16↔Opus, flag-gated)

morozsm:
## Summary

Step 19 of epic MOR-562 (ADR `docs/plans/2026-06-09-target-audio-architecture.md` §3.6) — the headline WAN feature: **Opus engages ONLY on detected slow/lossy links**, automatically and per connection, with dwell/hysteresis. `[BC]` — PCM16 default unchanged; **flag-gated**, default off.

## Controller state machine (server, `web/handlers/audio.py`)

Per-client `_AdaptiveEgressState` in `AudioBroadcaster`, evaluated on every `audio_stats` uplink **and** per relay-loop frame (the latter keeps the server-side signal live for congested clients whose stats uplink stalls). Signals are the MOR-585 surfaces: client-reported playback `underruns` + server-side `ws_queue_drops` (drop-oldest evictions of the bounded WS queue). A rising cumulative counter is one unit of degradation *evidence*; a degradation *episode* is continuous evidence with gaps shorter than the degrade window.

| Transition | Condition |
|---|---|
| PCM16 → Opus | evidence episode sustained ≥ `DEGRADE_WINDOW_S = 3.0` s |
| Opus → PCM16 | zero evidence for `CLEAN_WINDOW_S = 30.0` s |
| any switch | only after `DWELL_S = 10.0` s since the previous switch (anti-flap) |

Windows/dwell are named module constants, copied per-instance (injectable for tests); the clock is `time.monotonic` (injectable). On a switch the client's effective codec re-routes through `_client_egress_codec`, so the **MOR-584 encoder pool** lazily constructs a FRESH per-client transcoder on degrade and tears it down on upgrade — no encoding machinery was rewritten. A fresh `audio_format` ack (fire-and-forget, advisory) mirrors every switch.

**Eligibility:** only browser WS clients that declared Opus decode capability (`preferred_rx_codec=opus`) adapt; under the flag that declaration is a *capability*, not a static choice — initial codec is always PCM16 (ADR: "Opus is never the initial codec"). Explicit-PCM16 clients (no WebCodecs), no-preference/legacy clients, and Opus-native radios (pass-through, #762) keep MOR-584 static behavior.

## Flag

`WebConfig.audio_adaptive_egress: bool = False` → plumbed to `AudioBroadcaster(adaptive_egress=...)`. **Off (default) = controller inert**: egress is exactly the static MOR-584 per-connection choice; a client's codec never changes mid-stream; the PCM spine + bridge + scope are untouched (controller sits strictly at the WS-client egress). Pinned by `TestFlagOff` (opus client static-Opus + zero adaptive acks; pcm16 client static under a drop storm; `WebConfig` default + plumbing).

## T1a conformance pin (`tests/contracts/test_adaptive_egress_conformance.py`)

While the controller actively degrades a WS client to Opus, the bridge bus subscriber (digital path → WSJT-X) and the post-DSP PCM tap (FFT scope) observe **bit-identical spine PCM**; every encoder construction is keyed to the WS client id only — no encoder is constructible on a lossless-class path. Plus: flag ON + healthy link constructs zero encoders (zero cost in the common case).

## Frontend mid-stream switch (`frontend/src/lib/audio/rx-player.ts`)

The codec byte travels in every frame header. On a PCM16↔Opus change, the player closes and lazily re-initialises the WebCodecs Opus decoder (the server starts a fresh encoder stream per switch) with `opusTs` restarted at 0; the playback schedule (`nextPlayTime`) is intentionally NOT rebased, so the switch is seamless (no gap, no underrun count). Codec tracking resets on `stop()`. 5 new vitest cases cover fresh-decoder init, stale-decoder close, re-init on Opus→PCM16→Opus, schedule continuity, and stop/restart.

## Falsification (deterministic — injected clock, no real 3 s/30 s waits)

RED first (no flag/controller existed: import + `TypeError` failures), then GREEN:

- flag ON + sustained underrun evidence past `DEGRADE_WINDOW_S` → flips PCM16→Opus + fresh `audio_format: opus` ack
- flag ON + `ws_queue_drops` alone (client never sends `audio_stats`) → flips via relay-loop evaluation
- isolated burst (episode gap > window) → no switch
- 30 s clean → flips back to PCM16 + encoder torn down; 29 s clean → stays Opus
- sustained re-degradation within `DWELL_S` of a switch → blocked; same evidence past dwell → switches
- exemptions: pcm16-pinned, no-preference, Opus-native radio
- flag OFF → never switches (static MOR-584 pins)
- cleanup: unsubscribe + dead-client reap clear controller state

## Gate

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7563 passed**, 10 skipped, 3 deselected, 11 xfailed, 1 xpassed (incl. MOR-567 conformance, MOR-583 smoke, MOR-584/585 suites)
- `uv run ruff check src/ tests/` → clean; `ruff format --check` → clean
- `uv run mypy src/` → **21 errors in 8 files** (baseline exact, no new)
- `uv run lint-imports` → **4 kept / 0 broken**
- frontend: `npm run lint` clean · `npm run check` 0 errors/0 warnings · `npm run test` **2068 passed (119 files)** · `npm run build` ✓ built

## Notes

- File count: 6 (2 src + 1 frontend src + 2 new test files + 1 frontend test) — exceeds the 3-file guardrail, expected for a flagged feature spanning server+frontend+conformance; each hunk kept minimal.
- Deferred: live WAN re-validation (real degraded-link behavior with the flag enabled) — hardware/network-gated, to be exercised when the flag is first turned on in a real remote session.

### DIFF
--- a/frontend/src/lib/audio/__tests__/rx-player.test.ts
+++ b/frontend/src/lib/audio/__tests__/rx-player.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { RxPlayer } from '../rx-player';
-import { AUDIO_HEADER_SIZE, MSG_TYPE_RX, CODEC_PCM16, SAMPLE_RATE, FRAME_DURATION_MS } from '../constants';
+import { AUDIO_HEADER_SIZE, MSG_TYPE_RX, CODEC_OPUS, CODEC_PCM16, SAMPLE_RATE, FRAME_DURATION_MS } from '../constants';
 
 let ctx: any;
 beforeEach(() => {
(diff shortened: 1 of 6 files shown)


morozsm:
Agent Review: BLOCKED — implementation is correct, but the falsification suite has a discrimination hole on the headline anti-flap semantic, found by mutation testing against PR head 2ba0e917 (merged with current main 878ff355, clean/disjoint).

**Problem (test gap, not an implementation bug):** two independent mutations of the degrade-side controller survive the ENTIRE suite (17/17 green): (1) deleting the DEGRADE_WINDOW gate so the controller switches PCM16→Opus on the FIRST evidence unit, and (2) deleting the episode-reset (`degrade_since = None`) so evidence gaps never end an episode. Both should be caught by `TestAdaptiveDegrade::test_isolated_burst_does_not_flip` (tests/test_web_audio_adaptive_egress.py:177), but its assertions (:193–194, `frame[1] == AUDIO_CODEC_PCM16` + `_client_opus_transcoders == {}`) are satisfied by the MOR-584 no-libopus fallback in `_encode_client_rx_frame` (src/rigplane/web/handlers/audio.py:699–707): with no native opus backend the wire frame falls back to PCM16 and the pool entry is popped EVEN WHEN the controller wrongly switched to Opus. I verified the controller state IS `AUDIO_CODEC_OPUS` under mutation (1) while the test passes. Native libopus is absent on the dev Mac (`uv sync --all-extras` → "Audio codec backend unavailable") and CI installs no system libopus — so this test discriminates in NO environment the suite runs in. The repo convention (all sibling tests patch `create_pcm_opus_transcoder`) confirms native opus must never be assumed.

**Risk:** low today (I independently verified the un-mutated controller stays PCM16 on isolated bursts at the state level), but the central MOR-588 promise — "PCM16→Opus ONLY on a continuous ≥3 s evidence episode" — has no effective regression pin; a future refactor breaking exactly this semantic would merge green.

**Required fix (small):** in `test_isolated_burst_does_not_flip`, assert controller state directly, e.g. `assert broadcaster._client_adaptive[id(handler._frame_queue)].codec == AUDIO_CODEC_PCM16` (and ideally that no `audio_format` ack with `opus` was emitted); optionally patch `_TRANSCODER_PATCH` like the sibling tests so the wire assertion also discriminates. Re-verify by re-running mutations (1) and (2) — both must fail the test.

**Everything else verified and HOLDS:**
- Flag-off = static MOR-584 (the [BP] safety): structurally (`_client_adaptive` never populated, evaluate no-ops, `_client_egress_codec` falls through to the static ladder) and by mutation (forcing the flag on internally → 2 `TestFlagOff` pins fail). Default byte-identical to before.
- Dwell + clean-window mutations ARE caught (`test_no_flap_within_dwell`, `test_clean_window_not_met_stays_opus`); 29 s-clean-vs-30 s discrimination genuine; suite is deterministic via injected clock (17 tests in 3.5 s, no real window sleeps).
- T1a: encoder constructible only inside `_encode_client_rx_frame` on the WS-client path, pool keyed by WS client id; conformance test pins bit-identical PCM on the post-DSP tap AND the bridge bus subscriber through an active switch, pool keys == {ws_client_id}, zero constructions when healthy. Bridge/digital path untouchable by adaptation.
- Encoder lifecycle: lazy build on switch, teardown on Opus→PCM16 / unsubscribe / reap / dead-client; ack fire-and-forget with task retention.
- Frontend: `resetDecoder()` on codec-byte change, old decoder closed (no leak), `opusTs` from 0, `nextPlayTime` NOT rebased (test pins 0.06 s continuity, 0 underruns), reset in `stop()`.
- Full gate (worktree merged with main): pytest 7576 passed / 0 failed (10 skipped, 11 xfailed, 1 xpassed); ruff check + format clean; mypy exactly 21 errors / 8 files (baseline); lint-imports 4 kept / 0 broken; frontend lint + check (0 errors) + tests 2068/2068 (stable across 4 runs; one initial 3-test flake never reproduced, incl. 3× isolated rx-player 40/40) + build green.

**Checks to run after the fix:** `uv run pytest tests/test_web_audio_adaptive_egress.py tests/contracts/test_adaptive_egress_conformance.py -q` + re-run the two mutations above (must now fail) + the standard gate. Live-WAN re-validation deferred as planned.

morozsm:
Agent Review: PASS — independent re-review of the test-gap fix at head bb2429d0. The delta since the prior review (2ba0e917 → bb2429d0) is test-only: +17 LOC in tests/test_web_audio_adaptive_egress.py adding controller-state assertions (broadcaster._client_adaptive[...].codec) to the isolated-burst, sustained-degradation, and clean-window tests; no assertion weakened, no production change. Independently re-verified both previously surviving degrade-side mutations against web/handlers/audio.py: (1) neutering the DEGRADE_WINDOW_S continuous-episode gate and (2) removing the degrade_since episode reset — each now FAILS test_isolated_burst_does_not_flip on the new controller-state assertion (the wire-level PCM16 fallback no longer masks a wrong controller decision). Full gate on a merged-main worktree: pytest 7576 passed / 0 failed (integration excluded), ruff check + format clean, mypy exactly 21 errors in 8 files (baseline), lint-imports 4 kept / 0 broken; the prior PASS on implementation/T1a/frontend stands.

PR #1778, opened 2026-06-10
feat(MOR-591): PcmFrame carrier + LAN decode-at-ingress (dual-publish, additive)

morozsm:
## Summary

Step 15 (**PR-1 of 2**) of the MOR-562 universal audio path epic — ADR `docs/plans/2026-06-09-target-audio-architecture.md` §3.5, tenet T1 (PCM spine). Groundwork for #762 (FFT scope dark on Opus-native radios). Runtime benefit is dormant by design (no shipping radio delivers Opus by default; uLaw is non-default) — the value is architectural: decode-at-ingress, so the spine can carry PCM.

### PcmFrame carrier (additive)

New `rigplane.audio.pcm.PcmFrame` frozen dataclass: `sample_rate`, `channels`, `payload` (ALWAYS interleaved s16le), `seq` (monotonic per RX session, never wraps — unlike the uint16 `AudioPacket.send_seq`). Exported additively from `rigplane.audio` via the lazy map; the Pro export **superset pin** (`tests/contracts/test_audio_transport_conformance.py`) stays green.

### LAN decode-at-ingress + dual-publish

`AudioRuntimeMixin` (the LAN transport adapter) now decodes the radio-negotiated RX wire codec ONCE at ingress:

- **PCM16 (1/2ch)** — zero-copy passthrough: `PcmFrame.payload` IS the wire payload object (no decode, no copy).
- **uLaw (1/2ch)** — via the same production `decode_ulaw_to_pcm16` the broadcaster uses, so ingress output is bit-exact with what the per-consumer branch would have produced.
- **Opus (1/2ch)** — via a DEDICATED session-cached `PcmOpusTranscoder` at the negotiated `(sample_rate, channels, 20 ms)` (never evicts the user-facing `_get_pcm_transcoder` cache; stateful decoder preserved).

**Dual-publish:** the legacy `AudioPacket` callback handed to `AudioStream.start_rx` is never wrapped or replaced (callback identity preserved — existing identity-asserting tests untouched). The decoded `PcmFrame` flows in parallel through a single `AudioStream` RX tap to listeners registered via the new duck-typed `add_pcm_rx_tap` / `remove_pcm_rx_tap` (gap placeholders forwarded as `None`). The `AudioTransport` protocol pin (10 members) and `AudioCapable` freeze (14 members) are untouched.

**Safety:** with no PCM taps registered the decode is skipped entirely — the default path is byte- AND cost-identical (pinned by test). Opus backend init failure dead-flags the ingress for the session (one warning, no per-frame retry); decode/tap exceptions are contained — the legacy carrier keeps flowing no matter what.

### Falsification tests (`tests/test_pcm_ingress_decode.py`, RED→GREEN)

- fake LAN radio negotiating **uLaw** → published `PcmFrame` is correctly decoded s16le (2 bytes/sample, known table anchor `0x00 → -32124`, bit-exact with the production decoder), legacy `AudioPacket` still delivered as the same object;
- fake LAN radio negotiating **Opus** (stub transcoder + libopus-gated real-codec round-trip for CI) → decoded at the negotiated `(sr, ch, 20 ms)`, legacy carrier still ships the compressed wire payload;
- **PCM16-native (IC-7610 LAN default `PCM_2CH_16BIT`) byte-identical**: zero-copy payload identity + no-taps → no decode work + `start_rx` callback identity preserved;
- monotonic seq across uint16 wrap, per-session reset, gap (`None`) forwarding, tap add/remove idempotency, tap-exception isolation, 8-bit-codec no-mapping, Opus-backend-dead fallback.

`FakeAudioStream` (shared test double) gains `add_rx_tap` / `remove_rx_tap` / `emit_rx`, mirroring the real `AudioStream` tap surface.

### Deferred to PR-2 ([BC] follow-up — the actual #762 fix)

Per the staging note in MOR-591, NOTHING is removed here. PR-2, once consumers read `PcmFrame`: remove the broadcaster uLaw decode branch (`web/handlers/audio.py`), the Opus gates that skip DSP/taps for Opus-native radios (the #762 FFT-scope-dark crutch), and the bridge's own opuslib decoder (`audio/bridge.py`), with the fake-Opus-radio scope falsification.

### Gate

- `pytest tests/ -q --ignore=tests/integration`: **7589 passed**, 0 failed (conformance superset, AudioCapable freeze, MOR-583 smoke, MOR-567 lifecycle all green)
- `mypy src/`: baseline exactly **21 errors in 8 files** (unchanged)
- `lint-imports`: **4 kept / 0 broken**

⚠️ Live IC-7610 LAN re-validation (incl. a forced uLaw-negotiated session if feasible) is operator-gated — radio offline; flagged per the issue acceptance.

### DIFF
--- a/src/rigplane/audio/__init__.py
+++ b/src/rigplane/audio/__init__.py
@@ -58,6 +58,8 @@ _LAZY_MAP: dict[str, tuple[str, str]] = {
     "Limiter": ("rigplane.audio.dsp", "Limiter"),
     "NoiseGate": ("rigplane.audio.dsp", "NoiseGate"),
     "RmsNormalizer": ("rigplane.audio.dsp", "RmsNormalizer"),
+    # PCM spine carrier (MOR-591, ADR §3.5)
+    "PcmFrame": ("rigplane.audio.pcm", "PcmFrame"),
     # Resampling
     "PcmResampler": ("rigplane.audio.resample", "PcmResampler"),
     "SampleRateNegotiation": (
@@ -139,6 +141,8 @@ __all__ = [
     "Limiter",
     "NoiseGate",
     "RmsNormalizer",
+    # PCM spine carrier
+    "PcmFrame",
     # Resampling
     "PcmResampler",
     "SampleRateNegotiation",
--- a/src/rigplane/audio/pcm.py
+++ b/src/rigplane/audio/pcm.py
@@ -1,0 +1,38 @@
+"""PcmFrame — codec-neutral s16le carrier for the PCM audio spine.
+
+ADR §3.5 (``docs/plans/2026-06-09-target-audio-architecture.md``, tenet
+T1): transport adapters decode the radio-negotiated wire codec ONCE at
+ingress and publish PCM s16le frames, so the spine (bus / DSP / taps /
+bridge) never has to know a compressed codec exists. During the
+migration the ingress dual-publishes: the legacy
+:class:`~rigplane.audio.lan_stream.AudioPacket` (Pro-pinned export)
+keeps flowing unchanged alongside this carrier (MOR-591).
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+__all__ = ["PcmFrame"]
+
+
+@dataclass(frozen=True, slots=True)
+class PcmFrame:
+    """One decoded PCM audio frame on the spine (MOR-591, ADR §3.5).
+
+    Attributes:
+        sample_rate: Sample rate in Hz of the decoded payload.
+        channels: Interleaved channel count (1 or 2).
+        payload: Interleaved signed 16-bit little-endian PCM bytes —
+            ALWAYS s16le, regardless of the radio's negotiated wire
+            codec. For PCM16-native radios this is the wire payload
+            itself (zero-copy passthrough, no decode).
+        seq: Locally counted monotonic frame sequence for the current
+            RX session (starts at 0 per session; never wraps — unlike
+            the uint16 ``AudioPacket.send_seq``).
+    """
+
+    sample_rate: int
+    channels: int
+    payload: bytes
+    seq: int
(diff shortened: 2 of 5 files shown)


morozsm:
Agent Review: PASS — Additive PR-1 confirmed: 600 additions / 0 deletions across exactly the 5 declared files; no per-consumer decoder removed (broadcaster uLaw branch at web/handlers/audio.py:808, bridge opuslib decoder at audio/bridge.py:505, and session.py all untouched — PR-2 scope intact). Default path is byte+cost-identical: the bomb-decoder test pins that no decode runs without a PCM tap, callback identity to AudioStream.start_rx is pinned unwrapped, and the legacy AudioPacket is dual-published as the same object; lan_stream.py is not modified, the pre-existing add_rx_tap dedupes (idempotent re-arm), the real RX loop forwards None gaps to taps with no exception guard, and the ingress tap never raises. Decode falsification: reverting the mixin to main turns tests/test_pcm_ingress_decode.py RED (11 failed), restoring → GREEN (13 passed, 1 skip = libopus-absence only); uLaw is bit-exact by function identity with the production decode_ulaw_to_pcm16 the broadcaster imports, PCM16-native payload IS the wire bytes (zero-copy), Opus decodes via a dedicated session-cached transcoder at the negotiated (sr, ch, 20ms) — distinct attrs from the _get_pcm_transcoder cache — with backend-failure dead-flagging that leaves the legacy carrier intact; recovery covered (both legacy replay and session reestablish route through radio.start_rx → _arm_pcm_ingress). Gate green: pytest 7589 passed / 0 failed (no-integration), ruff check + format clean, mypy exactly 21 errors in 8 files (baseline), lint-imports 4 kept / 0 broken, export-surface pins (test_exports/test_public_api_surface/test_tap_surface, 68 passed) green with the additive PcmFrame lazy-map entry. PR-2 decoder-removal/#762 deferred; live-LAN revalidation deferred. Minor non-blocking note: stop_rx jitter-buffer flush delivers tail packets to the legacy callback only, not to PCM taps — an acceptable end-of-session asymmetry for the additive spine.

PR #1780, opened 2026-06-10
fix(MOR-596): per-client PCM→20ms reframing for Opus egress (no PCM16 fallback on non-Opus radios)

morozsm:
## What

Fix **MOR-596** (High): the per-client / adaptive Opus egress (MOR-584/MOR-588) never produced Opus on non-Opus-native radios (PCM16/uLaw) — it silently fell back to PCM16 on every frame because it transcoded **per radio packet** and `PcmOpusTranscoder` requires an exact 10/20/40/60 ms frame. Live IC-7610 LAN packets are ~1280 B ≈ 6.67 ms → `Unsupported frame_ms=6` → PCM16 fallback.

## How

- `_encode_client_rx_frame` now returns `list[(codec, frame_ms, payload)]` (0..N). PCM16 / Opus-native clients stay strictly 1:1 (single passthrough frame, preserving the per-codec `shared_frames` cache).
- Opus egress on PCM radios appends to a per-client PCM **reframing accumulator** (`_client_opus_pcm_buffers`) and pops exact fixed frames (`OPUS_EGRESS_FRAME_MS = 20`, 3840 B @48k/2ch) to encode; remainder carries across packets. Header codec=0x01, frame_ms=20.
- Relay loop enqueues each returned frame; drop-oldest QueueFull eviction + MOR-588 adaptive evaluation unchanged.
- Buffer lifecycle: cleared on `unsubscribe`, `reap_dead_clients`, relay-loop dead-client cleanup, `_adaptive_switch` to PCM16, and `_refresh_codec_state` (sample-rate/channel/codec flips never concatenate mismatched PCM). On a genuine libopus-missing encode failure the existing one-shot-warned PCM16 fallback is preserved.
- `_transcoder.py` untouched; the tap/bridge/FFT path upstream is untouched (T1a lossless digital invariant).

## Tests

New `tests/test_web_audio_opus_egress_reframing.py` (7 tests): reframing accumulation (spy transcoder always receives exactly 3840 B; N×1280 B → floor(total/3840) Opus frames; remainder retained; no PCM16 fallback), PCM16 1:1 passthrough unaffected, codec-switch flushes the buffer, Opus-native pass-through unchanged, and a `skipif`-gated real-libopus encode/decode round-trip. Falsification confirmed: the accumulation + real-encode tests fail on the unfixed code (PCM16 fallback) and pass after.

## Gate

- pytest (no integration, libopus present): 7599 passed, 8 skipped, 11 xfailed, 1 xpassed — 0 failures
- ruff check + format: clean · mypy: 21 errors in 8 files (baseline) · lint-imports: 4 kept / 0 broken

Found during the live IC-7610 Echelon-A OPUS validation. Related: MOR-588, MOR-584, MOR-562.

### DIFF
--- a/src/rigplane/web/handlers/audio.py
+++ b/src/rigplane/web/handlers/audio.py
@@ -54,6 +54,12 @@ DEGRADE_WINDOW_S = 3.0
 CLEAN_WINDOW_S = 30.0
 DWELL_S = 10.0
 
+# Fixed Opus egress frame duration (MOR-596).  Radio RX packets are not
+# Opus-frame-aligned (IC-7610 LAN @48k stereo ≈1280 B ≈6.67 ms), while
+# ``PcmOpusTranscoder`` accepts exactly one fixed frame per encode call —
+# per-client PCM is therefore reframed into exact 20 ms frames before encode.
+OPUS_EGRESS_FRAME_MS = 20
+
 
 @dataclass
 class _AdaptiveEgressState:
@@ -159,6 +165,11 @@ class AudioBroadcaster:
         self._client_opus_transcoders: dict[
             int, tuple[PcmOpusTranscoder, tuple[int, int, int]]
         ] = {}
+        # Per-client PCM reframing accumulators for Opus egress (MOR-596):
+        # s16le bytes buffered until a full 20 ms frame is available.
+        # Cleared on unsubscribe/reap, adaptive switch to PCM16, and any
+        # codec-state refresh (sample rate / channel changes).
+        self._client_opus_pcm_buffers: dict[int, bytearray] = {}
         # Per-client link-quality state (MOR-585, ADR §3.6): the latest
         # client-reported ``audio_stats`` snapshot plus the server-side
         # drop-oldest eviction counter for the bounded WS queue.  Stats
@@ -255,6 +266,7 @@ class AudioBroadcaster:
             self._client_ws.pop(client_id, None)
             self._client_rx_codec.pop(client_id, None)
             self._client_opus_transcoders.pop(client_id, None)
+            self._client_opus_pcm_buffers.pop(client_id, None)
             self._client_link_quality.pop(client_id, None)
             self._client_queue_drops.pop(client_id, None)
             self._client_adaptive.pop(client_id, None)
@@ -366,6 +378,7 @@ class AudioBroadcaster:
                 self._client_ws.pop(cid, None)
                 self._client_rx_codec.pop(cid, None)
                 self._client_opus_transcoders.pop(cid, None)
+                self._client_opus_pcm_buffers.pop(cid, None)
                 self._client_link_quality.pop(cid, None)
                 self._client_queue_drops.pop(cid, None)
                 self._client_adaptive.pop(cid, None)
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review at PR head 5d067ce0 (worktree, detached). Falsification-first.

**Scope** — exactly the 2 expected files; no strays. Production +79/-37 (=+42 net, as ticketed); test file +245 (7 ticket-mandated scenarios, no padding). Within guardrails (<=3 files).

**Correctness (adversarial)** — verified:
- Accumulator drains in exact frame_bytes chunks (`del buf[:frame_bytes]`); remainder carried. frame_bytes = sample_rate*20//1000*channels*2 == PcmOpusTranscoder.frame_bytes exactly (48k/2ch=3840, 24k/1ch=960). All _VALID_SAMPLE_RATES (8/12/16/24/48k) are %50==0 → integer-exact, channel-aligned, always >0. No off-by-one / partial-frame leakage.
- No unbounded growth: the drain `while` runs unconditionally every relay iteration before enqueue, independent of WS-queue state; residual bounded by <1 frame (<3840 B). Pre-queue buffer cannot accumulate.
- Lifecycle clears complete: unsubscribe (269), reap_dead_clients (381), relay dead-client (946), _adaptive_switch→PCM16 (642), _refresh_codec_state clear-all (459) + transcode-fail except (742). No stale-PCM concatenation path; no path clears a still-needed buffer.
- Concurrency: same single-event-loop pattern as the pre-existing _client_opus_transcoders dict; lifecycle mutations under self._lock, relay loop unchanged — no NEW race.
- seq: all N reframed sub-frames share the packet seq; browser rx-player.ts decodes per-frame by codec/payload and does NOT dedup/reorder by seq → correct, documented.
- 1:1 PCM16/Opus-native passthrough + shared_frames per-codec cache survived via `payload is audio_data` identity check; no double-send/drop.
- T1a: tap/DSP/FFT PCM fan-out upstream untouched (egress is downstream-only).

**Falsification** — checked out origin/main audio.py + ran the NEW tests: 7/7 FAIL on unfixed code. Behavior discriminators fail for the right reason — real-libopus test emits [2,2,2,2,2,2] (PCM16 fallback/packet) vs expected [1,1]; log shows `Unsupported frame_ms=6` (documented root cause). Mutation test (drop remainder via buf.clear()) → 3 tests FAIL, confirming carry-over is genuinely asserted. Restored file = pristine.

**Gate (independent)**:
- pytest (no integration): 7598 passed, 8 skipped, 11 xfailed, 1 xpassed, **1 failed** = test_audio_session_health::test_healthy_heartbeat_never_false_positives — a wall-clock heartbeat timing flake in audio/session.py (RX silent 0.1s under heavy parallel load), OUTSIDE the PR's 2-file scope; passes 8/8 in isolation. Not a regression.
- ruff check: All checks passed · ruff format --check: 572 files already formatted
- mypy src/: 21 errors in 8 files (baseline; none in audio.py)
- lint-imports: 4 kept / 0 broken

Diff matches the ticket, no substantiated correctness defect, tests genuinely discriminate. PASS.

PR #1781, opened 2026-06-10
fix(MOR-608): route web RX broadcaster through AudioSession so the T3 watchdog covers browser-only listeners

morozsm:
## What

Fix **MOR-608** (High): `AudioBroadcaster._start_relay` subscribed RX **directly** to `radio.audio_bus`, bypassing the radio-owned `AudioSession`. For browser-only listening (no bridge, no TX) the session's `rx_demand` stayed 0 → session state IDLE → the MOR-581 health watchdog never ran, so tenet **T3 "no silent audio death"** did not cover the most common usage. Reconnect for that case also flowed through the legacy `_replay_legacy` rather than `AudioSession.reestablish()`.

## How (ADR §3.2-correct, option a)

- `_start_relay`: when the radio has a session (`getattr(self._radio, "audio_session", None)`, the MOR-580 pattern), acquire RX via `await session.subscribe_rx("web-audio")` → demand registered, session → RX_ONLY, watchdog runs, reconnect routes through `reestablish()`. Radios WITHOUT a session keep the exact legacy `bus.subscribe("web-audio")` + `start()` path.
- Relay loop body unchanged — `RxSubscription.__aiter__` yields the same `AudioPacket | None` stream.
- New private `_release_subscription()` helper handles sync-stop (legacy `AudioSubscription`) vs async-release (`RxSubscription`) at all three teardown sites (`_stop_relay`, dead-relay-restart cleanup, and a previously-leaky stale-handle overwrite in `ensure_relay`). Lock-safe: `release()` takes only session→bus→radio locks; nothing calls back into the broadcaster, so awaiting it under the broadcaster `_lock` cannot deadlock.
- Tap lifetime preserved: session RX demand is held exactly while the relay runs (FFT-scope-tap-only operation keeps it).
- Type: only the private `self._subscription` annotation widened. `_replay_legacy`/`_SESSION_MANAGED_SNAPSHOT`/TX handler/bridge/shims untouched.

## Tests

New `tests/test_web_audio_session_rx.py` (7 tests): session demand created (RX_ONLY), watchdog covers web listening (RECOVERING on silence), teardown releases demand, tap keeps demand, no-session legacy fallback preserved, `_session_with_demand()` non-None for a web relay. Falsification: 6/7 fail on unmodified code (incl. the two key falsifiers — session stayed IDLE, watchdog never ran); test 5 (legacy pin) passes pre-fix as intended.

## Gate

- pytest (no integration, libopus present): 7606 passed, 8 skipped, 11 xfailed, 1 xpassed — 0 failures (incl. `test_audio_recovery_session.py`, lifecycle conformance, audio-path smoke, all broadcaster/relay suites)

Found by the MOR-562 post-epic cleanup audit. Serves the "no silent problems" principle (MOR-594). Related: MOR-581, MOR-586, MOR-562.

### DIFF
--- a/src/rigplane/web/handlers/audio.py
+++ b/src/rigplane/web/handlers/audio.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any, Callable, Protocol, cast
 from ..._audio_codecs import decode_ulaw_to_pcm16
 from ..._audio_transcoder import PcmOpusTranscoder, create_pcm_opus_transcoder
 from ...audio.bus import STAGE_RX_POST_DSP
+from ...audio.session import RxSubscription
 from ...audio.usb_driver import AudioAlreadyStartedError
 from ...dsp.tap_registry import TapHandle, TapRegistry
 from ...env_config import (
@@ -151,7 +152,12 @@ class AudioBroadcaster:
         self._clients: dict[int, asyncio.Queue[bytes]] = {}
         self._client_ws: dict[int, Connection] = {}
         self._client_rx_codec: dict[int, int] = {}
-        self._subscription: _AudioSubscription | None = None
+        # RX source handle: a session-routed RxSubscription when the radio
+        # owns an AudioSession (MOR-608 — registers RX demand so the
+        # MOR-581 watchdog covers browser-only listeners), or a legacy
+        # bus AudioSubscription for radios without one. Both yield the
+        # same AudioPacket|None stream via ``async for``.
+        self._subscription: _AudioSubscription | RxSubscription | None = None
         self._relay_task: asyncio.Task[None] | None = None
         self._seq: int = 0
         self._web_codec: int = AUDIO_CODEC_PCM16
@@ -249,8 +255,7 @@ class AudioBroadcaster:
                     if self._relay_task is not None:
                         self._relay_task.cancel()
                         self._relay_task = None
-                    self._subscription.stop()
-                    self._subscription = None
+                    await self._release_subscription()
                 if self._radio:
                     await self._start_relay()
         self._notify_client_count_change()
@@ -364,6 +369,9 @@ class AudioBroadcaster:
         async with self._lock:
             relay_alive = self._relay_task is not None and not self._relay_task.done()
             if not relay_alive and self._radio:
+                # A dead relay may leave a stale handle behind — release it
+                # so session RX demand never leaks across a restart.
+                await self._release_subscription()
                 await self._start_relay()
                 logger.info("audio-broadcaster: relay started for PCM tap")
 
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-608 at head 27c9379c (worktree, detached HEAD).

**Scope** — exactly 2 files: `src/rigplane/web/handlers/audio.py` (+47/-9) and `tests/test_web_audio_session_rx.py` (+237). Production net +38. Within guardrails (≤3 files, modest delta). Diff matches the ticket: `_start_relay` routes through `session.subscribe_rx("web-audio")` with a `getattr(radio, "audio_session", None)` guard falling back to `bus.subscribe`; three sync `.stop()` teardown sites (subscribe dead-relay path, `ensure_relay` stale-handle, `_stop_relay`) converted to the async `_release_subscription()` helper; only the private `_subscription` annotation widened to `_AudioSubscription | RxSubscription | None`.

**Deadlock analysis (NO deadlock).** Lock order is strictly broadcaster `_lock` → session `_lock` → bus lock → radio. Every site that awaits `RxSubscription.release()` while holding `_lock` (subscribe:258, ensure_relay:374, _stop_relay via :1006) follows that order. Verified no inverse path: `_release_rx`/`reestablish`/watchdog only call into `self._radio` and `self._bus`; the sole outward callback is `_emit` → `WebServer._on_audio_session_event` → `broadcast_event`, which is a lock-free synchronous `put_nowait` (server.py:1163) — never re-enters the broadcaster lock.

**Demand-leak analysis (NO leak / no double-release).** `subscribe_rx` is paired with exactly one `release()` across start/stop/restart: dead-relay restart releases the stale handle before re-arming; `ensure_relay` releases-then-arms; `_stop_relay` releases. `_release_subscription` nulls `self._subscription` BEFORE awaiting, and all callers hold `_lock`, so no interleave; `_release_rx` is idempotent (`if sub not in self._rx_subs: return`) — double-release safe. `subscribe_rx` fully unwinds on its own failure (session.py:282-285). Minor non-blocking note: `_start_relay`'s except handler (audio.py:815) nulls `_subscription` without releasing if an exception fires AFTER `subscribe_rx` succeeds but during `asyncio.create_task` — that window is unreachable in practice (create_task can't fail with a running loop), so no action needed.

**Tap-keeps-relay & no-session paths verified.** `test_pcm_tap_keeps_relay_and_session_demand` confirms both directions (tap holds demand at RX_ONLY with zero WS clients; detach+stop drops to IDLE). No-session fallback: `LanLikeRadio` has no `audio_session` attr, so the public-property guard returns None and `_session_with_demand` reads the absent private `_audio_session` slot — guard cannot return a truthy-but-broken object. Reconnect: web demand makes `_session_with_demand()` non-None → `reestablish()` (re-arms via `bus.restart_rx()` on the surviving subscription, no orphan/double-arm); legacy replay preserved for session-less radios.

**Falsification (tests genuinely discriminate).** Reverted production file to origin/main: 6/7 new tests FAIL (incl. T3 watchdog and recovery-path), only the no-session test passes (path unchanged) — confirms the bug. Restored: 7/7 pass. Mutation (leak demand: `_release_subscription` never calls `release()`): 4 tests FAIL on `rx_demand==0`/IDLE post-teardown assertions. Restored clean.

**Full gate (all green):**
- pytest tests/ --ignore=tests/integration → 7606 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failed (236s); flagged audio suites `test_audio_recovery_session`/`test_audio_lifecycle_conformance`/`test_audio_path_smoke` → 62 passed, 1 skipped; `test_info_endpoint_status` passed in-suite and in isolation (no flake observed).
- ruff check → All checks passed; ruff format --check → 573 files already formatted.
- mypy src/ → exactly 21 errors in 8 files (baseline; none in the changed files — type widening clean).
- lint-imports → 4 kept / 0 broken.

No deadlock or leak substantiable; tests discriminate; no-session + reconnect intact; gate green. Approved to merge once required checks are green and this PASS is fresh for the head.

PR #1783, opened 2026-06-10
test(MOR-607): inject monotonic clock into AudioSession watchdog for deterministic health tests

morozsm:
## What

Fix **MOR-607** (Low): `test_audio_session_health.py::test_healthy_heartbeat_never_false_positives` flaked under `pytest-xdist` heavy parallel load — the watchdog's heartbeat-feeder coroutine starved past `rx_liveness_timeout`, so by wall clock the heartbeat genuinely went stale and the watchdog fired per its contract (a test-timing artifact, NOT a production false-positive).

## How

- `AudioSession.__init__` gains a keyword-only `monotonic: Callable[[], float] = time.monotonic`; all 5 wall-clock reads in the watchdog/liveness path (`_emit` timestamp, `_rx_armed_at` in `_sync_watchdog` + `reestablish` ×2, `now` in `_check_liveness_locked`) route through it. Default `time.monotonic` → **zero production behavior change**; `rx_liveness_timeout` NOT loosened.
- Tests rewritten deterministically with a `_ManualClock` (counts reads so tests advance on "watchdog observed this state" rather than real time) + `_stamp_heartbeat` (delivers a real fan-out frame, re-stamps the bus heartbeat from the same clock). The sibling `test_frames_resuming_returns_to_demand_state` shared the flake shape and was rewritten too; added `test_watchdog_overrun_with_fresh_heartbeat_does_not_misfire` pinning the exact jitter shape.

## Falsification + stability

The 3 rewritten tests fail with `TypeError: unexpected keyword argument 'monotonic'` on unmodified code (red), pass after (green). Touched tests 50× → 50/50 green; health file under `-n auto` ×3 → all green (and ~3× faster, no real-time burn).

## Gate

- pytest (no integration): 7607 passed, 0 failed

`git diff --stat origin/main`: 2 files, +109/−21. Related: MOR-581, MOR-562, MOR-596.

### DIFF
--- a/src/rigplane/audio/session.py
+++ b/src/rigplane/audio/session.py
@@ -118,8 +118,10 @@ _WATCHED_STATES: frozenset[AudioSessionState] = frozenset(AudioSessionState) - {
 
 @dataclass(frozen=True, slots=True)
 class AudioSessionEvent:
-    """Local liveness event (NO telemetry — open-core). ``timestamp`` is
-    ``time.monotonic()``, comparable to ``AudioBus.last_rx_frame_monotonic``."""
+    """Local liveness event (NO telemetry — open-core). ``timestamp`` comes
+    from the session's monotonic clock (``time.monotonic`` by default) and is
+    comparable to ``AudioBus.last_rx_frame_monotonic`` — see the
+    ``AudioSession`` ``monotonic`` argument for the clock-injection caveat."""
 
     state: AudioSessionState
     reason: str
@@ -188,6 +190,13 @@ class AudioSession:
             :class:`AudioBus` — no import-matrix change).
         bus: Existing bus to wrap; defaults to ``radio.audio_bus`` or a
             fresh :class:`AudioBus`.
+        monotonic: Monotonic clock used for the liveness watchdog and the
+            :class:`AudioSessionEvent` timestamps; defaults to
+            :func:`time.monotonic` — production behavior is unchanged.
+            Test-only seam (MOR-607): liveness compares this clock against
+            ``AudioBus.last_rx_frame_monotonic``, and the bus stamps real
+            ``time.monotonic()`` on its own — callers injecting a fake
+            clock MUST also feed the bus heartbeat from that SAME clock.
     """
 
     def __init__(
@@ -197,6 +206,7 @@ class AudioSession:
         bus: AudioBus | None = None,
         watchdog_interval: float = WATCHDOG_INTERVAL_S,
         rx_liveness_timeout: float = RX_LIVENESS_TIMEOUT_S,
+        monotonic: Callable[[], float] = time.monotonic,
     ) -> None:
         self._radio = radio
         self._bus: AudioBus = (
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

**Scope** (git diff origin/main...HEAD): exactly `src/rigplane/audio/session.py` (+19/-6) and `tests/test_audio_session_health.py` (+91/-14). 109 insertions, 21 deletions, 2 files. No collateral.

**Production behavior provably unchanged.**
- New `monotonic` param is keyword-only (after `*`), default `time.monotonic`. Verified via `inspect.signature`: `default is time.monotonic == True`, `kind == KEYWORD_ONLY`.
- Grep of `session.py`: ZERO remaining bare `time.monotonic()` runtime calls — all 5 runtime reads (`_emit` timestamp, 3× `_rx_armed_at`, `_check_liveness_locked.now`) route through `self._monotonic`; remaining `time.monotonic` occurrences are the docstring + the `import time`.
- Constants unchanged vs origin/main: `WATCHDOG_INTERVAL_S=1.0`, `RX_LIVENESS_TIMEOUT_S=3.0` (byte-diff of the constants region empty). Not loosened.
- Cross-clock contract sound: in production both `self._monotonic()` and `AudioBus._last_rx_frame_monotonic` (bus.py:352) are real `time.monotonic()` → same-clock comparison. Only tests inject, and `_stamp_heartbeat` re-stamps the bus from the SAME fake clock. No real mismatch possible. Docstring caveat is accurate.
- All 3 runtime call sites (`runtime/radio.py:1037`, `yaesu_cat/radio.py:413`, `bridge.py:533`) construct `AudioSession` positionally without `monotonic` → pick up the default. Keyword-only param cannot break positional callers.

**Determinism — genuine, not just passing.** Liveness math reads ONLY the manual clock (moves solely on `clock.advance()`); `_wait_for` is a 'watchdog has run' barrier (predicate = state or `clock.calls > reads`) with a 2 s real-time failsafe, no assertion depends on sleep duration. Stability: touched file 30/30 green serial; 5/5 green under `-n auto`; full audio group (1168 tests) 2/2 green under `-n auto`.

**Discriminating.** Mutation `ref = self._rx_armed_at` (ignore the fresh `last` heartbeat) → `test_healthy_heartbeat_never_false_positives` AND the new `test_watchdog_overrun_with_fresh_heartbeat_does_not_misfire` both FAIL (spurious RECOVERING). The new test exercises the exact jitter shape (clock jumps 10×timeout, fresh same-clock heartbeat → must stay RX_ONLY) and catches the regression. (Restored after.)

**Falsification.** Against origin/main `session.py`, the 3 clock-injecting tests fail with `TypeError: __init__() got an unexpected keyword argument 'monotonic'` — they genuinely depend on the seam. Restored; worktree clean at 14eab36d.

**Gate.**
- pytest tests/ --ignore=tests/integration: 7607 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failed.
- ruff check: All checks passed. ruff format --check: 573 files already formatted.
- mypy src/: Found 21 errors in 8 files (baseline; none in audio/session.py).
- lint-imports: Contracts: 4 kept, 0 broken.

PR #1785, opened 2026-06-10
fix(MOR-595): fully unwind commander/keepalive tasks on failed connect() to stop orphaned future warnings

morozsm:
## What

Fix **MOR-595** (Low): calling `radio.connect()` again on the same `IcomRadio` after a discovery `TimeoutError` (the IC-7610 stale-LAN-slot case) left background tasks/futures from the failed attempt whose exceptions were never retrieved → a flood of `Future exception was never retrieved: ConnectionError('Commander stopped')` at GC. Surfaced during the MOR-586 live re-validation.

## What was leaking + the fix

1. **Keepalive tasks** (`IcomTransport._ping_loop`/`_idle_loop`/`_retransmit_loop`): `disconnect()` cancelled them but never awaited, so on the failed-connect cooldown path the cancelled-but-unawaited tasks lingered. `disconnect()` now collects → cancels → **awaits** each (suppressing `CancelledError`, debug-logging other exceptions), idempotent and `current_task`-safe.
2. **Orphaned commander futures**: `IcomCommander` failed fire-and-forget BACKGROUND sends (`wait_dispatch=False`, the live poller) with `ConnectionError('Commander stopped')` in `stop()` and the worker cancel path, but nobody awaits those futures. `_QueueItem` now carries a `fire_and_forget` flag and all three failure sites route through a new `_fail_item()` that calls `future.exception()` on fire-and-forget items to mark the exception retrieved. "Commander stopped" still propagates normally to real awaiters (NORMAL/IMMEDIATE/`wait_dispatch=True`).

Success path is byte-for-byte unchanged — only failure/teardown branches touched.

## Tests / falsification

New `tests/test_connect_teardown_hygiene.py` (3 tests): keepalive tasks `done()` after `disconnect()`; zero `Future exception was never retrieved` after a failed-connect→retry; no leaked `_ping_loop` task. All RED on unfixed `src/` (tasks not done, 5 captured loop-handler events, leaked task), GREEN after.

## Gate

- pytest (no integration): 7609 passed, 0 failed

Commit touches 3 files (`core/transport.py`, `commands/commander.py`, the test). Related: MOR-586, MOR-562.

### DIFF
--- a/src/rigplane/commands/commander.py
+++ b/src/rigplane/commands/commander.py
@@ -41,6 +41,26 @@ class _QueueItem:
     # True only for fire-and-forget BACKGROUND sends counted against the
     # ``_MAX_BG_INFLIGHT`` cap; the worker decrements the counter for these.
     counts_bg_inflight: bool = False
+    # True for any fire-and-forget send (``wait_dispatch=False``): no caller
+    # ever awaits ``future``, so whoever fails it must retrieve the exception
+    # itself or GC logs "Future exception was never retrieved" (MOR-595).
+    fire_and_forget: bool = False
+
+
+def _fail_item(item: _QueueItem, exc: BaseException) -> None:
+    """Fail a queue item's future, retrieving the exception for orphans.
+
+    Fire-and-forget futures have no consumer, so teardown (``stop()`` and
+    worker cancellation) marks their exception as retrieved immediately to
+    keep a failed ``connect()``/disconnect from flooding the log with
+    ``Future exception was never retrieved: ConnectionError('Commander
+    stopped')`` warnings at GC time (MOR-595).
+    """
+    if item.future.done():
+        return
+    item.future.set_exception(exc)
+    if item.fire_and_forget:
+        item.future.exception()
 
 
 class IcomCommander:
@@ -85,8 +105,7 @@ class IcomCommander:
                     break
                 if item.key is not None:
                     self._pending_by_key.pop(item.key, None)
-                if not item.future.done():
-                    item.future.set_exception(ConnectionError("Commander stopped"))
+                _fail_item(item, ConnectionError("Commander stopped"))
                 self._queue.task_done()
 
         if self._worker is not None and not self._worker.done():
@@ -157,6 +176,7 @@ class IcomCommander:
             key=key,
             wait_response=wait_response,
             counts_bg_inflight=counts_bg_inflight,
+            fire_and_forget=not wait_dispatch,
         )
 
         if key is not None:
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-595 (full unwind of commander/keepalive tasks on failed connect()). Scope = exactly 3 files (transport.py, commander.py, tests/test_connect_teardown_hygiene.py), +315/-13 LOC, disjoint from the 2 commits merged after the branch point — reviewed against merge-base e2d5160.

Correctness — current-task safety (transport.disconnect):
- The `task is not asyncio.current_task()` filter is applied to the single `tasks` list used for BOTH the cancel loop and the await loop, so disconnect can never await/cancel the task it runs inside. Verified no keepalive loop (_ping/_idle/_retransmit) calls disconnect() — defensive guard only, but correct.
- Idempotent: filter excludes None + already-done tasks; a second call sees all done() -> empty list -> no-op. Safe on partial-start and double-call.
- Rest of disconnect (DISCONNECT pkt, socket close, state=DISCONNECTED) all runs BEFORE the await loop; each `await task` is individually wrapped (CancelledError->pass, Exception->debug log), so a raising task cannot abort socket close, state transition, or the other awaits. No hang risk: all 3 loops swallow CancelledError immediately and yield at await sleep, so cancel+await resolves promptly.

Correctness — fire-and-forget-only silencing (commander):
- `fire_and_forget = not wait_dispatch` is correct at the single enqueue site. _fail_item calls set_exception always, but future.exception() (mark-retrieved) ONLY when fire_and_forget. wait_dispatch=True (NORMAL/IMMEDIATE/user) keeps fire_and_forget=False -> real awaiter at `await fut`/wait_for still raises 'Commander stopped'. future.exception() suppresses only the GC warning; it never clears the exception, so even a hypothetical dedupe cross-path awaiter still sees the error (and no dedupe=True + wait_dispatch=False call sites exist anyway).
- All 3 failure sites route through _fail_item consistently: stop() drain (L108), worker CancelledError teardown (L284), worker generic Exception (L287). Success path untouched — set_result stays in the try at L280; _fail_item only reachable on teardown/exception branches. No double-cancel, no new task leak, no swallowed real error.

Falsification (reverted both src files to merge-base):
- All 3 new tests FAIL for the exact right reasons: keepalive tasks not done() (cancelled-not-awaited); 5x captured 'Future exception was never retrieved: ConnectionError(Commander stopped)'; leaked cancelling _retransmit_loop task pending. Restored HEAD -> 3 pass, tree clean.
- Bonus mutation (make _fail_item skip future.exception()) -> TestCommanderStopHygiene FAILS (catches 5 unretrieved warnings). Mutation reverted, tree clean.

Gate:
- pytest tests/ (no integration): 7609 passed, 8 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failures (234s) — no flake reruns needed.
- ruff check: All checks passed. ruff format --check: 574 files already formatted.
- mypy src/: Found 21 errors in 8 files (matches baseline; none in touched files).
- lint-imports: 4 kept, 0 broken.
- PR not draft, MERGEABLE; quick + grep-gate green.

Teardown is current-task-safe + idempotent, real awaiters still see 'Commander stopped', success path unchanged, falsification confirmed, gate green. No concerns.

PR #1787, opened 2026-06-10
fix(MOR-620): truthful PTT + degraded-session indicator

morozsm:
## Problem (observed live on IC-7610 LAN, 2026-06-10)

After a long session the LAN link degraded: `hello` reported `connected:true` /
`radio_ready:false`, yet a control-WS PTT command returned
`{"ok":true,"result":{"state":true}}` while the radio never keyed. PTT
"success" meant "command enqueued", not "radio keyed", and the degraded
session was invisible in the UI.

## Why PTT lied

`ControlHandler._enqueue_rc_power` (`src/rigplane/web/handlers/control.py`)
ACKs PTT as soon as `PttOn()`/`PttOff()` is `q.put(...)` into the command
queue — RadioPoller executes `radio.set_ptt()` later and any failure never
reaches the WS response. No readiness check existed on the keying path.

## Changes

**Backend — truthful PTT (`web/handlers/control.py`):**
- New `RadioNotReadyError` + `_ensure_tx_session_ready()` gate on `ptt`
  (`state:true`) and `ptt_on`: rejected with the new `radio_not_ready` error
  envelope when the backend explicitly reports `radio_ready: False` (CI-V
  stream unhealthy) or a connect/reconnect cycle is in flight
  (`_backend_recovering()`, same predicate `radio_connect` already uses).
- **Unkeying always goes through**: `ptt state:false` / `ptt_off` bypass the
  gate — never strand the radio transmitting on a half-dead link.
- Radios that expose no `radio_ready` bool are not gated (no behavior change
  for simple/legacy backends and test fakes).
- An `ok:false` response already surfaces as an error toast via the existing
  `WsChannel` response-to-notification plumbing, so the rejection is visible.

**Frontend — degraded-session indicator (extends MOR-594, no new UI element):**
- `ws-client.ts` now consumes the MOR-594 `connection_status` WS event
  (previously broadcast but unconsumed — the deferred fast-follow) and feeds
  `setRadioStatus()`, so the StatusBar radio dot turns yellow during
  backend reconnect cycles in real time.
- `StatusBar.svelte`: the radio indicator degrades (yellow) when the session
  is `connected` but `radioReady=false`, with a tooltip label
  (`core.statusbar.health.radioNotReady`, added to en/ru/ja locales).

**Docs:** `docs/api/web.md` notes the `radio_not_ready` error code and the
PTT-OFF exemption.

## Tests

- `tests/test_handlers_coverage.py`: 6 new tests (TDD — written first,
  failing) — rejection on `radio_ready:false`, rejection during
  `conn_state=reconnecting`, `ptt_on` alias, PTT OFF allowed while degraded,
  PTT ON allowed when ready.
- `ws-client.test.ts`: `connection_status` → `setRadioStatus` routing +
  malformed-payload tolerance.
- `RadioLayout.test.ts` / `top-row-visual-regression.test.ts`: connection
  store mocks extended with `getRadioReady` (StatusBar now reads it).

## Gates

- `pytest` (no integration): 0 failures
- `mypy src/`: 21 errors / 8 files (baseline, unchanged)
- `lint-imports`: 4 kept / 0 broken
- frontend: `npm run i18n:check`, `npm run check` (0 errors), `npx vitest run`
  (2069 passed), `npm run lint` — all clean

Part of MOR-614 (T6: MOR-620).

### DIFF
--- a/docs/api/web.md
+++ b/docs/api/web.md
@@ -1083,6 +1083,12 @@ Error response:
 {"type":"response","id":"42","ok":false,"error":"command_failed","message":"..."}
 ```
 
+PTT keying (`ptt` with `state:true`, or `ptt_on`) on a degraded session — the
+backend reports `radio_ready:false` or a connect/reconnect cycle is in flight —
+is rejected with `"error":"radio_not_ready"` instead of a silent enqueue-ACK
+(MOR-620). Unkeying (`ptt` with `state:false`, or `ptt_off`) always goes
+through: it is the safe direction.
+
 Rate-limited high-frequency `set_*` commands are ACKed with:
 
 ```json
--- a/frontend/src/components-v2/layout/StatusBar.svelte
+++ b/frontend/src/components-v2/layout/StatusBar.svelte
@@ -12,6 +12,7 @@
     getHttpConnected,
     getRadioPowerOn,
     getRigConnected,
+    getRadioReady,
     getRadioHealth,
   } from '$lib/stores/connection.svelte';
   import { getFrequency } from '$lib/stores/radio.svelte';
@@ -65,6 +66,7 @@
   let audioState = $derived(isPoweredOff ? 'disconnected' : (isAudioConnected() ? 'connected' : 'disconnected'));
   let httpState = $derived(getHttpConnected() ? 'connected' : 'disconnected'); // server link — always real
   let rigConnected = $derived(getRigConnected());
+  let radioReady = $derived(getRadioReady());
   let radioHealth = $derived(getRadioHealth());
   // Effective radio indicator: downgrade to 'disconnected' when rigCtld reports radio offline
   let radioIndicatorState = $derived.by(() => {
@@ -74,7 +76,9 @@
     if (radioHealth?.readiness === 'delayed' || radioHealth?.readiness === 'stalled') {
       return 'degraded';
     }
-    return radioState === 'connected' && !rigConnected ? 'degraded' : radioState;
+    // MOR-620: a session that is connected but not radio-ready (CI-V link
+    // degraded) must be visibly degraded, not silently green.
+    return radioState === 'connected' && (!rigConnected || !radioReady) ? 'degraded' : radioState;
   });
   let radioHealthLabel = $derived.by(() => {
     switch (radioHealth?.likelyCause) {
@@ -89,9 +93,13 @@
       case 'server_unreachable':
         return t('core.statusbar.health.serverUnreachable');
       default:
-        return !rigConnected && radioState === 'connected'
-          ? t('core.statusbar.health.rigOffline')
-          : '';
+        if (!rigConnected && radioState === 'connected') {
+          return t('core.statusbar.health.rigOffline');
+        }
+        if (!radioReady && radioState === 'connected') {
+          return t('core.statusbar.health.radioNotReady');
+        }
+        return '';
     }
   });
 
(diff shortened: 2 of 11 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-620 (T6 of epic MOR-614) at head `a59a8aff`. Verified design correctness, ran all gates in a detached worktree, no regressions.

## Design — correct & safe
- **PTT OFF always bypasses the gate.** `_ensure_tx_session_ready()` is called only on the keying direction: `ptt` with `state:true` (guarded by `if on:`) and the `ptt_on` alias. `ptt_off` and `ptt state:false` have no gate call — confirmed in the diff and by `test_ptt_off_allowed_when_radio_not_ready` (covers both the `state:false` form and the `ptt_off` alias). Never strands the radio transmitting on a half-dead link.
- **Predicate matches the `radio_connect` convention.** The gate reuses the exact same `_backend_recovering()` predicate `_handle_radio_connect` already uses, plus an explicit `radio_ready is bool False` check. Consistent, no ad-hoc logic.
- **No regression for fakes/simple backends.** Gating fires only when `getattr(radio, "radio_ready", None)` is a real `bool False`, or `_backend_recovering()` is true. A backend with no `radio_ready` bool: `explicitly_not_ready=False`, and `_backend_recovering()` for it reduces to `connected and not connected` = always False (via the `radio_ready()` helper fallback to `connected`). So simple/legacy backends and SimpleNamespace/Mock fakes are never gated. Helper semantics (`runtime_helpers.radio_ready`) match the gate's conservatism.
- **Error envelope is the existing convention.** `RadioNotReadyError` is raised in `_enqueue_rc_power`, propagates `_enqueue_legacy_command` → `_ControlCommandExecutor.execute` → `CommandService.execute` (generic `except Exception: ... raise`, re-raised, not swallowed) → `_handle_command`'s except block, which now maps it to the `radio_not_ready` code in the standard `ok:false` response. Read-only PermissionError still rejects first (stronger constraint, correct ordering).
- **No new hang/deadlock.** The gate is purely synchronous — it returns or raises immediately, no `await`/sleep/queue-wait. Healthy session keys normally (`test_ptt_on_allowed_when_radio_ready`).
- **Audio byte path untouched. Layering respected.** lint-imports 4 kept / 0 broken. StatusBar.svelte imports only from `$lib/stores/connection.svelte` (no transport/audio import) — the transport consumer is correctly `ws-client.ts`. Frontend wiring is real, not dead: backend emits `broadcast_event("connection_status", {state, ...})` (server.py:1928, sourced from `radio_reconnect._emit_reconnect_status`), and ws-client reads `ev.data?.state` (string-guarded, malformed payloads ignored). The degraded yellow + `radioNotReady` tooltip is driven by `radioReady` (state-snapshot `state.connection.radioReady`, pre-existing pipeline) AND `radioState='connected'`; the new `connection_status`→`setRadioStatus` wiring feeds the dot's base state. Both legit and complementary.

## Gates (detached worktree at PR head)
- New tests: `tests/test_handlers_coverage.py` — **143 passed** (incl. the 6 new MOR-620 tests).
- mypy `src/`: **exactly 21 errors / 8 files** (baseline; none in the touched `control.py` — all pre-existing in radio_poller/yaesu/rigctld/runtime).
- lint-imports: **4 kept / 0 broken**.
- ruff `src/ tests/`: clean.
- Frontend: ws-client + RadioLayout + top-row-visual-regression — **69 passed**. New i18n key `core.statusbar.health.radioNotReady` present in all 3 locales (en/ru/ja).

## CI
`quick` PASS, `grep-gate` PASS, `Update Agent Review Gate status` PASS. `Agent Review Gate` was failing only for the missing PASS comment — this comment satisfies it. No RED attributable to this PR.

No blocking issues. Safe to merge once required checks are green.

Issue #771, opened 2026-04-17
test(frontend): flaky keyboard-wiring tests under vitest isolate:false


morozsm:
## Symptom

\`frontend/src/components-v2/wiring/__tests__/keyboard-wiring.test.ts\` intermittently fails ~6-15 tests when the full frontend suite runs under the current vitest config.  Back-to-back runs on the same commit give different results:

\`\`\`
Run 1: 1522 passed
Run 2:  1507 passed / 15 failed
Run 3:  1517 passed / 5 failed
\`\`\`

Failing cases all live in \`makeKeyboardHandlers\` — \"cycles to the next band by index\", \"cycles preamp values from capabilities\", \"toggles split using radio state\", etc.  Error is always \`Number of calls: 0\` for \`sendCommand\` / \`patchRadioState\`.

## Root cause (likely)

PR #707 turned on \`isolate: false\` for the 10× wall-clock speedup.  Under that config, module cache is shared across test files within a worker.  \`keyboard-wiring.test.ts\` uses \`vi.mock('$lib/transport/ws-client', …)\` and \`vi.mock('$lib/stores/radio.svelte', …)\`, which only take effect on a fresh module load — if any sibling test loaded those modules un-mocked first, the mocks are no-ops and \`makeKeyboardHandlers\` uses the real implementations.

Under alphabetical-ish test ordering the sibling file that triggers the un-mocked load varies, hence the intermittency.

## Evidence

Reproduced on pristine \`main\` (no local changes) — flakiness predates any recent work.  Not caused by #754 or #770.

## Proposed fixes (pick one)

1. Re-enable \`isolate: true\` for this specific file via \`test.server.deps.inline\` or a per-file \`isolate\` override, if vitest supports it (need to check).
2. Switch all \`vi.mock(...)\` in \`keyboard-wiring.test.ts\` to \`vi.doMock(...)\` inside \`beforeEach\` — forces re-mock on each test.
3. Add \`vi.resetModules()\` in \`beforeEach\` + re-import the module under test dynamically inside each test — standard vitest pattern for shared-cache scenarios.
4. Revert the module-scope imports in \`keyboard-wiring.test.ts\` to dynamic imports inside each \`it(...)\`.

Option 2 or 3 seems lowest-friction.

## Scope

- 1 file: \`frontend/src/components-v2/wiring/__tests__/keyboard-wiring.test.ts\`.
- ~20 LOC refactor.
- No production code changes.

## Acceptance

[...]

PR #1790, opened 2026-06-10
feat(MOR-617): warn + one-click Set LAN preflight guard for network voice TX

morozsm:
Part of MOR-614 (T3).

## What

When the user keys **network voice TX** from the web UI and the active mode group's MOD-input source is **not LAN (5)**, the UI now shows a clear, actionable warning with a **one-click "Set LAN"** fix. Warn-only: TX is never blocked, the rig is never changed silently (auto-set is a separate opt-in, T4/MOR-618).

## Why

If the active DATA group's MOD input isn't LAN, network voice TX modulates from MIC/ACC/USB instead of the network audio stream — open-mic feedback or dead air. This is the root cause the MOR-614 epic addresses; T1 (MOR-615) exposed the state, T2 (MOR-616) added the helpers + ModePanel control, this T3 adds the actual protection at the TX-start moment.

## How

Frontend-only — the backend audio byte path is untouched.

- **TX-start hook**: `lib/runtime/adapters/tx-adapter.ts` — `armModInputTxGuard()` is called inside `getTxAudioControl().startTx()`, the single keying entry shared by desktop `TxPanel` (PTT button) and `MobileRadioLayout` (FAB + landscape strip). Synchronous, never blocks/delays `runtime.startTx()`.
- **Guard**: new `lib/runtime/adapters/mod-input-tx-guard.svelte.ts` — `$state` latch armed at TX start; visibility is **derived reactively** from live state, so the warning clears as soon as the source becomes LAN (optimistic patch or readback) and reappears if the radio rejects the change. Resolves the ACTIVE receiver's group via T2's `modInputStateKey(dataMode)`; one-click reuses T2's `onModInputChange(5)` command path (`set_data_off/1/2/3_mod_input {source:5}` + optimistic patch).
- **Gating** identical to the T2 ModePanel control: `data_mode` capability + `fieldStatus` for the group not `missing` + source known (not null) — non-IC-7610 radios and unread groups never warn.
- **UI**: new presentation-only `components-v2/panels/ModInputTxWarning.svelte` (`role="alert"`, Set LAN + dismiss buttons), mounted inline in `TxPanel` under the PTT button and as a pointer-transparent floating overlay in `MobileRadioLayout` (mobile keying happens outside the TX sheet).
- **i18n**: `core.txGuard.modInputNotLan` / `setLan` / `dismiss` in en-US / ru-RU / ja-JP (glossary tokens MOD/LAN/TX preserved).

## Tests (TDD — written first)

`mod-input-tx-guard.test.ts` (13) + `ModInputTxWarning.test.ts` (4):
- fires on TX-start when source known and != LAN (incl. SUB receiver / D2 group)
- does NOT fire when source == LAN, unknown (null), fieldStatus missing, no `data_mode` cap, no state
- one-click sends the active group's SET command with `source=5` (D1 and DATA-OFF cases)
- clears on readback (source becomes LAN) and on dismiss
- tx-adapter still delegates to `runtime.startTx()` in both armed/clean cases
- banner renders/hides, buttons wire to adapter handlers

### Test-pool note

`vite.config.ts`: the two new `vi.mock` test files are routed through the **isolated** vitest pool per the documented #771 policy. Additionally, the pre-existing T2 `mod-input-wiring.test.ts` was observed flaking on unmodified main (~2/8 full-suite runs on a 16-core macOS host, same `isolate: false` shared-cache pinning) and is moved to the isolated pool too. With this change: 6/6 consecutive full-suite runs green vs 6/8 on base.

## Gates

- `npm run check` — 0 errors / 0 warnings
- `npx vitest run` — 2110 tests passed (incl. 17 new), 6/6 consecutive runs green
- `npm run lint` — clean
- `npm run i18n:check` — OK (3 locales, 164 keys)
- No Python touched.


morozsm:
Agent Review: PASS

Independent review of MOR-617 (T3) at head `2d476fb6`. Verdict: **PASS** — warn-only contract holds, gating mirrors T2, layering clean, flake-handling legit, gates green and deterministic.

**Warn-only (critical) — CONFIRMED.** `tx-adapter.ts` calls `armModInputTxGuard()` (synchronous, void) then unconditionally `return runtime.startTx()`. No conditional, no await before it, no throw path (`offendingSource` is null-safe: early `if (!state) return null`, optional chaining throughout). TX can never be blocked or delayed by the guard. Both `tx-adapter TX-start hook` tests assert `runtime.startTx` called once in armed AND clean cases.

**Gating — CONFIRMED.** `offendingSource` is a faithful copy of T2's `toModeProps` gate (`panel-props.ts:426-429` / `state-adapter.ts:356-359`): `data_mode` cap + `state !== null` + `getFieldAvailability(state, modInputStateKey(dataMode)) !== 'missing'`, plus source-known and source!=LAN(5). Fires only when known & non-LAN; silent on LAN / null / missing-fieldStatus / no-cap / no-state — all branch-tested.

**Active-group consistency — CONFIRMED.** `offendingSource` resolves `active==='SUB' ? sub : main`; one-click `onSetLan` → T2 `onModInputChange(5)` reads `getActiveReceiver()` (`radio.svelte.ts:364`) with the identical resolution, so the SET (`set_data{N}_mod_input {source:5}` + optimistic patch) targets exactly the warned group. Clears reactively on readback (source→LAN) and on dismiss; reappears on radio rejection.

**No backend / audio-path change — CONFIRMED.** Zero Python; zero audio/transport/backend/dsp files. Frontend-only.

**Layering — clean.** `ModInputTxWarning.svelte` (panel) imports only `$lib/i18n` + the runtime adapter — passes `FORBIDDEN_PANEL_IMPORTS` (no stores/transport/audio-manager). Guard logic lives in `lib/runtime/adapters/`. `MobileRadioLayout` reuses pre-existing `txCapable`. `npm run lint` clean.

**Flake/pool change — legit, not masking a new bug.** New module-scope `vi.mock` files + the pre-existing T2 `mod-input-wiring.test.ts` are routed to the documented #771 isolated pool. The guard's module-global `armed = $state` is contained: the only fast-pool importer of `tx-adapter` (`TxPanel.test.ts`) now mocks the guard, so no fast-pool test touches the real latch. The isolation addresses test determinism, not a leak affecting siblings.

**Gates (detached worktree @ PR head):**
- `npm run check` — 0 errors / 0 warnings
- `npm run lint` — clean
- `npm run i18n:check` — OK (3 locales, 164 keys)
- `npx vitest run` — 2110 passed, **3/3 consecutive runs deterministic** (123 files each)
- 17 new tests (13 guard + 4 banner) pass and assert real behavior

No blocking issues. CI quick.yml frontend block is green; this PASS satisfies the Agent Review Gate for head `2d476fb6`.

— Independent reviewer (did not author this PR)

PR #1791, opened 2026-06-10
feat(MOR-618): opt-in auto-set LAN MOD input for network voice TX

morozsm:
Part of MOR-614 (T4, final task of the epic).

## What

An **opt-in** (OFF by default) auto-configure for network voice TX: when the toggle is enabled, keying TX from the web auto-switches the active DATA group's MOD input to **LAN(5)** (remembering the previous source) and restores the previous source when TX stops. This is Option 3 of the epic — the default UX remains the T3 (MOR-617) warn + one-click "Set LAN" guard; this is purely an optional convenience.

## How

- **New runtime adapter** `frontend/src/lib/runtime/adapters/mod-input-auto.svelte.ts`
  - Preference persisted in localStorage (`rigplane:auto-lan-mod-input`, `'true'`/`'false'`, default OFF), exposed reactively via `deriveAutoLanModInputProps()` (gated by `data_mode` capability + observed group, same as the ModePanel control).
  - `autoSetLanModInputForTx()` — reuses the T2 per-group SET command + optimistic top-level patch path (`modInputCommand`/`modInputStateKey`, MOR-615/616). Quiet-gating identical to the T3 guard: no state / no capability / group not read / source unknown / already LAN → no change, no pending.
  - `restoreModInputAfterTx()` — one-shot restore of the remembered source; skipped (pending dropped) if the user moved the group off LAN mid-TX.
- **TX hook** (`tx-adapter.ts`, same path T3 uses): `startTx` runs auto-set **before** `armModInputTxGuard()`, so the optimistic LAN patch preempts the T3 warning (auto preempts the warn). A failed `startTx` (e.g. mic capture error) restores immediately. `stopTx` restores after stopping TX audio.
- **Toggle UI**: TxPanel settings modal (⚙ LEVELS popover), checkbox + help text, rendered only when the radio exposes MOD-input routing. i18n keys `core.txGuard.autoLanLabel`/`autoLanHelp` in en/ru/ja.
- **Best-effort disconnect recovery**: the pending "restore to previous" is persisted (`rigplane:mod-input-tx-restore:v1`). `FrontendRuntime.bootstrap()` calls `applyPendingModInputRestoreOnConnect()`, which waits (via the radio-state subscription) for live state on a connected control channel and then restores — only if the group is still on LAN; if it is no longer LAN (someone changed it), the pending record is dropped.

## Known limitation (by design — not over-engineered)

Clean TX stop is the primary restore path. If the session dies mid-TX (crash, power loss, browser killed) the radio is left on LAN until the **next** connect of this web app repairs it best-effort. Being left on LAN is benign — LAN is the correct source for network operation, and the T3 ModePanel control / warning still allow one-click changes at any time. A fully-robust variant (backend-tracked restore independent of the browser session, applied on radio-side TX-end or server reconnect) would need backend lifecycle changes and is proposed as a follow-up if desired.

## Invariants

- Audio byte path untouched.
- Toggle OFF (default) == exact T3/MOR-617 behavior (verified by test).
- Frontend layering respected: logic in `lib/runtime/adapters`, TxPanel stays presentation-only (imports the adapter, no transport/audio-manager imports).

## Tests (TDD — written first)

`frontend/src/lib/runtime/adapters/__tests__/mod-input-auto.test.ts` (20 tests, registered in the vitest **isolated** pool per the #771 `vi.mock` policy, same as the T3 file):
- toggle default OFF + localStorage persistence + capability/fieldStatus availability gating
- OFF → no auto-set, T3 warning still arms
- ON → set LAN on TX start (active group routing incl. SUB/D2), remembers previous source, suppresses the T3 warning, persists pending
- no-ops: already LAN / unknown source / no capability / group missing
- restore on stop (one-shot), no stomping of a manual mid-TX change, restore even if toggle flipped OFF mid-TX, restore after failed `startTx`
- best-effort reconnect path: applies persisted pending when state shows LAN, waits for state, drops when off LAN, ignores malformed data, defers to the in-session clean-stop while TX is active

Plus 3 TxPanel component tests for the toggle UI and a mock-completeness update in `frontend-runtime.test.ts`.

## Gate

- `npm run check` — 0 errors, 0 warnings (4256 files)
- `npx vitest run` — 2133 passed / 124 files, run 3× (deterministic)
- `npm run lint` — clean
- `npm run i18n:check` — OK (3 locales, 166 keys)
- Python untouched (no Python gate needed)


morozsm:
Agent Review: PASS

Independent review of PR #1791 (MOR-618, T4 — final task of epic MOR-614). Reviewed full diff (11 files, +766/-3, frontend-only) and confirmed all gates at PR head `8b77149` in a detached worktree.

**OFF == byte-for-byte T3.** `autoSetLanModInputForTx()` returns immediately on `!enabled` (default OFF, `readStoredPref()` reads `rigplane:auto-lan-mod-input` !== 'true'), so `startTx` runs `armModInputTxGuard()` → `runtime.startTx()` exactly as in T3, and `restoreModInputAfterTx()` is a no-op (pending null). Verified by the dedicated 'OFF behavior is exactly MOR-617' test: no `sendCommand`, T3 warning still arms, no pending persisted.

**TX never blocked or delayed.** The auto-set is fully synchronous and non-throwing: `sendCommand()` is guarded (returns `false` when radio degraded, optimistic patch wrapped in try/catch) and the optimistic `patchRadioState` is a plain store write. It runs *before* `await runtime.startTx()` but cannot await-block or raise; the keying path always proceeds. A failed `startTx` immediately calls `restoreModInputAfterTx()` (tested).

**Restore lifecycle sound.** ON: sets LAN on the *active* receiver group (MAIN/SUB, correct `set_dataN_mod_input` per data_mode — tested incl. SUB on D2), remembers the previous source, suppresses the T3 warning via the optimistic LAN patch. Quiet-gating identical to the T3 guard (no state / no data_mode cap / fieldStatus missing / source null / already LAN → no change, no pending). Restore is one-shot (consumes `pending`, clears persisted) and SKIPPED when the group was moved off LAN mid-TX — a deliberate user change wins (tested: manual USB(3) mid-TX → no clobber; toggle flipped OFF mid-TX still restores).

**Pending-restore on connect is safe.** `applyPendingModInputRestoreOnConnect()` (bootstrap step 6): waits for live state on a connected control channel, restores only if still on LAN, drops otherwise, drops if the group is provably `missing`, and defers while an in-session `pending` TX owns the restore. No infinite loop (every terminal branch unsubscribes), no double-restore (clean in-session stop clears persisted; the lingering subscription then reads null → finishes), no wrong-group restore (record carries its own command+key), and the synchronous-emit race is handled by the `if (finished) unsubscribe()` tail. All six reconnect cases tested incl. malformed-data ignore.

**No backend / no audio-byte-path change.** Zero Python, no `audio-manager`/`transport` byte-path edits — confirmed via `git diff --name-only`. Module never touches the audio path.

**Layering clean.** Logic lives in `lib/runtime/adapters`; TxPanel toggle is presentation-only (imports only the adapter + `$lib/i18n`, no transport/store/audio). Targeted eslint on the 3 changed files exits 0 — no `no-restricted-imports` violation.

**Flake/pool change legit.** New `mod-input-auto.test.ts` added to the #771 isolated pool (and excluded from the default pool) — correct for a `vi.mock` file, and necessary because the module-level `$state` pref/pending would otherwise leak across files in the shared pool. Standalone run = 20/20; not masking a real cross-test leak. `frontend-runtime.test.ts` gains only a `subscribeRadioState` mock-completeness stub.

**Disconnect limitation benign + documented.** Crash mid-TX leaves the radio on LAN (the correct network-voice source) until the next connect repairs it best-effort; T3 one-click control remains available. Documented in the module header, PR body, and i18n help text. No backend dependency.

**Gates (detached worktree at PR head, `npm ci`):**
- `npm run check` — 0 errors / 0 warnings (4256 files)
- `npx vitest run` — 124 files / 2133 tests passed, run 3× (deterministic)
- `npm run lint` — clean
- `npm run i18n:check` — OK (3 locales, 166 keys; en/ru/ja `autoLanLabel`/`autoLanHelp` present)
- CI: `quick` pass, `grep-gate` pass (only the Agent Review Gate was red, awaiting this PASS)

No blocking issues. Approving the final task of epic MOR-614.

PR #1792, opened 2026-06-11
fix(#1719): render FM/AM waterfall passband symmetric about the carrier

### LINKED ISSUE #1719: [bug] Asymmetric FM passband stripe on waterfall
### Description

In FM mode, the waterfall passband display erroneously renders asymmetrically to the right of the carrier frequency, resembling an upper sideband (USB) signal, instead of displaying symmetrically about the center carrier frequency as expected.

**Rigplane v2.8.0 renders like this:**

Image

```

```
**It should render like this:**

Image

### Steps to Reproduce

```python
IC-9700 connected using LAN set to FM modulation.
```

### Radio Model

IC-9700

### rigplane Version

2.8.0

### Python Version

_No response_

### OS

_No response_

### Debug Log

```text

```

### Analysis (filled during fix)

_No response_

morozsm:
## Root cause

`getPassbandEdgesHz()` in `frontend/src/components/spectrum/passband-geometry.ts` classified only `CW`, `CW-R`, `RTTY`, `RTTY-R`, and `AM` as symmetric (carrier-centered) modes. **FM was not in that branch**, so it fell through to the default USB-like computation (`leftHz = shift`, `rightHz = shift + width`), drawing the waterfall/spectrum passband stripe entirely to the right of the carrier — looking like a USB signal. The same mis-bucketing existed in `getFilterWidthFromRightEdgePx()` (right-edge drag-resize).

The carrier-relative passband shape is computed entirely client-side from mode + filter width; the backend only supplies the width and mode string, so no backend change is needed. An existing unit test even codified the wrong behavior ("places FM passband above carrier (same as USB)") — it has been replaced.

## Change

- Introduced a single `SYMMETRIC_PASSBAND_MODES` set shared by `getPassbandEdgesHz()` and `getFilterWidthFromRightEdgePx()` (replacing the two duplicated equality chains, no new abstraction).
- Added the FM/AM family to the symmetric branch: `FM`, `FM-N`, `WFM`, `DATA-FM`, `DATA-FM-N`, `C4FM-DN`, `C4FM-VW`, `DV`, `AM-N` (mode vocabulary taken from the shipped `rigs/*.toml` profiles; `FM-D`/`AM-D` strings do not exist — data mode is a separate flag and the geometry helper receives the raw `rx.mode`).
- `USB`/`LSB` stay one-sided; `CW`/`RTTY` keep their existing centered convention.
- Display geometry only — audio/filter behavior untouched.

## Before / after

| Mode | Before | After |
|---|---|---|
| FM 15 kHz | carrier .. +15 kHz (USB-like) | −7.5 kHz .. +7.5 kHz (centered) |
| AM | centered (unchanged) | centered |
| USB/LSB | one-sided (unchanged) | one-sided |

TDD: failing tests first (FM/FM-N/WFM/DATA-FM/DV/C4FM/AM-N symmetry + FM right-edge resize), then the fix.

## Gate results

- `npm run check` — 0 errors, 0 warnings (4256 files)
- `npx vitest run` — 2137 passed / 124 files, run 3× (deterministic)
- `npm run lint` — clean
- `npm run i18n:check` — OK
- Python untouched

Fixes #1719

### DIFF
--- a/frontend/src/components/spectrum/__tests__/passband-geometry.test.ts
+++ b/frontend/src/components/spectrum/__tests__/passband-geometry.test.ts
@@ -28,8 +28,25 @@ describe('getPassbandEdgesHz', () => {
     expect(getPassbandEdgesHz('CW-R', 500, 0)).toEqual({ leftHz: -250, rightHz: 250 });
   });
 
-  it('places FM passband above carrier (same as USB)', () => {
-    expect(getPassbandEdgesHz('FM', 15000, 0)).toEqual({ leftHz: 0, rightHz: 15000 });
+  it('centers FM passband around carrier (#1719)', () => {
+    expect(getPassbandEdgesHz('FM', 15000, 0)).toEqual({ leftHz: -7500, rightHz: 7500 });
+  });
+
+  it('centers narrow/wide FM variants around carrier (#1719)', () => {
+    expect(getPassbandEdgesHz('FM-N', 10000, 0)).toEqual({ leftHz: -5000, rightHz: 5000 });
+    expect(getPassbandEdgesHz('WFM', 200000, 0)).toEqual({ leftHz: -100000, rightHz: 100000 });
+    expect(getPassbandEdgesHz('DATA-FM', 15000, 0)).toEqual({ leftHz: -7500, rightHz: 7500 });
+    expect(getPassbandEdgesHz('DATA-FM-N', 10000, 0)).toEqual({ leftHz: -5000, rightHz: 5000 });
+  });
+
+  it('centers digital FM-based modes around carrier (#1719)', () => {
+    expect(getPassbandEdgesHz('DV', 7000, 0)).toEqual({ leftHz: -3500, rightHz: 3500 });
+    expect(getPassbandEdgesHz('C4FM-DN', 12500, 0)).toEqual({ leftHz: -6250, rightHz: 6250 });
+    expect(getPassbandEdgesHz('C4FM-VW', 12500, 0)).toEqual({ leftHz: -6250, rightHz: 6250 });
+  });
+
+  it('centers narrow AM around carrier (#1719)', () => {
+    expect(getPassbandEdgesHz('AM-N', 6000, 0)).toEqual({ leftHz: -3000, rightHz: 3000 });
   });
 
   it('applies IF shift to both passband edges', () => {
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of PR #1792 (fix #1719 — FM/AM waterfall passband symmetric about carrier). Frontend display-geometry only; 2 files, +48/−16.

**Correctness — confirmed**
- Root cause is right: `getPassbandEdgesHz()` and `getFilterWidthFromRightEdgePx()` each carried a duplicated CW/CW-R/RTTY/RTTY-R/AM equality chain; FM fell through to the default USB-like return (`left=shift, right=shift+width`), drawing a one-sided stripe. Replaced by a single shared `SYMMETRIC_PASSBAND_MODES` set used by BOTH functions — the drag-resize path is fixed too, not just the render path.
- Symmetric formula is correct: `left=shift-width/2, right=shift+width/2` (centered on carrier+shift). FM 15 kHz → −7500..+7500. The right-edge resize inverse `width=(rightEdgeHz-shift)*2` is the correct counterpart; the new `getFilterWidthFromRightEdgePx('FM',0,12000,600,375,15000)=3000` assertion checks out (75 px → 1500 Hz → ×2 = 3000), matching the adjacent AM case — a real symmetry assertion, not inverted noise. The replaced test now asserts the correct centered result.
- USB/LSB untouched (LSB own branch, USB default one-sided); CW/RTTY keep their existing centered convention. No regression for those modes.

**Mode vocabulary — correct**
- FM/FM-N/WFM/DATA-FM/DATA-FM-N/AM-N are double-sideband / FM-family symmetric (correct). DV/C4FM-DN/C4FM-VW are FM-based narrowband digital voice, symmetric about carrier (correct). Cross-checked against shipped `rigs/*.toml`: ftx1 ships DATA-FM, FM-N, AM-N, DATA-FM-N, C4FM-DN, C4FM-VW; ic9700 (the reported rig) ships plain FM in its mode list (so the reported bug is directly covered). No symmetric mode that appears in a profile is missing. (`WFM` as a *mode* isn't in any profile mode-list — the ftx1 `WFM` token is a freq-range label, 88–108 MHz — but including it is harmless and radio-correct.)
- No spurious one-sided mode was added to the symmetric set. `FM-D`/`AM-D` correctly absent (data is a separate flag; helper gets raw `rx.mode`). Note (non-blocking, pre-existing, out of #1719 scope): FTX-1's CW-U/CW-L/RTTY-L/RTTY-U variant strings are not in the symmetric set, but they were not before either and #1719 is FM-only.

**Layering**: pure helper, no transport/store imports added. Both consumers (SpectrumPanel.svelte, lib/renderers/spectrum-renderer.ts) reach it via the single `getPassbandGeometry` → `getPassbandEdgesHz` path. No backend/audio/filter change.

**Gates — independently run in a detached worktree at PR head ea6dc25 (npm ci):**
- `npm run check` — 0 errors, 0 warnings (4256 files)
- `npx vitest run` — 124 files / 2137 tests passed, run 3× deterministically (no #771-class flake)
- passband-geometry.test.ts alone — 27/27 pass
- `npm run lint` — clean (exit 0)
- `npm run i18n:check` — OK (166 keys)

CI: `quick` was pending and `Agent Review Gate` was waiting on this comment; `grep-gate` green. No RED attributable to this PR.

Verdict: correct, complete, and safe. Approving.

PR #1796, opened 2026-06-11
feat: backend session-teardown MOD-input restore for abnormal disconnect (MOR-624)

morozsm:
## Summary

The MOR-618 opt-in auto-LAN feature (T4 of epic MOR-614) restores the radio's previous MOD-input source on a clean TX stop, but on an abnormal disconnect mid-TX (browser closed / network drop) the restore depended on the SAME browser reconnecting (`applyPendingModInputRestoreOnConnect` via localStorage). If that browser never came back, the radio stayed on LAN.

This change tracks the armed restore on the BACKEND session:

- **Frontend** (`mod-input-auto.svelte.ts`): when the auto-LAN SET fires at TX start, additionally send `arm_mod_input_restore { command, source }`; on a clean TX stop (which owns the restore) send `disarm_mod_input_restore` — fired whether or not the restore SET itself is skipped by the mid-TX manual-change guard. The localStorage reconnect path is kept unchanged as a complementary best-effort for the full-server-restart case.
- **Backend** (`web/handlers/control.py`): `arm`/`disarm` are session-local bookkeeping — NOT radio commands — intercepted in `_handle_command` before the `_COMMANDS` gate, so the command catalog (and its docs equality check) stays untouched. The armed `(command, previous_source)` tuple lives on the `ControlHandler`; `run()`'s `finally` calls `_restore_mod_input_on_teardown()`, which enqueues the restore on the shared command queue (which outlives the session) if still armed.

## PttOff-then-restore ordering

A still-armed restore at teardown means the client vanished mid-TX, i.e. the rig may still be keyed. Restoring a non-LAN source (e.g. MIC) while keyed would re-trigger the exact open-mic feedback footgun MOR-614 fixed. The teardown therefore enqueues `PttOff()` FIRST, then the previous-source SET. Best-effort: if the shared poller is shutting down too, the failure is logged at debug level.

## Safety properties

- **Opt-in only** — nothing arms unless the user enabled the auto-LAN toggle and an auto-set actually fired.
- **No behavior change when the toggle is OFF** — the OFF path sends no arm/disarm at all (covered by the existing OFF-behavior test).
- **Audio byte path untouched** — only session bookkeeping + two enqueued poller actions on teardown.

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ --ignore=tests/integration` | 7640 passed, 8 skipped, 3 deselected, 11 xfailed, 1 xpassed |
| `ruff check` + `ruff format --check` | All checks passed · 575 files already formatted |
| `mypy src/` | 21 errors in 8 files — exactly the pre-existing baseline, no new errors |
| `lint-imports` | Contracts: 4 kept, 0 broken |
| `npm run check` | 4256 files, 0 errors, 0 warnings |
| `vitest` (full) | 124 files / 2137 tests passed (touched file: 20/20) |
| `npm run lint` | clean |

## Tests

- New `tests/test_web_mod_input_restore.py`: arm/disarm state machine, unknown-command rejection, teardown enqueues exactly `[PttOff, SetData*ModInput(source)]`, no-op when not armed / after disarm, and `_handle_command` routing with an `ok: true` ack.
- Extended `frontend/.../mod-input-auto.test.ts`: arm sent alongside the auto-LAN SET (active-group routing included), disarm on clean stop (including the manual-mid-TX-change path where the restore SET is skipped), OFF/already-LAN/no-pending paths still send nothing.

Closes MOR-624

### DIFF
--- a/frontend/src/lib/runtime/adapters/__tests__/mod-input-auto.test.ts
+++ b/frontend/src/lib/runtime/adapters/__tests__/mod-input-auto.test.ts
@@ -201,6 +201,11 @@ describe('auto-set at TX start (MOR-618)', () => {
       key: 'data1ModInput',
       source: 0,
     });
+    // MOR-624: backend session-teardown net armed alongside the LAN set.
+    expect(sendCommand).toHaveBeenCalledWith('arm_mod_input_restore', {
+      command: 'set_data1_mod_input',
+      source: 0,
+    });
   });
 
   it('routes to the ACTIVE receiver group (SUB on D2)', async () => {
@@ -216,6 +221,11 @@ describe('auto-set at TX start (MOR-618)', () => {
     await getTxAudioControl().startTx();
 
     expect(sendCommand).toHaveBeenCalledWith('set_data2_mod_input', { source: 5 });
+    // MOR-624: the armed backend net carries the same group + previous source.
+    expect(sendCommand).toHaveBeenCalledWith('arm_mod_input_restore', {
+      command: 'set_data2_mod_input',
+      source: 3,
+    });
   });
 
   it('does nothing when the source is already LAN', async () => {
@@ -272,6 +282,8 @@ describe('restore at TX stop (MOR-618)', () => {
 
     expect(runtime.stopTx).toHaveBeenCalledTimes(1);
     expect(sendCommand).toHaveBeenCalledWith('set_data1_mod_input', { source: 0 });
+    // MOR-624: a clean stop owns the restore — backend teardown net cleared.
+    expect(sendCommand).toHaveBeenCalledWith('disarm_mod_input_restore', {});
     expect(getRadioState()?.data1ModInput).toBe(0);
     expect(pendingInStorage()).toBeNull();
 
@@ -293,7 +305,11 @@ describe('restore at TX stop (MOR-618)', () => {
 
     getTxAudioControl().stopTx();
 
-    expect(sendCommand).not.toHaveBeenCalled();
+    // The manual choice wins: no restore SET — but the backend teardown net
+    // is still cleared (MOR-624), the clean stop owns the restore decision.
+    expect(sendCommand).toHaveBeenCalledTimes(1);
+    expect(sendCommand).toHaveBeenCalledWith('disarm_mod_input_restore', {});
+    expect(sendCommand).not.toHaveBeenCalledWith('set_data1_mod_input', expect.anything());
     expect(pendingInStorage()).toBeNull();
   });
 
--- a/frontend/src/lib/runtime/adapters/mod-input-auto.svelte.ts
+++ b/frontend/src/lib/runtime/adapters/mod-input-auto.svelte.ts
@@ -171,6 +171,12 @@ export function autoSetLanModInputForTx(): void {
   // (MOR-616); the backend confirms via write-through readback (MOR-615).
   patchRadioState({ [key]: LAN_MOD_INPUT_SOURCE } as Partial<ServerState>);
   sendCommand(pending.command, { source: LAN_MOD_INPUT_SOURCE });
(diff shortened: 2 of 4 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-624 (backend session-teardown MOD-input restore). Verified at PR head `c4d309a5`.

**Backend (`src/rigplane/web/handlers/control.py`)**
- Intercept placement is correct: `arm_mod_input_restore`/`disarm_mod_input_restore` are caught in `_handle_command` (L867-874) AFTER `cmd_id`/`name`/`params` are read and BEFORE the `set_`-prefix rate-limit block (L879), the `_COMMANDS` membership gate (L907), and the `self._radio is None` check (L921). It ACKs and returns. The two names are NOT in `_COMMANDS` — so `tests/test_command_catalog_docs.py` (strict set-equality, L35-38) stays green. Hard-constraint #3 satisfied.
- `_apply_mod_input_restore_cmd`: valid arm records `(command, source)` → `{armed:True}`; unknown command clears to None → `{armed:False}` (no stale arm); disarm clears → `{}`.
- `_restore_mod_input_on_teardown`: clears `self._mod_input_restore` BEFORE any early-return (no double-fire), no-ops on not-armed / no server / no radio / no queue / unknown command, and on the happy path enqueues `PttOff()` THEN `action(source)`, wrapped in try/except so a queue failure cannot crash teardown. Called from `run()` `finally` (L534) AFTER `unregister_control_event_queue` (L533).
- `_MOD_INPUT_RESTORE_ACTIONS` maps exactly the 4 group SET names to the 4 action classes (all already imported). No new mypy errors in `control.py`.
- Verified the shared `CommandQueue.put` path preserves ordering: in a coalesced segment `entries()` extends `ptt` before `dedup` (`runtime/_poller_types.py` L989-995), so `put(PttOff())` then `put(SetData*())` drains as `[PttOff, SetData*]` in production — PttOff-before-SET holds, not just in the stdlib-Queue test.

**Frontend (`mod-input-auto.svelte.ts`)**
- `arm_mod_input_restore` sent only inside `autoSetLanModInputForTx`, after all opt-in guards (enabled / data_mode cap / group read / off-LAN) and after the LAN SET, carrying the PREVIOUS source (`pending.source`), never LAN.
- `disarm_mod_input_restore` sent in `restoreModInputAfterTx` only after `if (!p) return;` and BEFORE the mid-TX-change guard, so it fires even when the restore SET is skipped. `applyPendingModInputRestoreOnConnect` unchanged.

**Constraints**
- Opt-in preserved: a disabled / already-LAN / no-pending TX path sends no arm and triggers no teardown enqueue (gated by the same early returns).
- Audio byte path untouched (only 4 files changed: control.py + 1 frontend adapter + 2 test files).
- Manual-mid-TX-change test not weakened — now asserts exactly 1 call = `disarm`, AND `set_data1_mod_input` NOT called.

**Gates (run in /tmp/pr-1796-review at head)**
- `pytest test_web_mod_input_restore + test_web_ptt_readonly + test_command_catalog_docs`: 24 passed
- Full `pytest tests/ --ignore=tests/integration`: 7640 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failures
- `ruff check src/ tests/`: All checks passed
- `mypy src/`: Found 21 errors in 8 files (== pre-existing baseline; none in control.py)
- frontend `vitest mod-input-auto.test.ts`: 20 passed

**Noted (non-blocking):** the teardown net enqueues PttOff under a single-operator assumption (armed ⇒ THIS session auto-keyed). In a hypothetical multi-client setup an abnormal disconnect could unkey/retune for a co-connected operator. Acceptable for a Low-priority safety net that only fires on an abnormal mid-TX disconnect of the auto-keying session; flagging for awareness, not a blocker.

PR #1799, opened 2026-06-11
fix(#MOR-252): refresh radio credentials on reconnect

morozsm:
## Problem (MOR-252)

A radio password is captured ONCE in `CoreRadio.__init__` (`src/rigplane/runtime/radio.py`) and never re-read. It is consumed only at initial auth — `_control_phase.py` builds the login packet from `h._username`/`h._password`. Token renewal re-sends the existing token, and `soft_reconnect` reconnects CI-V without a fresh login. Result: a rotated credential is ignored until process restart.

## Fix

Add `CoreRadio.update_credentials(*, username: str | None = None, password: str | None = None)` which mutates the stored `_username`/`_password`. The login path in `_control_phase._connect_once` already reads `h._password`, so a subsequent full `connect()` authenticates with the new value — no protocol, transport, or audio-path changes.

The docstring explicitly notes that `soft_reconnect()` intentionally reuses the existing session token (does NOT re-auth), so a credential refresh only takes effect on a full reconnect.

Scope: +20 LOC in one source file + one new test file. No new abstractions.

## Tests

`tests/test_credential_refresh.py` (TDD — written first, red, then green):

- `update_credentials` mutates stored values; partial update keeps the other field; no-arg call is a no-op.
- Full-reconnect path: construct a real `IcomRadio`, call `update_credentials(password="new-pass")`, drive `_control_phase._connect_once()` with a minimal fake control transport, and spy on `build_login_packet` — the spy invokes the REAL builder (catching signature drift) and asserts the packet is built with the NEW password.
- `soft_reconnect` does not re-auth: healthy CI-V noop path; `build_login_packet` never called, token unchanged.

## Verification

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **7645 passed, 8 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failed**
- `uv run ruff check src/ tests/` → All checks passed; `ruff format` clean
- `uv run mypy src/` → Success: no issues found in 266 source files
- `uv run lint-imports` → Contracts: 4 kept, 0 broken
- `git diff` → no unintended changes

### DIFF
--- a/src/rigplane/runtime/radio.py
+++ b/src/rigplane/runtime/radio.py
@@ -1110,6 +1110,26 @@ class CoreRadio(ScopeRuntimeMixin, AudioRuntimeMixin, DualRxRuntimeMixin):
         """
         return self._civ_request_tracker.snapshot_stats()
 
+    def update_credentials(
+        self, *, username: str | None = None, password: str | None = None
+    ) -> None:
+        """Update the stored login credentials for future full reconnects.
+
+        The new values are used the next time a full ``connect()`` performs
+        authentication (the login packet is built from the stored
+        credentials). ``soft_reconnect()`` intentionally reuses the existing
+        session token and does NOT re-authenticate, so a rotated credential
+        only takes effect on a full reconnect.
+
+        Args:
+            username: New username, or ``None`` to keep the current one.
+            password: New password, or ``None`` to keep the current one.
+        """
+        if username is not None:
+            self._username = username
+        if password is not None:
+            self._password = password
+
     async def connect(self) -> None:
         """Open connection to the radio and authenticate.
 
--- a/tests/test_credential_refresh.py
+++ b/tests/test_credential_refresh.py
@@ -1,0 +1,114 @@
+"""Tests for CoreRadio.update_credentials (MOR-252).
+
+A rotated radio password must be honoured on the next full reconnect.
+``soft_reconnect`` intentionally reuses the existing session token and
+must NOT re-authenticate.
+"""
+
+import time
+from typing import Any
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from rigplane.core.auth import build_login_packet as real_build_login_packet
+from rigplane.runtime.radio import IcomRadio
+
+
+class _AbortAfterLogin(Exception):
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-252 (credential refresh on reconnect) at head e8ed985d, verified in a fresh detached worktree with `uv sync --all-extras`.

**Acceptance met:**
- `CoreRadio.update_credentials(*, username, password)` mutates the stored `_username`/`_password`. Traced that the full-reconnect path `_control_phase._connect_once()` (radio.py logic at `_control_phase.py:188-193`) reads `h._username`/`h._password` live at login-packet build time, so a subsequent full `connect()` authenticates with the NEW password. Correct and load-bearing.
- `soft_reconnect()` (`_control_phase.py:547+`) never builds a login packet and reuses the existing token; the docstring documents this intentional no-re-auth behavior.
- Test `test_full_reconnect_builds_login_packet_with_new_password` spies on `build_login_packet` but the spy invokes the REAL `build_login_packet` and asserts `len(pkt)==0x80` — signature drift would surface, not be masked by MagicMock. It asserts the captured args are `{username: user, password: new-pass}`. `test_soft_reconnect_reuses_token_and_does_not_reauth` asserts `build_login_packet` is never called and the token is unchanged.
- Thread/async safety: two independent string attribute assignments, no `await` between them — no half-updated state observable under the GIL/asyncio. No lock needed.
- Scope: 2 files, +134/-0. No transport/protocol/audio-byte-path changes.

**Checks (fresh env):**
- `pytest tests/ -q --ignore=tests/integration` → 7645 passed, 8 skipped, 3 deselected, 11 xfailed, 1 xpassed, 0 failed
- `ruff check src/ tests/` → All checks passed
- `mypy src/` → Success: no issues found in 266 source files
- Targeted `tests/test_credential_refresh.py` → 5 passed

Non-blocking follow-up: `update_credentials` is not yet exposed on the `Radio` protocol nor wired to any reconnect call-site (e.g. web). The ticket only asks for the refresh path + tests, so this is acceptance-complete; integration can be a follow-up.

Approved to merge once required checks are green.

PR #1802, opened 2026-06-11
test(#MOR-590): fix 3.11 CI-V-timeout + serial RX failures (B3 + serial B4)

morozsm:
## Summary

Fixes the Python-3.11-only failures in the CI-V-timeout cluster (MOR-632 / MOR-590-B3) and the `test_serial_backend_smoke.py` half of MOR-650 (B4). Test-only changes — no `src/` modifications were needed.

## Root cause (common mechanism, verified per cluster)

Both clusters are downstream of the same interpreter delta: **pre-3.12 `asyncio.wait_for` wraps its awaitable in an extra `Task`** (rewritten in 3.12 via gh-96764 to await inside `asyncio.timeout`). The extra scheduling hop changes task interleaving on 3.11 versus 3.13.

**This is NOT the B1 (#1800) runtime-Protocol/Mock mechanism.** That was explicitly checked: all test doubles here are real classes (`MockTransport` in `tests/test_radio.py`, `FakeAudioBackend`, `_FakeUsbAudioDriver`, `_FakeSerialCivLink`) — no `MagicMock`/`AsyncMock` is involved, so the 3.11 `hasattr`-based `runtime_checkable` `isinstance` delta (gh-102433) cannot steer any code path. Both failures are genuine async-interleaving test fragility.

## Cluster 1 — CI-V timeout (`test_per_receiver_vfo_roundtrip.py`, `test_icom_receiver_tier.py`)

The select-SUB → op → restore-MAIN fallback tests pre-queued **both** ACKs into `MockTransport` upfront. The CI-V rx pump (`_civ_rx_loop`) then races delivery of the second ACK between two consumers:

- the **fire-and-forget ACK sink** registered by the middle frame (`wait_response=False`, `_civ_rx.py:2698`), and
- the **restore-MAIN ACK waiter** (`_civ_rx.py:2733`).

Evidence from `--log-cli-level=DEBUG` runs of the same test on both interpreters:

- **3.13 (pass):** `Dropped 1 stale ACK sink waiter(s) before blocking command` — the orphan ACK landed in the tracker's ACK backlog (`core/civ.py:372`) before the sink registered; the restore attempt still timed out once and only the **retry** (`_dual_rx_runtime.py:144`) consumed the backlog entry. I.e. the test was marginal even on 3.13 (`set_vfo_slot: timeout restoring VFO receiver to MAIN, retrying once` appears in the *passing* run).
[...]

PR #1803, opened 2026-06-11
test(#MOR-590): fix 3.11 web-server + audio Protocol/async-mock failures (B2 + audio B4)

morozsm:
## Linear

MOR-631 (MOR-590-B2: web-server cluster) + the `test_audio_rx_tx_transition.py` half of MOR-650 (B4). Sibling of #1800 (B1), same root-cause family.

## Diagnosis

On Python 3.11, `isinstance(obj, <runtime_checkable Protocol>)` uses `hasattr()`, which a bare `MagicMock` always satisfies. On 3.12+ (gh-102433) it uses `inspect.getattr_static()`, which does not trigger `Mock.__getattr__`, so the same mock fails the check. This makes 3.11 take a different branch than 3.13 in production glue — the failures were 3.11-only:

1. **`test_web_server.py` (30 errors/failures), `test_web_server_coverage.py` (2), `test_security_headers.py` (6)** — in `src/rigplane/web/web_startup.py:97`, the MagicMock radio falsely satisfies `ObservationPollable` on 3.11, so startup calls the auto-generated `radio.create_observation_poller()` and hands its MagicMock `.start()` to `WebServer._spawn` → `TypeError: a coroutine was expected`. On 3.13 the mock fails `ObservationPollable` *and* `StatePollable` (`web_startup.py:116`) and falls through to the `RadioPoller` branch (`web_startup.py:134`), which these tests were written against. (Deleting only `create_observation_poller` would just shift 3.11 onto the equally broken legacy `StatePollable` branch, hence both markers are removed.)

2. **`test_audio_rx_tx_transition.py` (1)** — the inverse direction: the real fake class `_SingleSlotAudioRadio` *should* satisfy `AudioCapable` but failed on 3.11. Its `audio_codec` property returned `self.profile.audio_codec`, and `RadioProfile` has no `audio_codec` attribute (it has `codec_preference`/`tx_codec`), so the getter raised `AttributeError`. 3.11's `hasattr`-based check invokes the getter → member missing → `isinstance` False; 3.13's `getattr_static` sees the property on the class without calling it → True. A latent bug in the fake, exposed only on 3.11.

## Fix (test layer only)

- `tests/test_web_server.py` — new `_del_poller_factory_attrs()` helper (B1's `make_mock_radio()` philosophy): `del radio.create_observation_poller` / `del radio.create_state_poller` so the mock fails both poller Protocols on both interpreters; applied in the `mock_radio` fixture and the one inline server-starting test.
- `tests/test_security_headers.py` — same two deletions in its local `_make_radio()`.
- `tests/test_web_server_coverage.py` — `_StateNotifyRadio.__init__` deletes the same two markers, so both interpreters reach the `StateNotifyCapable`/`RadioPoller` branch the fake is designed for.
- `tests/test_audio_rx_tx_transition.py` — `_SingleSlotAudioRadio.audio_codec` now returns `AudioCodec[self.profile.codec_preference[0]]` (a real enum value from the resolved IC-7610 profile) instead of raising.

**Production source untouched** — `git diff` is 4 test files, +45/−1.

## Verification

- 3.11 target files: `uv run --python 3.11 pytest tests/test_web_server.py tests/test_web_server_coverage.py tests/test_security_headers.py tests/test_audio_rx_tx_transition.py -q` → **294 passed, 2 skipped, 0 failed** (was 4 failed + 35 errors).
- 3.11 full suite (`--ignore=tests/integration`): the B2/audio clusters are gone; the only remaining failures are the known out-of-scope sibling clusters — `test_rigctld_handler.py` (B1, fixed in open #1800), `test_per_receiver_vfo_roundtrip.py` / `test_icom_receiver_tier.py` (B3), `test_serial_backend_smoke.py` (B4, other agent). No new failures introduced.
- 3.13 full suite: **7608 passed, 0 failed** — no regression.
- `ruff check` + `ruff format` clean, `mypy src/` Success, `lint-imports` 4 contracts kept.

### DIFF
--- a/tests/test_audio_rx_tx_transition.py
+++ b/tests/test_audio_rx_tx_transition.py
@@ -374,7 +374,13 @@ class _SingleSlotAudioRadio:
 
     @property
     def audio_codec(self) -> Any:
-        return self.profile.audio_codec
+        # NOTE: must not raise. ``RadioProfile`` has no ``audio_codec``
+        # attribute, so returning ``self.profile.audio_codec`` raised
+        # AttributeError — invisible on Python 3.12+ where the AudioCapable
+        # isinstance() check uses inspect.getattr_static() (gh-102433), but
+        # fatal on 3.11 where the hasattr()-based check invokes this getter
+        # and the AttributeError makes the Protocol check fail.
+        return AudioCodec[self.profile.codec_preference[0]]
 
     @property
     def audio_sample_rate(self) -> int:
--- a/tests/test_security_headers.py
+++ b/tests/test_security_headers.py
@@ -23,6 +23,14 @@ from rigplane.web.server import WebConfig, WebServer
 
 def _make_radio() -> MagicMock:
     radio = MagicMock(name="radio")
+    # Python 3.11 runtime-checkable Protocol isinstance() uses hasattr(), which
+    # a bare MagicMock always satisfies — start_web_server would then spawn the
+    # auto-generated MagicMock poller.start() as a coroutine (TypeError).
+    # Python 3.12+ uses inspect.getattr_static() (gh-102433) and is immune.
+    # Deleting the factory attrs makes both interpreters take the RadioPoller
+    # branch.
+    del radio.create_observation_poller
+    del radio.create_state_poller
     radio.connected = True
     radio.radio_ready = True
     radio.control_connected = True
--- a/tests/test_web_server.py
+++ b/tests/test_web_server.py
@@ -255,6 +255,22 @@ async def _close_ws(writer: asyncio.StreamWriter) -> None:
 # ---------------------------------------------------------------------------
 
 
+def _del_poller_factory_attrs(radio: MagicMock) -> MagicMock:
+    """Make a MagicMock radio fail ObservationPollable/StatePollable checks.
+
+    Python 3.11's runtime-checkable Protocol ``isinstance()`` uses
+    ``hasattr()``, which a bare MagicMock always satisfies, so
+    ``start_web_server`` would take the observation-poller branch and try to
+    spawn the auto-generated MagicMock ``poller.start()`` as a coroutine
+    (``TypeError: a coroutine was expected``). Python 3.12+ uses
+    ``inspect.getattr_static()`` (gh-102433) and is immune. Deleting the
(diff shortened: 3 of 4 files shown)


morozsm:
Agent Review: PASS — Independent review at head 95998ed9 (detached worktree /tmp/review-1803).

**Test-only confirmed:** diff is +45/-1 across 4 test files only (tests/test_audio_rx_tx_transition.py, tests/test_security_headers.py, tests/test_web_server.py, tests/test_web_server_coverage.py). Zero src/ changes.

**Root cause verified against production glue** (src/rigplane/web/web_startup.py:97-149): on 3.11 a bare MagicMock falsely satisfies runtime-checkable ObservationPollable (hasattr-based isinstance), steering into create_observation_poller().start() → _spawn(non-coroutine) TypeError; 3.12+ uses inspect.getattr_static (gh-102433) and is immune. Deleting create_observation_poller/create_state_poller makes both interpreters fall through to the sync RadioPoller branch — matching real Icom CI-V backends. The audio fix returns AudioCodec[profile.codec_preference[0]] (a validated enum-name tuple) instead of a non-existent profile.audio_codec attribute; coherent, not assertion-suppression.

**3.11 target files:** 294 passed, 2 skipped, 0 failed (tests/test_web_server.py tests/test_web_server_coverage.py tests/test_security_headers.py tests/test_audio_rx_tx_transition.py).

**3.11 full** (--ignore=tests/integration): the four target clusters are gone. Remaining failures are out-of-scope only: 92 test_rigctld_handler (B1, fixed on main by #1800 commit 930a7e7b which is AFTER this branch's merge-base 7ad07425 — pure branch staleness), and the B3 siblings (#1802): 2 test_per_receiver_vfo_roundtrip, 2 test_icom_receiver_tier, 1 test_serial_backend_smoke. No NEW failures introduced.

**3.13 full:** 7608 passed, 69 skipped, 0 failed — no regression.

**Lint/type/imports:** ruff clean, mypy Success (267 files), lint-imports 4 kept / 0 broken.

Fix is genuinely test-only, makes the isinstance/branch behavior interpreter-stable, and does not mask a production bug (3.13 startup glue verified correct).

PR #1806, opened 2026-06-11
fix(#MOR-549): normalize ALSA device names so Linux USB-audio topology pairing matches

morozsm:
## Problem (MOR-549)

The Linux no-override topology path `_resolve_linux` in `src/rigplane/audio/_usb_resolve.py` matches the sysfs USB `product` string against `sounddevice` device names with an EXACT string equality (`_pair_audio_cluster_by_name_rank`). But the two sides never agree on real Linux hardware:

- sysfs `product` (via `_sysfs_usb_audio_cards`) yields the bare card name, e.g. `"USB Audio CODEC"`;
- `sounddevice.query_devices()` on ALSA/PortAudio decorates it as `"USB Audio CODEC: Audio (hw:2,0)"`.

So the cluster match always failed, `_resolve_linux` returned `None`, and topology pairing silently fell back — untested because the existing tests mocked plain (undecorated) names.

## Fix

New `_normalize_alsa_device_name()` applied to the sounddevice-side name before the equality check in `_pair_audio_cluster_by_name_rank`:

- strips the trailing ` (hw:X,Y)` / ` (plughw:X,Y)` suffix, reusing the existing `_ALSA_HW_RE` from `rigplane.audio.backend` (anchored variant `_ALSA_NAME_SUFFIX_RE`);
- strips the `": <pcm descriptor>"` tail **only when** the ALSA suffix was present — conservative, never over-strips legitimate names;
- names without the suffix pass through unchanged, so macOS/Windows resolution and existing plain-name behavior are untouched.

No changes to the audio byte path, macOS/Windows resolvers, transport, or protocol.

## Tests (TDD — written first, red → green)

- `TestNormalizeAlsaDeviceName`: full ALSA form, plughw form, suffix without descriptor, plain-name passthrough, colon-without-suffix passthrough (over-strip guard).
- `TestResolveLinuxAlsaDeviceNames`: `_resolve_linux` against the existing sysfs fixture tree with sounddevice mocked using the REALISTIC decorated names (`"USB Audio CODEC: Audio (hw:2,0)"`), asserting each serial port pairs to the correct device — closing the prior test gap. Existing plain-name cases keep passing (no regression).

## Verification

- `uv run pytest tests/ -q --ignore=tests/integration` — 7678 passed, 0 failures
- `uv run ruff check src/ tests/` + `ruff format` — clean
- `uv run mypy src/` — Success: no issues in 267 files
- `uv run lint-imports` — 4 contracts kept, 0 broken

## Note

FINAL acceptance is hardware-gated on a real Linux host with a USB-audio CODEC attached (MOR-549 remains hw-gated); the code fix itself is unit-tested with realistic ALSA enumeration shapes.

### DIFF
--- a/src/rigplane/audio/_usb_resolve.py
+++ b/src/rigplane/audio/_usb_resolve.py
@@ -82,6 +82,8 @@ import subprocess
 from dataclasses import dataclass
 from typing import Any
 
+from rigplane.audio.backend import _ALSA_HW_RE
+
 logger = logging.getLogger(__name__)
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS — Independent review of head 8c87ff11.

**Verification (detached worktree, fresh `uv sync --all-extras`):**
- `pytest tests/ -q --ignore=tests/integration` → **7678 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failures** (known `test_hardware.py` ordering pollution did not appear).
- `ruff check src/ tests/` → All checks passed.
- `mypy src/` → Success, no issues in 267 files.
- `lint-imports` → 4 contracts kept, 0 broken (the new `from rigplane.audio.backend import _ALSA_HW_RE` is intra-`audio/` layer, allowed).

**Correctness assessment:**
- Normalization is **conservative**: `_ALSA_NAME_SUFFIX_RE` is `$`-anchored, so the `": <pcm>"` tail is stripped *only* when a trailing `(hw:X,Y)`/`(plughw:X,Y)` is present. Verified empirically: plain names, macOS/Windows names, and colon-bearing names without an ALSA suffix (`"Radio: Special Edition"`, `"Foo: Bar (hw:2,0) Live"`, `"Card (notawindow:1,2)"`) pass through untouched. No over-stripping of legitimate product names — real ALSA card names do not contain colons; the colon is ALSA-injected `card: pcm` separator.
- **No macOS/Windows regression**: both `_resolve_macos` and `_resolve_linux` route through `_pair_audio_cluster_by_name_rank`; on plain names the helper is a no-op, comparison identical to pre-fix. Existing macOS/plain-name Linux resolve tests stay green.
- `_is_usb_audio_codec` uses substring matching, so decorated names are still detected before clustering — pairing comparison is the only place normalization is applied. Correct placement.
- **Realistic-name test closes the gap**: `TestResolveLinuxAlsaDeviceNames` drives the real `_resolve_linux` with `"USB Audio CODEC: Audio (hw:2,0)"` / `"USB Audio Device: USB Audio (hw:1,0)"` against the bare sysfs `product`, asserting correct rx/tx indices (2 for FTDI, 1 for X6200) — the exact form that pre-fix exact-equality would have missed. Plus a plain-name unit case stays green. Fakes are realistic device dicts via `_make_mock_sd_two_usb_named`, not bug-hiding behavior mocks.

No blockers. Recommend merge once required checks are green.

PR #1815, opened 2026-06-11
fix(validation): contain per-check exceptions so one un-encodable RMVR value can't abort the run (MOR-659)

morozsm:
## Root cause

The first live IC-7610 validation run (`rigplane --model IC-7610 validate --hardware --provider native`, epic MOR-634) crashed with exit 1 and an artifact of 0 checks. The radio's current repeater tone frequency read back as **16.5 Hz** (tone not configured), which is below the encoder's settable band: `commands/tone.py::_encode_tone_freq` requires **67.0-254.1 Hz** and raises a bare `ValueError` otherwise. During the RMVR **restore** step, `set_tone_freq(16.5)` raised that `ValueError`, which:

1. escaped `_guard` (it only mapped rigplane/asyncio exception types),
2. escaped the restore `finally` (`_RESTORE_ERRORS` omitted `ValueError`/`TypeError`),
3. escaped `_run_one_check` and aborted `execute_hardware_checks` — losing the **entire** run's results.

## Fixes (defense in depth — all four)

1. **`_guard`** also catches `ValueError`/`TypeError` → per-check `COMMAND_EXECUTION` FAIL (same shape as the `CommandError` branch). `KeyboardInterrupt`/`SystemExit` still propagate.
2. **`_RESTORE_ERRORS`** now includes `ValueError`/`TypeError`, so the restore-in-`finally` (both RMVR and `mode.set` usages) can never escape.
3. **`execute_hardware_checks`** wraps each `_run_one_check` in a backstop `try/except Exception` that records an errored FAIL result (`failure_domain=COMMAND_EXECUTION`, `error=repr(exc)`) via `_base_result`, so the matrix always completes and the artifact is always produced. `BaseException` (incl. `CancelledError`) still propagates; the per-check INFO log is still emitted.
4. **RMVR restore-safety**: `_read_modify_verify_restore` gains an optional `restorable: Callable[[T], bool] | None` predicate evaluated on the original value **before any write**. If the current value is un-encodable the check returns SKIP with evidence `{original, reason: "current value outside settable range; not mutated to keep the run non-destructive"}` — the harness never writes a test value it cannot restore from. Wired via the value-rule layer (`_VALUE_RULE_RESTORABLE[TONE_FREQ_CYCLE] = 67.0 <= v <= 254.1`), which covers both `tone_freq.set` and `tsql_freq.set`; bounds match the real shared encoder `_encode_tone_freq` (used by both `set_tone_freq` and `set_tsql_freq`).

## Tests (TDD — written first, observed red, exact live crash reproduced)

`tests/validation/test_hardware_crash_containment.py` (8 tests, stateful fake radio with an encoder-faithful setter — no internal mocks):

- un-encodable 16.5 Hz original: run does not raise, every template entry produces a result, `tone_freq.set`/`tsql_freq.set` SKIP with the non-destructive reason, and the setters are **never** called (radio not mutated);
- in-band original (88.5 Hz): RMVR still runs and PASSes (gate doesn't over-skip);
- `_guard` maps `ValueError`/`TypeError` coroutines to `COMMAND_EXECUTION` FAIL; `KeyboardInterrupt` still propagates;
- `_RESTORE_ERRORS` contains `ValueError`/`TypeError`;
- backstop: a handler raising `RuntimeError` becomes an errored FAIL with timestamps while the other entries still run; `CancelledError` propagates.

## Verification

- `pytest tests/ --ignore=tests/integration`: **7751 passed, 0 failed**
- `mypy src/`: Success, no issues (283 files)
- `ruff check` + `ruff format`: clean
- `lint-imports`: 5 kept, 0 broken
- Dry-run golden gates IC-7610 / X6200 / FTX-1: all exit 0

## Goldens

**No golden regenerated** — the restorable predicate is wired in the hardware execution path only (value-rule layer in `hardware.py`); the registry and dry-run declarations are unchanged.

Not in scope: the `preamp.set` DIGI-SEL `CommandError` (radio.py:3269) — intentional and correctly recorded as environmental.

### DIFF
--- a/src/rigplane/validation/hardware.py
+++ b/src/rigplane/validation/hardware.py
@@ -72,6 +72,8 @@ _LOGGER = logging.getLogger("rigplane.validation")
 T = TypeVar("T")
 
 # Exceptions handled by the best-effort restore path; never escapes ``finally``.
+# ``ValueError``/``TypeError`` are included because command encoders raise them
+# for out-of-band values (MOR-659: a 16.5 Hz tone readback aborted a whole run).
 _RESTORE_ERRORS = (
     asyncio.TimeoutError,
     RigTimeoutError,
@@ -80,6 +82,8 @@ _RESTORE_ERRORS = (
     AuthenticationError,
     CommandError,
     OSError,
+    ValueError,
+    TypeError,
 )
 
 # Capability tag -> capability Protocol used for runtime ``isinstance`` checks.
@@ -139,14 +143,26 @@ async def execute_hardware_checks(
     by_level: dict[ValidationLevel, list[CheckResult]] = {}
     for entry in template.entries:
         started = _utcnow_iso()
-        result = await _run_one_check(
-            radio,
-            entry,
-            safety,
-            allow_writes=allow_writes,
-            per_check_timeout=per_check_timeout,
-            write_only_capabilities=write_only_capabilities,
-        )
+        try:
+            result = await _run_one_check(
+                radio,
+                entry,
+                safety,
+                allow_writes=allow_writes,
+                per_check_timeout=per_check_timeout,
+                write_only_capabilities=write_only_capabilities,
+            )
+        except Exception as exc:
+            # Backstop (MOR-659): one raising check must never abort the
+            # matrix — the artifact is ALWAYS produced. ``BaseException``
+            # (KeyboardInterrupt/SystemExit/CancelledError) still propagates.
+            _LOGGER.exception("validate check %s raised unexpectedly", entry.check_id)
+            result = _base_result(
+                entry,
+                CheckStatus.FAIL,
+                failure_domain=FailureDomain.COMMAND_EXECUTION,
+                error=repr(exc),
+            )
         finished = _utcnow_iso()
         result = dataclasses.replace(result, started_at=started, finished_at=finished)
         _LOGGER.info(
@@ -360,6 +376,15 @@ async def _guard(
             failure_domain=FailureDomain.COMMAND_EXECUTION,
             error=str(exc),
         )
+    except (ValueError, TypeError) as exc:
+        # Command encoders raise these for out-of-band values (MOR-659:
+        # set_tone_freq(16.5) -> ValueError); contain them per-check.
+        return None, _base_result(
+            entry,
+            CheckStatus.FAIL,
+            failure_domain=FailureDomain.COMMAND_EXECUTION,
+            error=str(exc),
+        )
 
 
 def _default_equal(a: T, b: T) -> bool:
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of PR #1815 (MOR-659 crash containment). Verified in an isolated worktree on PR head 75046478, not self-review.

## Scope
\`git diff origin/main...HEAD --stat\`: exactly 2 files, +334/-8 — \`src/rigplane/validation/hardware.py\` (+71) and the new \`tests/validation/test_hardware_crash_containment.py\` (+271). No registry/golden/schema changes, as claimed. The restorable predicate lives in the execution path, not the declaration, so no golden regen is required (confirmed below).

## Four containment layers verified
- **Layer 1 — \`_guard\`** (hardware.py:379) maps \`ValueError\`/\`TypeError\` -> per-check COMMAND_EXECUTION \`CheckStatus.FAIL\`, same shape as the \`CommandError\` branch. \`KeyboardInterrupt\`/\`SystemExit\`/\`CancelledError\` are not in any except clause -> still propagate.
- **Layer 2 — \`_RESTORE_ERRORS\`** (hardware.py:77) adds \`ValueError\`/\`TypeError\`; both \`finally\` restore paths (generic RMVR :530 and mode.set :774) reference the same tuple. Neither can escape.
- **Layer 3 — backstop** (hardware.py:155): \`except Exception\` (NOT BaseException) wraps the \`_run_one_check\` call, builds an errored result via \`_base_result\` (check_id/level/declaration preserved), keeps the INFO log, lets CancelledError/KeyboardInterrupt propagate, and one raising entry does not skip subsequent entries.
- **Layer 4 — restore-safety** (hardware.py:473): when \`restorable(original)\` is False the check returns SKIP with a non-destructive reason BEFORE \`make_changed\`/\`write\` — setter never called. Predicate \`67.0 <= float(v) <= 254.1\` (hardware.py:1165) MATCHES the real encoder \`commands/tone.py::_encode_tone_freq:28\` exactly. BOTH \`tone_freq.set\` and \`tsql_freq.set\` use \`ValueRule.TONE_FREQ_CYCLE\` (registry/_tone.py:45,77) and both \`set_tone_freq\`/\`set_tsql_freq\` route through the SAME \`_encode_tone_freq\` (tone.py:161,210) — no encoder-bounds mismatch. In-band 88.5 still runs full RMVR (not over-skipped).

## Tests real, not vacuous
8 tests reproduce the live 16.5 Hz crash (no exception escapes, all 4 entries produce results, tone/tsql SKIP, setters never called), plus \`_guard\` mapping, \`_RESTORE_ERRORS\` membership, backstop loop-continuation, and CancelledError propagation. Spot-check: reverting the Layer-4 wiring (\`restorable=None\`) in my worktree turned \`test_unencodable_tone_freq_never_crashes_the_run\` red (FAIL vs expected SKIP) while the run still didn't crash (Layers 1-3 independent) — then restored via \`git checkout\`, tree clean.

## Gates (all green)
- \`pytest tests/ -q --ignore=tests/integration\`: \`7751 passed, 8 skipped, 3 deselected, 11 xfailed, 1 xpassed in 233.90s\`
- \`mypy src/\`: \`Success: no issues found in 283 source files\`
- \`ruff check src/ tests/\`: \`All checks passed!\`
- \`lint-imports\`: \`Contracts: 5 kept, 0 broken.\`
- Dry-run golden gates: IC-7610 exit=0, X6200 exit=0, FTX-1 exit=0; no golden file in the diff.

Correct, minimal, well-tested, non-destructive. Clear to merge once required checks are green.

morozsm:
Agent Review: PASS

Focused re-confirmation of the single new commit `ddb8bc9c` (the prior PASS already covered the production crash-containment logic in `src/rigplane/validation/hardware.py` at `75046478`, which is unchanged). This refreshes the gate for head `ddb8bc9c`.

**Delta is test-only.** `git diff 75046478...HEAD --stat` shows ONLY:
```
tests/validation/test_hardware_crash_containment.py | 17 ++++++++++++++---
```
No production file changed since the prior PASS, so that PASS still covers the fix.

**The invariant is still genuinely tested, not weakened to a tautology.** The renamed `test_guard_does_not_swallow_base_exception_interrupts` raises a custom `_ControlInterrupt(BaseException)` and asserts `pytest.raises(_ControlInterrupt)` propagates out of `_guard`. `_guard` catches only a closed set of `Exception`-derived rig errors (`asyncio.TimeoutError`, `RigTimeoutError`, `RigConnectionError`, `AuthenticationError`, `CommandError`, and the broadest `RigplaneError` — which is `class RigplaneError(Exception)`). Since `_ControlInterrupt` derives from `BaseException` and not `Exception`, no clause catches it and it propagates — exactly mirroring KeyboardInterrupt/SystemExit-tier behavior. The substitution is sound: it preserves the BaseException-non-swallow invariant while avoiding xdist's interception of a real KeyboardInterrupt as a worker shutdown signal.

**xdist worker crash is resolved — ran the exact CI condition.**
- Target file: `pytest tests/validation/test_hardware_crash_containment.py -n auto -q` → `8 passed in 0.70s` (10 workers, no worker crash)
- Broader group: `pytest tests/validation/ -n auto -q` → `170 passed in 17.84s` (10 workers, no worker crashes)

**ruff:** `uv run ruff check tests/validation/test_hardware_crash_containment.py` → All checks passed.

The `quick` CI `-n auto` failure cause (real KeyboardInterrupt crashing the xdist worker) is gone.

PR #1816, opened 2026-06-11
fix(scope): send range+edge selector in get_scope_fixed_edge (MOR-662)

morozsm:
## What
Live IC-7610 validation FAILed `scope_fixed_edge.read` with `timeout after 8.0s`. Raw CI-V trace root-caused it: `get_scope_fixed_edge` sent a **bare** `0x27 0x1E` query, which the radio **NAKs (0xFA)**; `_send_civ_expect` then waited 8s for a data frame that never came. The IC-7610 requires a `<range><edge>` selector (same keying as `scope_set_fixed_edge`):

```
TX fe fe 98 e0 27 1e fd        -> RX fe fe e0 98 FA fd                              (NAK — bare rejected)
TX fe fe 98 e0 27 1e 01 01 fd  -> RX fe fe e0 98 27 1e 01 01 00 00 50 00 00 00 00 50 01 00 fd  (data)
TX fe fe 98 e0 27 1e 01 fd     -> RX fe fe e0 98 FA fd                              (NAK — edge-only rejected)
```

## Fix
- `commands/scope.py::get_scope_fixed_edge` — add `range_index: int = 1`, `edge: int = 1` kwargs; emit the 2-byte BCD selector so the frame is `... 27 1e <range> <edge> fd`.
- `runtime/_scope_runtime.py::get_scope_fixed_edge` — pass the selector.
- `runtime/_civ_rx.py` — the GET/SET frame-classification heuristic for sub `0x1E` now treats a `<=2`-byte payload as the GET (the SET carries the 12-byte bounds payload), so the outgoing selector-bearing GET is classified correctly.

## Verification
- pytest (xdist, no integration): **7754 passed, 0 failed**
- mypy: Success (283 files) · ruff: clean · lint-imports: 5 kept, 0 broken
- All three dry-run golden gates exit 0; **no golden changed** (read-path only — dry-run does not execute reads)
- New unit tests: builder emits `... 27 1e 01 01 fd`; parser decodes the real captured response; out-of-range edge raises.

## Scope
`commands/scope.py`, `runtime/_scope_runtime.py`, `runtime/_civ_rx.py`, tests. No overlap with the sibling MOR-660/661 PR (hardware.py / ic7610.toml / goldens).

Will be confirmed on hardware in the post-merge live IC-7610 re-run.

### DIFF
--- a/src/rigplane/commands/scope.py
+++ b/src/rigplane/commands/scope.py
@@ -694,13 +694,32 @@ def scope_set_vbw(
 
 
 def get_scope_fixed_edge(
-    to_addr: int, from_addr: int = CONTROLLER_ADDR, cmd_map: CommandMap | None = None
+    to_addr: int,
+    from_addr: int = CONTROLLER_ADDR,
+    cmd_map: CommandMap | None = None,
+    *,
+    range_index: int = 1,
+    edge: int = 1,
 ) -> bytes:
+    # The IC-7610 NAKs (0xFA) a bare 0x27 0x1E query — it requires a
+    # <range_index><edge> selector (1-byte BCD each). Mirror the setter's
+    # encoding so the radio answers with the fixed-edge start/end (MOR-662).
+    _validate_scope_range("scope fixed edge range", range_index, 1, 99)
+    _validate_scope_range("scope fixed edge", edge, 1, 4)
+    selector = bcd_encode_value(range_index, byte_count=1) + bcd_encode_value(
+        edge, byte_count=1
+    )
     if cmd_map is not None:
         return _build_from_map(
-            cmd_map, "get_scope_fixed_edge", to_addr=to_addr, from_addr=from_addr
+            cmd_map,
+            "get_scope_fixed_edge",
+            to_addr=to_addr,
+            from_addr=from_addr,
+            data=selector,
         )
-    return _scope_query(_SUB_SCOPE_FIXED_EDGE, to_addr=to_addr, from_addr=from_addr)
+    return build_civ_frame(
+        to_addr, from_addr, _CMD_SCOPE, sub=_SUB_SCOPE_FIXED_EDGE, data=selector
+    )
 
 
 def scope_set_fixed_edge(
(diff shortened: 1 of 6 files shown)


morozsm:
Agent Review: PASS

Independent review of `codex/mor-662-scope-fixed-edge-read` (head faaa7e2e). Verified, not trusted.

**Scope** — diff touches only the 4 declared areas (`commands/scope.py`, `runtime/_civ_rx.py`, `runtime/_scope_runtime.py`, + 3 test files; 6 files, +69/-13). No `validation/hardware.py`, no `rigs/*.toml`, no golden files — no overlap with the concurrent MOR-660/661 work.

**Builder** — `get_scope_fixed_edge(to_addr=0x98, range_index=1, edge=1)` emits exactly:
`fe fe 98 e0 27 1e 01 01 fd` (ends with `27 1e 01 01 fd`). Selector is 1-byte BCD each, mirroring `scope_set_fixed_edge`. Out-of-range edge (0,5) and range (0,100) both raise ValueError. Default kwargs produce the same range=1/edge=1 frame. `cmd_map` branch passes `data=selector` correctly.

**Parser** — fed the real captured bytes `fe fe e0 98 27 1e 01 01 00 00 50 00 00 00 00 50 01 00 fd`:
`ScopeFixedEdge(range_index=1, edge=1, start_hz=500000, end_hz=1500000)`. My independent BCD decode of `00 00 50 00 00` / `00 00 50 01 00` gives 500000 / 1500000 — parser agrees.

**_civ_rx.py heuristic** — the `sub == 0x1E -> len(data) <= 2` branch is the GET (expects-response) classifier. Verified at runtime: (a) 2-byte GET selector -> True; (b) 12-byte SET payload -> False; (c) other subs unaffected — prefix subs (0x1D: 1B->True, 2B->False) and bare subs (0x1B: 0B->True, 1B->False) classify unchanged. Boundary `<= 2` is correct: the only legitimate 0x1E frames are the 2-byte GET and the 12-byte SET (range+edge+5+5 BCD); nothing exists in the 3-11 byte gap, so `<= 2` cannot collide with a SET. `== 2` would also work; `<= 2` is the safe lenient choice.

**Suite (xdist, foreground, to completion):**
`7754 passed, 8 skipped, 11 xfailed, 1 xpassed, 1 warning in 74.14s` — 0 failures, no worker crashes. Integration scope test: 14 passed.

**Types/lint/imports:** mypy `Success: no issues found in 283 source files`; ruff `All checks passed!`; lint-imports `Contracts: 5 kept, 0 broken`.

**Goldens:** none in diff; all three dry-run gates exit 0 (IC-7610=0, X6200=0, FTX-1=0).

Clean, minimal, correct fix matching the captured CI-V trace. Approved to merge once required checks are green.

PR #1818, opened 2026-06-11
fix(validation): coerce non-primitive check evidence to JSON-safe types so artifact emission never crashes (MOR-663)

morozsm:
## The bug (live IC-7610 run, epic MOR-634)

The full live IC-7610 matrix ran all 70 checks but **crashed while writing the artifact**:

```
TypeError: Object of type ScopeFixedEdge is not JSON serializable
```

The MOR-662 fix made `scope_fixed_edge.read` return a real `ScopeFixedEdge` dataclass (instead of timing out). That object lands in the check's `evidence` dict, and the final `json.dumps(artifact.to_dict())` had no `default=` handler — so it raised and aborted artifact emission **after the whole matrix had already run**.

## Root cause

`CheckResult.to_dict()` in `src/rigplane/validation/schema.py` did `payload["evidence"] = dict(self.evidence)` — a **shallow** copy. Non-JSON-primitive evidence values (a dataclass like `ScopeFixedEdge`, an `Enum`, `bytes`) passed through unconverted, so `json.dumps` could not encode them.

## Fix (defense in depth — the harness must ALWAYS be able to emit its artifact)

1. **Recursive `_jsonify()` helper** in `schema.py`, applied per evidence value in `CheckResult.to_dict()`:
   - dataclass → `to_dict()` if present, else `dataclasses.asdict` (then recurse);
   - `Enum` → `.value`;
   - `bytes`/`bytearray` → hex string;
   - `set`/`tuple`/`frozenset` → list (recurse elements);
   - `dict` → recurse values (non-str keys coerced via `str`);
   - `list` → recurse elements;
   - primitives (str/int/float/bool/None) → **unchanged** (so primitive-only evidence keeps its exact prior shape and goldens do not move).
2. **Belt-and-suspenders**: the three `json.dumps(...)` calls in `cli/_validate.py` (regen-golden write, `--output`, `--json`) now pass `default=_json_default` (dataclass → `asdict`, else `str`), so an unforeseen object can never again abort emission (mirrors MOR-659's "always produce output" principle).
3. `scope_fixed_edge.read` itself is **unchanged** — the general coercion is the fix.

## Tests (`tests/validation/test_schema_jsonify.py`, TDD)

- `CheckResult` evidence with a real `ScopeFixedEdge` + `Enum` + `bytes` → `to_dict()` is `json.dumps`-able; ScopeFixedEdge becomes a dict of its exact fields, Enum its `.value`, bytes its `.hex()`.
- Nested containers (list/set/tuple, nested dataclass/bytes) recurse correctly.
- A full `ValidationArtifact` carrying such a check → `json.dumps(artifact.to_dict())` does not raise (the exact live-run path).
- Regression: primitive-only evidence is byte-identical to the previous shallow-copy output.

## Verification (all foreground, to completion)

- `pytest tests/ --ignore=tests/integration -n auto` → **7763 passed, 8 skipped, 11 xfailed, 1 xpassed**
- `mypy src/` → **Success: no issues found in 283 source files**
- `ruff check` → **All checks passed!**; `ruff format` → 600 files unchanged
- `lint-imports` → **5 kept, 0 broken**
- Dry-run golden gates IC-7610 / X6200 / FTX-1 each **exit 0**; **no golden changed** (`git diff tests/golden/` empty) — dry-run produces no non-primitive evidence.

### DIFF
--- a/src/rigplane/cli/_validate.py
+++ b/src/rigplane/cli/_validate.py
@@ -543,6 +543,19 @@ async def _run_hardware_both(
     return _apply_golden_flags(art_n, args, rc_n if rc_n else rc_h)
 
 
+def _json_default(obj: object) -> object:
+    """Last-resort JSON coercion so artifact emission can never crash.
+
+    ``CheckResult.to_dict`` already coerces evidence to JSON-safe types; this
+    is belt-and-suspenders for any unforeseen object that still reaches
+    ``json.dumps`` (mirrors the "always produce output" principle): a
+    dataclass becomes its ``asdict``, anything else its ``str``.
+    """
+    if dataclasses.is_dataclass(obj) and not isinstance(obj, type):
+        return dataclasses.asdict(obj)
+    return str(obj)
+
+
 def _apply_golden_flags(
     artifact: ValidationArtifact, args: argparse.Namespace, exit_code: int
 ) -> int:
@@ -560,7 +573,10 @@ def _apply_golden_flags(
         normalized = normalize_artifact(artifact.to_dict())
         path = Path(regen_path)
         path.parent.mkdir(parents=True, exist_ok=True)
-        path.write_text(json.dumps(normalized, indent=2) + "\n", encoding="utf-8")
+        path.write_text(
+            json.dumps(normalized, indent=2, default=_json_default) + "\n",
+            encoding="utf-8",
+        )
         print(f"Golden written to: {regen_path}", file=sys.stderr)
 
     gate_path = getattr(args, "gate", None)
@@ -597,11 +613,11 @@ def _stamp_overrides(
 def _emit_artifact(artifact: ValidationArtifact, args: argparse.Namespace) -> None:
     """Render the artifact: file on --output, JSON on --json, else human text."""
     if args.output:
-        text = json.dumps(artifact.to_dict(), indent=2)
+        text = json.dumps(artifact.to_dict(), indent=2, default=_json_default)
         Path(args.output).write_text(text + "\n", encoding="utf-8")
         print(f"Artifact written to: {args.output}", file=sys.stderr)
     elif args.json:
-        print(json.dumps(artifact.to_dict(), indent=2))
+        print(json.dumps(artifact.to_dict(), indent=2, default=_json_default))
     else:
         print(human_summary(artifact))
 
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-663 artifact-evidence serialization fix. Verified on a clean worktree at PR HEAD `854fe552` (Python 3.13.5).

**Scope** — exactly 3 files, nothing stray: `validation/schema.py` (+43), `cli/_validate.py` (+22, -3), new `tests/validation/test_schema_jsonify.py` (+137).

**`_jsonify` correctness** — verified by reading + empirical exec against the real `ScopeFixedEdge` (core/types.py:140) and `AgcMode` enum:
- dataclass -> `to_dict()` if present else `dataclasses.asdict` (recursed); Enum -> `.value`; bytes/bytearray -> hex; set/tuple/frozenset -> list; dict -> recursed (non-str keys coerced); list -> recursed; primitives unchanged.
- **Byte-identity confirmed**: primitive-only evidence (`dict({...})` vs `{k:_jsonify(v)}`) produces identical `json.dumps` output, so no golden can shift. StrEnum/IntEnum are caught by the first primitive branch and pass through unchanged (`"pass"`/`1`) — matching the old `dict(evidence)` path exactly. tuple -> list serializes to the same JSON array `json.dumps` already produced. The regression test `test_primitive_evidence_shape_is_unchanged` asserts BOTH dict-equality and `sort_keys` byte-identity.

**Real bug fixed** — independently built a `CheckResult` with `evidence={"readback": ScopeFixedEdge(...)}`, called `.to_dict()` + `json.dumps`: no raise, ScopeFixedEdge -> `{range_index, edge, start_hz, end_hz}`. Same verified for plain Enum -> value and bytes -> hex.

**Belt-and-suspenders** — all THREE `json.dumps` sites in `_validate.py` carry `default=_json_default` (regen-golden L577, --output L616, --json L620); no other sites. `default=` coerces dataclass->asdict / else str, and is only invoked by CPython for otherwise-unserializable objects, so it cannot alter already-serializable payloads.

**Gates / suites**
- Full suite (xdist, `-n auto`, no integration): **7763 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failed** (72.4s).
- `mypy src/`: Success, no issues (283 files). `ruff check src/ tests/`: All checks passed. `lint-imports`: 5 kept, 0 broken.
- Goldens UNCHANGED: `git diff origin/main...HEAD -- tests/golden/` is empty.
- Dry-run golden gates: **IC-7610 exit 0, X6200 exit 0, FTX-1 exit 0**.

No regressions, no scope creep, the fix is minimal and the key safety property (no golden movement) holds. Approving.

PR #1819, opened 2026-06-11
fix(validation): preamp.set clears and restores the DIGI-SEL prerequisite before testing PREAMP (MOR-665)

morozsm:
### The live FAIL
The live IC-7610 validation run (epic MOR-634) FAILed `preamp.set`:

> `Cannot set preamp level 1: DIGI-SEL (IP+) is ON. PREAMP and DIGI-SEL are mutually exclusive`

The radio's preflight in `runtime/radio.py` correctly enforces PREAMP/DIGI-SEL mutual exclusion (setting PREAMP while DIGI-SEL is ON raises `CommandError`). So the **harness**, not the radio, must adapt: it should temporarily clear the blocking prerequisite, test PREAMP, then restore it — rather than just reporting FAIL.

### Clear-and-restore approach
`preamp.set` (the named handler in `_SUPPORTED_HANDLERS`) is now prerequisite-aware, localized to that one check:

- Before the PREAMP read-modify-verify-restore, read DIGI-SEL via `radio.get_digisel()`.
- **DIGI-SEL ON** → `set_digisel(False)`, run the normal PREAMP RMVR, then restore `set_digisel(True)` in a `finally` that **never raises** (best-effort, same spirit as the RMVR restore); the restore is verified by re-reading.
- **DIGI-SEL OFF** → no toggle; behaviour is exactly as before.
- **Unsupported/erroring** `get_digisel`/`set_digisel` (missing attr, raises, NAK) → fall back to today's plain RMVR; no crash (MOR-659 containment still applies, but this is a clean handled path).

No generic precondition framework was added (preamp/DIGI-SEL is the only known case), and **`runtime/radio.py`'s preflight is untouched** — that mutual-exclusion is correct radio behaviour.

### Evidence keys
- `digisel_was_on: <bool>` — always recorded.
- `digisel_restored: <bool>` — recorded only when DIGI-SEL was toggled, verified by re-read.
- Best-effort error breadcrumbs when relevant: `digisel_clear_error`, `digisel_restore_error`, `digisel_restore_read_error`.

### Tests (TDD, `tests/validation/test_hardware.py`)
- DIGI-SEL ON → clears it, PREAMP RMVR PASSes, DIGI-SEL restored ON; evidence `digisel_was_on: true`, `digisel_restored: true`; `set_digisel` called `False` then `True`.
- DIGI-SEL OFF → no toggle (`set_digisel` not called), RMVR runs as today; evidence `digisel_was_on: false`.
- Restore-on-failure → PREAMP write/readback fails mid-check, DIGI-SEL is STILL restored to ON (`False` then `True`).
- Unsupported `get_digisel` (absent, and raising) → graceful fallback, no crash, no DIGI-SEL evidence, `set_digisel` not called.

### Verification (all green)
1. `pytest tests/ --ignore=tests/integration -n auto` — **7763 passed**, 8 skipped, 11 xfailed, 1 xpassed
2. `mypy src/` — Success (283 files)
3. `ruff check` + `ruff format` — clean (599 files)
4. `lint-imports` — 5 kept, 0 broken
5. Dry-run golden gates (`tests/test_validation_gating.py`, incl. IC-7610/X6200/FTX-1 exit-zero + committed IC-7610 golden) — 18 passed

### No golden changed
This affects **live execution only** — dry-run does not actuate, so no golden should change. Confirmed: `git diff tests/golden/` is empty. Only `src/rigplane/validation/hardware.py` and `tests/validation/test_hardware.py` changed (no overlap with the sibling MOR-663 PR's `schema.py`/`_validate.py`).

### DIFF
--- a/src/rigplane/validation/hardware.py
+++ b/src/rigplane/validation/hardware.py
@@ -958,24 +958,32 @@ async def _check_preamp_set(
     try:
         result = await _preamp_rmvr(radio, entry, per_check_timeout=per_check_timeout)
     finally:
+        # Best-effort restore that must never raise (mirrors
+        # ``_read_modify_verify_restore``): ``_guard`` maps the rig-error family,
+        # and the surrounding ``except _RESTORE_ERRORS`` additionally contains a
+        # bare ``OSError`` from the UDP send path so a mid-restore LAN drop can
+        # never escape and leave DIGI-SEL in the wrong state.
         restored = False
-        _, restore_fail = await _guard(
-            cast(Awaitable[None], set_digisel(True)),
-            entry,
-            per_check_timeout=per_check_timeout,
-        )
-        if restore_fail is not None:
-            extra["digisel_restore_error"] = restore_fail.error
-        else:
-            ctrl, ctrl_fail = await _guard(
-                cast(Awaitable[bool], get_digisel()),
+        try:
+            _, restore_fail = await _guard(
+                cast(Awaitable[None], set_digisel(True)),
                 entry,
                 per_check_timeout=per_check_timeout,
             )
-            if ctrl_fail is not None:
-                extra["digisel_restore_read_error"] = ctrl_fail.error
+            if restore_fail is not None:
+                extra["digisel_restore_error"] = restore_fail.error
             else:
-                restored = bool(ctrl)
+                ctrl, ctrl_fail = await _guard(
+                    cast(Awaitable[bool], get_digisel()),
+                    entry,
+                    per_check_timeout=per_check_timeout,
+                )
+                if ctrl_fail is not None:
+                    extra["digisel_restore_read_error"] = ctrl_fail.error
+                else:
+                    restored = bool(ctrl)
+        except _RESTORE_ERRORS as exc:
+            extra["digisel_restore_error"] = str(exc)
         extra["digisel_restored"] = restored
 
     return _merge_evidence(result, extra)
--- a/src/rigplane/validation/hardware.py
+++ b/src/rigplane/validation/hardware.py
@@ -871,16 +871,22 @@ async def _check_af_level_set(
     )
 
 
-async def _check_preamp_set(
+def _merge_evidence(result: CheckResult, extra: dict[str, object]) -> CheckResult:
+    """Return ``result`` with ``extra`` merged into its evidence dict."""
+    if not extra:
+        return result
+    merged = dict(result.evidence)
+    merged.update(extra)
+    return dataclasses.replace(result, evidence=merged)
+
+
+async def _preamp_rmvr(
     radio: Radio,
     entry: CapabilityDeclarationEntry,
     *,
-    allow_writes: bool,
     per_check_timeout: float,
 ) -> CheckResult:
-    gate = _write_gate(radio, entry, allow_writes=allow_writes)
-    if gate is not None:
-        return gate
+    """The plain PREAMP read-modify-verify-restore cycle (no prerequisites)."""
     antenna = cast(AntennaControlCapable, radio)
     return await _read_modify_verify_restore(
         radio,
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of `codex/mor-665-preamp-digisel-precondition` @ 084911ad (MOR-665). I did not write this; everything below was verified in an isolated worktree.

## Scope
`git diff origin/main...HEAD --name-only` = exactly two files: `src/rigplane/validation/hardware.py`, `tests/validation/test_hardware.py`. **`runtime/radio.py` and its PREAMP/DIGI-SEL mutual-exclusion preflight are untouched.** Tree clean.

## Restore safety (the critical property) — verified
- (a) When DIGI-SEL was ON, the PREAMP RMVR runs inside `try` and `set_digisel(True)` always runs in the `finally` — restored even if the RMVR write/readback fails, and even on cancellation (the `finally` runs before propagation).
- (b) `_preamp_rmvr` (= `_read_modify_verify_restore`) maps all radio I/O through `_guard`, which converts the documented error families to CheckResults and never raises — so `result` is always bound at the final `return`; no UnboundLocalError path.
- (c) DIGI-SEL OFF originally → `set_digisel` is NOT called (test asserts `assert_not_called()`).
- (d) `get_digisel` missing/raising → graceful fallback to plain RMVR, no `set_digisel`, no crash (broad `except Exception` guards the read).
- Evidence records `digisel_was_on` always and `digisel_restored`/error keys when toggled. Uses the real `radio.get_digisel()/set_digisel()` protocol methods.

## Tests — real, not vacuous (4 new)
- clears+restores when ON: PASS, `set_digisel` call order asserted `[False, True]`, final state ON, RMVR original=0/changed=1/restored.
- no-toggle when OFF: `set_digisel.assert_not_called()`, state stays OFF.
- **restore-on-failure**: no-op PREAMP write → FAIL/READBACK, yet `digisel_restored=True`, final ON, order `[False, True]`. Restore-on-failure IS covered.
- unsupported get_digisel (absent + raising): fallback, no DIGI-SEL evidence, no `set_digisel`.
All would fail against main (old handler emits no DIGI-SEL evidence and lacks the prerequisite clear).

## Gates / suite (foreground, to completion)
- `pytest tests/ --ignore=tests/integration -n auto`: **7763 passed, 8 skipped, 11 xfailed, 1 xpassed, 0 failed** (81s).
- `mypy src/`: Success, no issues (283 files). `ruff check`: All checks passed. `lint-imports`: 5 kept, 0 broken.
- Goldens: `git diff origin/main...HEAD -- tests/golden/` empty.
- Dry-run gates: **IC-7610 exit=0, X6200 exit=0, FTX-1 exit=0**.

## Non-blocking recommendation (hardening)
The new `finally` restore block (set_digisel(True) + readback) relies solely on `_guard`, which does NOT catch `OSError` — whereas the established sibling `_read_modify_verify_restore` wraps its restore body in `try/except _RESTORE_ERRORS` (which includes `OSError`). A bare synchronous `OSError` from the UDP `sendto` path during restore could escape the `finally` and mask `result`. Dominant LAN failures (RigConnectionError/RigTimeoutError/CommandError) ARE guarded, so this is an edge case, but for consistency with the file's own pattern consider wrapping the finally body in `try/except _RESTORE_ERRORS` (and recording `digisel_restore_error`). Not required to merge.

Verdict: **PASS** — may merge once required checks are green and this PASS is fresh for the current head.
🤖 Independent agent review (Claude Opus 4.8)

morozsm:
Agent Review: PASS

Focused re-confirmation of the post-PASS hardening delta (new head `deda9885`). The prior independent PASS covers everything else; this refreshes the gate for the current head.

**Delta scope — confined to exactly the two expected files:**
- `src/rigplane/validation/hardware.py` (preamp DIGI-SEL `finally` block in `_check_preamp_set`)
- `tests/validation/test_hardware.py` (one new OSError test)

No other files touched (+49/-13).

**Hardening is correct.** The restore body in the `finally` is now wrapped in `try: ... except _RESTORE_ERRORS as exc:`. `_RESTORE_ERRORS` includes `OSError` (hardware.py:84). `restored = False` is set before the `try`, so a bare `OSError` from the `set_digisel(True)` UDP send path during restore is caught, recorded as `digisel_restore_error = str(exc)`, leaves `digisel_restored` False, and the handler returns its result via `_merge_evidence` rather than propagating out of the `finally` and aborting the run. Happy-path (no OSError) behavior is byte-for-byte unchanged — the existing logic is the new try-block body verbatim.

**New test is real.** `test_preamp_digisel_restore_oserror_never_escapes` makes the restore `set_digisel(True)` raise `OSError`, asserts `execute_hardware_checks` does NOT raise, `digisel_restored is False`, `digisel_restore_error` present, and `calls == [False, True]` (cleared then restore attempted). Without the try/except the OSError would escape the finally and abort `execute_hardware_checks`, so the test would fail — it genuinely exercises the fix.

**Gates (fresh worktree at `deda9885`):**
- pytest (xdist, foreground, to completion): `7764 passed, 8 skipped, 11 xfailed, 1 xpassed`, 0 failed
- `ruff check src/ tests/`: All checks passed
- `mypy src/`: Success, no issues (283 files)
- `lint-imports`: 5 kept, 0 broken
- Goldens: `git diff origin/main...HEAD -- tests/golden/` empty

Approved to merge once required checks are green.

PR #1828, opened 2026-06-11
feat(#MOR-678): add IC-7610 MOD-input routing (1A 05 0089-0094) RMVR validation guard

morozsm:
## What

Adds an RMVR round-trip validation check for the **IC-7610 DATA-OFF MOD-input routing** (CI-V `1A 05 0091-0094`) — the surface behind "web voice TX = noise/squeal" (documented fix: DATA-OFF MOD=LAN). A regression that silently flipped MOD back to MIC was invisible to the harness; now it round-trips and restores.

Backend already exposes `get/set_data_off_mod_input` — this is pure registry wiring (no backend gap).

## Check

- id `mod_input.set` — `RMVR_SAFE_WRITE`, capability `mod_input_routing`, ops `get/set_data_off_mod_input`, `FailureDomain.AUDIO`.
- New value rule `MOD_SRC_FLIP` = flip USB(2) ↔ LAN(3) (two always-valid digital sources), restore original; write-only test value = 3 (LAN, the documented fix). Never writes out of the 0–5 source range.

## Changes (12 files, +166/-2)

`core/capabilities.py` (new `CAP_MOD_INPUT_ROUTING`), `validation/registry/_types.py` (ValueRule), `validation/hardware.py` (mutation + test value), `validation/registry/_system.py` (CheckSpec), `rigs/ic7610.toml` (declare cap), 3 regenerated goldens, registry/generator/IC-7610 guard tests.

## Goldens (only the new row added per profile, +5 each; all `--gate` exit 0)

- `ic7610`: `mod_input.set` → supported (skip in dry-run)
- `x6200` / `ftx1`: `mod_input.set` → unsupported_pending_evidence (cap undeclared)

## Tests (tests/validation/)

`test_mod_input_set_rmvr_roundtrip_flips_and_restores`, `test_mod_input_set_flips_away_from_lan_when_already_lan` (asserts every write in-range 0–5), `test_mod_input_check_unsupported_when_radio_lacks_op`, `test_mod_input_supported_when_cap_declared`. Registry guards updated (count 54→55, frozen id-set, IC-7610 cap parity 46→47).

## Validation

- `pytest tests/ --ignore=tests/integration` → 7825 passed, 12 skipped, 11 xfailed
- `ruff check` + `format --check` → clean
- `mypy src/` → no issues

## Note

Live IC-7610 round-trip verification not run (no hardware in worktree) — dry-run goldens + fake-radio RMVR tests green. Live check pending hardware.

Closes MOR-678.

### DIFF
--- a/rigs/ic7610.toml
+++ b/rigs/ic7610.toml
@@ -100,6 +100,10 @@ features = [
     # repeater-tone / TSQL family is intentionally NOT declared (MOR-661).
     # Data
     "data_mode",
+    # Modulation input routing — USB/LAN/DATA MOD source select (1A 05
+    # 0089-0094). Gates the mod_input.set RMVR guard behind the "web voice
+    # TX = noise" surface (DATA-OFF MOD must be LAN, not MIC; MOR-678).
+    "mod_input_routing",
     # AGC
     "agc",
     # System
--- a/src/rigplane/core/capabilities.py
+++ b/src/rigplane/core/capabilities.py
@@ -51,6 +51,7 @@ __all__ = [
     "CAP_SQL_TYPE",
     "CAP_VOICE_TX",
     "CAP_DATA_MODE",
+    "CAP_MOD_INPUT_ROUTING",
     "CAP_AGC",
     "CAP_POWER_CONTROL",
     "CAP_DIAL_LOCK",
@@ -155,6 +156,11 @@ CAP_SQL_TYPE = "sql_type"
 # Data
 CAP_DATA_MODE = "data_mode"
 
+# Modulation input routing — USB/LAN/DATA MOD source select (Icom 1A 05
+# 0089-0094). Guards the TX-audio routing surface behind the "web voice TX =
+# noise" failure mode (the documented fix is DATA-OFF MOD = LAN; MOR-678).
+CAP_MOD_INPUT_ROUTING = "mod_input_routing"
+
 # AGC
 CAP_AGC = "agc"
 
@@ -235,6 +241,8 @@ KNOWN_CAPABILITIES: frozenset[str] = frozenset(
         CAP_VOICE_TX,
         # Data
         CAP_DATA_MODE,
+        # Modulation input routing
+        CAP_MOD_INPUT_ROUTING,
         # AGC
         CAP_AGC,
         # System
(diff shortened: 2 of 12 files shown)


morozsm:
Agent Review: PASS

Independent adversarial review of MOR-678 (IC-7610 MOD-input routing RMVR guard). Verified, not trusted:

- **Backend ops real.** `get_data_off_mod_input`/`set_data_off_mod_input` exist end-to-end; the CI-V builders in `commands/config.py` map to **1A 05 0091** (DATA-OFF sub-addr `_CTL_MEM_DATA_OFF_MOD_INPUT = b"\x00\x91"` in `_frame.py:99`) — precisely the documented "web voice TX = noise" surface. `set_data_off_mod_input` itself guards `0 <= source <= 5` and raises ValueError otherwise (defense in depth).
- **VALID-source invariant proven.** `MOD_SRC_FLIP = lambda v: 3 if int(v) != 3 else 2` returns only 2 (USB) or 3 (LAN) for ANY input — both always in-range, both digital sources, neither keys TX (it is a routing select). Original is read, written-away-from, and restored. `test_mod_input_set_flips_away_from_lan_when_already_lan` asserts every write is `0 <= v <= 5` and that LAN->USB picks a different valid source.
- **Gating correct.** SUPPORTED only on ic7610 (declares `mod_input_routing`); UNSUPPORTED_PENDING_EVIDENCE on x6200/ftx1. `test_mod_input_supported_when_cap_declared` (declared + undeclared paths) and `test_mod_input_check_unsupported_when_radio_lacks_op` (cap declared but op missing) both cover it.
- **Goldens add only the new row.** Diff of all three dry-run JSONs adds exactly one `mod_input.set` row each (ic7610 supported/skip; x6200+ftx1 unsupported_pending_evidence) — no other row drifted. All three `--gate` runs exit 0.
- **Registry guards deliberate.** 54->55 entry count, frozen id-set gains exactly `mod_input.set`, IC-7610 cap-parity 46->47. Spot-check row matches spec.
- **Load-bearing.** Reverting the `_system.py`+`hardware.py` hunks to main fails 5 of the new tests (KeyError mod_input.set / count 54!=55).
- **Suite green at PR head (b338da1):** `pytest --ignore=integration` = 7825 passed, 12 skipped, 11 xfailed, 1 xpassed; `ruff check` clean; `ruff format --check` clean; `mypy src/` = no issues (no new errors). No scope creep — diff vs main is exactly the 12 intended files.

Note: LIVE IC-7610 hardware verification is still pending — acceptable, recommend as a follow-up (the RMVR write is non-TX and self-restoring), not a block.

PR #1829, opened 2026-06-11
fix(#MOR-664): correct scope vbw/rbw/dual receiver-prefix on SET (live-proven IC-7610 round-trip)

morozsm:
## What

Backend-only fix in `runtime/_scope_runtime.py` for the GET/SET receiver-prefix mismatch that made three scope `.set` checks fail to round-trip on the live IC-7610 (MOR-664, found by the harness):

- **`set_scope_vbw`** (`0x27 0x1D`) — getter passes `receiver=`, setter didn't → **added** it.
- **`set_scope_rbw`** (`0x27 0x1F`) — same → **added** it.
- **`set_scope_dual`** (`0x27 0x13`) — getter uses no receiver (global setting), setter wrongly added one → **removed** it (frame is now `0x27 0x13 0x01`, not `0x27 0x13 0x00 0x01`).

## Live-proven root cause

Raw CI-V traces on the real IC-7610 (2026-06-11): vbw/rbw with the receiver prefix → readback == written value (REACTED); dual WITHOUT the receiver prefix → REACTED (dual_watch state irrelevant). Builders already accept `receiver=None`; only the runtime call sites changed.

## Tests

`TestScopeSetterReceiverPrefix` in `tests/test_radio.py` (real `MockTransport`, no MagicMock) asserts exact CI-V frame bytes, parametrized over scope receiver 0/1: vbw/rbw include the receiver byte; dual has NO receiver byte. Confirmed failing before / passing after.

## Validation

- `pytest tests/ --ignore=tests/integration` → 7839 passed, 12 skipped, 11 xfailed
- `ruff check` + `format --check` → clean · `mypy src/` → no issues
- Goldens UNCHANGED (backend-only, dry-run-inert).

**Live verification:** re-running the IC-7610 hardware validation from this branch (in progress) — expect `scope_dual/vbw/rbw.set` to move FAIL → PASS.

Closes MOR-664.

### DIFF
--- a/src/rigplane/runtime/_scope_runtime.py
+++ b/src/rigplane/runtime/_scope_runtime.py
@@ -221,9 +221,11 @@ class ScopeRuntimeMixin(_MixinBase):  # type: ignore[misc]
     async def set_scope_dual(self, dual: bool) -> None:
         """Enable or disable dual scope mode."""
         self._check_connected()
-        receiver = self._scope_controls().receiver
+        # Single/dual is a GLOBAL scope setting — get_scope_dual sends no
+        # receiver prefix, so the setter must not either (MOR-664). A stray
+        # receiver byte malforms the write and the radio ignores it.
         await self._send_civ_raw(
-            _scope_single_dual_cmd(dual, to_addr=self._radio_addr, receiver=receiver),
+            _scope_single_dual_cmd(dual, to_addr=self._radio_addr),
             wait_response=False,
         )
         self._scope_controls().dual = dual
@@ -423,8 +425,12 @@ class ScopeRuntimeMixin(_MixinBase):  # type: ignore[misc]
     async def set_scope_vbw(self, narrow: bool) -> None:
         """Enable or disable narrow scope VBW."""
         self._check_connected()
+        # get_scope_vbw reads the selected receiver (sends a receiver prefix),
+        # so the setter must address the same receiver or the write never
+        # reflects back (MOR-664).
+        receiver = self._scope_controls().receiver
         await self._send_civ_raw(
-            _scope_set_vbw_cmd(narrow, to_addr=self._radio_addr),
+            _scope_set_vbw_cmd(narrow, to_addr=self._radio_addr, receiver=receiver),
             wait_response=False,
         )
         self._scope_controls().vbw_narrow = narrow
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent review of MOR-664 (scope vbw/rbw/dual receiver-prefix). Verified against builders in `commands/scope.py` and the runtime getter/setter pairs in `runtime/_scope_runtime.py`:

- **Each setter now matches its getter.** `get_scope_vbw`/`get_scope_rbw` call `_scope_query(..., receiver=receiver)` (prefix sent); the setters now pass `receiver=self._scope_controls().receiver` — symmetric. `get_scope_single_dual` calls `_scope_query` with NO receiver (single/dual is global); `set_scope_dual` now drops `receiver=` and its now-unused local — symmetric. Audited all 12 scope get/set pairs: every pair is receiver-symmetric, no other setter touched. Builders untouched.
- **Frame bytes exact.** New `TestScopeSetterReceiverPrefix` (real MockTransport, raw `sent_packets` bytes, parametrized receiver 0/1) asserts vbw `0x27 0x1D <r> 0x01`, rbw `0x27 0x1F <r> 0x02`, dual `0x27 0x13 0x01` (no receiver byte) AND that the prefixed dual frame `0x27 0x13 <r> 0x01` is NOT sent. NOT-sent assertion is meaningful even at receiver=0 (corrected `...13 01 fd` vs buggy `...13 00 01 fd`).
- **Tests genuinely catch the regression.** Stash-and-revert of all three setters to the buggy form → all 6 new tests FAIL (raw traces confirm buggy dual emits the stray receiver byte, buggy vbw/rbw omit the prefix). Restored fix → 6/6 pass.
- **Goldens unchanged.** Zero diff under `tests/golden/` in both the PR diff vs main and the working tree — backend fix is dry-run-inert.
- **Suite green at PR head:** 7839 passed, 12 skipped, 11 xfailed, 1 xpassed, 0 failed; `ruff check` clean; `ruff format --check` clean (605 files); `mypy src/` Success, 0 issues (284 files).

State-cache writes (`vbw_narrow`/`rbw`/`dual`) after each set are intact. Live IC-7610 run on this branch confirmed 3 PASS / 0 FAIL (orchestrator evidence). Clean, in-scope, no collateral.

PR #1883, opened 2026-06-19
feat(MOR-881): canonical state-payload schema + TypeScript codegen (reference contract pattern)

morozsm:
## MOR-881 (re-scoped: payload-canonical) — parent epic MOR-707

Reference implementation of the contract-codegen pattern: the public radio-state
WIRE payload (`build_public_state_payload` / `…_from_snapshot`) becomes the single
source of truth for the frontend TypeScript `state.ts`. This is the worked example
later contract migrations copy.

**Boundary:** `open-core candidate` — generic radio-state contract + codegen
toolchain in the public MIT `rigplane-core`. No proprietary material.

### What landed (3 commits)

1. **`src/rigplane/web/state_schema.py` + conformance test** — pydantic v2 model
   tree mirroring the public payload (union/superset of both emit paths). Pins all
   literal unions (`active`, `radioLink`, `readiness`, `likelyCause`,
   freshness/availability). `extra="forbid"` → `additionalProperties:false` (no TS
   index signature). `tests/web/test_state_schema_conformance.py` validates REAL
   payloads from BOTH producers (dataclass + snapshot, 1 and 2 receivers) against
   `ServerStatePublic`. Producers are unchanged — the test is the tie-in.
2. **`scripts/gen_state_types.py` + `state.ts` split** — `model_json_schema()` →
   `json2ts` → spliced into a `// BEGIN/END GENERATED` block. The hand-written tail
   holds stable aliases (`ReceiverState`, `ScopeControls`, `FieldStatus`,
   `FieldFreshness`, `FieldAvailability`) over the generated `*Public` interfaces,
   plus the client-only `ServerState` augmentation and `UiState`/`PendingCommand`.
3. **CI drift gate + dev-dep wiring** — `state-types-gate.yml` runs
   `gen_state_types.py --check` and fails on drift. `pydantic` in a `codegen`
   extra + dev group; `json-schema-to-typescript` in frontend devDeps.

### Zero-runtime-dep guarantee (held)

`pydantic` is **not** in core runtime `dependencies`. `state_schema.py` is imported
ONLY by the conformance test and the generator — never by `runtime_helpers.py` /
`server.py`. Verified: `grep state_schema src/` matches nothing outside the module.

### The 7 latent disagreements (spec §8) — how each was handled

1. **`radioDetail.uptimeSeconds`** (TS required, never emitted) — **contract
   correction + latent bug found.** Generated `RadioDetailPublic` carries only
   `status`. Types-only; no runtime change.
2. **`meterSource`** (in TS `ServerState`, never server-sent) — **contract
   correction + latent bug found.** Removed from the generated server-sent
   contract; re-added as a hand-written client-only `ServerState` augmentation so
   `layout-utils.ts` (which reads `radioState.meterSource`) still compiles.
3. **`FieldStatus.quality: string[]`** (emitted, absent from TS) — **additive,
   included** as optional on `FieldStatusPublic`.
4. **`vfoA`/`vfoB`/`activeSlot`/`filterNum`** (emitted, absent from TS) —
   **additive, included** as optional (`VfoSlotPublic`). Safe — existing consumers
   ignore unknown optional fields.
5. **Redundant active-slot scalars vs slot view** — both modeled; the slot view is
   documented as canonical, the scalars as derived back-compat. Both stay in the
   generated `ReceiverStatePublic`.
6. **Receiver scalars required vs always-present** — kept the old required set
   (`freqHz/mode/filter/dataMode/sMeter/att/preamp/nb/nr/afLevel/rfGain/squelch`)
   so consumers (`rx.att > 0`) compile; everything additive is optional.
7. **Top-level `?optional` vs always-present** — kept the old required set; the
   extra always-present synthetic objects (`scopeControls`/`radioDetail`/
   `radioHealth`/`wsClients`/`txBandEdges`) stay optional to match the old contract
   and the defensive `?.` consumers (benign tightening relaxed per spec §8.7).

Also: `sub` is honestly optional in the generated wire contract (dropped for
single-RX radios) and narrowed to required in the hand-written client `ServerState`
(consumers assume it present). `transportSeq` is modeled on the envelope and added
to the client `ServerState` augmentation (the frontend hoists it).

### Hard guard — no frontend consumer broken

`npm run check` (svelte-check + tsc): **0 errors, 0 warnings**. No `.svelte`/`.ts`
consumer or test fixture was edited — the entire reconciliation lives in the schema
model, the generator `required` policy, and the hand-written alias section.

### Verification

- Conformance: `tests/web/test_state_schema_conformance.py` — **4 passed** (both
  paths, receiver_count 1 and 2).
- Generator idempotent: committed `state.ts` matches `--check` (exit 0).
- **Drift gate proven:** adding a model field without regenerating →
  `gen_state_types.py --check` exits 1; reverted to clean.
- `npm run check` — 0 errors; `npx vitest run` — **2178 passed (126 files)**.
- `uv run mypy --strict src/rigplane/web` — clean (codegen-only schema gets a
  targeted `misc` override for the pydantic-Any-under-`follow_imports=skip` case).
- `uv run ruff check src/ tests/` clean; `uv run lint-imports` — 5 kept, 0 broken.

### Out of scope / follow-ups

- Producers still return `dict` (conformance test is the tie-in; not wiring
  `model_validate` into the live path, per spec).
- The 7 disagreements are documented + corrected in the *types*; no frontend
  runtime-logic change. Aligning runtime consumers to the now-honest `sub`
  optionality (or surfacing `vfoA`/`vfoB` in the UI) is future work.
- **`txBandEdges` wire-format inventory gap:** entries in `txBandEdges` use
  snake_case keys (`start_hz`, `end_hz`) on the wire while the rest of the
  state payload uses camelCase. The schema models the entries as
  `list[dict[str, int]]` to remain accurate without prescribing key casing.
  Aligning to camelCase is a separate contract change (affects existing
  consumers) and is left for a dedicated follow-up.

https://claude.ai/code/session_01MFjAT974fbZ48EBA9DxAiS

### DIFF
--- a/.github/workflows/state-types-gate.yml
+++ b/.github/workflows/state-types-gate.yml
@@ -1,0 +1,71 @@
+name: State types gate
+
+# MOR-881: the server-sent portion of `frontend/src/lib/types/state.ts` is
+# GENERATED from the pydantic contract in `src/rigplane/web/state_schema.py`
+# via `scripts/gen_state_types.py` (pydantic -> JSON Schema -> json2ts). This
+# gate regenerates the block and fails on drift, so the TypeScript contract can
+# never silently diverge from the Python state-payload producer. Mirrors the
+# corpus-drift / rebrand-gate staleness pattern.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "src/rigplane/web/state_schema.py"
+      - "scripts/gen_state_types.py"
+      - "frontend/src/lib/types/state.ts"
+      - "frontend/package.json"
+      - "frontend/package-lock.json"
+      - "pyproject.toml"
+      - ".github/workflows/state-types-gate.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "src/rigplane/web/state_schema.py"
+      - "scripts/gen_state_types.py"
+      - "frontend/src/lib/types/state.ts"
+      - "frontend/package.json"
+      - "frontend/package-lock.json"
+      - "pyproject.toml"
+      - ".github/workflows/state-types-gate.yml"
+
+concurrency:
+  group: state-types-gate-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  state-types-gate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 8
+    env:
+      UV_PYTHON: "3.11"
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          enable-cache: true
+          cache-dependency-glob: |
+            pyproject.toml
+            uv.lock
+
+      - name: Set up Python 3.11
+        run: uv python install 3.11
+
+      - name: Install Python deps (pydantic for codegen)
+        run: uv sync --extra codegen
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '20'
+          cache: 'npm'
+          cache-dependency-path: frontend/package-lock.json
+
+      - name: Install json-schema-to-typescript
+        run: npm --prefix frontend ci
+
+      - name: Regenerate state.ts and fail on drift
+        run: uv run python scripts/gen_state_types.py --check
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,8 @@ mockups/
 scripts/*
 # But the release-validation runner is tracked (MOR-648).
 !scripts/validate-release.sh
+# And the public-state TypeScript codegen generator is tracked (MOR-881).
+!scripts/gen_state_types.py
 # But CI scripts under .github/ are tracked.
 !.github/scripts/
 # And frontend tooling scripts (e.g. i18n lint) are tracked.
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -22,6 +22,7 @@
         "eslint": "^10.2.0",
         "eslint-plugin-svelte": "^3.17.0",
         "jsdom": "^28.1.0",
+        "json-schema-to-typescript": "^15.0.4",
         "svelte": "^5.45.2",
         "svelte-check": "^4.3.4",
         "svelte-eslint-parser": "^1.6.0",
@@ -54,6 +55,24 @@
       },
       "peerDependencies": {
         "ajv": ">=8"
+      }
+    },
+    "node_modules/@apidevtools/json-schema-ref-parser": {
+      "version": "11.9.3",
+      "resolved": "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-11.9.3.tgz",
+      "integrity": "sha512-60vepv88RwcJtSHrD6MjIL6Ta3SOYbgfnkHb+ppAVK+o9mXprRtulx7VlRl3lN3bbvysAfCS7WMVfhUYemB0IQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@jsdevtools/ono": "^7.1.3",
+        "@types/json-schema": "^7.0.15",
+        "js-yaml": "^4.1.0"
+      },
+      "engines": {
+        "node": ">= 16"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/philsturgeon"
       }
     },
     "node_modules/@asamuzakjp/css-color": {
@@ -2520,6 +2539,13 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@jsdevtools/ono": {
+      "version": "7.1.3",
+      "resolved": "https://registry.npmjs.org/@jsdevtools/ono/-/ono-7.1.3.tgz",
+      "integrity": "sha512-4JQNk+3mVzK3xh2rqd6RB4J46qUR19azEHBneZyTZM+c456qOrbbM/5xcR8huNCCcbVt7+UmizG6GuUvPvKUYg==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@playwright/test": {
       "version": "1.58.2",
       "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.58.2.tgz",
@@ -3129,6 +3155,13 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/lodash": {
+      "version": "4.17.24",
+      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.17.24.tgz",
+      "integrity": "sha512-gIW7lQLZbue7lRSWEFql49QJJWThrTFFeIMJdp3eH4tKoxm1OvEPg02rm4wCCSHS0cL3/Fizimb35b7k8atwsQ==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@types/node": {
       "version": "24.12.0",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-24.12.0.tgz",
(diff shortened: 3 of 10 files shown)


morozsm:
Agent Review: BLOCKED — independent review (I did not author this PR). Re-ran every check in an isolated worktree on PR head a623cf0. One real, reproducible schema-fidelity defect blocks this from becoming the reference contract; everything else holds up well.

## Blocker — schema types four wire fields as `int`, but the canonical snapshot producer emits normalized `float`

`src/rigplane/web/state_schema.py:107-109` (`afLevel`/`rfGain`/`squelch`) and `:246` (`powerLevel`) are typed `int`. The live snapshot path normalizes exactly these via `_normalize_public_level_snapshot_value` (`src/rigplane/web/runtime_helpers.py:759-768`, applied at `:823-825` for `af_level`/`rf_gain`/`squelch` and `:864-868` for `power_level` — see `_NORMALIZED_RECEIVER_OPERATOR_CONTROL_FIELDS` and `_NORMALIZED_GLOBAL_OPERATOR_CONTROL_FIELDS`). That function returns `float(value)/max_value`, i.e. a fraction in [0,1], for any int input.

I reproduced the failure end-to-end. Observing `af_level=200`, `rf_gain=128`, `squelch=64`, `power_level=100` through `build_public_state_payload_from_snapshot` then `ServerStatePublic.model_validate(payload)` raises 4 `int_from_float` errors:
```
powerLevel  -> 0.392..., type=int_from_float
main.afLevel -> 0.784..., type=int_from_float
main.rfGain  -> 0.501..., main.squelch -> 0.251...
```
The dataclass path leaves these as raw `int` (0), so the field is genuinely `int | float` across the two producers and the canonical model must accommodate `float`.

Why this blocks the REFERENCE PR specifically: the contract claims `int` for a value that is canonically a normalized fraction. The producer the schema is supposed to validate against fails `model_validate` the moment a real backend reports AF/RF/squelch/power level. The generated TS happens to compile (it types these `number`), so the frontend masks the wrong Python contract — but the canonical model is lying, and any later migration copying this pattern inherits a "tests green, schema wrong" footgun.

The conformance test reinforces the trap: `tests/web/test_state_schema_conformance.py` claims to validate "real payloads from BOTH producers", but no test case observes these four normalized fields, so the suite passes while the actual divergence between the two paths goes uncovered.

### Required fixes
1. Type the four fields `float` (or `int | float`) in `state_schema.py`. They are normalized fractions on the canonical path.
2. Add a snapshot conformance case that OBSERVES `af_level`/`rf_gain`/`squelch`/`power_level` with raw ints so the test would have caught this (the test currently asserts it covers both producers but does not exercise the path-divergent fields).
3. Regenerate `state.ts` (`python scripts/gen_state_types.py`) and re-run `--check`.
4. Re-run `uv run --extra codegen python -m pytest tests/web/test_state_schema_conformance.py`.

## Non-blocking concerns (fix or document; not merge-blockers)
- `txBandEdges` entries are snake_case on the wire (`{start_hz,end_hz}`, `core/radio_state.py:378-380`) — never camel-cased because the camel pass does not recurse top-level list values. `state_schema.py:297` `list[dict[str,int]]` accepts it, but the keys diverge from the otherwise-camelCase contract. Consider a typed `TxBandEdgePublic{startHz,endHz}` or a documented exception.
- `active` is pinned `Literal["MAIN","SUB"]` but `RadioState.active` is a plain `str` with no producer-side guard; an out-of-domain value would fail validation. Acceptable as the contract, worth a note.

## What I independently re-ran (all PASS except where noted)
- Zero-runtime-dep: `grep -rn state_schema src/` -> only `state_schema.py` (+ a docstring mention); only that module imports pydantic; `pydantic` is in the `codegen` extra + dev group, NOT in `[project] dependencies`. HOLDS.
- Gate works: `gen_state_types.py --check` PASS on committed `state.ts`; regen is byte-idempotent (no git drift); injected a throwaway model field -> `--check` exits 1; injected a hand-edit INSIDE the generated block -> `--check` exits 1; reverted, clean.
- Frontend: `npm --prefix frontend run check` -> 0 errors, 0 warnings, 4260 files. `meterSource` reads off the hand-written `ServerState` augmentation (`layout-utils.ts:44`); `radioDetail.uptimeSeconds` has ZERO remaining consumers so its removal is safe; `radioDetail.status` consumer (`http-client.ts:167`) intact. Generated/hand-written split is cleanly marked.
- Literal unions for `radioHealth.radioLink/readiness/likelyCause`: I enumerated every return of `classify_radio_health` (`runtime_helpers.py:560-675`) — all values fall within the pinned sets; no out-of-union value reachable. `freshness`/`availability` match the source enums.
- `uv run lint-imports` -> 5 contracts kept, 0 broken. `uv run ruff check` on touched paths -> clean. `.gitignore` correctly force-tracks the generator (`!scripts/gen_state_types.py` under the `scripts/*` ignore).
- The 7 latent disagreements (spec §8) are handled as designed (additive fields optional, `radioDetail`/`meterSource` corrections types-only). The bug above is a NEW, 8th disagreement the migration missed — the float-normalized level fields.

The pattern, toolchain, gate, and split are sound; fix the level-field types + add the covering conformance case and this is a strong reference. Do not merge until then.

morozsm:
## Blocker addressed — commit 7eac12c7

### What was fixed

**`int_from_float` errors on snapshot payloads (blocker)**

`afLevel`, `rfGain`, `squelch` and `powerLevel` were typed `int` but
`_normalize_public_level_snapshot_value` converts them to `float` in [0, 1]
on the snapshot path. `model_validate` raised `int_from_float` on any real
snapshot payload that had observed values for those fields.

Fix: typed all four as `float` in `state_schema.py`. Pydantic lax mode
coerces the `int(0)` default from the dataclass path, so `float` is correct
for both producers.

**New conformance case** (`test_snapshot_path_normalized_level_fields_conform`)

Drives the snapshot producer with raw integer observations (128 / 200, as a
hardware driver emits), asserts the emitted payload contains float values in
[0, 1], and calls `ServerStatePublic.model_validate`. Coverage gap closed.

Failure proof: temporarily reverting `afLevel` back to `int` gives:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ServerStatePublic
main.afLevel
  Input should be a valid integer, got a number with a fractional part [type=int_from_float, ...]
```

### Non-blocking concerns

**`active: Literal["MAIN", "SUB"]`** (concern 4): Audited all three write
sites (`_civ_rx.py`, `_dual_rx_runtime.py`, `RadioState` default). Only
`"MAIN"` and `"SUB"` are ever assigned. Added an inline comment justifying
the Literal. No schema change needed.

**`txBandEdges` snake_case entries** (concern 5): Wire format uses `start_hz` /
`end_hz` while the rest of the contract is camelCase. Documented in the PR
description as a known follow-up / inventory gap (not changed here — that's a
separate wire-format contract change).

### Verification

- `uv run pytest tests/web/test_state_schema_conformance.py`: **5 passed**
- New test fails against old `int` typing: confirmed (`int_from_float`)
- `python scripts/gen_state_types.py --check`: exit 0 (state.ts unchanged — float and int both map to `number` in JSON schema)
- Drift gate trips: deliberately corrupted state.ts → exit 1; restored
- `npm --prefix frontend run check`: **0 errors, 0 warnings** (4260 files)
- `uv run mypy --strict src/rigplane/web`: **clean** (25 files)
- `uv run ruff check src/rigplane/web/state_schema.py tests/web/test_state_schema_conformance.py`: **clean**
- `uv run lint-imports`: 5 kept, 0 broken

morozsm:
Agent Review: PASS — blocker (float-normalized level fields) resolved at 7eac12c7.

**Re-ran at HEAD (7eac12c7):**

1. `afLevel`, `rfGain`, `squelch` (lines 111-113) and `powerLevel` (line 255) in `src/rigplane/web/state_schema.py` are now typed `float`.
2. New conformance case `test_snapshot_path_normalized_level_fields_conform` drives the snapshot producer with raw integer observations (128/200), asserts the four payload fields are `float` in [0,1], then calls `ServerStatePublic.model_validate(payload)` — this would raise `int_from_float` under the old `int` typing.
3. `uv run pytest tests/web/test_state_schema_conformance.py` — **5/5 passed** (2×dataclass-path, 2×snapshot-path, 1×new level-fields case).
4. `uv run python scripts/gen_state_types.py --check` — exits 0 (up to date). Note: the drift gate cannot distinguish `int` vs `float` because both map to TS `number`; the conformance test is the correct enforcement layer.
5. `npm --prefix frontend run check` — 0 errors, 0 warnings (4260 files).
6. `uv run lint-imports` — 5/5 contracts kept.
7. `uv run ruff check` on touched paths — all checks passed.
8. `uv run mypy --strict src/rigplane/web/state_schema.py src/rigplane/web/runtime_helpers.py` — success, no issues.
9. Generated TS (`frontend/src/lib/types/state.ts`) shows all four fields as `number` — no TS surface change.

morozsm:
Agent Review: PASS — head 06bcfa1c; delta since prior PASS (7eac12c7) is ruff-format-only in the conformance test with no logic change; quick gate green, all gates pass.

PR #1888, opened 2026-06-19
feat(MOR-911): promote build_hamlib_discovery_payload to public API

morozsm:
## Summary

- Moves `build_hamlib_discovery_payload` and its private helper functions from `rigplane.cli._discover_hamlib` (internal Tier-3 module) to `rigplane.backends.discovery` (public Tier-2 module, alongside `build_setup_discovery_payload`).
- Adds a PEP 562 lazy entry in `rigplane/__init__.py` so it is importable as `from rigplane import build_hamlib_discovery_payload`.
- Keeps a re-export shim in `rigplane.cli._discover_hamlib` — existing internal callers are unbroken without modification.
- Updates `docs/api/public-api-surface.md`: symbol listed in both the "Hamlib-assisted discovery" table and the Tier-2 section.

**No behavior change** — this is a pure export/visibility promotion.

Canonical public import path:
```python
from rigplane.backends.discovery import build_hamlib_discovery_payload
# or equivalently (lazy):
from rigplane import build_hamlib_discovery_payload
```

Closes MOR-911. Part of epic MOR-707 (Hamlib assisted discovery public surface).

**open-core candidate** — no proprietary logic, safe to expose publicly.

Unblocks rigplane-pro from importing `rigplane.cli._discover_hamlib.build_hamlib_discovery_payload` directly; the follow-up pro PR will drop the corresponding import-linter whitelist entry.

## Test plan

- [x] `from rigplane.backends.discovery import build_hamlib_discovery_payload` resolves correctly
- [x] `from rigplane import build_hamlib_discovery_payload` resolves correctly (lazy PEP 562 path)
- [x] `from rigplane.cli._discover_hamlib import build_hamlib_discovery_payload` still works (backward-compat shim)
- [x] `uv run lint-imports` — all 5 contracts KEPT, 0 broken
- [x] `uv run ruff check` — clean on all touched files
- [x] `uv run pytest tests/test_discovery.py` — 91/91 passed
- [x] `uv run pytest tests/test_cli_coverage.py` — 51/51 passed (includes hamlib CLI tests)
- [x] `uv run pytest -m "not slow"` — full fast suite green


morozsm:
Agent Review: PASS — Move-only promotion, no behavior change. All three import paths verified green: `from rigplane.backends.discovery import build_hamlib_discovery_payload`, `from rigplane import build_hamlib_discovery_payload`, and the backward-compat shim `from rigplane.cli._discover_hamlib import build_hamlib_discovery_payload`. Internal CLI callers in `cli/__init__.py` continue to import via the shim and are unaffected. `uv run lint-imports`: 5 contracts kept, 0 broken — `backends.discovery` is a correct layering home (lower than `cli`). `uv run ruff check src/`: clean. Full fast suite: 8249 passed, 129 skipped, 0 failures. Function body diff confirms verbatim move with a docstring addition only. `docs/api/public-api-surface.md` updated correctly with canonical import path and MOR-911 provenance note.

Issue #1215, opened 2026-04-29
feat(telemetry): audit ?ui=v1 usage before removal


morozsm:
## Parent epic
#874

## Goal

Before deleting v1 UI code, determine whether any users still hit the `?ui=v1` URL fallback. If telemetry exists, run a query and report counts; if no telemetry, document the absence and rely on time-based deprecation (one release window per epic plan).

## Files / areas affected

- Frontend telemetry hooks (if any) — grep `frontend/src/` for analytics, posthog, plausible, sentry breadcrumbs, or custom event sinks.
- Server-side request logs (`src/icom_lan/web/server.py`, access log middleware) — query for `?ui=v1` query string occurrences.
- Output: short markdown report under `docs/plans/2026-04-23-ui-v1-telemetry-audit.md` (or similar dated file) summarising findings + recommendation.

## Acceptance

- [ ] Inventoried all telemetry / logging surfaces that could capture `?ui=v1`.
- [ ] If data exists: report query results + sample size + time window.
- [ ] If no data exists: explicitly document the gap and confirm time-based deprecation is the chosen path.
- [ ] Markdown report committed.
- [ ] `uv run pytest tests/ -q --tb=short` zero failures.
- [ ] `uv run ruff check src/ tests/` zero violations.

## Dependencies

None — this is a research spike. Runs in parallel with sub-issues #2/#4/#5/#6.

## Out of scope

- Adding new telemetry instrumentation.
- Any code deletion (handled in sibling sub-issues).
- SpectrumToolbar `hideSourceControls` cleanup (#873) — explicitly punted by epic body.

PR #1897, opened 2026-06-22
fix(MOR-934): AudioSession declarative reconciler — fixes TX-only strand after reconnect (#1226, #1212)

### LINKED ISSUE #1226: fix(sync): add get_alc_meter to sync.IcomRadio (Codex P1 on #1213)
**Codex P1 review on PR #1213** (post-merge).

## Problem

PR #1213 removed deprecated `sync.IcomRadio.get_alc` but did NOT add `get_alc_meter` sync wrapper. The async-side `MetersCapable.get_alc_meter` exists (added in #1129 / v0.19), but sync users have **no ALC meter API at all** now. Code that previously read ALC via sync gets `AttributeError` when migrating to the documented replacement.

Symmetric to PR #1184 which added `sync.IcomRadio.get_swr_meter` after the same kind of removal. Same fix pattern needed.

## Fix

In `src/icom_lan/sync.py` — add `get_alc_meter() -> int` sync wrapper mirroring async-side `MetersCapable.get_alc_meter`. Pattern matches existing `get_swr_meter` at line ~200.

## Tests

- `test_sync_get_alc_meter_returns_int` — wrapper returns int 0-255.

## Acceptance

- `sync.IcomRadio.get_alc_meter()` works.
- Tests pass; ruff + mypy clean.

## Reference

- Codex inline: PR #1213, `sync.py:214`
- Sibling pattern: `sync.IcomRadio.get_swr_meter` (added in PR #1184)

morozsm:
Root cause: scattered imperative TX arming plus a reestablish path that never handled the `TX_ONLY` state, leaving the audio session stranded TX-only after a reconnect.

Fix: a pure `_desired()` function computes the target state, and a single idempotent `_converge()` drives toward it. Push events converge rather than reject, and TX arming is intent-gated so the MOR-556 ordering invariant is preserved.

Closes #1226
Closes #1212

Independently verified PASS (adversarial Opus review): RED confirmed on main, invariants I1-I4 verified strong, MOR-556/559/574 order tests byte-for-byte unchanged, public API unchanged.

Out of scope:
- #1215 RX stale-FD (separate issue)
- CAT-PTT first-class plumbing (deferred)

Release gate: live-hardware SSB<->FT8 TX re-test required before the pro beta.9 public flip.

### DIFF
--- a/docs/architecture/2026-06-22-audio-session-reconciler-PLAN.md
+++ b/docs/architecture/2026-06-22-audio-session-reconciler-PLAN.md
@@ -1,0 +1,356 @@
+# AudioSession reconciler refactor — executable, test-first implementation plan (Option B)
+
+**Date:** 2026-06-22
+**Status:** approved-design implementation plan (read-only on production code; builder agent executes)
+**Scope:** `rigplane-core` — `src/rigplane/audio/session.py` only (+ test files)
+**Boundary:** **open-core.** Generic LAN/USB audio TX state machine and recovery. No licensing/commercial/customer logic. The public demand API (`subscribe_rx` / `acquire_tx` / `TxLease.push` / `reestablish`) is unchanged.
+**Primary input:** `docs/architecture/2026-06-22-audio-session-state-architecture.md` (the architectural verdict — read it first).
+**Decision:** Option B (FIXED). Do not relitigate Options A or C.
+
+---
+
+## A. Plain-language restatement (the thermostat)
+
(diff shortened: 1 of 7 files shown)


morozsm:
Agent Review: PASS — independent adversarial review completed (RED confirmed on main, invariants I1-I4 verified strong, MOR-556/559/574 order tests unchanged, public API unchanged, deviation from eager-arm judged sound & faithful). Live-hardware TX re-test tracked as a release gate before the pro beta.9 public flip.

PR #1898, opened 2026-06-22
fix: core quick-wins — broadcast flood (#1876), diag log default (#1879), idle state-build gate (#1877)

### LINKED ISSUE #1876: Control-state broadcast floods logs and never recovers when a control subscriber stalls
**rigplane 2.10.0** · web/server.py · severity: medium (degrades over uptime, only cleared by restart)

> **Motivation — RX audio glitches.** One of a linked set of findings from chasing audible RX **audio
> dropouts** on a single-event-loop deployment. The per-broadcast log + queue churn this flood causes
> steals CPU from the audio egress path on the shared loop, contributing to the dropouts. (Companion
> issues: #1877, #1878, #1879 — same root cause: CPU starvation of the single audio event loop.)

## Summary
When a `/api/v1/ws` (control) subscriber stops draining its event queue — a slow client, a
backgrounded browser tab, a half-open socket — its per-connection `BoundedQueue(maxsize=100)` fills
and **stays full**. Every subsequent state broadcast (~20/s) then hits `asyncio.QueueFull` and emits
a `logger.warning("control state queue full; dropping state_update")`. The result is thousands of
warning lines/min plus continual queue churn, accumulating over uptime and **only cleared by a
service restart**. (Same family as the broadcaster-desync behavior some of us have hit before.)

## Reproduce
1. Connect a client to `/api/v1/ws` and **stop reading** from it (don't drain).
2. After its queue fills (~5 s at 20 Hz), watch:
   `journalctl -u <svc> --since '1 min ago' | grep -c 'control state queue full'`

Measured on our deployment: **12 stalled subscribers → 5754 `control state queue full` warnings in
30 s** (~190/s), and backend CPU steal rose 10% → 24% from the log/queue churn. A healthy client at
the same time still degrades because the loop is busier.

## Root cause
`web/server.py`, the broadcast fan-out:

```python
        target_queues = self._control_event_queues if queues is None else queues
        for q in list(target_queues):
            try:
                q.put_nowait(event)
            except asyncio.QueueFull:
                logger.warning("control state queue full; dropping state_update")
```

`put_nowait` drops the **newest** event and logs — but a `state_update` is a full/delta snapshot, so
the *newest* is the one worth keeping; the stale queued ones are worthless. The queue never recovers
on its own, and nothing reaps a persistently-full control subscriber.

## Proposed fix (smallest blast radius)
`BoundedQueue` already implements `put_drop_oldest` (used elsewhere). Coalesce-to-latest instead of
drop-newest-and-log:

```python
        target_queues = self._control_event_queues if queues is None else queues
        for q in list(target_queues):
            # A state_update supersedes older queued state; on a stalled/slow
            # consumer, evict the oldest instead of dropping the freshest + logging.
            q.put_drop_oldest(event)
```

Why it can't regress a healthy client: a draining client keeps its queue near-empty, so
`put_drop_oldest` sees `full() == False` and behaves exactly like `put_nowait` — eviction only
triggers at the maxsize=100 stalled case this targets.

Notes / options for your consideration:
- The control queue is mixed-type (`state_update`, `notification`, `dx_spot`, …). Plain
  `put_drop_oldest` could, in principle, evict a queued `notification`/`dx_spot` sitting ahead of a
  state backlog. In practice those are enqueued at separate low-rate call sites and the queue only
  backs up under state_update flood, so we judged the one-liner safe; a stricter variant could
  scan-and-drop only the oldest `state_update`. Your call on which you'd prefer.
- Optionally also reap a subscriber whose queue has been full for N seconds.

**Validated:** with this change, the 12-stalled-subscriber repro produced **0** queue-full warnings
(was 5754/30 s) and a healthy control client still received state normally (~40 `state_update`/6 s).

## Patch

```diff
--- a/rigplane/web/server.py
+++ b/rigplane/web/server.py
@@ -1277,10 +1277,9 @@
 
         target_queues = self._control_event_queues if queues is None else queues
         for q in list(target_queues):
-            try:
-                q.put_nowait(event)
-            except asyncio.QueueFull:
-                logger.warning("control state queue full; dropping state_update")
+            # A state_update supersedes older queued state; coalesce-to-latest on a
+            # stalled/slow consumer instead of dropping the freshest + log-flooding.
+            q.put_drop_oldest(event)
 
     async def _delayed_state_broadcast(self, delay: float) -> None:
         try:
```
*Applies to pristine 2.10.0 (`patch -p1` / `git apply`). Source files are CRLF; apply with
`--ignore-whitespace` if your working tree is LF.*

---
*Reported by Leon Toorenburg (WW0R). Diagnosis/patch prototyping by Claude (Anthropic Claude Code).
Measured on an Oracle Always-Free micro (single-event-loop, CPU-steal-constrained); patch validated
by A/B on the live deployment but not PR'd here pending your guidance.*

### LINKED ISSUE #1877: State-broadcast pipeline does heavy demand-independent work (~40% of a core at idle, zero clients)
**rigplane 2.10.0** · web/server.py + runtime/_civ_rx.py · severity: high on constrained hosts

> **Motivation — RX audio glitches.** This was the **single biggest contributor to audible RX audio
> dropouts** on our deployment: the demand-independent rebuild work competes with audio egress on the
> one event loop, so under CPU contention the audio stream loses frames. (Companion issues: #1876,
> #1878, #1879.)

## Summary
`_broadcast_state_update` rebuilds the **entire public-state payload** — `copy.deepcopy` ×3 + a full
recursive camelCase re-walk + delta-encode — up to **~20×/sec**, and it:
1. runs **even when no control client is subscribed**, and
2. is triggered on **freshness-only meter re-observations** (value unchanged).

`py-spy` of the **idle** loop with **zero clients connected** shows this consuming **~40% of one
core**, with `copy.deepcopy`/`_deepcopy_dict` in nearly every sample. On a single-event-loop backend
this competes directly with audio egress; on a CPU-contended host it's the headroom that audio loses
→ RX dropouts. Likely a side effect of the MOR-334 state-pipeline rewrite.

## Evidence
- py-spy idle (0 clients), top cumulative frames: `deepcopy` ≫ `_broadcast_state_update` →
  `_build_public_state_from_snapshot` → `build_public_state_payload_from_snapshot` → `_camel_keys`,
  driven by `_civ_rx._notify_state_store_changed` → `_on_radio_state_change`.
- Idle CPU (no clients): **~43% of a core**. After the fixes below: **10%**.
- Moderate-stress (1 pinned core) RX frame deficit: **18% → ~1%** with these fixes applied.

### 2a. Build is not gated on subscriber count
The full build + delta-encode runs before the fan-out, regardless of whether `_control_event_queues`
is empty. Proposed gate (placed **after** the FFT-scope sync so audio-scope clients are unaffected):

```python
        snapshot = self.command_state_store.snapshot()
        self._update_fft_scope_freq(snapshot)
        self._update_fft_scope_mode(snapshot)
        if not force and not self._control_event_queues:
            return                     # no control consumer -> skip build/delta/fan-out
        body = self._build_public_state_from_snapshot(snapshot)
        ...
```
A connecting client gets initial state via its own `hello`/`state_update` on connect, so it isn't
starved. (Validated: true-idle CPU 43% → 10%; a connected client still gets ~7 Hz state.)

### 2b. Freshness-only re-observations trigger full rebuilds for no semantic change
`runtime/_civ_rx.py` `_notify_state_store_changed`:

```python
    def _notify_state_store_changed(self, changeset: ChangeSet) -> None:
        paths = {change.path for change in changeset.changes}
        paths.update(changeset.freshness_paths)     # <-- meters re-FRESH at 20Hz, value unchanged
        if not paths:
            return
```
Meters age past `max_age` (~0.6 s) under the 50 ms freshness tick and re-observe FRESH every poll
with the **same value**, producing a freshness-only changeset that still fires a full broadcast.
Proposed: drop freshness-only flips for the high-rate meter family only (preserves stale/fresh UI
signaling for non-meter fields). The `path.family.value != "meters"` idiom is already used elsewhere
in this file:

```python
        paths.update(
            p for p in changeset.freshness_paths if p.family.value != "meters"
        )
```
(Validated: moderate-stress deficit 18% → 8% from this change alone.)

### 2c. Redundant deepcopies + a cache key that always misses
- The build path does ~3 stacked `copy.deepcopy`s (e.g. `deepcopy(_EMPTY_STATE_DICT)` then
  `deepcopy(state)`), plus the delta-encoder and the cache-hit path each deepcopy again — where the
  snapshot path's input is freshly allocated and not aliased.
- The build cache key in `_public_state_delivery_key` embeds `snapshot.freshness_revision`, which the
  freshness flips (2b) bump every tick — so the cache **always misses** at idle.

We did **not** patch 2c (correctness risk without your test coverage — some copies may be
defensive), but flag it as the highest-value remaining optimization: collapse the redundant copies
and/or exclude `freshness_revision` from the cache key when only freshness changed.

## Impact
2a + 2b together took idle CPU from ~43% to ~20% (then 10% with 2a's idle gate), and moderate-stress
RX deficit from 18% to ~1%. 2c would cut per-build cost further.

## Patch (covers 2a + 2b; 2c is description-only)

```diff
--- a/rigplane/runtime/_civ_rx.py
+++ b/rigplane/runtime/_civ_rx.py
@@ -1464,7 +1464,11 @@
 
     def _notify_state_store_changed(self, changeset: ChangeSet) -> None:
         paths = {change.path for change in changeset.changes}
-        paths.update(changeset.freshness_paths)
+        # Skip freshness-only re-FRESH flips for high-rate meters (value unchanged):
+        # they otherwise force a full public-state rebuild/broadcast per poll.
+        paths.update(
+            p for p in changeset.freshness_paths if p.family.value != "meters"
+        )
         if not paths:
             return
         self._notify_change(
--- a/rigplane/web/server.py
+++ b/rigplane/web/server.py
@@ -1246,6 +1246,10 @@
         # from the same canonical snapshot used for Web delivery.
         self._update_fft_scope_freq(snapshot)
         self._update_fft_scope_mode(snapshot)
+        # Skip the public-state build/delta/fan-out when no control client is
+        # subscribed (a connecting client gets initial state on connect).
+        if not force and not self._control_event_queues:
+            return
         body = self._build_public_state_from_snapshot(snapshot)
         state_key = self._public_state_delivery_key(
             snapshot,
```
*Applies to pristine 2.10.0 (`patch -p1` / `git apply`). Source files are CRLF; apply with
`--ignore-whitespace` if your working tree is LF.*

---
*Reported by Leon Toorenburg (WW0R). Diagnosis/patch prototyping by Claude (Anthropic Claude Code).
Measured on an Oracle Always-Free micro; 2a/2b validated by A/B on the live deployment, 2c proposed
only. Not PR'd here pending your guidance on defaults and the defensive-copy semantics.*

### LINKED ISSUE #1879: Diagnostic logging is forced to DEBUG by default → ~40–57 disk writes/sec on the event loop
**rigplane 2.10.0** · diagnostics/_logging.py · type: default question / minor perf

> **Motivation — RX audio glitches.** These are synchronous disk writes on the **same single event
> loop that ships the audio**; eliminating them removed a steady source of loop stalls contributing to
> audible RX **audio dropouts**. (Companion issues: #1876, #1877, #1878.)

## Summary
The diagnostic file handler forces the `rigplane` logger to **DEBUG** whenever the host app hasn't
set a level (`level == NOTSET`). In normal operation this means every per-CI-V-frame / per-state-
change `logger.debug(...)` actually **formats and writes to disk**. On our deployment that's
**~40–57 journald lines/sec at idle** (dominated by per-meter-read `civ-rx: cmd=0x14 …` and
`notify state_store_changed` lines) — steady I/O on the single event loop, and worse on an
I/O-constrained micro.

## Current code
`diagnostics/_logging.py`:
```python
        handler.setLevel(logging.DEBUG)
        ...
        if icom_logger.level == logging.NOTSET:
            icom_logger.setLevel(logging.DEBUG)
```

## Measured
- Idle journald rate (rigplane unit): **~850 lines / 15 s** (~57/s) at DEBUG.
- Setting the `rigplane` logger to **INFO** dropped that to **~1 line / 15 s** and shaved several
  percent off idle CPU (the per-frame `logger.debug` calls then short-circuit at `isEnabledFor`
  before formatting — they're %-style, so no arg work is done).

## Proposed (any one of these)
1. Default the `rigplane` logger to **INFO** rather than DEBUG when `NOTSET` (keep DEBUG opt-in via
   an env var, e.g. `ICOM_DEBUG`/`ICOM_LOG_LEVEL`). Smallest change, biggest effect.
2. And/or `isEnabledFor`-guard the hottest per-frame debug sites (e.g. the per-`notify` and
   per-meter-read lines in `runtime/_civ_rx.py`) so the call is free at INFO regardless.

We applied (1) on our box and lost nothing operationally — forensic DEBUG is still one env var away.

## Question
Is the always-on DEBUG file handler intended as a default for end-user deployments, or a dev/debug
convenience? If the former, an INFO default + env opt-in seems safer for production use on modest
hardware.

## Patch (option 1 — INFO default)

```diff
--- a/rigplane/diagnostics/_logging.py
+++ b/rigplane/diagnostics/_logging.py
@@ -82,7 +82,7 @@
         # The handler's own level is DEBUG, so any record the logger doesn't
         # filter still hits the file — see spec §4.1.
         if icom_logger.level == logging.NOTSET:
-            icom_logger.setLevel(logging.DEBUG)
+            icom_logger.setLevel(logging.INFO)  # was DEBUG: avoids ~40-57 disk writes/s on the loop (see issue)
     except Exception as exc:  # noqa: BLE001 — best-effort init, swallow all
         sys.stderr.write(f"rigplane: diagnostic logging disabled: {exc}\n")
```
*Applies to pristine 2.10.0 (`patch -p1` / `git apply`). Source files are CRLF; apply with
`--ignore-whitespace` if your working tree is LF.*

---
*Reported by Leon Toorenburg (WW0R). Diagnosis by Claude (Anthropic Claude Code). Measured on an
Oracle Always-Free micro; INFO-default change validated on the live deployment.*

morozsm:
Three reporter-validated low-risk fixes, each with a RED->GREEN regression test.

- **#1876**: replace `put_nowait`+warning with `put_drop_oldest` for the control WS fanout, ending the broadcast flood.
- **#1879**: default the rigplane logger to INFO instead of DEBUG, plus a test-isolation fix that restores the logger level in diagnostics tests.
- **#1877 (2a)**: demand-gate `_broadcast_state_update` on subscriber count to skip idle state-builds.

Closes #1876
Closes #1879
Closes #1877

Independently verified PASS: #1876/#1877 PASS with drop-oldest-safety and connect-race confirmations; #1879 functional change correct; full fast suite green (8324 passed, 0 failed) after the isolation fix.

### DIFF
--- a/src/rigplane/diagnostics/_logging.py
+++ b/src/rigplane/diagnostics/_logging.py
@@ -82,7 +82,9 @@ def configure_diagnostic_logging() -> None:
         # The handler's own level is DEBUG, so any record the logger doesn't
         # filter still hits the file — see spec §4.1.
         if icom_logger.level == logging.NOTSET:
-            icom_logger.setLevel(logging.DEBUG)
+            icom_logger.setLevel(
+                logging.INFO
+            )  # was DEBUG: avoids ~40-57 disk writes/s on the loop (see issue #1879)
     except Exception as exc:  # noqa: BLE001 — best-effort init, swallow all
         sys.stderr.write(f"rigplane: diagnostic logging disabled: {exc}\n")
 
--- a/src/rigplane/web/server.py
+++ b/src/rigplane/web/server.py
@@ -1347,10 +1347,9 @@ class WebServer:
 
         target_queues = self._control_event_queues if queues is None else queues
         for q in list(target_queues):
-            try:
-                q.put_nowait(event)
-            except asyncio.QueueFull:
-                logger.warning("control state queue full; dropping state_update")
+            # A state_update supersedes older queued state; coalesce-to-latest on a
+            # stalled/slow consumer instead of dropping the freshest + log-flooding.
+            q.put_drop_oldest(event)
 
     async def _delayed_state_broadcast(self, delay: float) -> None:
         try:
--- a/src/rigplane/web/server.py
+++ b/src/rigplane/web/server.py
@@ -1316,6 +1316,10 @@ class WebServer:
         # from the same canonical snapshot used for Web delivery.
         self._update_fft_scope_freq(snapshot)
         self._update_fft_scope_mode(snapshot)
+        # Skip the public-state build/delta/fan-out when no control client is
+        # subscribed (a connecting client gets initial state on connect via force=True).
+        if not force and not self._control_event_queues:
+            return
         body = self._build_public_state_from_snapshot(snapshot)
         state_key = self._public_state_delivery_key(
             snapshot,
--- a/tests/test_diagnostics_logging.py
+++ b/tests/test_diagnostics_logging.py
@@ -132,7 +132,7 @@ def test_log_file_writes_to_platformdirs_cache(
     monkeypatch.setattr(platformdirs, "user_cache_path", lambda app: tmp_path)
     configure_diagnostic_logging()
     logger = logging.getLogger("rigplane.test")
-    logger.debug("hello")
+    logger.info("hello")  # use INFO: default level is now INFO (not DEBUG, see #1879)
     log_file = tmp_path / "logs" / "rigplane.log"
     assert log_file.exists()
     assert "hello" in log_file.read_text(encoding="utf-8")
@@ -147,7 +147,9 @@ def test_log_file_honors_rigplane_log_dir(
 
     configure_diagnostic_logging()
     logger = logging.getLogger("rigplane.test")
-    logger.debug("hello from env")
+    logger.info(
+        "hello from env"
+    )  # use INFO: default level is now INFO (not DEBUG, see #1879)
 
     log_file = log_dir / "rigplane.log"
     assert log_file.exists()
@@ -175,12 +177,17 @@ def test_preset_logger_level_preserved(
         icom_logger.setLevel(logging.NOTSET)
 
 
-def test_unset_logger_level_set_to_debug(
+def test_unset_logger_level_set_to_info(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
-    """When the logger level is NOTSET, init promotes it to DEBUG."""
+    """When the logger level is NOTSET, init promotes it to INFO (not DEBUG).
+
+    Changed in #1879: defaulting to DEBUG caused ~40-57 disk writes/sec on
+    the event loop at idle. INFO is the new production default; DEBUG remains
+    opt-in via an explicit setLevel call before configure_diagnostic_logging.
+    """
     monkeypatch.setattr(platformdirs, "user_cache_path", lambda app: tmp_path)
     icom_logger = logging.getLogger("rigplane")
     icom_logger.setLevel(logging.NOTSET)
     configure_diagnostic_logging()
-    assert icom_logger.level == logging.DEBUG
+    assert icom_logger.level == logging.INFO
--- a/tests/test_diagnostics_logging.py
+++ b/tests/test_diagnostics_logging.py
@@ -18,13 +18,20 @@ from rigplane.diagnostics._logging import (
 
 @pytest.fixture(autouse=True)
 def _enable_diagnostic_logging(monkeypatch: pytest.MonkeyPatch) -> Any:
-    """Re-enable diagnostic logging locally for each test, then clean up."""
+    """Re-enable diagnostic logging locally for each test, then clean up.
+
+    Also snapshots and restores the ``rigplane`` logger level so that tests
+    calling ``configure_diagnostic_logging()`` (which promotes NOTSET → INFO
+    since #1879) do not pollute the logger state for subsequent tests.
+    """
     monkeypatch.delenv("RIGPLANE_DISABLE_DIAGNOSTIC_LOGGING", raising=False)
-    yield
     icom_logger = logging.getLogger("rigplane")
+    saved_level = icom_logger.level
+    yield
     icom_logger.handlers = [
         h for h in icom_logger.handlers if not isinstance(h, SafeRotatingFileHandler)
     ]
+    icom_logger.setLevel(saved_level)
 
 
 def test_handler_attached_to_icom_lan_logger_not_root(
--- a/tests/test_quickwins_regression.py
+++ b/tests/test_quickwins_regression.py
@@ -1,0 +1,203 @@
+"""Regression tests for three quick-win infra fixes.
+
+Issues:
+- #1876: control-state broadcast floods logs when a subscriber stalls
+- #1879: diagnostic logging forced to DEBUG by default
+- #1877 (2a only): state-broadcast does heavy work even with zero subscribers
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import Any
+from unittest.mock import patch
+
+import platformdirs
+import pytest
+
+from rigplane._bounded_queue import BoundedQueue
+from rigplane.diagnostics._logging import (
+    SafeRotatingFileHandler,
+    configure_diagnostic_logging,
+)
+from rigplane.web.server import WebServer
+
+
+# ---------------------------------------------------------------------------
+# Fix A (#1876) — stalled subscriber must NOT produce log warnings
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(autouse=False)
+def _fresh_icom_logger() -> Any:
+    """Ensure the rigplane logger is clean between tests."""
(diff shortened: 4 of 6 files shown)


morozsm:
Agent Review: PASS — independent review completed; #1876/#1877 PASS with drop-oldest-safety and connect-race confirmations, #1879 functional change correct and full fast suite green after isolation fix.

PR #1940, opened 2026-07-27
Core: own managed TX runtime lifetime

morozsm:
## Summary

- add one backend-neutral `ManagedRadioRuntime` that privately owns a single `TxSafetySupervisor` and provider-generation counter per target runtime
- order provider-ready TX-effect servicing before the lifecycle seam returns
- bound shutdown de-key work and always release the provider without clearing unresolved TX state

Linear: [MOR-1007](https://linear.app/morozsm/issue/MOR-1007/core-own-one-tx-supervisor-per-managed-radio-runtime)
Accepted decision: [MOR-986](https://linear.app/morozsm/issue/MOR-986/decision-define-backend-tx-ownership-stale-queue-and-session-loss)
Closes #1935

## Scope

Exactly two new files and 200 net lines. This is lifetime ownership only. Production factory/CLI assembly, Web/rigctld ingress, managed SDK facade, provider PTT execution/readback, and IC-7610 authority remain later slices. No public API, CLI, config, rigctld wire, server, or frontend behavior changes.

## Verification

- `uv run pytest -q --tb=short tests/test_managed_radio_runtime.py` — 7 passed
- full non-integration pytest — 8273 passed, 73 skipped, 11 xfailed, 1 xpassed
- `uv run ruff check src/ tests/` — passed
- `uv run ruff format --check src/ tests/` — passed
- targeted runtime mypy + strict Web mypy — passed
- import-linter — 5/5 contracts kept
- full repository mypy baseline still has 13 pre-existing errors in three untouched runtime files
- diff check — exactly two files, 200 insertions

## Review notes

This PR is intentionally draft and must receive independent safety review before merge.

PR #1949, opened 2026-07-27
Core: bind managed TX effect service and request seam

### LINKED ISSUE #1945: Bind provider effect service and expose managed TX requests
Linear: https://linear.app/morozsm/issue/MOR-1040/core-bind-provider-effect-service-and-expose-managed-tx-requests-on
Parent blocker: https://linear.app/morozsm/issue/MOR-1009/core-route-managed-sdk-ptt-through-managedtxapi
Prerequisite merged in PR #1940 / issue #1935.

## Scope

Bind one private provider-effect service to each `ManagedRadioRuntime` and expose narrow async `request_on`, `request_off`, and `release_owner` methods through the runtime-owned `TxSafetySupervisor`.

Owned files only:

## Acceptance

- Runtime privately retains exactly one effect service for its lifetime/provider generation.
- Each managed request reduces through the owned supervisor exactly once and services the resulting effects exactly once.
- Returned transitions preserve owner/lease identity needed by later facades.
- Provider service errors propagate while durable OFF state remains pending.
- Provider replacement, readiness, and shutdown use the same bound executor and cannot recurse through a facade.
- No raw provider or bare `set_ptt` is exposed.
- No `radio_protocol.py`, sync, Web, rigctld, IC-7610, poller, frontend, or provider-specific integration changes.
- Focused/full non-integration tests, Ruff/format, mypy, import-linter, and diff check pass.

No more than 200 net lines.

morozsm:
Closes #1945

Linear: https://linear.app/morozsm/issue/MOR-1040/core-bind-provider-effect-service-and-expose-managed-tx-requests-on
Blocks parent: https://linear.app/morozsm/issue/MOR-1009/core-route-managed-sdk-ptt-through-managedtxapi
Prerequisite: #1940

## What changed

- Bind one private `TxService` when `ManagedRadioRuntime` is constructed.
- Route provider replacement, readiness, shutdown, and managed TX requests through that same executor.
- Add narrow async `request_on`, `request_off`, and `release_owner` methods that reduce once, service emitted effects once, and return the supervisor transition with owner/lease identity.
- Preserve durable OFF state when the provider service raises while propagating the error.

## Safety properties

- No provider object or bare `set_ptt` callback is exposed.
- The effect executor receives only the owned supervisor and emitted transition, so it cannot recurse through a facade.
- Supervisor reduction happens before service; service failures cannot roll back release state.
- Shutdown remains bounded and releases the provider in `finally`.

## Scope

Exactly two approved files, +130/-40 (net +90). No `radio_protocol.py`, sync facade, Web, rigctld, IC-7610, poller, frontend, or provider-specific integration.

## Verification

- focused managed runtime: 9 passed
- full non-integration: 8343 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed
- Ruff check: passed
- Ruff format check: passed
- targeted mypy for changed runtime: passed
- import-linter: 5/5 contracts kept
- diff check: passed

Full `mypy src/` retains the known baseline of 13 errors in three untouched files: `_control_phase.py`, `_audio_runtime_mixin.py`, and `_civ_rx.py`.

Independent safety review is required before this draft is made ready.

### DIFF
--- a/src/rigplane/runtime/managed_radio_runtime.py
+++ b/src/rigplane/runtime/managed_radio_runtime.py
@@ -7,6 +7,7 @@ from rigplane.core.tx_safety import (
     Clock,
     IdFactory,
     TxOutcome,
+    TxOwner,
     TxReleaseReason,
     TxSafetySupervisor,
     TxTransition,
@@ -21,6 +22,7 @@ class ManagedRadioRuntime:
         self,
         target_id: str,
         *,
+        service: TxService,
         clock: Clock | None = None,
         id_factory: IdFactory | None = None,
         shutdown_timeout_seconds: float = 3.0,
@@ -29,6 +31,7 @@ class ManagedRadioRuntime:
             raise ValueError("shutdown_timeout_seconds must be finite-positive")
         self.target_id = target_id
         self._tx_safety = TxSafetySupervisor(clock=clock, id_factory=id_factory)
+        self._service = service
         self._provider_generation = 0
         self._shutdown_timeout = shutdown_timeout_seconds
 
@@ -36,37 +39,56 @@ class ManagedRadioRuntime:
     def tx_safety(self) -> TxSafetySupervisor:
         return self._tx_safety
 
-    async def replace_provider(
-        self, *, ready: bool, service: TxService
-    ) -> TxTransition:
+    async def _service_effects(self, transition: TxTransition) -> None:
+        if transition.effects:
+            await self._service(self._tx_safety, transition)
+
+    async def replace_provider(self, *, ready: bool) -> TxTransition:
         self._provider_generation += 1
         transition = self.tx_safety.replace_provider(
             self._provider_generation, ready=ready
         )
-        if transition.effects:
-            await service(self.tx_safety, transition)
+        await self._service_effects(transition)
         return transition
 
-    async def set_provider_ready(
-        self, *, ready: bool, service: TxService
-    ) -> TxTransition:
+    async def set_provider_ready(self, *, ready: bool) -> TxTransition:
         transition = self.tx_safety.set_provider_ready(
             self._provider_generation, ready=ready
         )
-        if transition.effects:
-            await service(self.tx_safety, transition)
+        await self._service_effects(transition)
         return transition
 
-    async def shutdown(
-        self, *, dekey: TxService, release_provider: ProviderRelease
+    async def request_on(self, owner: TxOwner) -> TxTransition:
+        transition = self._tx_safety.request_on(owner)
+        await self._service_effects(transition)
+        return transition
+
+    async def request_off(
+        self,
+        owner: TxOwner,
+        lease_id: str,
+        *,
+        reason: TxReleaseReason = TxReleaseReason.OPERATOR_RELEASE,
     ) -> TxTransition:
+        transition = self._tx_safety.request_off(owner, lease_id, reason=reason)
+        await self._service_effects(transition)
+        return transition
+
+    async def release_owner(
+        self, owner: TxOwner, *, reason: TxReleaseReason
+    ) -> TxTransition:
+        transition = self._tx_safety.release_owner(owner, reason=reason)
+        await self._service_effects(transition)
+        return transition
+
+    async def shutdown(self, *, release_provider: ProviderRelease) -> TxTransition:
         transition = self.tx_safety.emergency_release(
             reason=TxReleaseReason.SERVER_SHUTDOWN
         )
         try:
             if transition.outcome is not TxOutcome.NOOP:
                 await asyncio.wait_for(
-                    dekey(self.tx_safety, transition),
+                    self._service_effects(transition),
                     timeout=self._shutdown_timeout,
                 )
         except TimeoutError:
--- a/src/rigplane/runtime/managed_radio_runtime.py
+++ b/src/rigplane/runtime/managed_radio_runtime.py
@@ -9,6 +9,7 @@ from rigplane.core.tx_safety import (
     TxOutcome,
     TxOwner,
     TxReleaseReason,
+    TxSafetySnapshot,
     TxSafetySupervisor,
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: BLOCKED

Reviewed exact head `e0aefad519e62fae3fa33ab63088c21d3b33b2c4` independently against GitHub issue #1945, MOR-1040, the accepted MOR-986 backend TX contract, and the merged supervisor model.

## [P1] Public mutable supervisor accessor bypasses both the bound effect service and target-owned generation

`ManagedRadioRuntime.tx_safety` (`src/rigplane/runtime/managed_radio_runtime.py:38-40`) returns the live mutable `TxSafetySupervisor`. A caller that receives the managed target host can therefore bypass every new async seam:

- `rt.tx_safety.request_on(owner)` emits a `WRITE_ON` while the bound `_service` is called zero times;
- `rt.tx_safety.replace_provider(999, ready=True)` advances the supervisor independently of `_provider_generation`; the next `await rt.replace_provider(...)` returns `STALE` and leaves generation `999` active.

I reproduced both on this head:

```text
effects_via_public_supervisor 1 bound_service_calls 0
public_generation 999 host_next_outcome stale host_generation 999
```

This contradicts this slice's one-entry/exactly-once service contract and MOR-986's rule that the managed API is the exposed ingress surface while provider/generation mutation stays internal. It also leaves no mechanical guard against a later managed facade, Web, rigctld, or SDK adapter bypassing ownership and effect execution.

Required remediation:

1. Remove the public mutable supervisor accessor from the managed host. If observation is needed for diagnostics/tests, expose a read-only snapshot; keep the actual supervisor private and pass it only to the trusted bound effect executor.
2. Add a regression proving managed callers cannot obtain a mutable supervisor to issue `request_on` or replace the provider generation outside the host seam.
3. Keep provider observation/settlement entry points controlled by the host in their approved later slice; do not reintroduce a generic mutable-supervisor escape hatch.

The current repository has no production consumer of `ManagedRadioRuntime`, so fixing this now does not break an established shipped call site. The class was introduced by the immediately preceding internal host slice.

## Verification completed

- focused managed-runtime tests: `9 passed`
- full non-integration backend suite: `8343 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed`
- Ruff check: passed
- Ruff format check: passed
- changed-source mypy: passed
- import-linter: `5 kept, 0 broken`
- full mypy: head and exact merge-base both retain the identical known `13` errors in the same three untouched files
- diff guard: exactly two owned files, `+130/-40` (net `+90`), no whitespace errors
- exact-head quick and grep CI: green

Concurrency note: allowing the bound executor to receive a cancellation while a provider attempt is in flight can require concurrent callback entry; therefore I am not requesting a lock around `await service` in this slice. The future effect-executor integration must define and test its reentrancy/serialized-lane contract. That is separate from the concrete public bypass above.

Do not mark ready or merge this head.

morozsm:
Remediation pushed for the independent P1 finding.

Exact head: 866b0d4a56e10614297fbbf235c40e68ecdc435f

Changes:
- removed the public mutable tx_safety supervisor accessor;
- exposed only the frozen TxSafetySnapshot through tx_snapshot;
- kept provider generation, readiness, managed request, and shutdown mutation on the private runtime-owned supervisor;
- added a red-before regression proving the managed host has no public mutable supervisor, the diagnostic snapshot has no request/provider-replacement methods, and its generation cannot be mutated.

Consumer scan found no production ManagedRadioRuntime consumer and no production .tx_safety access, so no supported caller was broken. Provider observation/settlement host entry points remain deferred to their approved later integration slice; tests use private white-box setup only.

Verification on this head:
- focused managed runtime: 10 passed
- full non-integration: 8283 passed, 73 skipped, 5 deselected, 11 xfailed, 1 xpassed
- full Ruff check and format check: passed
- changed-source mypy: passed
- import-linter: 5 kept, 0 broken
- diff check: passed
- scope from issue base: 2 files, +156/-50, net +106

PR remains draft and requires fresh independent review for this exact head.

morozsm:
Agent Review: PASS

Fresh independent review completed for exact head `866b0d4a56e10614297fbbf235c40e68ecdc435f` against issue #1945, MOR-1040, the accepted MOR-986 managed-TX boundary, and the prior P1 bypass report.

Evidence:
- The live mutable `tx_safety` accessor is gone. The only observation surface is `tx_snapshot`, a detached frozen value whose nested `TxOwner` and `ProviderAttempt` values are also frozen; it contains no supervisor, service, mutation method, or provider reference.
- I reproduced the old bypass attempts: no public supervisor can issue `request_on` or poison provider generation. Ordinary mutation of snapshot, nested owner, transition, and effect fields fails; stale snapshots remain unchanged across later host mutations; concurrent snapshot reads are stable.
- The complete public surface was enumerated. Managed request, replacement, readiness, and shutdown mutations all go through the private runtime-owned supervisor, and every emitted transition is handed once to the single bound service. No alternate public service/provider/supervisor escape route exists.
- Injected provider-service failure during a ready provider replacement propagates, retains the release obligation, and leaves host/supervisor generation synchronized so the next replacement advances normally rather than becoming stale.
- Production-consumer scan found no supported `ManagedRadioRuntime` consumer or `.tx_safety` caller, so the correction breaks no shipped call site.
- Atomic scope is exactly two approved files, `+156/-50` (net `+106`), with no diff-check errors.

Verification on this exact head:
- focused runtime: `10 passed`
- full non-integration: `8283 passed, 73 skipped, 5 deselected, 11 xfailed, 1 xpassed`
- Ruff check and format check: pass
- changed-source mypy: pass
- import-linter: `5 kept, 0 broken`
- exact-head quick and grep CI: green

Non-blocking follow-up: the future provider effect executor still needs an explicit serialized-lane/reentrancy contract for concurrent callbacks and cancellation settlement. This prerequisite seam does not add overlapping execution or weaken current safety guarantees.

Issue #1968, opened 2026-07-27
Project freshness-safe TX target into public runtime state


morozsm:
## Summary

Implement Linear MOR-1106 by projecting canonical `global.tx_state.tx_target` into both public ServerState producer paths without creating a second freshness authority.

## Acceptance criteria

- [ ] Dataclass and StateStore snapshot producers always emit camelCase `txTarget`.
- [ ] Missing target emits `{status:"unknown", reason:"not-observed"}`.
- [ ] Fresh known target maps receiver, slot, and frequency losslessly.
- [ ] A canonically stale target emits `{status:"unknown", reason:"stale"}` and leaks no stale receiver, slot, or frequency.
- [ ] Fresh valid `unsupported` and `contradiction` reasons pass through unchanged.
- [ ] Malformed impossible snapshot values fail closed and never leak partial target data.
- [ ] `fieldStatus.txTarget` reports the canonical store path/freshness coherently and remains diagnostic only.
- [ ] Focused tests cover the reason/freshness matrix and fresh-to-stale transition.

## Scope / non-goals

- `src/rigplane/web/runtime_helpers.py`
- `tests/test_web_runtime_helpers.py`
- Maximum 2 files and 200 net lines.
- Reuse `StateSnapshot` / `FieldSnapshot.freshness` and the merged TxTarget domain.
- No provider acquisition, schema/codegen, frontend selector/component, capability, model, split, selected-RX, or TX lifecycle changes.

## Compatibility

Additive population of the already-staged optional `ServerStatePublic.txTarget`; no CLI, config, rigctld wire, or public command API change.

## Test plan

Focused runtime helper and schema-conformance tests, full non-integration suite, Ruff/format, strict web mypy, import-linter, diff guard, and exact-head CI.

PR #1974, opened 2026-07-27
feat: publish FTX-1 native FT readback as TxTarget

### LINKED ISSUE #1969: MOR-1049: publish FTX-1 native FT readback as backend-neutral TxTarget
Linear: MOR-1049

Implement the bounded Yaesu CAT/FTX-1 acquisition slice for `global.tx_state.tx_target`.

Acceptance:
- Native FT readback is authoritative: FT0 -> MAIN, FT1 -> SUB.
- FTX-1 shared A/B scheme publishes `slot=null`.
- Frequency is included only when obtained coherently and freshly for the same receiver; otherwise it remains null.
- Invalid/missing/unsupported/stale/contradictory evidence publishes the matching `UnknownTxTarget`.
- Never infer TX target from split, selected RX, FR, active receiver, or a default VFO.
- Tests cover MAIN/SUB, missing frequency, stale/mismatched frequency evidence, invalid FT, reconnect/reset/unsupported, and no split fallback.

Scope is limited to the Yaesu observation adapter, FTX-1 acquisition profile if required, and focused tests. No web/schema/frontend/audio changes.

morozsm:
Closes #1969

Linear: MOR-1049
Depends on merged MOR-1119 / #1992.

## Summary
- publish FTX-1 native FT readback as the authoritative MAIN/SUB `TxTarget`
- bracket the selected receiver frequency with FT-before / selected-frequency / FT-after and reject a changed FT or CAT generation fail closed
- require exact `int` FT values 0/1, rejecting bool/float/out-of-range evidence; map missing/not-implemented/rejected FT support to explicit unknown
- keep `ab_shared` slot null and rely on the merged poller generation invalidation to expire prior authority on reconnect/provider changes

## Scope
- exact owned 3 files, 200 insertions / 1 deletion = 199 net LOC
- no schema, frontend, audio, poller, or public provider API changes

## Verification
- focused adapter + current poller + state contracts: 302 passed
- production Yaesu parser: 42 passed
- Ruff lint + format check
- strict mypy on changed production source
- import-linter: 5 contracts kept
- `git diff --check`
- exact-head GitHub quick + grep checks

Fresh independent exact-head review is still required before this Draft may become ready or merge.

### DIFF
--- a/rigs/ftx1.toml
+++ b/rigs/ftx1.toml
@@ -77,6 +77,7 @@ polling_only = [
     "receiver.sub.active.freq_mode.freq_hz",
     "receiver.sub.active.freq_mode.mode",
     "global.tx_state.ptt",
+    "global.tx_state.tx_target",
     "receiver.main.operator_controls.af_level",
     "receiver.main.operator_controls.rf_gain",
     "receiver.main.operator_controls.squelch",
--- a/src/rigplane/backends/yaesu_cat/observations.py
+++ b/src/rigplane/backends/yaesu_cat/observations.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Protocol, TypeVar
 from rigplane.core.observation_adapter import ProviderObservationAdapter
 from rigplane.core.state_acquisition_policy import RadioAcquisitionProfile
 from rigplane.core.state_pipeline_contracts import FieldPath, Observation
+from rigplane.core.tx_target import KnownTxTarget, TxTarget, UnknownTxTarget
 from rigplane.runtime.meter_cal import interpolate_meter
 
 from .parser import CatFormatError, CatParseError
@@ -32,6 +33,7 @@ _MAIN_MODE = FieldPath.active("main", "freq_mode", "mode")
 _SUB_FREQ = FieldPath.active("sub", "freq_mode", "freq_hz")
 _SUB_MODE = FieldPath.active("sub", "freq_mode", "mode")
 _PTT = FieldPath.global_("tx_state", "ptt")
+_TX_TARGET = FieldPath.global_("tx_state", "tx_target")
 _MAIN_AF = FieldPath.receiver("main", "operator_controls", "af_level")
 _MAIN_RF = FieldPath.receiver("main", "operator_controls", "rf_gain")
 _MAIN_SQL = FieldPath.receiver("main", "operator_controls", "squelch")
@@ -182,6 +184,8 @@ class YaesuObservationRadio(Protocol):
 
     async def read_ptt(self) -> bool: ...
 
+    async def get_tx_func(self) -> int: ...
+
     async def read_af_level(self, receiver: int = 0) -> int: ...
 
     async def read_rf_gain(self, receiver: int = 0) -> int: ...
@@ -313,6 +317,14 @@ class YaesuObservationAdapter:
                 observations.append(
                     adapter.observation(_SUB_MODE, result[0], native_id="read_mode")
                 )
+        if self._can_poll(_TX_TARGET):
+            observations.append(
+                adapter.observation(
+                    _TX_TARGET,
+                    await self._read_tx_target(),
+                    native_id="get_tx_func",
+                )
+            )
         if self._can_poll(_PTT):
             ok, value = await self._safe_read("ptt", self.radio.read_ptt())
             if ok:
@@ -336,6 +348,61 @@ class YaesuObservationAdapter:
                     )
                 )
         return tuple(observations)
+
+    async def _read_tx_func(self) -> int | UnknownTxTarget:
+        method = getattr(self.radio, "get_tx_func", None)
+        if not callable(method):
+            return UnknownTxTarget(reason="unsupported")
+        try:
+            tx_func = await method()
+        except (AttributeError, NotImplementedError, CatCommandRejected):
+            return UnknownTxTarget(reason="unsupported")
+        except (CatParseError, CatFormatError, ValueError, KeyError):
+            return UnknownTxTarget(reason="contradiction")
+        if tx_func is None:
+            return UnknownTxTarget(reason="not-observed")
+        if type(tx_func) is not int or tx_func not in (0, 1):
+            return UnknownTxTarget(reason="contradiction")
+        return tx_func
+
+    async def _read_tx_target(self) -> TxTarget:
+        """Bracket the selected frequency with stable native FT authority."""
+        if not self._has_runtime_capability("tx"):
+            return UnknownTxTarget(reason="unsupported")
+        stats = getattr(getattr(self.radio, "_transport", None), "stats", None)
+        reconnects = getattr(stats, "reconnects", 0)
+        generation = (
+            self.profile.provider,
+            reconnects if type(reconnects) is int else 0,
+        )
+        before = await self._read_tx_func()
+        if isinstance(before, UnknownTxTarget):
+            return before
+        ok, raw_frequency = await self._safe_read(
+            "tx_target.freq", self.radio.read_freq(before)
+        )
+        after = await self._read_tx_func()
+        current_reconnects = getattr(stats, "reconnects", 0)
+        current_generation = (
+            self.profile.provider,
+            current_reconnects if type(current_reconnects) is int else 0,
+        )
+        if generation != current_generation:
+            return UnknownTxTarget(reason="not-observed")
+        if isinstance(after, UnknownTxTarget):
+            return after
+        if after != before:
+            return UnknownTxTarget(reason="contradiction")
+        frequency = (
+            raw_frequency
+            if ok and type(raw_frequency) is int and raw_frequency > 0
+            else None
+        )
+        return KnownTxTarget(
+            receiver="MAIN" if before == 0 else "SUB",
+            slot=None,
+            frequency_hz=frequency,
+        )
 
     async def poll_rx_meters(
         self,
--- a/tests/test_yaesu_cat_observation_adapter.py
+++ b/tests/test_yaesu_cat_observation_adapter.py
@@ -5,11 +5,12 @@ from __future__ import annotations
 
 from dataclasses import replace
 from types import SimpleNamespace
-from unittest.mock import AsyncMock, MagicMock
+from unittest.mock import AsyncMock, MagicMock, call
 
 import pytest
 
 from rigplane.core.state_acquisition_policy import RadioAcquisitionProfile
+from rigplane.core.tx_target import KnownTxTarget, TxReceiver, UnknownTxTarget
 from rigplane.core.types import BreakInMode
 from rigplane.profiles import get_radio_profile
 from rigplane.radio_state import RadioState
@@ -17,6 +18,7 @@ from rigplane.radio_state import RadioState
 from rigplane.backends.yaesu_cat.observations import YaesuObservationAdapter
 from rigplane.backends.yaesu_cat.parser import CatParseError
 from rigplane.backends.yaesu_cat.radio import RadioConnectionError, YaesuCatRadio
+from rigplane.backends.yaesu_cat.transport import CatCommandRejected
 
 
 def _clock() -> float:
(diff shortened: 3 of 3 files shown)


morozsm:
Agent Review: BLOCKED

Reviewed exact head `5e4081cf8570d3c881a0c4a89120873bc30b8552` against MOR-1047, MOR-1049, MOR-446, the FTX-1 profile, and the exact three-file/+182 LOC diff.

Blocking findings:

1. **[P1] The target frequency is not coherently bound to FT authority** (`src/rigplane/backends/yaesu_cat/observations.py:293-338`). Production order is `FA -> mode -> FB -> mode -> FT`, with only one FT sample after both frequencies. An external front-panel/CAT change between `FB` and `FT` can therefore publish a mixed known target. Reproduction: `FB` returned 7 MHz; before the final FT read the radio changed FT MAIN->SUB and SUB to 8 MHz; the adapter published `KnownTxTarget(SUB, null, 7 MHz)` while the authoritative FT-time target was SUB/8 MHz. The poller's serial lock prevents local request interleaving, but it cannot prevent front-panel/external-CAT changes. Required fix: bracket the exact target-frequency read with authoritative FT before/after (or otherwise prove an equivalent generation-bound transaction); if FT changes, publish an explicit unknown or keep identity only with `frequencyHz:null`. Add an event-ordered production-poller test for FT toggling during the cycle.

2. **[P1] Reconnect/reset does not clear prior authority** (`observations.py:367-378`; `tests/test_yaesu_cat_observation_adapter.py:694-704`). The test named `...reconnect_and_stale_decay` proves only that `RadioConnectionError` propagates. After applying one known target, a subsequent FT connection/reset failure produces no observation and leaves the old StateStore entry `fresh` and known; it remains authoritative until the generic 8-second TTL expires. There is no provider generation on the observation and no reconnect/reset invalidation. Required fix: invalidate to `stale`/`not-observed` at the reconnect generation boundary and reject old-generation/late observations. If this cannot fit the frozen three-file scope, split the necessary poller/runtime invalidation slice before proceeding. Add a real StateStore + reconnect/reset regression, not only `pytest.raises`.

3. **[P1] Invalid/unsupported FT evidence is not fully fail-closed** (`observations.py:367-388`). Python key equality makes `True` and `1.0` map to SUB, and `False` and `0.0` map to MAIN. Reproductions publish known targets for all four invalid values. A missing `get_tx_func` raises `AttributeError`, and a provider `NotImplementedError` propagates instead of producing `unknown: unsupported`. Required fix: require `type(tx_func) is int` and exactly `{0,1}`; map missing/not-implemented provider support to `unsupported` while preserving true connection failures for reconnect handling. Add bool/float/negative/large/missing/not-implemented cases.

Test-evidence gap: the new fixtures default `get_tx_func` to `AsyncMock(return_value=0)`, so every unrelated medium-poll fixture is pre-shaped to produce the expected MAIN target. Add at least one production-faithful `YaesuCatRadio`/profile-command test and explicit call-order assertions; the current tests do not prove the real `FT;` parser path, generation reset, or coherent ordering.

Checks completed before verdict:

- focused adapter/profile/poller/domain/projection matrix: `420 passed`
- focused Ruff and format: PASS
- strict mypy on changed production source: PASS
- import-linter: 5 contracts kept
- diff check: PASS
- exact-head GitHub `quick` and `grep-gate`: PASS
- scope: exact 3 files, +182 net LOC

A full non-integration run was still progressing beyond 93% without failures when this safety verdict was posted; passing broad tests cannot cover the reproduced authority races above.

Please fix the three authority/fail-closed gaps and request a fresh independent review on the new head. PR remains draft.

morozsm:
Verification update for the BLOCKED review above: the full non-integration run completed cleanly (`8389 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed` in 287.88s). The reproduced TxTarget authority/fail-closed defects remain blocking.

morozsm:
Agent Review: PASS

Fresh independent review at exact head `05a4c2626e5920186ebbb449b22cd9a4f4b7b22e` (local `HEAD`, `origin/codex/1969-ftx1-tx-target`, and PR `headRefOid` all verified equal; worktree clean; single commit above `origin/main` = `15f958adf1d0990313865e50835b89154824da12`). I did not author this change.

## Invariants verified against the diff

**1. Strict built-in integer FT semantics — fails closed.**
`_read_tx_func` (`src/rigplane/backends/yaesu_cat/observations.py:352-366`) requires `type(tx_func) is not int or tx_func not in (0, 1)`. The type test is the *left* operand of the `or`, so short-circuit evaluation means `True`/`False`/`1.0`/`0.0` never reach the membership test — the `True == 1` / `1.0 == 1` equality trap that blocked the previous head is closed. Out-of-range (`-1`, `2`, `2**63`) → `contradiction`. Non-callable/missing attribute, `AttributeError`, `NotImplementedError`, `CatCommandRejected` → `unsupported`. `CatParseError`/`CatFormatError`/`ValueError`/`KeyError` → `contradiction`. `None` → `not-observed`. `RadioConnectionError`/`CatTimeoutError` are *not* caught and propagate, so a dead link still drives the poller's reconnect/backoff instead of being masked as a field skip.

**2. One coherent FT-before → selected-receiver frequency → FT-after observation.**
`_read_tx_target` (`observations.py:368-405`) reads FT, then `read_freq(before)` for exactly that receiver, then FT again, and returns `contradiction` if `after != before`. `before` is a strict int in `{0, 1}`, and the code branches on `isinstance(before, UnknownTxTarget)` rather than truthiness, so `FT0` (falsy) is handled correctly. `get_tx_func` → `0 = MAIN TX, 1 = SUB TX` (`radio.py:1561-1564`) and `read_freq(0/1)` → `MAIN (VFO-A) / SUB (VFO-B)` (`radio.py:833-841`), so the `0 → "MAIN"` / `1 → "SUB"` mapping is the codebase's own convention, not an inference. The bracket is atomic against *local* CAT traffic: `_run_poll_cycle` holds `self._lock` across `_drain_commands()` + `_poll_medium()` (`poller.py:329-332`, `668-671`), so a queued `set_tx_func` is applied before the bracket, never inside it; *external* front-panel/CAT changes are exactly what the FT-after check catches. A mismatched-template reply (e.g. `FB;` answered in main form) raises `CatParseError` at `parser.py:309-315` rather than being silently accepted.

**3. Frequency degrades to null, identity does not.**
A field-level failure of the frequency read yields `frequencyHz: null` while the receiver identity stays FT-bracket-verified. This matches issue #1969 verbatim: *"Frequency is included only when obtained coherently and freshly for the same receiver; otherwise it remains null."* `KnownTxTarget.__post_init__` and `KnownTxTargetPublic` both admit `frequencyHz: int(gt=0) | None`, and the bool/non-positive guard at `observations.py:396-400` prevents a bogus value reaching the dataclass.

**4. No split / FR / VS / active-receiver fallback.**
The new path calls only `get_tx_func` and `read_freq`. No `read_split`, `read_vfo_select`, or `get_rx_func` anywhere in `_read_tx_target`/`_read_tx_func`. `slot` is unconditionally `None`, per the FTX-1 shared A/B acceptance criterion.

**5. No independent authority; relies on the merged poller invalidation.**
Generation invalidation is the already-merged `d786d898` (MOR-1112, #1979): `_current_tx_target_generation()` = `(profile.provider, transport.stats.reconnects)`; `_emit_medium_observations` captures it before `poll_medium()`, drops any `tx_target` observation whose generation changed across the poll, and calls `_invalidate_tx_target()` (`poller.py:226-252`). `poll_medium` has exactly one caller — that method — so every publication is bracketed by it. The adapter's inline generation check re-derives the *same* tuple from the *same* two inputs over a window strictly contained in the poller's, cannot publish anything itself, and fails only conservatively; it is a redundant defensive copy, not a competing authority. Two notes for the record, both subsumed by the poller and therefore non-blocking: the adapter binds `stats` once before the bracket (the poller re-resolves `self._radio._transport` on each check), and it coerces int-subclasses to `0` via `type(...) is int` where the poller uses `isinstance(...) and not bool`. `YaesuCatRadio._transport` is assigned exactly once (`radio.py:194`) and `TransportStats.reconnects` is monotonic (`transport.py:215`), so the counter cannot silently reset within a radio's lifetime.

**6. Only a fresh, explicitly known target is published; in-flight completions cannot survive reconnect.**
A reconnect during the bracket bumps `reconnects`, and the poller's post-poll check drops the observation. A disconnect mid-poll raises `CatTransportError`/`RadioConnectionError`, which `_emit_medium_observations` catches to call `_invalidate_tx_target()` (emitting `stale`/`not-observed`) before re-raising. Cancellation propagates as `BaseException`, so no partial observation is emitted. After a clean stop, the field decays via the standard 8.0 s freshness TTL — identical to `ptt`/`freq`/`mode`, with the generation invalidation as an additional guard `tx_target` alone has.

## Commands run and results

- `uv run pytest tests/test_yaesu_cat_observation_adapter.py tests/test_yaesu_cat_poller.py tests/test_yaesu_cat_parser.py tests/test_web_runtime_helpers.py tests/test_state_pipeline_contracts.py tests/test_state_type_codegen.py tests/web/test_state_schema_conformance.py tests/test_yaesu_web_projection.py -q` → **454 passed** in 5.56 s
- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **8365 passed, 73 skipped, 5 deselected, 11 xfailed, 1 xpassed** in 182.46 s, zero failures
- `uv run ruff check src/ tests/` → All checks passed
- `uv run ruff format --check src/ tests/` → 627 files already formatted
- `uv run lint-imports` → 5 contracts kept, 0 broken (288 files, 804 dependencies)
- `uv run mypy src/` → 15 errors, **0 in any file this PR touches**; all are pre-existing in untouched `_control_phase.py` / `_audio_runtime_mixin.py` / `_civ_rx.py` / `web/runtime_helpers.py` (local mypy resolution differs from CI, which does not run mypy in `quick`)
- `git diff --check 15f958ad 05a4c262` → clean

Test quality: `test_real_ftx1_parser_brackets_selected_frequency` drives a real `YaesuCatRadio` through the real CAT parser and asserts the exact frame order `FT;` → `FA;`/`FB;` → `FT;` for both MAIN and SUB — independent evidence of the bracket rather than a mirror of the implementation. Note for readers: "real" there means the real parser/command-template path with a mocked transport `query`, **not** hardware.

## Scope

Exactly the 3 intended files: `rigs/ftx1.toml` (+1), `src/rigplane/backends/yaesu_cat/observations.py` (+67), `tests/test_yaesu_cat_observation_adapter.py` (+132/−1). **200 insertions / 1 deletion = 199 net LOC**, within the ≤3 files / ≤200 LOC guardrails. No schema, frontend, audio, poller, or public provider API changes.

## CI at exact head

`quick` **pass**, `grep-gate` **pass**, `Update Agent Review Gate status` **pass** — all reporting `head_sha = 05a4c2626e5920186ebbb449b22cd9a4f4b7b22e`. `Agent Review Gate` was failing only for "Missing Agent Review PASS comment for the current PR head"; this comment supplies it.

## Non-blocking observations

1. **Medium-lane CAT traffic grows ~50%.** The 5 Hz medium lane goes from 6 to 9 queries per cycle (`FT;` + a frequency read + `FT;`), and for `FT0` the `FA;` read duplicates the `read_freq(0)` already issued at the top of `poll_medium`. The duplicate is *justified* — reusing the earlier unbracketed value would break the coherence requirement — but the added per-query turnaround is worth measuring against the 200 ms budget during live acceptance.
2. **`_read_tx_target` is not gated on `dual_rx`.** `FT1` would publish `KnownTxTarget(SUB, …)` on a hypothetical single-RX Yaesu profile. Unreachable today: `ftx1` is the only `provider = "yaesu_cat"` rig and it declares `dual_rx`.
3. **Pre-existing, unrelated to this PR:** running the Python suite rewrites the committed golden `frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json` (adds a `txTarget` entry), because `tests/test_field_status_frontend_fixture.py:76-77` self-heals the fixture before asserting and therefore can never fail. The drift comes from the already-merged #1968 (`4f90dd87`/`ad97c6b5`), derives from `StateSnapshot.empty()` with `radio=None` — no input from `rigs/ftx1.toml` or the Yaesu adapter — and is equally present on `main`. I restored the file so the tree is back at exact head. Worth its own ticket; out of scope here.

## Hardware-evidence limitation (explicit)

**This review establishes the software contract only. There is no real FTX-1 hardware evidence in this PR, and none is claimed** — the PR body and commit message list software verification exclusively, and every new test uses a mocked transport. Live-radio acceptance of the FT readback (including FT toggling from the front panel during a poll cycle, cross-band split routing, and the medium-lane timing impact in observation 1 above) remains **separately outstanding** and is not covered by this PASS.

Verdict: no material issue found. PR may proceed once required checks are green. Not merging — that decision stays with the maintainer.

PR #1993, opened 2026-07-27
MOR-1109: bind authoritative PTT reads to observer epoch

### LINKED ISSUE #1990: MOR-1109: bind authoritative PTT reads to observer lifecycle epoch
Linear: MOR-1109
Parent: MOR-1103
Depends on: #1965 / merged PR #1967

## Outcome

Bind each authoritative IC-7610 PTT read to a lifecycle epoch distinct from provider generation and MOR-1010 observation sequence. Every bind and unbind increments the epoch, including same-generation rebinding of the identical callback. Publication succeeds only while the captured observer, provider generation, binding epoch, transport, and CI-V source generation remain current.

## Exact scope

- `src/rigplane/runtime/_civ_rx.py`
- `src/rigplane/runtime/radio.py`
- one focused existing radio test module
- maximum 3 files and 200 net LOC

No ManagedRadioRuntime state machine, executor, SDK routing, Web/poller/frontend, cache authority, or public provider API. This consumes the exact transaction primitive from #1965; it does not claim to complete the remaining MOR-1103 wrapper work.

## Acceptance

- [ ] Observer binding epoch increments on every bind and unbind, including same callback/same provider generation.
- [ ] Same-provider observation sequence remains monotonic across rebinds.
- [ ] A fresh read captures epoch, observer, provider generation, transport, and source generation at dispatch.
- [ ] Rebind/unbind/reconnect before dispatch, after send, or before decode invalidates the read fail closed.
- [ ] Callback identity reuse cannot conceal lifecycle change.
- [ ] Only the exact transaction completion suppresses ordinary MOR-1010 authority emission.
- [ ] Matching NAK, timeout, cancellation, reconnect, rebind, and callback failure publish no stale current truth.
- [ ] Red-first regressions use real CivRuntime dispatch behavior.

## Verification

Focused authority/radio matrix, full non-integration suite, Ruff/format, strict changed-source mypy, import-linter, exact scope/diff guards, exact-head CI, and independent review.

morozsm:
## Summary
- add an observer binding epoch that advances on every bind and unbind without resetting same-provider observation sequence
- bind each fresh PTT read to one observer/provider/source/transport snapshot and revalidate it before dispatch and publication
- attach the fresh read to MOR-1108 exact data-transaction ownership so its response cannot publish through the ordinary observation path

## Verification
- red-first lifecycle tests for identical callback unbind/rebind, pre-dispatch invalidation, callback failure, reconnect, and sequence continuity
- `uv run pytest tests/test_raw_civ_transaction.py tests/test_civ_transaction_ownership.py tests/test_radio.py -q` (259 passed)
- `uv run pytest -q -m "not integration" -n auto` (8562 passed, 73 skipped, 11 xfailed, 1 xpassed)
- `uv run ruff check src/rigplane/runtime/_civ_rx.py src/rigplane/runtime/radio.py tests/test_radio.py`
- `uv run ruff format --check src/rigplane/runtime/_civ_rx.py src/rigplane/runtime/radio.py tests/test_radio.py`
- import smoke for `CivRuntime` and `CoreRadio`
- `git diff --check`

Draft pending CI and independent review. Full MOR-1103 provider-attempt wrapper remains out of scope.

Closes #1990
Linear: MOR-1109

### DIFF
--- a/src/rigplane/runtime/_civ_rx.py
+++ b/src/rigplane/runtime/_civ_rx.py
@@ -641,7 +641,7 @@ class CivRuntime:
         return (
             self._active_ptt_read in (None, read)
             and self._ptt_observer_binding_epoch == read.binding_epoch
-            and self._ptt_observer == read.observer
+            and self._ptt_observer is read.observer
             and self._ptt_observer_provider_generation == read.provider_generation
             and self._ptt_observer_civ_generation == read.source_generation
             and self._host._civ_epoch == read.source_generation
@@ -1352,7 +1352,7 @@ class CivRuntime:
         )
         try:
             observer(observation)
-        except Exception:  # noqa: BLE001 - observer failure must not break CI-V RX
+        except (Exception, asyncio.CancelledError):  # noqa: BLE001
             logger.exception("Authoritative PTT observer raised")
             return False
         return True
--- a/src/rigplane/runtime/_civ_rx.py
+++ b/src/rigplane/runtime/_civ_rx.py
@@ -165,11 +165,21 @@ class _CivDataTransaction:
 
     response: asyncio.Future[CivFrame]
     nak: asyncio.Future[CivFrame]
+    owner: object | None = None
     frame: CivFrame | None = None
 
     @property
     def futures(self) -> tuple[asyncio.Future[CivFrame], ...]:
         return (self.response, self.nak)
+
+
+@dataclass(frozen=True, slots=True)
+class _AuthoritativePttRead:
+    observer: Callable[[ProviderPttObservation], None]
+    provider_generation: int
+    binding_epoch: int
+    source_generation: int
+    transport: object
 
 
 _CMD14_RECEIVER_LEVEL_FIELDS = {
@@ -550,7 +560,10 @@ class CivRuntime:
         self._ptt_observer: Callable[[ProviderPttObservation], None] | None = None
         self._ptt_observer_provider_generation: int | None = None
         self._ptt_observer_civ_generation: int | None = None
+        self._ptt_observer_binding_epoch = 0
         self._ptt_observation_seq = 0
+        self._ptt_read_lock = asyncio.Lock()
+        self._active_ptt_read: _AuthoritativePttRead | None = None
         self._active_rx_source_generation: int | None = None
         self._civ_data_transaction: _CivDataTransaction | None = None
 
@@ -576,13 +589,64 @@ class CivRuntime:
             raise ValueError("provider_generation must be non-negative")
         if provider_generation != self._ptt_observer_provider_generation:
             self._ptt_observation_seq = 0
+        self._ptt_observer_binding_epoch += 1
         self._ptt_observer = observer
         self._ptt_observer_provider_generation = provider_generation
         self._ptt_observer_civ_generation = self._host._civ_epoch
 
     def unbind_ptt_observer(self) -> None:
         """Detach the callback while preserving this provider's sequence."""
+        self._ptt_observer_binding_epoch += 1
         self._ptt_observer = None
+
+    async def request_authoritative_ptt_read(
+        self,
+        civ_frame: bytes,
+        *,
+        provider_generation: int,
+        observer: Callable[[ProviderPttObservation], None],
+    ) -> bool:
+        """Publish one exact PTT response only while its binding stays current."""
+        async with self._ptt_read_lock:
+            transport = self._host._civ_transport
+            read = _AuthoritativePttRead(
+                observer=observer,
+                provider_generation=provider_generation,
+                binding_epoch=self._ptt_observer_binding_epoch,
+                source_generation=self._host._civ_epoch,
+                transport=transport,
+            )
+            if transport is None or not self._ptt_read_is_current(read):
+                return False
+            self._active_ptt_read = read
+            try:
+                result = await self.execute_civ_transaction(
+                    civ_frame, expect="data", owner=read
+                )
+                return (
+                    result.status == "response"
+                    and result.frame is not None
+                    and self._ptt_read_is_current(read)
+                    and self._emit_authoritative_ptt(
+                        result.frame,
+                        source_generation=read.source_generation,
+                        expected=read,
+                    )
+                )
+            finally:
+                if self._active_ptt_read is read:
+                    self._active_ptt_read = None
+
+    def _ptt_read_is_current(self, read: _AuthoritativePttRead) -> bool:
+        return (
+            self._active_ptt_read in (None, read)
+            and self._ptt_observer_binding_epoch == read.binding_epoch
+            and self._ptt_observer == read.observer
+            and self._ptt_observer_provider_generation == read.provider_generation
+            and self._ptt_observer_civ_generation == read.source_generation
+            and self._host._civ_epoch == read.source_generation
+            and self._host._civ_transport is read.transport
+        )
 
     async def stop_pump(self) -> None:
         """Stop CI-V receive pump and fail pending request futures."""
@@ -667,6 +731,7 @@ class CivRuntime:
         *,
         expect: RawCivExpectation,
         timeout: float | None = None,
+        owner: object | None = None,
     ) -> RawCivTransactionResult:
         """Execute one explicitly-scoped raw CI-V transaction.
 
@@ -692,7 +757,7 @@ class CivRuntime:
 
         if expect == "none":
             self.start_pump()
-            await self._send_civ_frame_now(civ_frame)
+            await self._send_civ_frame_now(civ_frame, owner=owner)
             return RawCivTransactionResult(status="sent")
 
         await self._drain_ack_sinks_before_blocking()
@@ -719,10 +784,11 @@ class CivRuntime:
                 pending_waiters.append(pending_or_token)
             else:
                 data_transaction = self._register_civ_data_transaction(request_key)
+                data_transaction.owner = owner
                 pending_waiters.extend(data_transaction.futures)
 
             self.start_pump()
-            await self._send_civ_frame_now(civ_frame)
+            await self._send_civ_frame_now(civ_frame, owner=owner)
             remaining = deadline_monotonic - time.monotonic()
             if remaining <= 0:
                 raise asyncio.TimeoutError("CI-V response timed out")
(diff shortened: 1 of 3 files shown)


morozsm:
Agent Review: BLOCKED

Reviewed exact head `427ca7c4df822f9388c5c5c9c621005ac2f16af2` independently against MOR-1109 / #1990.

Blocking findings:

1. **Observer identity is checked with equality, not identity** (`src/rigplane/runtime/_civ_rx.py:_ptt_read_is_current`). Two distinct callable objects whose `__eq__` returns true are accepted as the same binding. Reproduction: bind callable A, invoke `_request_authoritative_ptt_read(..., observer=B)` where `A is not B` but `A == B`; the request sends, returns successfully, and publishes through A. This violates “captures observer” and “callback identity reuse cannot conceal lifecycle change.” Require `self._ptt_observer is read.observer` and a regression test using equality-colliding callable instances.

2. **Observer-raised cancellation escapes instead of failing closed** (`_emit_authoritative_ptt` catches `Exception`, while `asyncio.CancelledError` is a `BaseException`). Reproduction: a bound synchronous observer raising `asyncio.CancelledError` causes the fresh-read task itself to raise `CancelledError` after the exact response rather than returning false / `CommandError`. This misses the explicit cancellation/callback-failure acceptance matrix. Add a focused regression proving no authoritative success/stale truth and intentional containment of observer-raised cancellation.

Evidence at this head:
- scope: 3 files, +210/-17 = **193 net LOC** (within cap)
- `git diff --check`: PASS
- Ruff + format check: PASS
- focused authority/radio tests: **259 passed**
- exact-head GitHub quick + grep checks: PASS
- changed-source mypy introduced no changed-line diagnostics; repository baseline diagnostics remain outside this patch

Do not mark ready or merge until both regressions are fixed and the independent exact-head review is rerun.

morozsm:
Official BLOCKED remediation pushed at exact head `3458df907e8ae2a0375c4a585e9815a6eacc5396` after a clean rebase onto current `origin/main@15f958ad`.

Addressed only the two review blockers:
- `_ptt_read_is_current()` now requires strict callback identity with `is`; an adversarial pair of distinct equality-colliding `MagicMock` callables is rejected fail-closed before dispatch/publication.
- synchronous observer-raised `asyncio.CancelledError` is contained at the callback boundary and returns fail-closed through the exact adapter as `CommandError`; transaction/task cancellation at await points remains unchanged and is covered by the existing raw transaction cancellation regressions.

Evidence:
- red-before: both adversarial cases failed on prior behavior
- focused exact transaction/authority/radio: `259 passed`
- full non-integration: `8562 passed, 73 skipped, 11 xfailed, 1 xpassed`
- Ruff check/format, import-linter 5/0, import smoke, and diff check: PASS
- targeted mypy reports the same six repository-baseline `_civ_rx.py` diagnostics outside changed symbols; no changed-line diagnostic
- scope: exact 3 owned files, `+218/-19`, net `+199` LOC

PR remains Draft. No self-review, ready transition, merge, or Linear status change; fresh independent review is required for this exact head.

morozsm:
Agent Review: PASS

Independent fresh review of exact head `c54d87566112ad60bf7690b631c040a419b06ce0` (local worktree head, `origin/codex/1990-observer-binding-epoch`, and PR `headRefOid` all verified identical). Merge base `15f958ad`. I did not author this change and did not self-review.

## Invariants verified

| Acceptance criterion | Verdict | Evidence |
|---|---|---|
| Binding epoch increments on **every** bind/unbind, incl. identical callback + identical generation; sequence never resets | PASS | `_civ_rx.py:592`/`:599` increment unconditionally before assignment; probe: 3× bind/unbind of one callable at `provider_generation=4` → epochs `[1,2,3,4,5,6]` strictly increasing, `ptt_observation_seq` `[1,2,3]` continuous. Seq reset stays keyed to provider-generation change only. |
| Active read captures epoch + observer object identity + provider/source generation; stale completion rejected | PASS | `_AuthoritativePttRead` (`_civ_rx.py:176`) captures observer, provider_generation, binding_epoch, source_generation, transport. Revalidated at **three** points: pre-dispatch (`_send_civ_frame_now:922`), pre-emit (`_emit_authoritative_ptt:1338`), post-completion (`:625`). Probes: unbind mid-flight, rebind-to-different-observer mid-flight, and `advance_generation` mid-flight each yield zero publications and a fail-closed raise. |
| Freshness compares callable identity with `is`, never equality | PASS | `self._ptt_observer is read.observer` (`:645`). **Mutation-proven**: globally swapping that one `is` for `==` makes `test_fresh_ptt_read_rejects_rebind_and_equal_distinct_observer` FAIL, and the equality-colliding observer both gets accepted and receives a publication. The test is load-bearing, not a mirror of the implementation. |
| Sync observer raising `asyncio.CancelledError` contained at the callback boundary, fail-closed, fresh-read task not cancelled | PASS | `except (Exception, asyncio.CancelledError)` → `return False` → `CommandError` (`radio.py:2979`). `observer(...)` is a fully synchronous call site with no `await`, so a `CancelledError` caught there is necessarily observer-raised, never delivered external cancellation. **Mutation-proven**: reverting to `except Exception` makes `test_fresh_ptt_read_fails_closed_on_callback_cancel_and_reconnect` FAIL with `asyncio.exceptions.CancelledError: observer` escaping the read task. `issubclass(CancelledError, Exception) is False` confirms the old clause could not contain it. |
| External cancellation still governed by the transaction contract, not broadly swallowed | PASS | Probe: `task.cancel()` while awaiting `execute_civ_transaction` → `CancelledError` propagates out of `_request_authoritative_ptt_read`; `_active_ptt_read` cleared, `_ptt_read_lock` released, and a subsequent read completes and publishes normally. No except-clause wraps the awaits. |
| Reconnect invalidation remains the authority; no parallel freshness source | PASS | `_civ_epoch` is still the sole reconnect signal; `_ptt_read_is_current` checks it twice (`_ptt_observer_civ_generation` and live `_host._civ_epoch`), and `_emit_authoritative_ptt` gained `source_generation != self._host._civ_epoch`, which **tightens** the pre-existing check rather than competing with it. `binding_epoch` covers a strictly different axis (bind/unbind), never reconnect. |
| New adversarial test: plain callable recorder, explicit unbind, pump stopped, no leak | PASS | `MagicMock` removed in `baab0290` in favour of `observations.append` and a real `EqualObserver` class (honours the CLAUDE.md "MagicMock hides signature bugs" rule); the test unbinds explicitly and ends with `await radio._civ_runtime.stop_pump()`. No leak observed — see below. |
| Exact scope: 3 files, net ≤200 LOC | PASS | `git diff --numstat origin/main...HEAD` → `src/rigplane/runtime/_civ_rx.py`, `src/rigplane/runtime/radio.py`, `tests/test_radio.py`; **+217 / -18 = net +199**. No unintended files. |

Also checked and clean: the authoritative response cannot double-publish (`_route_civ_frame` suppresses the ordinary emit via `claimed.owner` `isinstance` — probe shows exactly one publication per successful read); `_publish_civ_event` only enqueues, so moving the emit after publish/resolve introduces no re-entrancy; `resolve()` uses `set_result`, whose callbacks are `call_soon`-scheduled, so the pump always finishes suppressing before the awaiting coroutine resumes; the `transport is None` guard prevents tripping `execute_civ_transaction`'s assert; `_close_civ_data_transaction`'s new `future.exception()` retrieval is correctly guarded by `not future.cancelled()` under an `elif done`.

## What was run (all at `c54d8756`)

- `uv run pytest tests/ -q --tb=short -m "not integration" -n auto` → **8623 passed, 12 skipped, 11 xfailed, 1 xpassed** (94s)
- `uv run pytest tests/test_radio.py::TestPtt tests/test_raw_civ_transaction.py -q` → 30 passed
- Repetition/flake: 5× `tests/test_radio.py tests/test_raw_civ_transaction.py tests/test_civ_rx_coverage.py -n 4` → 605 passed each, zero variance
- Leak probe (autouse async fixture enumerating live tasks at teardown): the two new tests that omit `stop_pump()` leave one `CivRuntime._civ_rx_loop` pending — **identical to three pre-existing tests in the same class** (`test_set_ptt_on`, `test_set_ptt_off`, `test_set_ptt_does_not_update_state_cache_optimistically`). No `__del__` "Radio collected with active connection/tasks" warning from this module (0 occurrences with and without the new tests), no xdist worker instability, in-memory fake transport only. Not a regression.
- Mutation testing (production methods patched via an out-of-tree `-p` plugin, repo files untouched): mutant A `is`→`==` and mutant B `except Exception` each turn the corresponding new test red; both are green unmutated.
- `uv run ruff check src/ tests/` → All checks passed · `ruff format --check` on the 3 files → already formatted
- `uv run lint-imports` → Contracts: 5 kept, 0 broken
- `uv run mypy src/rigplane/runtime/_civ_rx.py src/rigplane/runtime/radio.py` → 6 errors, all at lines `935, 1686, 1919, 1923, 2268, 2269`, every one **outside** the changed hunks and byte-identical on `origin/main`. No changed-line diagnostic.
- `git diff --check` → clean · `git status` → clean at exact head
- Exact-head CI: `quick` **success**, `grep-gate` **success**, `Update Agent Review Gate status` **success**. `Agent Review Gate` was failing only with "Missing Agent Review PASS comment for the current PR head" — resolved by this comment.

## Non-blocking observations (follow-up, not gating)

1. `_ptt_read_is_current` line 643 uses `self._active_ptt_read in (None, read)`, and `_AuthoritativePttRead` is `@dataclass(frozen=True)` with default `eq=True`, so `in` can fall back to `__eq__`. I confirmed two distinct records with equality-colliding observers satisfy `r1 in (None, r2)`. It is **unreachable today** — every mutation of `_active_ptt_read` happens inside `_ptt_read_lock`, so the field is only ever `None` or the exact object under test — but given this PR's own thesis, `(self._active_ptt_read is None or self._active_ptt_read is read)` would close the latent trap. Related: the generated `__hash__` would raise on an unhashable observer if these records ever entered a set/dict.
2. A coroutine-function observer returns `SUCCESS` while its body never runs (`RuntimeWarning: coroutine ... was never awaited` at `_civ_rx.py:1354`). The declared contract is sync-only (`Callable[[ProviderPttObservation], None]`) and the ordinary path behaves identically on `main`, so this is pre-existing — but the new boolean return now converts it into a false *authoritative-success* signal. Worth an explicit guard when the MOR-1103 wrapper lands.
3. `_request_authoritative_ptt_read` has no production caller yet, consistent with the PR body deferring the MOR-1103 provider-attempt wrapper.
4. `c54d8756` ("keep PR scope at review limit") drops `@pytest.mark.asyncio` from one new test purely to land at 199 net LOC. Harmless under `asyncio_mode = "auto"` — I verified the test executes and stays load-bearing under mutation — but it is now stylistically inconsistent with its two siblings and would read better restored once the LOC cap is not binding.
5. Unrelated to this PR: running the Python suite rewrites `frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json`, adding a `txTarget` leaf. It traces to the already-merged TxTarget work in the base, not to these 3 files. I restored the file; the base fixture appears stale on `main`.

Both previously-BLOCKED findings are genuinely fixed, mutation-proven, and covered by tests that fail when the fix is removed. Scope, static gates, and exact-head CI are clean. Marking ready for review; **not merging** — merge remains gated on required checks going green for this head.

PR #1997, opened 2026-07-27
MOR-1120: bind managed provider generation fail closed

### LINKED ISSUE #1996: MOR-1120: bind managed provider generation fail-closed before readiness
Linear: MOR-1120
Parent: MOR-1104 / #1953

## Outcome

Add the private consumer-defined managed provider lifecycle boundary. `ManagedRadioRuntime` remains the sole provider-generation owner and admits readiness, ON, and authoritative observations only for the exact successfully bound generation.

## Acceptance

- [ ] Exact keyword-only CoreRadio lifecycle methods satisfy the typed port without MagicMock.
- [ ] Effectless first replacement advances host/supervisor together, not-ready, and binds before optional readiness.
- [ ] Partial bind failure leaves the same host/supervisor generation unbound and not-ready; callbacks and ON fail closed.
- [ ] Unbind failure still advances invalidation, clears bound state, and rejects stale callbacks/readiness/ON.
- [ ] Old-generation callbacks cannot mutate current authority.
- [ ] `set_provider_ready(true)` and `request_on` reject unbound generations.
- [ ] Focused/full non-integration tests, Ruff/format, strict changed mypy, import-linter, diff guard, exact-head CI and independent review pass.

## Scope

Exactly:

Maximum 200 net LOC. No fresh-read await/serialization or terminal shutdown (MOR-1121), CI-V/token changes, executor, SDK, Web/frontend, public API or raw provider exposure.

morozsm:
## Summary

- add a private, consumer-defined provider lifecycle port to `ManagedRadioRuntime`
- bind an exact provider generation before readiness or ON can be admitted
- invalidate host authority even when provider unbind fails
- reject partial-bind and old-generation callbacks fail closed

## Why

MOR-1104's full lifecycle boundary exceeded the repository's 200 net LOC guardrail. This PR is the first bounded child slice: generation binding and fail-closed readiness. Fresh-read serialization and terminal shutdown remain in MOR-1121.

## Verification

- `uv run pytest tests/test_managed_radio_runtime.py tests/test_tx_safety.py tests/test_radio.py -q -n auto --tb=short` — 311 passed
- `uv run pytest tests/ --ignore=tests/integration -n auto --tb=short --timeout=300 --timeout-method=thread -q` — 8433 passed, 12 skipped, 11 xfailed, 1 xpassed
- `uv run ruff check src/ tests/`
- `uv run ruff format --check src/ tests/`
- strict mypy for both changed files
- `git diff --check`

## Scope

- 2 files
- 191 net LOC
- no fresh-read await/serialization, terminal shutdown, CI-V/token, executor, SDK, Web/frontend, or public API changes

Linear: MOR-1120

Closes #1996

### DIFF
--- a/src/rigplane/runtime/managed_radio_runtime.py
+++ b/src/rigplane/runtime/managed_radio_runtime.py
@@ -2,10 +2,12 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Awaitable, Callable
+from typing import Literal, Protocol
 
 from rigplane.core.tx_safety import (
     Clock,
     IdFactory,
+    ProviderPttObservation,
     TxOutcome,
     TxOwner,
     TxReleaseReason,
@@ -16,6 +18,20 @@ from rigplane.core.tx_safety import (
 
 TxService = Callable[[TxSafetySupervisor, TxTransition], Awaitable[None]]
 ProviderRelease = Callable[[], Awaitable[None]]
+_PttObserver = Callable[[ProviderPttObservation], None]
+_ProviderLifecycleState = Literal["unbound", "bound", "invalidating"]
+
+
+class ProviderTxLifecycle(Protocol):
+    def _bind_authoritative_ptt_observer(
+        self, *, provider_generation: int, observer: _PttObserver
+    ) -> None: ...
+
+    def _unbind_authoritative_ptt_observer(self) -> None: ...
+
+    async def _request_authoritative_ptt_read(
+        self, *, provider_generation: int, observer: _PttObserver
+    ) -> None: ...
 
 
 class ManagedRadioRuntime:
@@ -24,6 +40,7 @@ class ManagedRadioRuntime:
         target_id: str,
         *,
         service: TxService,
+        provider_lifecycle: ProviderTxLifecycle | None = None,
         clock: Clock | None = None,
         id_factory: IdFactory | None = None,
         shutdown_timeout_seconds: float = 3.0,
@@ -33,7 +50,10 @@ class ManagedRadioRuntime:
         self.target_id = target_id
         self._tx_safety = TxSafetySupervisor(clock=clock, id_factory=id_factory)
         self._service = service
+        self._provider_lifecycle = provider_lifecycle
         self._provider_generation = 0
+        self._bound_generation: int | None = None
+        self._provider_state: _ProviderLifecycleState = "unbound"
         self._shutdown_timeout = shutdown_timeout_seconds
 
     @property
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

**Head reviewed:** `6d5a10744cb2fd3c77bdd49583c9adcdda49ca45` (base `96b94e4254cff8e6c13465b3b2bf9b514386d4d5` == current `origin/main`; `mergeable: MERGEABLE`). Independent review — I did not author this change.

## Invariants verified

Verified by reading `src/rigplane/runtime/managed_radio_runtime.py` against `core/tx_safety.py` and `runtime/_civ_rx.py`, then by 64 adversarial probes driven through the production `_observe_ptt` callback (probe script kept outside the repo tree; every assertion below is an executed result, not a reading).

1. **Single generation authority, no duplicated freshness.** `ManagedRadioRuntime` is the only writer of `_provider_generation` / `_bound_generation` / `_provider_state`. It re-implements no CI-V token, binding-epoch, `civ_generation`, or observation-sequence logic — `grep -nE "epoch|observation_seq|civ_generation|token|binding_epoch"` over the changed file returns nothing; that freshness stays where MOR-1108/1109 merged it (`_civ_rx._ptt_observer_binding_epoch`, `_ptt_observer_civ_generation`, `_ptt_observation_seq`) and in `TxSafetySupervisor.observe_ptt`. `_observe_ptt` (`managed_radio_runtime.py:70`) adds exactly one new gate: the `observation.provider_generation == _bound_generation == _provider_generation` triple.

2. **Typed port accepts the exact production signature; no permissive mock.** `tests/test_managed_radio_runtime.py:249` binds a real `CoreRadio` and asserts `radio._civ_runtime._ptt_observer_provider_generation == 1`, so the real keyword-only `radio.py:2950` bind is exercised. `mypy --strict` on the test file reports no `arg-type` error at the `lifecycle=radio` call site (only 12 `untyped-decorator` artifacts from `@pytest.mark.asyncio` under bare `--strict`), confirming `CoreRadio` structurally satisfies `ProviderTxLifecycle`. The fake at `:28` is hand-written with keyword-only params. Probe: a provider with a renamed keyword (`generation=`) raises `TypeError` and leaves `_provider_state == "unbound"`, `_bound_generation is None`, `request_on -> NOT_READY`.

3. **Deterministic transitions; every authority replacement invalidates stale reads.** Same-provider rebind (A→A) bumps 1→2, unbinds first, and the gen-1 callback is rejected. The gen-1 and gen-2 bound methods are **equal but not identical** (`obs1 == obs2 and obs1 is not obs2`) — identity is provably not the discriminator, the generation is. A→B→A: the gen-1 callback still cannot publish at gen 3. `invalidate_provider(99)` and `invalidate_provider(0)` return `STALE` and mutate nothing.

4. **First bind has no false-ready side effect.** `replace_provider(ready=False)` leaves `provider_ready False`; `replace_provider(ready=True)` fabricates no `ProviderPttObservation` (`radio_tx` stays `UNKNOWN`), emits zero effects, and `request_on` returns `RADIO_NOT_OFF` — never `ACCEPTED`.

5. **Unbound/replaced generations cannot regain readiness or publish.** After bind failure, unbind failure, or partial lifecycle: leaked old callbacks are inert, `set_provider_ready(ready=True) -> NOT_READY`, `request_on -> NOT_READY`. `set_provider_ready(ready=False)` is deliberately ungated — the fail-safe direction.

6. **Bind/unbind exceptions fail closed.** Unbind raising `RuntimeError`, `asyncio.CancelledError`, or `KeyboardInterrupt` all propagate *and* still advance invalidation (host gen 2, supervisor gen 2, `provider_ready False`, `_bound_generation None`). Bind failure runs the compensating unbind exactly once; when that unbind also raises, the **bind** error is the one surfaced and the runtime stays unbound. Reentrancy is covered by the third `"invalidating"` state: an observation fired synchronously from inside `unbind` is rejected, as is one fired from inside `bind`. Generation arithmetic is unbounded-int safe (verified past `2**63`) and monotonic.

7. **No safety evidence compressed away.** Zero tests deleted — the 2 test-file deletions are import lines. The 2 src deletions removed `await self._service_effects(transition)` from the shared replace path; that is **provably lossless**, because `TxSafetySupervisor.replace_provider(gen, ready=False)` can never emit effects (`_service_release` early-returns on `not self._ready`, `tx_safety.py:517-531`). Probe confirms empirically: replacing/invalidating **while keyed** yields `effects == ()` and services nothing, while `replace_provider(ready=True)` while keyed still services the `WRITE_OFF` dekey (now via the `set_provider_ready(gen, ready=True)` + `_service_effects` path at `:114-116`). The pre-existing `test_ready_provider_services_pending_off_before_return` still guards the resume-ready ordering.

## What I ran

- `uv run pytest tests/test_managed_radio_runtime.py tests/test_tx_safety.py tests/test_radio.py -q --tb=short` — **311 passed**
- `uv run pytest tests/ --ignore=tests/integration -n auto --tb=short --timeout=300 --timeout-method=thread -q` — **8433 passed, 12 skipped, 11 xfailed, 1 xpassed**
- 64 adversarial probes (same-provider rebind, equal-but-distinct callbacks, wrong keyword signatures, sync/async mismatch, partial lifecycle, exception/cancellation, generation overflow, reentrancy, A→B→A, shutdown boundary, no-lifecycle default) — **64 pass / 0 fail**
- `uv run ruff check src/ tests/` — clean · `uv run ruff format --check src/ tests/` — 627 files formatted
- `uv run mypy src/rigplane/runtime/managed_radio_runtime.py --strict` — **no issues**; `uv run mypy src/` — 15 pre-existing errors in 4 unrelated files, none in the changed file
- `uv run lint-imports` — 5 contracts kept, 0 broken · `git diff --check` — clean
- Exact-head CI (`6d5a1074`): `Tests (quick)` **success**, `Rebrand grep gate` **success**, `Agent Review Gate` red only for the missing PASS comment this review supplies.

Transparency note: one unrelated test (`TestDspLevelParity::test_get_direct_dsp_level[get_anti_vox_gain-23-108]`, a CI-V response timeout) failed once while I was running the focused set concurrently with the full suite under CPU contention. It passed 3× in isolation, in the clean focused run, and in the full suite. Load-sensitive flake, unrelated to this change.

## Scope

Exactly the 2 files named in #1996 — `src/rigplane/runtime/managed_radio_runtime.py`, `tests/test_managed_radio_runtime.py` — at **+195 / -4 = 191 net LOC** (≤ 3 files, ≤ 200 LOC). No CI-V/token change, no executor (MOR-1043), no SDK, no Web/frontend, no public API, no raw provider exposure. All seven acceptance boxes in #1996 are met.

## Explicit MOR-1121 remainder (not defects of this slice)

Stated so the boundary is not silently assumed closed:

- **`shutdown()` does not unbind or bump the generation.** Probed: after `shutdown(release_provider=...)` the provider is still bound (`unbinds == 0`, gen unchanged) and a further `emit` still publishes into the supervisor. Terminal shutdown invalidation is MOR-1121; this PR correctly does not claim that boundary early (the binding paths reference neither `emergency_release` nor `release_provider`).
- **No fresh-read serialization.** An observation fired synchronously from inside `bind` is dropped (conservative), so the first post-bind `request_on` is `RADIO_NOT_OFF` until the provider next reports. The awaited fresh read that closes this is MOR-1121; `_request_authoritative_ptt_read` is declared on the port here but intentionally never called.
- **Bind failure while keyed leaves a pending release un-serviceable** (`ready=False`, so no `WRITE_OFF` can be emitted) until a provider is ready again. Correct fail-closed behaviour for this slice; physical dekey on a dead provider is MOR-1121.
- **Transport-epoch loss does not self-invalidate the host binding.** After a `_civ_epoch` bump, `_emit_authoritative_ptt` silently stops publishing while the host stays `bound`/`ready`. Closing that requires the owner to call `invalidate_provider(gen)` — the seam this PR adds — which is MOR-1043/MOR-1121 integration duty.

Two non-blocking notes for those follow-ups, neither reachable from a type-checked production caller today:

- An `async def` binder would return a discarded coroutine and the host would mark itself `bound`/`ready` with nothing bound. Unreachable via mypy (the port declares a sync `-> None`) and `CoreRadio` is sync, but a `runtime_checkable`/awaitable-result guard would be cheap insurance when the executor wires real providers.
- `provider_lifecycle=None` (the constructor default) makes `request_on` and `set_provider_ready(True)` permanently `NOT_READY`. Fail-closed and currently harmless — `grep` confirms no production construction site exists yet — but no test documents it; consider making the parameter required when MOR-1043 lands.

Marking ready for review. Not merging — merge is the author's call once required checks are green.

Issue #1999, opened 2026-07-27
Atomically project and enforce browser TX-audio snapshots


morozsm:
Linear: MOR-1127
Parent: #1958
Blocked by: #1998

## Scope

Consume the dormant immutable snapshot from #1998 atomically in both public capability serialization and the audio WS lifecycle.

Exact files:
- `src/rigplane/web/handlers/audio.py`
- `src/rigplane/web/server.py`
- one focused lifecycle/endpoint test file

Maximum 3 files and 200 net LOC. Preserve all unrelated documentation/comments.

## Acceptance

- [ ] `/api/v1/capabilities` first introduces `audioTx`, `audioTxRoute`, and `audioTxRequiredModInputSource` from `BrowserTxAudioFacts`; only those fields serialize.
- [ ] In the same slice, `audio_start direction=tx` evaluates facts once and uses the exact snapshot for transcoder-before-arm, acquire/start, push, codec routing, stop, re-arm cleanup, and teardown.
- [ ] No public `audioTx=true` path exists without matching WS enforcement.
- [ ] Transcoder failure reports denial, starts/acquires nothing, leaves TX inactive, and keeps reader/RX alive.
- [ ] Denial, start failure, cancellation, and BaseException during re-arm de-key prior lease-less TX exactly once; session cleanup remains lease-owned.
- [ ] Neutral lease-less stop without start, double stop, and stop after denial are safe/idempotent: no legacy AttributeError, reader death, or RX teardown.
- [ ] Post-arm mutable radio/session/callable swaps cannot change the captured path.
- [ ] Repeated stop/final teardown release a session lease exactly once.
- [ ] Real IC-7610 and FTX-1 endpoint tuples match the snapshot helper.
- [ ] Focused/full/static/CI and fresh independent exact-head review pass.

PR #2000, opened 2026-07-27
feat: compute dormant browser TX-audio facts

### LINKED ISSUE #1998: Compute dormant fail-closed BrowserTxAudioFacts snapshots
Linear: MOR-1126
Parent: #1958
Consumer follow-up: #1999

## Scope

Define and test the pure immutable `BrowserTxAudioFacts` computation only.

Exact files:
- `src/rigplane/web/handlers/audio.py`
- `tests/test_web_capability_guards.py`

Maximum 3 files and 200 net LOC. Preserve every unrelated `origin/main` comment/docstring.

## Acceptance

- [ ] Add immutable `BrowserTxAudioFacts` containing the future public tuple plus validated codec, selected lifecycle path, and exact concrete callables.
- [ ] Facts computation is dormant: no server/endpoint/WS consumer, no wire change, and no production behavior change.
- [ ] Accept only class-defined bound async lifecycle methods owned by the evaluated object and compatible with the exact future invocation shape.
- [ ] Protocol/abstract/partial/sync/wrong-arity/foreign-bound/unbound lifecycle evidence fails closed.
- [ ] Missing/unsupported codec and malformed, throwing, unavailable, or contradictory route/policy evidence fail closed.
- [ ] Each codec, route, policy, session, and lifecycle attribute is evaluated once per snapshot; no `start_tx` double read.
- [ ] Real `IcomRadio` IC-7610 resolves `true/lan/5/PCM_1CH_16BIT/session`.
- [ ] Real `YaesuCatRadio` FTX-1 resolves `true/usb/null/PCM_1CH_16BIT/session`.
- [ ] Focused/full/static/CI evidence passes.

## Boundary

#1999 owns the first `/api/v1/capabilities` projection and all audio WS snapshot consumption in one atomic enforced slice. No public `audioTx` field is introduced here.

morozsm:
Closes #1998
Linear: MOR-1126
Parent: #1958 / MOR-1050
Consumer follow-up: #1999 / MOR-1127

## Summary

- define one immutable, fail-closed `BrowserTxAudioFacts` snapshot containing the future public tuple plus the validated codec, selected path, and concrete async lifecycle callables
- accept only class-defined bound async methods owned by the evaluated object and compatible with the exact future call shape
- reject Protocol/abstract/partial/synchronous/wrong-arity/foreign-bound/unbound lifecycle evidence, unsupported codecs, and malformed or contradictory routes
- evaluate each lifecycle attribute exactly once per snapshot
- pin real IC-7610 and FTX-1 provider snapshots

## Dormant boundary

This slice has no production consumer and makes no behavior or wire change. `browser_tx_audio_facts` is defined only in `src/rigplane/web/handlers/audio.py`; `/api/v1/capabilities` remains unchanged and audio WS still uses its existing behavior.

MOR-1127/#1999 atomically owns both the first public `audioTx` projection and the WS snapshot enforcement. Public capability cannot appear before enforcement.

## Scope

Exactly 2 files, `+198/-1`, net `+197` LOC. All unrelated protocol comments/docstrings remain intact.

## Verification

- focused facts test file: 29 passed
- focused web/audio matrix: PASS
- full non-integration: 8433 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed
- Ruff check/format, strict web mypy, import-linter 5/5, and diff check: PASS
- exact-head CI pending

PR remains Draft pending fresh independent Opus 5 max review. No readiness or merge action.

### DIFF
--- a/src/rigplane/web/handlers/audio.py
+++ b/src/rigplane/web/handlers/audio.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 import logging
 import time
 from collections.abc import AsyncIterator, Awaitable
@@ -11,6 +12,7 @@ from typing import TYPE_CHECKING, Any, Callable, Protocol, cast
 
 from rigplane.audio._codecs import decode_ulaw_to_pcm16
 from rigplane.audio._transcoder import PcmOpusTranscoder, create_pcm_opus_transcoder
+from ...audio import route as audio_route
 from ...audio.bus import STAGE_RX_POST_DSP
 from ...audio.session import RxSubscription
 from ...audio.usb_driver import AudioAlreadyStartedError
@@ -39,6 +41,7 @@ if TYPE_CHECKING:
     from ...radio_protocol import Radio
 
 from ...capabilities import CAP_AUDIO, CAP_LAN_DUAL_RX_AUDIO_ROUTING
+from ...capabilities import CAP_MOD_INPUT_ROUTING, CAP_TX
 
 __all__ = ["AudioBroadcaster", "AudioHandler", "RxPcmTapSource"]
 
@@ -107,6 +110,114 @@ class RxPcmTapSource:
 
     codec: AudioCodec
     sample_rate: int
+
+
+@dataclass(frozen=True, slots=True)
+class BrowserTxAudioFacts:
+    available: bool
+    route: str | None
+    required_mod_input_source: int | None
+    codec: AudioCodec | None = None
+    path: str | None = None
+    lifecycle: tuple[Any, ...] | None = None
+
+
+def _tx_codec_for_radio(radio: object) -> AudioCodec | None:
+    codec = getattr(radio, "audio_tx_codec", None)
+    if codec is None:
+        codec = getattr(getattr(radio, "audio_stream_contract", None), "tx_codec", None)
+    if codec is None:
+        codec = getattr(radio, "audio_codec", None)
+    supported = (AudioCodec.PCM_1CH_16BIT, AudioCodec.OPUS_1CH)
+    return codec if isinstance(codec, AudioCodec) and codec in supported else None
+
+
+def _concrete_async_method(owner: object, name: str, method: object) -> Any | None:
+    if not callable(method) or not asyncio.iscoroutinefunction(method):
+        return None
+    if name in getattr(owner, "__dict__", {}):
+        return None
+    base = next((cls for cls in type(owner).__mro__ if name in cls.__dict__), None)
+    if (
+        base is None
+        or getattr(base, "_is_protocol", False)
+        or getattr(base.__dict__[name], "__isabstractmethod__", False)
+    ):
+        return None
+    if getattr(method, "__self__", None) is not owner:
+        return None
+    args = (b"",) if "push" in name else ("web",) if name == "acquire_tx" else ()
+    kwargs = {"sample_rate": 48_000} if name == "start_audio_tx_pcm" else {}
+    try:
+        inspect.signature(method).bind(*args, **kwargs)
+    except (TypeError, ValueError):
+        return None
+    return method
+
+
+def _lifecycle(
+    owner: object, names: tuple[str, ...]
+) -> tuple[tuple[Any, ...] | None, bool]:
+    raw = tuple(getattr(owner, name, None) for name in names)
+    methods = tuple(
+        _concrete_async_method(owner, name, method)
+        for name, method in zip(names, raw, strict=True)
+    )
+    return (methods if all(methods) else None), any(item is not None for item in raw)
+
+
+def _tx_path(radio: object, codec: AudioCodec) -> tuple[str, tuple[Any, ...]] | None:
+    neutral, has_neutral = _lifecycle(radio, ("start_tx", "push_tx", "stop_tx"))
+    session = getattr(radio, "audio_session", None)
+    if session is not None:
+        acquire_raw = getattr(session, "acquire_tx", None)
+        acquire = _concrete_async_method(session, "acquire_tx", acquire_raw)
+        return ("session", (acquire, *neutral[1:])) if acquire and neutral else None
+    if has_neutral:
+        return ("neutral", neutral) if neutral is not None else None
+    suffix = "pcm" if codec == AudioCodec.PCM_1CH_16BIT else "opus"
+    lifecycle, _ = _lifecycle(
+        radio, tuple(f"{op}_audio_tx_{suffix}" for op in ("start", "push", "stop"))
+    )
+    return (suffix, lifecycle) if lifecycle is not None else None
+
+
+def browser_tx_audio_facts(radio: "Radio | None") -> BrowserTxAudioFacts:
+    """Resolve the exact fail-closed structural snapshot for HTTP and audio WS."""
+    unavailable = BrowserTxAudioFacts(False, None, None)
+    caps = runtime_capabilities(radio)
+    if radio is None or not {CAP_AUDIO, CAP_TX}.issubset(caps):
+        return unavailable
+    try:
+        codec = _tx_codec_for_radio(radio)
+        if codec is None or (path := _tx_path(radio, codec)) is None:
+            return unavailable
+        route = audio_route.resolve_audio_route(radio)
+        source, policy = route.tx_audio_source, route.data_mode_policy
+        if (source, policy) not in {
+            (audio_route.TxAudioSource.LAN, audio_route.DataModePolicy.DATA2_LAN),
+            (audio_route.TxAudioSource.LAN, audio_route.DataModePolicy.LEGACY),
+            (audio_route.TxAudioSource.USB, audio_route.DataModePolicy.DATA1_USB),
+            (audio_route.TxAudioSource.ACC, audio_route.DataModePolicy.LEGACY),
+        }:
+            return unavailable
+        required = (
+            audio_route.rigctld_wsjtx_policy(route)[1]
+            if CAP_MOD_INPUT_ROUTING in caps
+            else None
+        )
+        if required is not None and (
+            source is not audio_route.TxAudioSource.LAN
+            or not isinstance(required, int)
+            or isinstance(required, bool)
+            or not 0 <= required <= 9_007_199_254_740_991
+        ):
+            return unavailable
+        return BrowserTxAudioFacts(
+            True, source.value, required, codec, path[0], path[1]
+        )
+    except Exception:
+        return unavailable
 
 
 def _is_benign_tx_restart(exc: RuntimeError) -> bool:
--- a/tests/test_web_capability_guards.py
+++ b/tests/test_web_capability_guards.py
@@ -6,10 +6,16 @@ TDD: these tests were written FIRST, then the backend was modified to pass them.
 from __future__ import annotations
 
 import json
-from dataclasses import replace
+from dataclasses import astuple, replace
+from types import SimpleNamespace
 
 import pytest
 
+from rigplane.backends.yaesu_cat.radio import YaesuCatRadio
+from rigplane.radio_protocol import AudioCapable
+from rigplane.runtime.radio import IcomRadio
+from rigplane.types import AudioCodec
+from rigplane.web.handlers.audio import browser_tx_audio_facts
 from rigplane.web.server import WebServer
 
 
@@ -290,6 +296,86 @@ class TestCapabilitiesEndpoint:
         await srv._serve_capabilities(writer)  # noqa: SLF001
         data = _parse_json_body(writer)
         assert data["antennas"] == 1
+
+
+def _yaesu(cls=YaesuCatRadio):
+    return cls("/dev/null", audio_driver=SimpleNamespace())
+
+
+def test_browser_tx_audio_facts_fail_closed():
+    no_tx = type(
+        "NoTx",
+        (YaesuCatRadio,),
+        {"capabilities": property(lambda _: {"audio"})},
+    )
+    bad_codec = type(
+        "BadCodec",
+        (YaesuCatRadio,),
+        {
+            "audio_tx_codec": property(lambda _: None),
+            "audio_codec": property(lambda _: None),
+        },
+    )
+    variants = (
+        _yaesu(no_tx),
+        _yaesu(bad_codec),
+        _yaesu(type("Partial", (YaesuCatRadio,), {"push_tx": None})),
+        _yaesu(type("SyncPush", (YaesuCatRadio,), {"push_tx": lambda *_: None})),
+        _yaesu(
+            type("WrongArity", (YaesuCatRadio,), {"push_tx": YaesuCatRadio.stop_tx})
+        ),
+    )
+    assert all(not browser_tx_audio_facts(radio).available for radio in variants)
+
+
+def test_browser_tx_audio_facts_reject_stubs_and_instance_callables():
+    class StubRadio(AudioCapable):
+        capabilities = {"audio", "tx"}
+        backend_id = "yaesu_cat"
+        has_usb_audio = True
+        audio_tx_codec = AudioCodec.OPUS_1CH
+
+    assert browser_tx_audio_facts(StubRadio()).available is False
+    donor, victim = _yaesu(), _yaesu()
+    victim.start_tx = donor.start_tx
+    assert browser_tx_audio_facts(victim).available is False
+
+    victim = _yaesu()
+    victim.push_tx = YaesuCatRadio.stop_tx
+    assert browser_tx_audio_facts(victim).available is False
+
+
+class _CountingRadio(YaesuCatRadio):
+    reads: dict[str, int]
+
+    def __getattribute__(self, name: str):
+        if name in {"start_tx", "push_tx", "stop_tx"}:
+            reads = object.__getattribute__(self, "reads")
+            reads[name] += 1
+        return super().__getattribute__(name)
+
+
+def test_browser_tx_audio_facts_reads_each_lifecycle_method_once():
+    radio = _yaesu(_CountingRadio)
+    radio.reads = dict.fromkeys(("start_tx", "push_tx", "stop_tx"), 0)
+    facts = browser_tx_audio_facts(radio)
+    assert facts.available is True
+    assert radio.reads == {"start_tx": 1, "push_tx": 1, "stop_tx": 1}
+
+
+def test_browser_tx_audio_facts_pin_real_provider_snapshots():
+    icom = browser_tx_audio_facts(IcomRadio("192.168.55.40", model="IC-7610"))
+    yaesu = browser_tx_audio_facts(
+        YaesuCatRadio("/dev/null", audio_driver=SimpleNamespace())
+    )
+    assert astuple(icom)[:-1] == (True, "lan", 5, AudioCodec.PCM_1CH_16BIT, "session")
+    assert astuple(yaesu)[:-1] == (
+        True,
+        "usb",
+        None,
+        AudioCodec.PCM_1CH_16BIT,
+        "session",
+    )
 
 
 # ── /api/v1/state tests ───────────────────────────────────────


morozsm:
Builder evidence for exact head `17170f0a2f4982f36657b636f02e27afda41c5c7`:

- scope: exactly 3 files, `+202/-2`, net `+200`
- all unrelated `origin/main` documentation/comments preserved; the only deletions replace the prior capability import and `__all__` declaration
- focused capability suite: 34 passed
- focused web/audio matrix: 514 passed, 2 skipped
- full non-integration: 8377 passed, 73 skipped, 5 deselected, 11 xfailed, 1 xpassed
- Ruff check/format, strict web mypy, import-linter 5/5, and `git diff --check`: PASS
- exact-head CI: quick, consumer-contracts-gate, and grep-gate PASS
- Agent Review Gate is intentionally red pending a fresh independent exact-head PASS

The full suite rewrites the known generated `txTarget` fixture locally; that unrelated change was reverted and is absent from the commit/PR.

PR remains Draft and explicitly non-mergeable until lifecycle child #1999 is implemented and ready. No self-review or readiness action.

morozsm:
Agent Review: BLOCKED

Reviewed exact head `17170f0a2f4982f36657b636f02e27afda41c5c7` (local `HEAD`, `origin/codex/1998-browser-tx-facts`, and PR `headRefOid` all verified equal before and after review). Independent review — I did not author this change. Base `96b94e42`; `origin/main` has advanced to `53b8e550` but neither of the two intervening commits touches the three reviewed files, so no rebase is needed for correctness.

The fact computation itself is close to right, and all four production backends I could construct resolve exactly as claimed. It is blocked because it is **not independently mergeable** (finding 1), plus one real gap in the lifecycle validator (finding 2).

---

## [Blocking, chain] Merging this slice alone publishes `audioTx` that the current WS neither reads nor honors

`browser_tx_audio_facts` has exactly one consumer on this head — the capability projection:

https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/server.py#L2726-L2763

`codec`, `path`, and `lifecycle` have no consumer at all (`grep -rn browser_tx_audio_facts src/` → definition + `server.py` only). Every WS TX decision still uses independent, looser authority:

| step | this head | snapshot |
|---|---|---|
| admission | [`audio.py:1432`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L1431-L1433) — `CAP_AUDIO` only, no `CAP_TX`, no codec, no path | `audio` + `tx` + codec + path + route |
| arm | [`audio.py:1463`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L1462-L1464) — `_tx_active = True` unconditionally | — |
| push | [`audio.py:1838`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L1837-L1841) re-derives codec via `_tx_codec()` | captured `codec` |
| stop | [`audio.py:1532`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L1530-L1538) re-derives via `getattr` | captured `lifecycle[2]` |

This is not only theoretical drift. A legacy per-codec OPUS radio — exactly the shape `_tx_path` validates as `path="opus"` at [`audio.py:187-191`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L186-L192) — gets advertised as available and then transmits nothing. Reproduced on this SHA through the real `AudioHandler._reader_loop`:

```text
ENDPOINT publishes: {'audioTx': True, 'audioTxRoute': 'usb', 'audioTxRequiredModInputSource': None}
                    (internal path=opus codec=OPUS_1CH)
audio_start tx      -> start_audio_tx_opus()          # radio is KEYED
binary TX frame     -> WARNING audio: push TX error
                       AttributeError: 'LegacyOpusRadio' object has no attribute
                       'push_audio_tx_pcm'. Did you mean: 'push_audio_tx_opus'?
READER: exited cleanly; radio calls = ['start_opus']
```

`_handle_tx_audio` hardcodes `legacy_method="push_audio_tx_pcm"` for browser PCM16 frames regardless of the radio's codec, and [`audio.py:1891-1892`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L1890-L1893) swallows the `AttributeError`. Net operator-visible result of an advertised `audioTx: true`: **keyed transmitter, no modulation, warning-only log.** The converse also holds — a radio with `audio` but not `tx` gets `audioTx: false` while the WS still admits and keys it.

Both directions are exactly the "HTTP/WS authority drift" that #1958 says must never ship as a supported state, and #1998 already records "Current main WS admission is not yet bound to the snapshot." The code is locally correct; the *contract* is not deliverable until #1999 / MOR-1127 binds admission/start/push/stop/re-arm to this snapshot.

**Required:** PR stays Draft. Do not `gh pr ready`, merge, or release ahead of #1999. Repository policy has no rule that would hold this back once a PASS exists and Draft is cleared, so a PASS cannot be issued for this head — Draft is the only guard, and `Agent Review: PASS` per `CLAUDE.md` explicitly authorizes merge once checks are green and the PR is not draft. Land as a stacked pair (or gate slice B on the same head).

## [Medium] `_concrete_async_method` short-circuits on the instance `__dict__`, defeating its own owner/Protocol/abstract checks

https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L155-L169

Line 159-160 returns before the MRO/`_is_protocol`/`__isabstractmethod__` checks at 161-167 **and** before the `__self__ is owner` check at 168. Every one of those guards is unreachable for any attribute present in the instance dict. Reproduced on this SHA:

```text
wrong owner:  victim.start_tx = donor.start_tx (etc.)
  -> available=True route=usb path=neutral
  -> captured lifecycle __self__ is victim: [False, False, False]

unbound zero-arg async fn assigned to instance:
  -> available=True path=neutral lifecycle=(loose, loose, loose)
  -> slice B's push(data) would raise TypeError
```

This is the criterion "wrong-owner / malformed lifecycle evidence cannot establish availability", and slice B will `await` these captured callables — so it is cheaper to close now than after #1999 depends on them.

Not reachable from shipped backends (`grep -rn '\.start_tx *=\|\.push_tx *=\|\.stop_tx *=' src/` → no hits; instance assignment appears only in `tests/`), which is why I am not calling it High. But it materially weakens the evidence, because **both positive cases in the primary matrix survive only because of this bypass.** Hardening the helper in-memory and re-running:

```text
PR fixture _tx_radio()   [({}, (True,"usb",None))]   as-is=True path=neutral | +owner-check=False path=None
PR fixture lan/5 variant [(..., (True,"lan",5))]     as-is=True path=neutral | +owner-check=False path=None
production IcomRadio IC-7610                         as-is=True path=session | +owner-check=True  path=session
production YaesuCatRadio                             as-is=True path=session | +owner-check=True  path=session
```

`_tx_radio` is a `SimpleNamespace` of `AsyncMock`s ([`tests/test_web_capability_guards.py:303-337`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/tests/test_web_capability_guards.py#L302-L338)), so the parametrized matrix never exercises the validator it is nominally testing. Production radios are unaffected by hardening.

**Fix:** apply the owner check uniformly — resolve the definer as instance dict first, then MRO, but require `getattr(method, "__self__", None) is owner` on every path. That will force the positive fixtures onto class-defined lifecycles, which is what #1998's "production-faithful evidence" asks for anyway.

## [Low, non-gating] `_tx_path` re-reads `radio.start_tx` twice per snapshot

[`audio.py:177`](https://github.com/rigplane/rigplane-core/blob/17170f0a2f4982f36657b636f02e27afda41c5c7/src/rigplane/web/handlers/audio.py#L176-L192) builds `neutral` from `start_tx`/`push_tx`/`stop_tx`; lines 182-185 read the same attributes again for the marker-only check. One `browser_tx_audio_facts()` call against a property-backed radio yields `{'start_tx': 2, 'push_tx': 1, 'stop_tx': 1}`. Both outcomes of a mid-computation flip are fail-closed here, and #1998's own criterion names only codec/route/policy (each verified read exactly once), so this is not gating — but it contradicts the "no mutable authority is reread" framing and evaluates a property-backed attribute twice per capabilities request. Derive the marker check from the values already in `neutral`.

## [Low, non-gating] The IC-7610 `true/lan/5` assertion is not production-faithful

#1998 requires "IC-7610 reports `true/lan/5`; FTX-1 reports `true/usb/null` using production-faithful evidence." FTX-1 is genuine (`_StatefulRadio(YaesuCatRadio)`), but the IC-7610 row is a hand-built `SimpleNamespace(backend_id="rigplane", profile=SimpleNamespace(data_mode_count=3), ...)`, not `IcomRadio`. I confirmed manually that real `IcomRadio("192.168.55.40", model="IC-7610")` does resolve `True / lan / 5 / path=session / PCM_1CH_16BIT`, so the behavior is right — the test just doesn't pin it. Nothing asserts `path` or `codec` either, so a `session` → `neutral` regression would pass silently.

---

## Verified sound on this head

- **Production sweep.** `IcomRadio` IC-7610 `True/lan/5/session`; IC-7300, IC-9700, IC-705 `True/lan/None/session` (correct — `data_mode_count < 2` → `LEGACY`, so `rigctld_wsjtx_policy` returns `None`); `YaesuCatRadio` `True/usb/None/session`; `radio=None` `False/None/None`.
- **No serializer leakage.** `as_payload()` emits exactly `audioTx`, `audioTxRoute`, `audioTxRequiredModInputSource`; `codec`/`path`/`lifecycle` never serialize. `server.py` has a single capability serialization site (`grep -n '"tx":' src/rigplane/web/server.py` → one hit, `:2761`) and there is no WS capability frame to drift from.
- **Immutability.** `@dataclass(frozen=True, slots=True)`; assignment raises.
- **Fail-closed set.** Protocol-stub `AudioCapable` subclass, unsupported codec (`PCM_1CH_8BIT`, `None`), sync `push_tx`, missing `tx` cap, unknown backend without USB audio, and `UNAVAILABLE`/incoherent `(source, policy)` pairs all return `False/None/None`; the blanket `except Exception` covers throwing `resolve_audio_route`/`rigctld_wsjtx_policy`.
- **Side effects.** `resolve_audio_route` / `rigctld_wsjtx_policy` are pure `getattr`+branch (`src/rigplane/audio/route.py:257-301`); `browser_tx_audio_facts` performs no writes. Codec/route/policy each read exactly once (pinned by `test_browser_tx_audio_facts_freeze_provider_evidence`); the only reread is finding 3.
- **Unrelated documentation restored.** The whole diff against base contains exactly two deletions — the `capabilities` import line and the old `__all__`. Zero comment or docstring lines removed, which closes the "protocol rationale still stripped" note from the #1973 review.
- **Scope.** Exactly 3 files, `+202/-2`, net `+200` — at the guardrail ceiling, within it.

## Checks run

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` — **8377 passed, 73 skipped, 5 deselected, 11 xfailed, 1 xpassed** (exit 0)
- Focused web/audio matrix (`test_web_capability_guards`, `test_web_audio_tx_session`, `test_web_audio_session_rx`, `test_handlers_coverage`, `test_web_server`, `test_web_server_coverage`, `test_audio_session`, `test_audio_duplex`, `test_web_mod_input_restore`, `test_session_lifecycle`) — **560 passed, 2 skipped**
- `tests/test_web_capability_guards.py` on Python **3.11 / 3.12 / 3.13** — 34 passed each (quick CI only covers 3.11; no `runtime_checkable` version skew)
- `uv run ruff check src/ tests/` — All checks passed; `ruff format --check` on the 3 files — already formatted
- `uv run mypy src/rigplane/web` — Success, 25 source files
- `git diff --check 96b94e42..17170f0a` — clean
- Exact-head CI on `17170f0a`: `quick` success, `consumer-contracts-gate` success, `grep-gate` success; `Agent Review Gate` red pending a fresh PASS
- Adversarial probes on this SHA: wrong-owner/unbound-callable acceptance, instance-dict bypass vs. hardened helper, TOCTOU read counting, full-backend fact sweep, and the endpoint-vs-WS divergence driven through the real `_reader_loop`

PR left Draft. No ready, merge, rebase, or Linear action taken.

morozsm:
Builder remediation evidence for exact head `07d0dd1b07055b655ac205bef60a97ea683ff3c8` after the BLOCKED review:

Boundary correction:
- removed the `/api/v1/capabilities` consumer and all `server.py` changes
- `browser_tx_audio_facts` now has zero production consumers: `rg` finds only its class/definition/returns in `audio.py`
- no public `audioTx` fields, wire change, WS behavior change, or other production behavior change in this slice
- #1999/MOR-1127 now explicitly owns both first public projection and WS enforcement atomically

Lifecycle validation correction:
- instance-dict lifecycle overrides are rejected, including foreign bound methods and unbound async functions
- only class-defined bound async methods whose `__self__` is the evaluated owner are accepted
- Protocol/abstract/partial/sync/wrong-arity evidence fails closed
- `inspect.signature(...).bind(...)` validates the exact future invocation shape, including session acquire, push payload, and PCM start sample-rate keyword
- neutral `start_tx`/`push_tx`/`stop_tx` are each read exactly once per snapshot; no second marker lookup
- real `IcomRadio("192.168.55.40", model="IC-7610")` pins `true/lan/5/PCM_1CH_16BIT/session`
- real `YaesuCatRadio` pins `true/usb/null/PCM_1CH_16BIT/session`

Scope and evidence:
- exactly 2 files, `+198/-1`, net `+197`
- all unrelated protocol documentation/comments preserved
- facts test file: 29 passed
- focused web/audio matrix: 563 passed, 2 skipped
- full non-integration: 8433 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed
- Ruff check/format, strict web mypy, import-linter 5/5, and diff check: PASS
- exact-head CI quick and grep gate: PASS; consumer-contracts workflow is not triggered by this dormant two-file diff
- Agent Review Gate intentionally red pending a fresh independent Opus 5 max exact-head review

PR remains Draft. No self-review, ready, merge, or Linear Done action.

morozsm:
Agent Review: PASS

Reviewed exact head `07d0dd1b07055b655ac205bef60a97ea683ff3c8` — local `HEAD`, `origin/codex/1998-browser-tx-facts`, and PR `headRefOid` verified equal before and after review. Independent review; I did not author this change. Base `96b94e42` (merge-base with `origin/main`); `origin/main` is at `53b8e550` and neither intervening commit touches the two reviewed files, so no rebase is needed for correctness.

Both blocking findings from the prior review on `17170f0a` are closed, and I re-derived the fix rather than trusting the remediation note.

---

## Independent dormancy — the finding that blocked `17170f0a`

`rg` over the whole tree: `browser_tx_audio_facts` / `BrowserTxAudioFacts` appear **only** in their definition site (`src/rigplane/web/handlers/audio.py`) and in `tests/test_web_capability_guards.py`. Zero hits in `src/rigplane/web/server.py`, `frontend/`, or `docs/`.

- `__all__` is byte-identical to base: `["AudioBroadcaster", "AudioHandler", "RxPcmTapSource"]`. Not re-exported from `rigplane/__init__.py` (`"BrowserTxAudioFacts" in rigplane.__all__` → `False`; `browser_tx_audio_facts in dir(rigplane)` → `False`).
- No serializer exists at all in this slice — there is no `as_payload()`, no `astuple` call in `src/`, and no hit for `BrowserTxAudioFacts` / `handlers.audio` in `web/state_schema.py`, `scripts/gen_state_types.py`, or `server.py`. Internals cannot leak because nothing serializes.
- No new import edge: `rigplane.audio.route` was already in `sys.modules` after the **pre-change** import set of this module (verified by importing only HEAD~1's imports). `lint-imports` 5/5 kept.
- New module-level code is definitions only; nothing executes at import.

Merging this alone changes no public capability, no wire format, and no WS admission. The `/api/v1/capabilities` payload and `audio_start direction=tx` are untouched, so the endpoint-vs-WS drift I reproduced on `17170f0a` (advertised `audioTx: true` → keyed transmitter, no modulation) cannot ship from this head.

## Lifecycle validator — the second blocking finding

The instance-`__dict__` short-circuit that made the owner / `_is_protocol` / `__isabstractmethod__` checks unreachable is gone; the rejection is now explicit (`audio.py:138`) and the owner check at `:147` is on every accepted path. Verified by construction, not by reading:

| adversarial input (class-level, on `push_tx`) | result |
|---|---|
| `staticmethod` async, `classmethod` async | rejected (no `__self__` / `__self__` is the class) |
| `AsyncMock()`, `MagicMock()` class attribute | rejected (`__self__` is dunder-blocked on Mock) |
| `functools.partial` of an async function | rejected |
| foreign bound method (donor instance) | rejected |
| sync fn, non-callable, `None`, plain object | rejected |
| `property` returning a **foreign** bound method | rejected |
| `__getattr__`-synthesised method (in no `cls.__dict__`) | rejected |
| bare Protocol body inherited without override | rejected |
| abstract `push_tx` | not instantiable / rejected |

Instance mutation matrix — all six `False`: `victim.start_tx = donor.start_tx`, `.push_tx`, `.stop_tx`, `victim.push_tx = <own bound>`, `victim.push_tx = <unbound async fn>`, and direct `victim.__dict__["push_tx"] = fn`. Class-level monkeypatch with a foreign bound method is also rejected. Arity is genuinely enforced: `LegacyPcmBadArity` (drops the `sample_rate` keyword) → `available=False`.

Critically, the positive fixtures no longer survive via the bypass — `_yaesu()` is a real `YaesuCatRadio`, not a `SimpleNamespace` of `AsyncMock`s, and I confirmed `audio_driver=None` (no stub at all) yields an identical snapshot, so the injected `SimpleNamespace` is not load-bearing.

## Single-read accounting

Instrumented `__getattribute__` over `{capabilities, backend_id, has_usb_audio, profile, audio_tx_codec, audio_stream_contract, audio_codec, audio_session, start_tx, push_tx, stop_tx}` plus a counting proxy on the session's `acquire_tx`:

```
YaesuCatRadio  capabilities:1 audio_codec:1 audio_tx_codec:1 start_tx:1 push_tx:1 stop_tx:1 audio_session:1 backend_id:1  -> >1 read: NONE
IcomRadio      (same) + profile:1                                                                                        -> >1 read: NONE
session.acquire_tx: 1
```

The `_tx_path` double read of `start_tx` from the prior head is gone. A property flipping `audio_tx_codec` between calls is read exactly once (`seq_reads=1`), so no mid-snapshot TOCTOU. `rigctld_wsjtx_policy(route)` takes the already-captured frozen `AudioRoute`, not the radio, so policy derives no second read.

## Codec / route / policy authority

`supported = (PCM_1CH_16BIT, OPUS_1CH)` is an explicit allowlist behind `isinstance(codec, AudioCodec)`. The route coherence set admits exactly `(LAN, DATA2_LAN)`, `(LAN, LEGACY)`, `(USB, DATA1_USB)`, `(ACC, LEGACY)`; an unknown backend without USB audio resolves `(UNAVAILABLE, LEGACY)` and fails closed. Exception fail-closed confirmed for a throwing `backend_id`, `audio_tx_codec`, `audio_session`, and `start_tx`.

## Real-provider snapshots

Constructed the actual production classes (no mocks) and swept every profile the loader knows:

```
IcomRadio IC-7610  True  lan  5     PCM_1CH_16BIT  session  (AudioSession.acquire_tx, AudioRuntimeMixin.push_tx, AudioRuntimeMixin.stop_tx)
IcomRadio IC-7300  True  lan  None  PCM_1CH_16BIT  session   # data_mode_count < 2 -> LEGACY -> no MOD-input requirement
IcomRadio IC-9700  True  lan  None  PCM_1CH_16BIT  session
IcomRadio IC-705   True  lan  None  PCM_1CH_16BIT  session
YaesuCatRadio      True  usb  None  PCM_1CH_16BIT  session  (AudioSession.acquire_tx, YaesuCatRadio.push_tx, YaesuCatRadio.stop_tx)
radio=None         False None None  None           None
```

Matches the pinned IC-7610 `true/lan/5/PCM_1CH_16BIT/session` and FTX-1 `true/usb/null/PCM_1CH_16BIT/session`. The legacy per-codec branch is reachable and correct (`path="pcm"` / `path="opus"`). Snapshots are frozen+slots (`FrozenInstanceError` on every field, no `__dict__`) and idempotent across repeated calls with stable callable identity. No threads or connections are created by construction or by the call.

## Scope and documentation

Exactly **2 files, `+198/-1`, net `+197`** — within the ≤3 files / ≤200 net LOC guardrail. The diff contains **zero deletions under `src/`**; the single deletion is `tests/…: -from dataclasses import replace`, replaced by `from dataclasses import astuple, replace`. No comment or docstring is stripped, closing the note carried over from the #1973 review.

## Checks run on this head

- `uv run pytest tests/ -q --tb=short --ignore=tests/integration` → **8433 passed, 12 skipped, 5 deselected, 11 xfailed, 1 xpassed** (exit 0)
- `uv run pytest tests/test_web_capability_guards.py -q` → **29 passed**
- `uv run ruff check src/ tests/` → All checks passed; `ruff format --check` on both files → already formatted
- `uv run mypy src/rigplane/web` → Success, 25 source files
- `uv run lint-imports` → 5 contracts kept, 0 broken
- `git diff --check 96b94e42..07d0dd1b` → clean
- Exact-head CI on `07d0dd1b`: `quick` success, `grep-gate` success (`consumer-contracts-gate` is not triggered by this two-file diff — correct, it has no contract paths)
- Adversarial probes: consumer-graph sweep, export/serializer leakage, descriptor/MRO/staticmethod/classmethod/partial/Mock/wrapper traps, instance and class mutation matrix, per-attribute read accounting, mid-snapshot property flips, route-coherence and exception fail-closed matrices, full production backend sweep

---

## Non-gating notes for #1999 / MOR-1127

None of these affect this dormant slice; all are reachable only once a consumer exists.

1. **`runtime_capabilities(radio)` at `audio.py:188` sits outside the `try`.** A `capabilities` property raising a non-`AttributeError` propagates instead of returning `unavailable` (reproduced with `RuntimeError`; `AttributeError` fails closed correctly). Unreachable today — every shipped backend exposes `capabilities` as a plain set — and it matches the existing un-guarded `runtime_capabilities` usage across `server.py`. Cheapest fix is moving line 188 inside the `try`.
2. **Reading `radio.audio_session` lazily materialises `radio._audio_bus` and `radio._audio_session`.** Idempotent and by design for those singletons, but it means the snapshot is not strictly pure — under #1999 the first `/api/v1/capabilities` request will construct them. Worth an explicit decision there.
3. **`inspect.signature` follows `__wrapped__`**, so a `functools.wraps`-decorated method can advertise an arity it does not have and pass the `bind` check. Unreachable — no `functools.wraps` exists on any lifecycle method in `backends/`, `runtime/`, or `audio/` — but `follow_wrapped=False` would close it.
4. **Test coverage is narrower than the behaviour.** Route-coherence rejection, the exception fail-closed matrix, and the legacy `pcm`/`opus` paths are all correct (verified above) but unpinned, so a regression there would pass silently.
5. **`astuple(icom)` deep-copies the live radio** through the `lifecycle` field before the `[:-1]` slice. It works and is deterministic on an unconnected radio, but it is a heavy idiom that would break on any radio holding a non-copyable member. Comparing the first three fields directly is equivalent and cheaper.

## Remainder — what this slice does not deliver

#1999 / MOR-1127 atomically owns the first public `audioTx` / `audioTxRoute` / `audioTxRequiredModInputSource` projection **and** audio-WS snapshot enforcement (transcoder-before-arm, acquire/start, push, codec routing, stop, re-arm cleanup, teardown), including the lease-less-safe `audio_stop` and de-key-exactly-once requirements. No public `audioTx=true` may exist without matching WS enforcement, so the two must land together or in that order.

PASS applies to this head only. No merge, ready-state change beyond clearing Draft, rebase, or Linear action taken by me.

PR #2004, opened 2026-07-27
feat: add resource demand foundation

### LINKED ISSUE #2001: MOR-1056 Foundation: pure resource-demand model baseline
## Purpose

Foundation implementation slice for Linear [MOR-1129](https://linear.app/morozsm/issue/MOR-1129/mor-1056-foundation-pure-resource-demand-model-baseline), under umbrella #1994 / MOR-1056.

## Scope and guard

- One new pure model under `frontend/src/lib/runtime/`, its focused tests, and at most one adjacent type-only file.
- **≤3 files and ≤200 net LOC.**
- Pure and unimported: no side effects or consumers; no App/Svelte/channel/audio/transport/TX/registry/overlay changes.
- Excludes retry, terminal teardown/ended state, queued dispose behavior, and advanced completion invalidation.

## Acceptance

- Opaque unique leases; first/shared/last demand declarations; duplicate/unknown release no-ops; unavailable selection does not start; replacement overlap avoids transient stop.
- Focused test, check, lint, build, file/LOC guards and exact-head CI pass.
- Draft PR must reference this issue but **must not close #1994**.

## Sequencing

GitHub Hardening issue will be blocked by this Foundation issue. Do not close umbrella #1994 from the Foundation PR.

morozsm:
## Summary

- add a pure, side-effect-free resource-demand foundation model
- declare exact opaque leases and first/shared/last demand operations
- qualify start/stop completions by issued operation identity and session epoch
- persist per-resource cleanup ownership for the full session using `Object.is(handle)` identity
- give adopted handles stop ownership and orphan handles at most one dispose ownership
- keep configuration limited to the public `available` and `selected` fields

This is a Foundation-only slice. It is intentionally unimported and **not integration-ready**: no runtime, Svelte, channel, audio, transport, TX, retry, teardown, or session-finalization consumer is included.

## Scope guard

- exactly 2 new files
- 200 added LOC versus `main` (129 model + 71 focused tests)
- no production consumers or imports
- closes only the Foundation issue; umbrella #1994 remains open

Linear: [MOR-1129](https://linear.app/morozsm/issue/MOR-1129/mor-1056-foundation-pure-resource-demand-model-baseline)

Closes #2001

## Cleanup ledger contract

Exact head: `73dd51041a4638693dd72923be0fb8022406d8ef`

- ledger lifetime is one `ResourceDemand` session and identity is `(resource, Object.is(handle))`
- a handle accepted active is reserved for cleanup by `stop`; late stale completions can never add a later `dispose`
- an orphan handle can claim `dispose` only once, including across separate operation drains
- equal handle identities under different resources are independent
- pending-start deferral and the live-handle guard preserve stable A→B→A stop priority when a stale completion arrives before the active completion

The ledger is intentionally session-bounded and retains one identity entry per distinct adopted or orphan-cleaned handle until the model is discarded. Retry, terminal teardown, and session finalization remain later hardening scope.

The non-blocking operation-freezing observation is explicitly deferred; issued-operation identity guards remain load-bearing and no mutable-operation behavior was expanded in this remediation.

## Regression evidence

- adopted stable handle: one stale completion is deferred before adoption, then two late same-handle completions after drained and acknowledged stop emit no dispose
- duplicate orphan identity: one dispose both when completions precede one drain and when they span separate drains
- distinct orphan identities: one dispose each
- equal identity across `rx-audio` and `hardware-scope`: one independent dispose per resource
- prior exact/foreign lease, fake/duplicate completion, first/shared/last, configuration whitelist, stable alias, and replacement-overlap guards remain in the consolidated scenarios

## Validation

- focused Vitest: 5/5 passed (one contract scenario plus four cleanup-ledger cases)
- `npm run check`: passed, 0 errors and 0 warnings
- `npm run lint`: passed
- `npm run build`: passed
- full frontend with a fresh local-storage file: 2290/2291; the pre-existing order-dependent `mod-input-tx-guard.test.ts` failed
- isolated `mod-input-tx-guard.test.ts` with a fresh local-storage file: 13/13 passed
- `git diff --check`: passed
- no-consumer/import guard: passed
- exact-head GitHub `grep-gate`: passed in 5s (run 30303942796)
- exact-head GitHub `quick`: passed in 1m35s (run 30303942536)

The PR remains Draft pending a fresh independent review. The Agent Review Gate is expectedly blocked because no PASS comment exists for this exact head.

### DIFF
--- a/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
+++ b/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
@@ -1,71 +1,71 @@
 import { describe, expect, it } from 'vitest';
-import { ResourceDemand } from '../resource-demand';
+import { ResourceDemand, type AppResource } from '../resource-demand';
 const one = (model: ResourceDemand<string>) => {
   const operations = model.takeOperations();
   expect(operations).toHaveLength(1);
   return operations[0];
 };
-const cleanupOrderings = ['stop-before-stale', 'stale-before-stop'] as const;
+const supersede = (model: ResourceDemand<string>, resource: AppResource, consumer: string) => {
+  const lease = model.acquire(resource, consumer);
+  const start = one(model);
+  model.release(lease);
+  return start;
+};
+const cleanupCases = [
+  [false, ['rx-audio', 'rx-audio'], ['H', 'H'], ['rx-audio:H']],
+  [true, ['rx-audio', 'rx-audio'], ['H', 'H'], ['rx-audio:H']],
+  [false, ['rx-audio', 'rx-audio'], ['A', 'B'], ['rx-audio:A', 'rx-audio:B']],
+  [false, ['rx-audio', 'hardware-scope'], ['H', 'H'], ['rx-audio:H', 'hardware-scope:H']],
+] as const;
 describe('ResourceDemand foundation', () => {
-  it('uses exact leases and emits only first start and last stop', () => {
+  it('uses exact leases and remembers adopted handles across cleanup drains', () => {
     const model = new ResourceDemand<string>('session-1');
-    model.configure('hardware-scope', { available: true, selected: true });
-    const first = model.acquire('hardware-scope', 'first');
-    const start = one(model);
-    const second = model.acquire('hardware-scope', 'second');
-    expect(second).not.toBe(first);
-    expect(model.takeOperations()).toEqual([]);
-    expect(model.completeStart(start, { handle: 'scope' })).toBe(true);
-    expect(model.release(first)).toBe(true);
-    expect(model.takeOperations()).toEqual([]);
-    expect(model.release({ ...second } as typeof second)).toBe(false);
-    expect(new ResourceDemand<string>('session-2').release(second)).toBe(false);
-    expect(model.release(second)).toBe(true);
-    expect(one(model)).toMatchObject({ kind: 'stop', handle: 'scope' });
-  });
-  it('keeps unavailable resources idle and copies only public configuration', () => {
-    const model = new ResourceDemand<string>('session-1');
-    const widerConfig = { available: false, selected: true, demand: 99, activeHandle: 'ghost' };
-    model.configure('audio-fft', widerConfig);
+    const wider = { available: false, selected: true, demand: 99, activeHandle: 'ghost' };
+    model.configure('audio-fft', wider);
     model.acquire('audio-fft', 'panel');
-    expect(model.takeOperations()).toEqual([]);
     expect(model.snapshot('audio-fft')).toEqual({
       available: false, selected: true, demand: 1, health: 'inactive',
     });
-    model.configure('audio-fft', { available: true, selected: true });
-    expect(one(model).kind).toBe('start');
-  });
-  it.each(cleanupOrderings)('dedupes stable-handle cleanup for %s ordering', (ordering) => {
-    const model = new ResourceDemand<string>('session-1');
-    model.configure('hardware-scope', { available: true, selected: true });
-    const first = model.acquire('hardware-scope', 'A');
-    const startA = one(model);
-    expect(model.completeStart({ ...startA }, { handle: 'fake' })).toBe(false);
-    model.release(first);
-    const second = model.acquire('hardware-scope', 'B');
-    const startB = one(model);
-    if (ordering === 'stop-before-stale') {
-      expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
-      expect(model.completeStart(startB, { handle: 'stable' })).toBe(false);
-      model.release(second);
-      expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
-    } else {
-      expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
-      expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
-      model.release(second);
-    }
-    expect(one(model)).toMatchObject({ kind: 'stop', handle: 'stable' });
-  });
-  it('rejects a stale stop when replacement demand overlaps it', () => {
-    const model = new ResourceDemand<string>('session-1');
     model.configure('rx-audio', { available: true, selected: true });
-    const first = model.acquire('rx-audio', 'A');
-    model.completeStart(one(model), { handle: 'A' });
-    model.release(first);
-    const oldStop = one(model);
-    model.acquire('rx-audio', 'B');
-    model.completeStart(one(model), { handle: 'B' });
-    expect(model.completeStop(oldStop)).toBe(false);
-    expect(model.snapshot('rx-audio')).toMatchObject({ activeHandle: 'B', health: 'streaming' });
+    const staleA = supersede(model, 'rx-audio', 'A');
+    const late = ['B', 'C'].map((consumer) => supersede(model, 'rx-audio', consumer));
+    const first = model.acquire('rx-audio', 'first');
+    const start = one(model);
+    const shared = model.acquire('rx-audio', 'shared');
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.release({ ...shared } as typeof shared)).toBe(false);
+    expect(new ResourceDemand<string>('session-1').release(shared)).toBe(false);
+    expect(model.completeStart({ ...start }, { handle: 'fake' })).toBe(false);
+    expect(model.completeStart(staleA, { handle: 'stable' })).toBe(false);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.completeStart(start, { handle: 'stable' })).toBe(true);
+    expect(model.completeStart(start, { handle: 'stable' })).toBe(false);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.release(first)).toBe(true);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.release(shared)).toBe(true);
+    const stop = one(model);
+    expect(model.completeStop(stop)).toBe(true);
+    for (const operation of late)
+      expect(model.completeStart(operation, { handle: 'stable' })).toBe(false);
+    expect(model.takeOperations()).toEqual([]);
+    const replacement = model.acquire('rx-audio', 'replacement');
+    model.completeStart(one(model), { handle: 'next' });
+    model.release(replacement);
+    const staleStop = one(model);
+    model.acquire('rx-audio', 'overlap');
+    model.completeStart(one(model), { handle: 'latest' });
+    expect(model.completeStop(staleStop)).toBe(false);
+  });
+  it.each(cleanupCases)('dedupes orphan cleanup %#', (drain, resources, handles, expected) => {
+    const model = new ResourceDemand<string>('session-1');
+    resources.forEach((resource) => model.configure(resource, { available: true, selected: true }));
+    const starts = resources.map((resource, index) => supersede(model, resource, String(index)));
+    const emitted = starts.flatMap((start, index) => {
+      model.completeStart(start, { handle: handles[index] });
+      return drain ? model.takeOperations() : [];
+    });
+    emitted.push(...model.takeOperations());
+    expect(emitted.map((op) => 'handle' in op && `${op.resource}:${op.handle}`)).toEqual(expected);
   });
 });
--- a/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
+++ b/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
@@ -6,48 +6,38 @@ const one = (model: ResourceDemand<string>) => {
   return operations[0];
 };
 const supersede = (model: ResourceDemand<string>, resource: AppResource, consumer: string) => {
-  const lease = model.acquire(resource, consumer);
-  const start = one(model);
+  const lease = model.acquire(resource, consumer), start = one(model);
   model.release(lease);
   return start;
 };
-const cleanupCases = [
-  [false, ['rx-audio', 'rx-audio'], ['H', 'H'], ['rx-audio:H']],
+const cleanupCases = [[false, ['rx-audio', 'rx-audio'], ['H', 'H'], ['rx-audio:H']],
   [true, ['rx-audio', 'rx-audio'], ['H', 'H'], ['rx-audio:H']],
   [false, ['rx-audio', 'rx-audio'], ['A', 'B'], ['rx-audio:A', 'rx-audio:B']],
-  [false, ['rx-audio', 'hardware-scope'], ['H', 'H'], ['rx-audio:H', 'hardware-scope:H']],
-] as const;
+  [false, ['rx-audio', 'hardware-scope'], ['H', 'H'], ['rx-audio:H', 'hardware-scope:H']]] as const;
 describe('ResourceDemand foundation', () => {
   it('uses exact leases and remembers adopted handles across cleanup drains', () => {
     const model = new ResourceDemand<string>('session-1');
     const wider = { available: false, selected: true, demand: 99, activeHandle: 'ghost' };
     model.configure('audio-fft', wider);
     model.acquire('audio-fft', 'panel');
-    expect(model.snapshot('audio-fft')).toEqual({
-      available: false, selected: true, demand: 1, health: 'inactive',
-    });
+    expect(model.snapshot('audio-fft')).toEqual({ available: false, selected: true, demand: 1, health: 'inactive' });
     model.configure('rx-audio', { available: true, selected: true });
     const staleA = supersede(model, 'rx-audio', 'A');
     const late = ['B', 'C'].map((consumer) => supersede(model, 'rx-audio', consumer));
-    const first = model.acquire('rx-audio', 'first');
-    const start = one(model);
+    const first = model.acquire('rx-audio', 'first'), start = one(model);
     const shared = model.acquire('rx-audio', 'shared');
     expect(model.takeOperations()).toEqual([]);
     expect(model.release({ ...shared } as typeof shared)).toBe(false);
     expect(new ResourceDemand<string>('session-1').release(shared)).toBe(false);
     expect(model.completeStart({ ...start }, { handle: 'fake' })).toBe(false);
     expect(model.completeStart(staleA, { handle: 'stable' })).toBe(false);
-    expect(model.takeOperations()).toEqual([]);
     expect(model.completeStart(start, { handle: 'stable' })).toBe(true);
     expect(model.completeStart(start, { handle: 'stable' })).toBe(false);
-    expect(model.takeOperations()).toEqual([]);
     expect(model.release(first)).toBe(true);
-    expect(model.takeOperations()).toEqual([]);
     expect(model.release(shared)).toBe(true);
     const stop = one(model);
     expect(model.completeStop(stop)).toBe(true);
-    for (const operation of late)
-      expect(model.completeStart(operation, { handle: 'stable' })).toBe(false);
+    for (const operation of late) expect(model.completeStart(operation, { handle: 'stable' })).toBe(false);
     expect(model.takeOperations()).toEqual([]);
     const replacement = model.acquire('rx-audio', 'replacement');
     model.completeStart(one(model), { handle: 'next' });
@@ -67,5 +57,15 @@ describe('ResourceDemand foundation', () => {
     });
     emitted.push(...model.takeOperations());
     expect(emitted.map((op) => 'handle' in op && `${op.resource}:${op.handle}`)).toEqual(expected);
+  });
+  it('defers singleton cleanup until pending adoption', () => {
+    const model = new ResourceDemand<string>('session-1');
+    model.configure('rx-audio', { available: true, selected: true });
+    const staleStart = supersede(model, 'rx-audio', 'stale');
+    const liveLease = model.acquire('rx-audio', 'live'), liveStart = one(model);
+    expect(model.completeStart(staleStart, { handle: 'stable' })).toBe(false);
+    expect([model.takeOperations(), model.takeOperations()]).toEqual([[], []]);
+    expect(model.completeStart(liveStart, { handle: 'stable' })).toBe(true);
+    model.release(liveLease); expect(model.takeOperations().map((op) => op.kind)).toEqual(['stop']);
   });
 });
--- a/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
+++ b/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
@@ -1,17 +1,15 @@
 import { describe, expect, it } from 'vitest';
-import { ResourceDemand, type AppResource } from '../resource-demand';
-const ready = (model: ResourceDemand<string>, resource: AppResource) => {
-  model.configure(resource, { available: true, selected: true });
-};
+import { ResourceDemand } from '../resource-demand';
 const one = (model: ResourceDemand<string>) => {
   const operations = model.takeOperations();
   expect(operations).toHaveLength(1);
   return operations[0];
 };
+const cleanupOrderings = ['stop-before-stale', 'stale-before-stop'] as const;
 describe('ResourceDemand foundation', () => {
   it('uses exact leases and emits only first start and last stop', () => {
     const model = new ResourceDemand<string>('session-1');
-    ready(model, 'hardware-scope');
+    model.configure('hardware-scope', { available: true, selected: true });
     const first = model.acquire('hardware-scope', 'first');
     const start = one(model);
     const second = model.acquire('hardware-scope', 'second');
@@ -31,34 +29,36 @@ describe('ResourceDemand foundation', () => {
     model.configure('audio-fft', widerConfig);
     model.acquire('audio-fft', 'panel');
     expect(model.takeOperations()).toEqual([]);
-    expect(model.snapshot('audio-fft')).toMatchObject({
+    expect(model.snapshot('audio-fft')).toEqual({
       available: false, selected: true, demand: 1, health: 'inactive',
     });
-    expect(model.snapshot('audio-fft').activeHandle).toBeUndefined();
-    ready(model, 'audio-fft');
+    model.configure('audio-fft', { available: true, selected: true });
     expect(one(model).kind).toBe('start');
   });
-  it('ignores fake and duplicate starts and protects a stable aliased handle', () => {
+  it.each(cleanupOrderings)('dedupes stable-handle cleanup for %s ordering', (ordering) => {
     const model = new ResourceDemand<string>('session-1');
-    ready(model, 'hardware-scope');
+    model.configure('hardware-scope', { available: true, selected: true });
     const first = model.acquire('hardware-scope', 'A');
     const startA = one(model);
     expect(model.completeStart({ ...startA }, { handle: 'fake' })).toBe(false);
     model.release(first);
     const second = model.acquire('hardware-scope', 'B');
     const startB = one(model);
-    expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
-    expect(model.takeOperations()).toEqual([]);
-    expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
-    expect(model.completeStart(startB, { handle: 'stable' })).toBe(false);
-    expect(model.takeOperations()).toEqual([]);
-    expect(model.snapshot('hardware-scope')).toMatchObject({ activeHandle: 'stable', health: 'streaming' });
-    model.release(second);
+    if (ordering === 'stop-before-stale') {
+      expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
+      expect(model.completeStart(startB, { handle: 'stable' })).toBe(false);
+      model.release(second);
+      expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
+    } else {
+      expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
+      expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
+      model.release(second);
+    }
     expect(one(model)).toMatchObject({ kind: 'stop', handle: 'stable' });
   });
   it('rejects a stale stop when replacement demand overlaps it', () => {
     const model = new ResourceDemand<string>('session-1');
-    ready(model, 'rx-audio');
+    model.configure('rx-audio', { available: true, selected: true });
     const first = model.acquire('rx-audio', 'A');
     model.completeStart(one(model), { handle: 'A' });
     model.release(first);
--- a/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
+++ b/frontend/src/lib/runtime/__tests__/resource-demand.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from 'vitest';
+import { ResourceDemand, type AppResource } from '../resource-demand';
+const ready = (model: ResourceDemand<string>, resource: AppResource) => {
+  model.configure(resource, { available: true, selected: true });
+};
+const one = (model: ResourceDemand<string>) => {
+  const operations = model.takeOperations();
+  expect(operations).toHaveLength(1);
+  return operations[0];
+};
+describe('ResourceDemand foundation', () => {
+  it('uses exact leases and emits only first start and last stop', () => {
+    const model = new ResourceDemand<string>('session-1');
+    ready(model, 'hardware-scope');
+    const first = model.acquire('hardware-scope', 'first');
+    const start = one(model);
+    const second = model.acquire('hardware-scope', 'second');
+    expect(second).not.toBe(first);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.completeStart(start, { handle: 'scope' })).toBe(true);
+    expect(model.release(first)).toBe(true);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.release({ ...second } as typeof second)).toBe(false);
+    expect(new ResourceDemand<string>('session-2').release(second)).toBe(false);
+    expect(model.release(second)).toBe(true);
+    expect(one(model)).toMatchObject({ kind: 'stop', handle: 'scope' });
+  });
+  it('keeps unavailable resources idle and copies only public configuration', () => {
+    const model = new ResourceDemand<string>('session-1');
+    const widerConfig = { available: false, selected: true, demand: 99, activeHandle: 'ghost' };
+    model.configure('audio-fft', widerConfig);
+    model.acquire('audio-fft', 'panel');
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.snapshot('audio-fft')).toMatchObject({
+      available: false, selected: true, demand: 1, health: 'inactive',
+    });
+    expect(model.snapshot('audio-fft').activeHandle).toBeUndefined();
+    ready(model, 'audio-fft');
+    expect(one(model).kind).toBe('start');
+  });
+  it('ignores fake and duplicate starts and protects a stable aliased handle', () => {
+    const model = new ResourceDemand<string>('session-1');
+    ready(model, 'hardware-scope');
+    const first = model.acquire('hardware-scope', 'A');
+    const startA = one(model);
+    expect(model.completeStart({ ...startA }, { handle: 'fake' })).toBe(false);
+    model.release(first);
+    const second = model.acquire('hardware-scope', 'B');
+    const startB = one(model);
+    expect(model.completeStart(startA, { handle: 'stable' })).toBe(false);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.completeStart(startB, { handle: 'stable' })).toBe(true);
+    expect(model.completeStart(startB, { handle: 'stable' })).toBe(false);
+    expect(model.takeOperations()).toEqual([]);
+    expect(model.snapshot('hardware-scope')).toMatchObject({ activeHandle: 'stable', health: 'streaming' });
+    model.release(second);
+    expect(one(model)).toMatchObject({ kind: 'stop', handle: 'stable' });
+  });
+  it('rejects a stale stop when replacement demand overlaps it', () => {
+    const model = new ResourceDemand<string>('session-1');
+    ready(model, 'rx-audio');
+    const first = model.acquire('rx-audio', 'A');
+    model.completeStart(one(model), { handle: 'A' });
+    model.release(first);
+    const oldStop = one(model);
+    model.acquire('rx-audio', 'B');
+    model.completeStart(one(model), { handle: 'B' });
+    expect(model.completeStop(oldStop)).toBe(false);
+    expect(model.snapshot('rx-audio')).toMatchObject({ activeHandle: 'B', health: 'streaming' });
+  });
+});
--- a/frontend/src/lib/runtime/resource-demand.ts
+++ b/frontend/src/lib/runtime/resource-demand.ts
@@ -56,7 +56,9 @@ export class ResourceDemand<H> {
       if (operation.kind !== 'dispose') return true;
       const state = this.state(operation.resource);
       if (state.pending?.kind !== 'start')
-        return !Object.is(state.activeHandle, operation.handle);
+        return !Object.is(state.activeHandle, operation.handle) && !queued.some((candidate) =>
+          candidate.kind === 'stop' && candidate.resource === operation.resource
+            && Object.is(candidate.handle, operation.handle));
       this.operations.push(operation);
       return false;
     });
@@ -103,7 +105,7 @@ export class ResourceDemand<H> {
     if (!wanted) {
       if (state.pending?.kind === 'start') state.pending = undefined;
       if (state.activeHandle !== undefined) {
-        const stop = this.operation('stop', resource, state, state.activeHandle);
+        const stop = this.operation('stop', resource, state.activeHandle);
         state.activeHandle = undefined;
         state.pending = stop;
         this.operations.push(stop);
@@ -112,14 +114,12 @@ export class ResourceDemand<H> {
     }
     if (state.health === 'failed' || state.activeHandle !== undefined || state.pending?.kind === 'start')
       return;
-    const start = this.operation('start', resource, state);
+    const start = this.operation('start', resource);
     state.pending = start;
     state.health = 'starting';
     this.operations.push(start);
   }
-  private operation(
-    kind: 'start' | 'stop', resource: AppResource, state: State<H>, handle?: H,
-  ): ResourceOperation<H> {
+  private operation(kind: 'start' | 'stop', resource: AppResource, handle?: H): ResourceOperation<H> {
     const base = { kind, resource, sessionEpoch: this.sessionEpoch };
     const operation = (kind === 'start' ? base : { ...base, handle: handle as H }) as ResourceOperation<H>;
     if (kind === 'start') this.issuedStarts.add(operation);
--- a/frontend/src/lib/runtime/resource-demand.ts
+++ b/frontend/src/lib/runtime/resource-demand.ts
@@ -24,8 +24,8 @@ export class ResourceDemand<H> {
   private readonly leases = new Set<ResourceLease>();
   private readonly issuedStarts = new WeakSet<object>();
   private readonly cleanupLedger: [AppResource, H][] = [];
+  private readonly adoptedLedger: [AppResource, H][] = [];
   private operations: ResourceOperation<H>[] = [];
-
   constructor(readonly sessionEpoch: string) {}
   configure(resource: AppResource, config: { available: boolean; selected: boolean }): void {
     const state = this.state(resource);
@@ -53,7 +53,7 @@ export class ResourceDemand<H> {
       if (operation.kind !== 'dispose') return true;
       const state = this.state(operation.resource);
       if (state.pending?.kind !== 'start')
-        return !Object.is(state.activeHandle, operation.handle);
+        return !this.adoptedLedger.some(([r, h]) => r === operation.resource && Object.is(h, operation.handle));
       this.operations.push(operation);
       return false;
     });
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: BLOCKED

**Head reviewed:** `ab62473c50718666cd1153ea636d633d5deb6d2a` (`gh pr view 2004 --json headRefOid` == local `git rev-parse HEAD`). Base `origin/main` = `53b8e5506132b5e3113927779f498a6dc22b3017`; `git diff origin/main...ab62473c --stat` = 2 files, +199/-0. Independent review; I did not author this change.

---

## Finding 1 — `dispose` escapes for a stable handle the model itself stopped (duplicate teardown)

- **Severity:** Medium — blocking for this PR, because stable-handle alias safety is the headline invariant this Foundation-only slice exists to establish. No production risk today (zero consumers), and model state is never corrupted.
- **Location:** `frontend/src/lib/runtime/resource-demand.ts:52-63` (`takeOperations`), interacting with `frontend/src/lib/runtime/resource-demand.ts:68-71` (`completeStart` stale-path dispose enqueue).
- **Category:** correctness / contract

### What is wrong

The alias guard at line 59 is a single point-in-time comparison, evaluated only at drain time:

```ts
if (state.pending?.kind !== 'start')
  return !Object.is(state.activeHandle, operation.handle);
```

It suppresses a queued `dispose(H)` only while `H` *is* `activeHandle` at the instant `takeOperations()` runs. It has no memory of `H` having been adopted, and no awareness of a `stop(H)` that supersedes it. `reconcile` clears `state.activeHandle` when it issues the stop (line 106), so from that instant the guard stops protecting `H` — while `H` is still an open handle from the executor's point of view.

Two reachable orderings, both producing a duplicate teardown of the same stable handle:

**(A) Late stale completion — reproduces under *any* drain discipline, including drain-after-every-mutation.**
A superseded `start` resolves with the same singleton handle *after* that handle has already been stopped. The dispose is queued after the stop, so no drain placement can suppress it.

**(B) Two mutations in one batch — `dispose` is emitted *before* the `stop`, i.e. against a handle the model reported `streaming` for the whole batch.**

### Repro (against the committed blob, not the working tree)

```bash
git show ab62473c:frontend/src/lib/runtime/resource-demand.ts > /tmp/rd/resource-demand.ts
# probe.ts:
```
```ts
import { ResourceDemand } from './resource-demand';

// (A) late stale completion, drain after every mutation but the last
const m = new ResourceDemand<string>('e1');
m.configure('rx-audio', { available: true, selected: true });
const l1 = m.acquire('rx-audio', 'A');
const s1 = m.takeOperations()[0];            // start #1, will be superseded
m.release(l1);                                // demand 0 -> pending start cancelled
const l2 = m.acquire('rx-audio', 'B');
const s2 = m.takeOperations()[0];
m.completeStart(s2, { handle: 'STABLE' });    // STABLE adopted
m.takeOperations();                           // drain
m.release(l2);                                // stop(STABLE) queued, activeHandle cleared
m.completeStart(s1, { handle: 'STABLE' });    // late stale completion, same singleton
console.log(JSON.stringify(m.takeOperations()));

// (B) same alias sequence, both completions + release in one undrained batch
const n = new ResourceDemand<string>('e1');
n.configure('hardware-scope', { available: true, selected: true });
const k1 = n.acquire('hardware-scope', 'A');
const t1 = n.takeOperations()[0];
n.release(k1);
const k2 = n.acquire('hardware-scope', 'B');
const t2 = n.takeOperations()[0];
n.completeStart(t1, { handle: 'STABLE' });    // queues dispose(STABLE)
n.completeStart(t2, { handle: 'STABLE' });    // STABLE adopted
n.release(k2);                                // stop(STABLE)
console.log(JSON.stringify(n.takeOperations()));
```
```
esbuild probe.ts --bundle --platform=node --format=cjs --outfile=probe.cjs && node probe.cjs
```

**Actual output**

```
(A) [{"kind":"stop","resource":"rx-audio","sessionEpoch":"e1","handle":"STABLE"},
     {"kind":"dispose","resource":"rx-audio","sessionEpoch":"e1","handle":"STABLE"}]

(B) [{"kind":"dispose","resource":"hardware-scope","sessionEpoch":"e1","handle":"STABLE"},
     {"kind":"stop","resource":"hardware-scope","sessionEpoch":"e1","handle":"STABLE"}]
```

**Expected:** exactly one teardown operation for `STABLE` in each batch (the `stop`); the superseded `dispose` must be suppressed. In (B) the `dispose` is additionally ordered *ahead* of the `stop`, so a consumer walking the batch in order tears down a handle that `snapshot()` reported as `streaming` for the entire batch.

### Why this is a Foundation defect, not MOR-1128 hardening

This is the model's own emission ordering for `takeOperations`, not retry, teardown, session finalization or failure handling. The PR body claims "preserve duplicate-completion and stable-handle alias safety" as a delivered property of this slice, and the acceptance for the slice is explicitly that a stale completion cannot dispose a live stable handle. As written the property holds only for one drain interleaving.

### Fix

Deduplicate teardown operations by `(resource, handle)` at emit time, preferring `stop` over `dispose`, instead of relying on the drain-time `activeHandle` comparison alone. This is the same merge rule the follow-up hardening work already applies inside `teardown()`'s `addCleanup`, so the semantics are settled — `takeOperations` is simply missing the analogue. A minimal, order-independent shape:

- when `reconcile` issues `stop(H)` for a resource, drop any already-queued `dispose` for the same `(resource, H)`; and
- in `takeOperations`, drop a `dispose` whose `(resource, handle)` matches a `stop` present in the same emitted batch or in the still-outstanding queue.

### Test gap

`frontend/src/lib/runtime/__tests__/resource-demand.test.ts:41-59` (`ignores fake and duplicate starts and protects a stable aliased handle`) passes only because of the drain at line 54, which sits at the exact point where the `activeHandle` comparison is favourable. It mirrors the implementation's happy path rather than pinning the invariant. Please add cases for orderings (A) and (B) above; both fail today.

---

## Non-blocking observations (please address or explicitly defer)

1. **Operations are mutable and alias private state** — `resource-demand.ts:118-127`. `acquire` freezes the lease (`resource-demand.ts:36-38`), but `operation()` does not freeze the operation, and the object stored in `state.pending` is the *same reference* handed to the caller by `takeOperations`. Probe: mutating `op.resource` on a returned start op changes `snapshot(...).pending` (`internal pending mutated by caller = true`). Given the slice claims "opaque exact lease and operation identity", the operation half is not opaque. `Object.freeze` on the operation would make it symmetric with the lease at no cost.
2. **A `dispose` deferred behind a start that never resolves is retained indefinitely** (`resource-demand.ts:58-61`) — no path emits it and there is no teardown yet. Acknowledged as MOR-1128 scope; flagging so it is not lost.
3. **Three lines exceed 100 chars** (65, 113, 124). Cosmetic; the overall density matches sibling modules such as `lib/runtime/adapters/presentation-capabilities.ts`, so I do not read the formatting as scope compression.

---

## What I verified and what passed

**Scope / dormancy — all confirmed:**
- 2 files, +199/-0 net vs `origin/main` — within the ≤3 files / ≤200 LOC guardrail.
- Zero imports in `resource-demand.ts` (no runtime, Svelte, channel, audio, transport or TX dependency).
- Zero production consumers: `grep -rn 'resource-demand|ResourceDemand|ResourceLease|AppResource' --include='*.ts' --include='*.svelte' --include='*.js' --include='*.json' frontend/src frontend/vite.config.ts` returns only the model and its own test.
- Public surface is exactly `configure, acquire, release, takeOperations, completeStart, completeStop, snapshot` — no `retry`, no `teardown`. No over-claim of later hardening. Failure latches closed (`health: 'failed'` is sticky, no retry path), which is the correct conservative default for a dormant slice.

**Adversarial probes run against the committed blob (13 groups, 12 clean, 1 failure = Finding 1):**
- Structural lease forgery (frozen clone, identical shape and epoch) rejected; demand not decremented. PASS
- Cross-instance release with *identical* epoch rejected; neither instance mutated. PASS
- Duplicate release rejected; demand never goes negative. PASS
- Cross-instance `completeStart` with identical epoch rejected, no state mutation, no dispose of the foreign handle; each instance still accepts its own op. PASS
- Duplicate `completeStart` / `completeStop` rejected; active handle not replaced; no dispose of the live handle. PASS
- `configure` with a prototype-backed, wider object carrying `demand`/`health`/`activeHandle`: only `available`/`selected` copied, getter evaluated exactly once, snapshot key set unchanged. PASS
- `snapshot()` is a copy — mutating it does not leak back. PASS
- Lease is frozen and unforgeable. PASS
- First/shared/last determinism: 5 acquires emit exactly one `start`, 4 releases emit nothing, the 5th emits exactly one `stop`. PASS
- Orphan handle from a cancelled start is disposed, not leaked. PASS
- Stable handle A→B→A **in the interleaving the test uses**. PASS
- Stable handle A→B→A **in orderings (A) and (B) above**. **FAIL — Finding 1**

**Gates:** exact-head CI is green — `quick` pass 1m38s (run 30300920534; runs `npm run i18n:check`, `npm run check`, `npx vitest run` over the full frontend suite, `npm run build`), `Rebrand grep gate` pass 5s (run 30300920418). `Agent Review Gate` fails only for the missing PASS comment on this head. I did **not** run the frontend gates locally: the review worktree, while at `ab62473c`, carries uncommitted MOR-1128 hardening edits to `frontend/src/lib/runtime/resource-demand.ts`, so a local `npm run check` / `vitest` would have exercised the wrong source. All findings above were derived from `git show ab62473c:...` only.

**Note (not a defect):** the review worktree is on branch `codex/mor-1128-demand-hardening` while the PR head branch is `codex/mor-1129-demand-foundation`; both resolve to `ab62473c`, and the contract judged here is the committed blob.

---

## Required to unblock

1. Fix Finding 1 in `takeOperations` / the stop-issuing path so teardown emission is order-independent.
2. Add regression tests for orderings (A) and (B).
3. Push, let `quick` + `grep-gate` go green on the new head, and request a fresh review. The remaining scope — retry, teardown/session finalization, unavailable/failure hardening, and any consumer integration — correctly stays with MOR-1128 and later children of MOR-1056; I am not asking for it here.

Leaving the PR in Draft.

morozsm:
Agent Review: BLOCKED

**Head reviewed:** `95f71d72ad178a66d63c4daa76f3cecc136b5e67` (`gh pr view 2004 --json headRefOid` == local `git rev-parse HEAD`; worktree clean and detached at that SHA). Base = merge-base with `origin/main` = `53b8e5506132b5e3113927779f498a6dc22b3017`. `git diff 53b8e550...95f71d72 --numstat` = 2 files, **+199/-0**. Fresh independent review; I did not author this change and did not review the previous head.

The remediation commit `95f71d72` is a genuine improvement — both orderings named in the prior review now emit exactly one `stop`, and I confirmed that directly. But the dedupe was implemented as a **filter over the transient `queued` array**, not as a property of handle ownership, so the invariant still holds only for the interleavings the new tests pin. Two same-class defects survive, one of which needs no drain trick at all.

---

## Finding 1 — cleanup dedupe is batch-local: a drained `stop` no longer suppresses a later stale `dispose`

- **Severity:** Medium — blocking. No production risk today (zero consumers, model state never corrupted), but "exactly one cleanup per handle across all interleavings and drains" is the headline invariant this Foundation slice exists to establish, and it is the same defect class the previous review blocked on.
- **Location:** `frontend/src/lib/runtime/resource-demand.ts:55-64` (`takeOperations`), interacting with `resource-demand.ts:70-73` (`completeStart` stale-path dispose enqueue) and `resource-demand.ts:105-113` (`reconcile` clearing `activeHandle` when it issues the stop).
- **Category:** correctness / contract

### What is wrong

The new guard suppresses a `dispose(H)` only if a matching `stop(H)` is **in the same `queued` array**:

```ts
return !Object.is(state.activeHandle, operation.handle) && !queued.some((candidate) =>
  candidate.kind === 'stop' && candidate.resource === operation.resource
    && Object.is(candidate.handle, operation.handle));
```

`queued` is the batch being drained right now. Once a `stop(H)` has been handed to the caller in an earlier `takeOperations()`, it is gone from the model's memory — and `reconcile` already cleared `state.activeHandle` at line 109. From that instant nothing suppresses a `dispose(H)`, even though `H` is a handle the model itself adopted and has already ordered torn down. `takeOperations()` is a *drain*, so a consumer that drains per reconcile tick — the natural usage — hits this on every late stale completion.

### Repro (committed blob only; `git show 95f71d72:...`)

```bash
mkdir -p /tmp/rd && git show 95f71d72:frontend/src/lib/runtime/resource-demand.ts \
  | sed 's/constructor(readonly sessionEpoch: string) {}/sessionEpoch;\n  constructor(sessionEpoch) { this.sessionEpoch = sessionEpoch; }/' > /tmp/rd/rd.ts
# (only transformation: the parameter property, which Node cannot type-strip; semantically identical)
```

```js
import { ResourceDemand } from './rd.ts';
const kinds = (o) => o.map((x) => `${x.kind}(${x.resource},${String(x.handle)})`).join(' ');

const m = new ResourceDemand('e1');
m.configure('rx-audio', { available: true, selected: true });
const l1 = m.acquire('rx-audio', 'A');
const startA = m.takeOperations()[0];          // superseded start, still outstanding
m.release(l1);
const l2 = m.acquire('rx-audio', 'B');
const startB = m.takeOperations()[0];
m.completeStart(startB, { handle: 'STABLE' }); // singleton handle adopted
m.takeOperations();
m.release(l2);                                 // stop(STABLE) queued, activeHandle cleared
console.log('batch1:', kinds(m.takeOperations()));   // <-- drain: stop leaves the queue
m.completeStart(startA, { handle: 'STABLE' }); // late stale completion, same singleton
console.log('batch2:', kinds(m.takeOperations()));
```

**Actual**
```
batch1: stop(rx-audio,STABLE)
batch2: dispose(rx-audio,STABLE)
```
**Expected:** exactly one cleanup for `STABLE` — the `stop`. The consumer has already torn `STABLE` down; the `dispose` is a second teardown of a dead handle.

Two further confirmed variants of the same root cause:

- **Stop acknowledged first.** Same sequence with `completeStop(stop)` called before the late stale completion — the model knows the resource is `inactive` — still emits `dispose(audio-fft,STABLE)`.
- **Multiple late stales.** Three generations, `STABLE` adopted by the last, stop drained, then two superseded starts resolve with `STABLE` → `stop(STABLE)` followed by **two** `dispose(rx-audio,STABLE)`.

---

## Finding 2 — `dispose` is never deduped against `dispose`, even inside a single batch

- **Severity:** Medium — blocking, same fix.
- **Location:** `frontend/src/lib/runtime/resource-demand.ts:59-61` — the `queued.some(...)` predicate only ever matches `candidate.kind === 'stop'`.

The PR claims cleanup is deduplicated "by resource and `Object.is(handle)`". It is not: two stale completions carrying the *same* handle produce two `dispose` operations in **one drain**, with no `stop` involved and no drain-between required.

### Repro

```js
const m = new ResourceDemand('e1');
m.configure('rx-audio', { available: true, selected: true });
const l1 = m.acquire('rx-audio', 'A'); const sA = m.takeOperations()[0];
m.release(l1);
const l2 = m.acquire('rx-audio', 'B'); const sB = m.takeOperations()[0];
m.release(l2);
m.acquire('rx-audio', 'C');            const sC = m.takeOperations()[0];
m.completeStart(sA, { handle: 'H' });  // deferred behind pending startC
m.completeStart(sB, { handle: 'H' });  // deferred, SAME handle
console.log('while pending :', kinds(m.takeOperations()) || '(empty)');
m.completeStart(sC, { handle: 'OTHER' });
console.log('after resolve :', kinds(m.takeOperations()));
```

**Actual**
```
while pending : (empty)
after resolve : dispose(rx-audio,H) dispose(rx-audio,H)
```
**Expected:** one `dispose(rx-audio,H)`.

---

## Why these are Foundation defects, not MOR-1128 hardening

This is `takeOperations`' own emission contract — not retry, teardown, session finalization, or failure handling, none of which this slice exposes. The acceptance for the slice is explicitly that cleanup dedupes by resource + `Object.is(handle)` across interleavings **and drains**, with `stop` taking priority over `dispose`, emitting exactly one cleanup for a singleton stable handle. I am not asking for any later-child scope here.

## Fix

Key the suppression on **handle ownership rather than queue contents**: record the handles the model itself adopted (per resource) and never emit a `dispose` for one — cleanup of an adopted handle is owned by its `stop`. That makes emission order-independent and drain-independent, and subsumes both findings.

I validated exactly this shape against the committed blob (added `owned: H[]` to `State`, pushed `result.handle` at the adopt site `resource-demand.ts:78`, and changed the stale-path condition at line 71 to `!state.owned.some((h) => Object.is(h, result.handle))`). Result: Findings 1 and 2 clear, and no probe regresses — cross-resource equal handles still both emit, genuinely distinct orphan handles are still disposed exactly once, and the deferred-dispose path still emits once the pending start resolves. The `queued.some(...)` clause becomes redundant under this rule. Please decide how `owned` should be bounded (it grows per distinct adopted handle per resource in a session) — a note in the module is fine for a Foundation slice.

## Test gap

`frontend/src/lib/runtime/__tests__/resource-demand.test.ts:38-58` is parameterized over two orderings, but both drain only once at the very end (`one(model)` at line 57) — the exact placement where the `queued.some` guard is favourable. It pins the implementation's batch shape, not the invariant. Please add:

1. drain **between** the stop and the late stale completion (Finding 1);
2. stop drained **and** `completeStop`-acknowledged before the stale completion;
3. two stale completions carrying the same handle, no stop in the batch (Finding 2).

All three fail today.

---

## What I verified, and what passed

**Scope / dormancy — all confirmed:**
- 2 files, +199/-0 vs base — within the ≤3 files / ≤200 LOC guardrail.
- Zero imports in `resource-demand.ts` (no runtime, Svelte, channel, audio, transport or TX dependency); no side effects at module load.
- Zero production consumers: `grep -rn 'resource-demand|ResourceDemand|ResourceLease|AppResource' --include='*.ts' --include='*.svelte' --include='*.js' --include='*.mjs' --include='*.json' frontend/src frontend/vite.config.ts` matches only the module itself and its own test.
- Public surface is exactly `configure, acquire, release, takeOperations, completeStart, completeStop, snapshot` — no `retry`, no `teardown`, no session finalization. **No overclaim** of MOR-1128/later scope. Failure latches closed (`health: 'failed'` sticky, no retry path), the correct conservative default for a dormant slice.
- `git status` clean, `git diff --check` clean.

**Adversarial probes against the committed blob (36 assertions, 31 pass):**
- Prior ordering (A) stop-before-stale, single batch → exactly one `stop`. **PASS (fixed)**
- Prior ordering (B) stale-before-stop, single batch → exactly one `stop`. **PASS (fixed)**
- Drain **between** stop and late stale. **FAIL — Finding 1**
- Stop drained + `completeStop` acknowledged, then late stale. **FAIL — Finding 1**
- Two late stale completions of the same handle after a drained stop. **FAIL — Finding 1**
- Two stale completions of the same handle, single batch, no stop. **FAIL — Finding 2**
- Cross-resource equal handles (`SHARED` on `rx-audio` and `hardware-scope`): both emitted, no cross-suppression. PASS
- Genuine distinct orphan handle disposed exactly once alongside the stop — no leak. PASS
- Structurally-equal but distinct object handles both handled; same object reference deduped — identity really is `Object.is`, not `==`. PASS
- Structural lease forgery (frozen clone, identical epoch) rejected; demand not decremented. PASS
- Cross-instance release with *identical* epoch rejected; neither instance mutated; demand never negative; duplicate release rejected. PASS
- Cross-instance `completeStart` with identical epoch rejected — no mutation, no dispose of the foreign handle. PASS
- Structural start forgery rejected; duplicate `completeStart`/`completeStop` rejected; live handle not replaced and not cleaned up. PASS
- `configure` with a prototype-backed wider object carrying `demand`/`health`/`activeHandle`/`pending` plus a getter on `selected`: only `available`/`selected` copied, prototype fields ignored, getter evaluated exactly once. PASS
- `snapshot()` is a copy — mutating it does not leak back. PASS
- First/shared/last determinism: 5 acquires → exactly one `start`; 4 releases → nothing; 5th → exactly one `stop` carrying the live handle. PASS
- Repeated/empty `takeOperations()` idempotent. PASS
- Deferred dispose behind a pending start stays deferred across repeated drains, then emits exactly once when the start resolves. PASS

**Gates run locally at exact head** (`node_modules` installed fresh via `npm ci` in this worktree; unlike the previous review there are no foreign edits, so these exercise the real source):
- focused `npx vitest run src/lib/runtime/__tests__/resource-demand.test.ts` → **5/5 passed**
- `npm run check` → **4266 files, 0 errors, 0 warnings**
- `npm run lint` → clean
- `npm run build` → built in 6.92s (only the pre-existing chunk-size advisory)

**Exact-head CI:** `Tests (quick)` success (run 30302379496, 1m37s — includes the full frontend vitest suite), `Rebrand grep gate` success (run 30302379445). `Agent Review Gate` fails solely for the missing PASS comment on this head.

## Non-blocking (please address or explicitly defer)

1. **Operations are mutable and alias private state** — `resource-demand.ts:122-127`. `acquire` freezes the lease (line 36) but `operation()` does not freeze the operation, and the object stored in `state.pending` is the *same reference* returned by `takeOperations()` and exposed via `snapshot().pending`. Probe: setting `op.resource = 'hardware-scope'` on a returned start op changes `snapshot('rx-audio').pending.resource`. It fails closed (`completeStop`'s `state.pending !== op` check then rejects), so no live handle is corrupted, but given the slice claims opaque operation identity, `Object.freeze` on the operation would make it symmetric with the lease at no cost. Carried over from the previous review, still open.
2. **A `dispose` deferred behind a start that never resolves is retained indefinitely** (`resource-demand.ts:58-63`) — acknowledged as MOR-1128 scope; flagging so it is not lost.
3. **Three lines exceed 100 chars** (65, 115, 124). Cosmetic; density matches sibling modules such as `lib/runtime/adapters/presentation-capabilities.ts`, so I do not read the formatting as scope compression.

## Required to unblock

1. Make cleanup suppression a property of handle ownership so it is order- and drain-independent (Findings 1 and 2).
2. Add the three regression cases listed under **Test gap**.
3. Push, let `quick` + `grep-gate` go green on the new head, and request a fresh review.

Retry, teardown/session finalization, unavailable/failure hardening, and consumer integration correctly remain with MOR-1128 and later children of MOR-1056 — not asked for here.

Leaving the PR in Draft.

morozsm:
Agent Review: BLOCKED

**Head reviewed:** `73dd51041a4638693dd72923be0fb8022406d8ef` (`gh pr view 2004 --json headRefOid` == local `git rev-parse HEAD`, detached worktree, `git status --short` clean). Merge-base with `origin/main` (`0c6c97ec9cc137fa57df4f5d43e2c3a7387ec162`) = `53b8e5506132b5e3113927779f498a6dc22b3017`, confirmed via `git merge-base`. `git diff --numstat 53b8e550..73dd5104` = 2 files, **+200/-0**. Fresh independent review; I did not author this change or either remediation, and I did not write the two prior reviews.

The `cleanupLedger` remediation is a real improvement and it **does** close both defects the previous review blocked on — I reproduced that independently (probes A1–A3, B1–B3 below all pass now). But the ledger is consulted at **claim** time, not at **emit** time, and the adopt path at line 73 **discards** `claimCleanup`'s return value. So the ownership rule holds only in the direction where the stale completion arrives *after* adoption. In the opposite direction — stale completion *before* adoption of the same handle — the queued `dispose` is still guarded solely by the transient `state.activeHandle` comparison, which `reconcile` nullifies the instant it issues the `stop`. This is the third occurrence of the same defect class, in the direction neither prior review probed.

---

## Finding 1 — an adopted handle still receives both a `dispose` and a `stop`, with the `dispose` ordered first

- **Severity:** Medium — blocking. No production risk today (zero consumers, model state never corrupted), but "an adopted handle is cleaned only by its `stop`" is the headline invariant of this Foundation slice and the explicit contract in the PR body ("a handle accepted active is reserved for cleanup by `stop`").
- **Location:** `frontend/src/lib/runtime/resource-demand.ts:55-56` (the `takeOperations` live-handle guard), interacting with `resource-demand.ts:73` (adopt path discards `claimCleanup`'s result) and `resource-demand.ts:108-111` (`reconcile` clears `activeHandle` when it issues the stop).
- **Category:** correctness / contract

### What is wrong

When a superseded start resolves *before* the live start, it legitimately claims the ledger — nothing had adopted that handle yet — and queues `dispose(H)`:

```ts
if (state.pending !== op) {
  if ('handle' in result && this.claimCleanup(op.resource, result.handle))
    this.operations.push({ ...op, kind: 'dispose', handle: result.handle });
  return false;
}
```

The live start then resolves with the **same singleton handle**. In the adopt branch the ledger call's result is thrown away:

```ts
this.claimCleanup(op.resource, result.handle);   // returns false here — ignored
state.activeHandle = result.handle;
```

So the model adopts a handle that already has a `dispose` in flight, and the only thing protecting it is the drain-time comparison:

```ts
if (state.pending?.kind !== 'start')
  return !Object.is(state.activeHandle, operation.handle);
```

`reconcile` sets `state.activeHandle = undefined` at line 110 the moment it issues the `stop`. From that instant the guard compares against `undefined` and the `dispose` escapes — in the *same batch* as the `stop`, and **ahead of it** in the emitted array.

The PR's own test passes only because line 40 (`expect(model.takeOperations()).toEqual([])`) drains between the stale completion and the adoption at line 41, and line 43 drains again while `activeHandle` is still set — the exact two placements where the guard is favourable. Remove either drain and the invariant breaks. That is drain-discipline dependence, the same property both prior reviews blocked on.

### Repro (against the committed blob at `73dd5104`; worktree source verified identical via `git diff --stat HEAD -- frontend/src/lib/runtime/resource-demand.ts` = empty)

`frontend/src/lib/runtime/__tests__/zz-minimal-repro.test.ts`:

```ts
import { describe, expect, it } from 'vitest';
import { ResourceDemand, type ResourceOperation } from '../resource-demand';

const kinds = (ops: ResourceOperation<string>[]) =>
  ops.map((o) => `${o.kind}(${o.resource},${'handle' in o ? String(o.handle) : '-'})`);

it('stale completion before adoption of the same singleton handle', () => {
  const m = new ResourceDemand<string>('e1');
  m.configure('rx-audio', { available: true, selected: true });

  const l1 = m.acquire('rx-audio', 'A');
  const staleStart = m.takeOperations()[0];
  m.release(l1);                                  // start superseded

  const l2 = m.acquire('rx-audio', 'B');
  const liveStart = m.takeOperations()[0];

  m.completeStart(staleStart, { handle: 'STABLE' });        // queues dispose(STABLE)
  expect(m.completeStart(liveStart, { handle: 'STABLE' })).toBe(true);  // adopts STABLE
  expect(m.snapshot('rx-audio').activeHandle).toBe('STABLE');
  expect(m.snapshot('rx-audio').health).toBe('streaming');

  m.release(l2);                                  // model orders stop(STABLE)

  expect(kinds(m.takeOperations())).toEqual(['stop(rx-audio,STABLE)']);
});
```

```
npx vitest run src/lib/runtime/__tests__/zz-minimal-repro.test.ts
```

**Actual**
```
ACTUAL BATCH: ["dispose(rx-audio,STABLE)","stop(rx-audio,STABLE)"]

AssertionError: expected [ 'dispose(rx-audio,STABLE)', …(1) ] to deeply equal [ 'stop(rx-audio,STABLE)' ]
- Expected
+ Received
  [
+   "dispose(rx-audio,STABLE)",
    "stop(rx-audio,STABLE)",
  ]
```

**Expected:** exactly one teardown operation for `STABLE` — the `stop`.

Note there is **no drain trick** here: the stale and live completions are adjacent, with no `takeOperations()` between them. A consumer walking the batch in order disposes a handle that `snapshot()` reported as `streaming` for the whole batch, and then stops the handle it just disposed.

A second confirmed variant (probe A5) inserts drains *while the start is pending* — the deferral holds correctly across repeated drains, the handle is then adopted, and the `dispose` still escapes alongside the `stop` once `release` clears `activeHandle`. So the deferral path does not rescue this either.

### Fix

Make the emit-time suppression a property of **adoption ownership** rather than of the transient `activeHandle` — which is what the previous review prescribed, and what the shared `cleanupLedger` does not express (it conflates "this identity was claimed for cleanup by whoever got there first" with "this identity is owned by a stop"). I validated a 3-line shape against the committed blob:

```ts
private readonly adoptedLedger: [AppResource, H][] = [];
// in completeStart's adopt branch, next to the existing claimCleanup call:
this.adoptedLedger.push([op.resource, result.handle]);
// in takeOperations, replacing the activeHandle comparison:
if (state.pending?.kind !== 'start')
  return !this.adoptedLedger.some(([r, h]) =>
    r === operation.resource && Object.is(h, operation.handle));
```

**Result: 33/33 green** — the minimal repro, all 27 of my adversarial probes, and the 5 existing PR tests, with no regression. Cross-resource equal handles still emit independently, genuinely distinct orphans still dispose exactly once each, and the pending-start deferral still emits once the start resolves. I reverted this patch and re-verified the tree against the committed blob before running the gates below.

### Test gap

`frontend/src/lib/runtime/__tests__/resource-demand.test.ts:38-43` covers stale-before-adoption only with a drain at line 40 and another at line 43. Please add the ordering above (stale completion → adoption → release → single drain), and the A5 variant (drains only while the start is pending). Both fail today.

---

## What I verified, and what passed

**Scope / dormancy / hygiene — all confirmed:**
- 2 files, **+200/-0** vs merge-base — within the ≤3 files / ≤200 LOC guardrail. Zero deletions anywhere in the diff (`git diff .. | grep -c '^-[^-]'` = 0), so **no unrelated docs or user-facing changes are removed**. `git diff --name-status` lists only the two added files.
- `git diff --check` clean.
- **Slice is dormant:** repo-wide `grep -rn -E 'resource-demand|ResourceDemand|ResourceLease|AppResource'` (excluding `node_modules` and `.git`) matches only the module and its own test. Zero production consumers.
- **Slice is pure:** zero `import`/`require` statements in `resource-demand.ts`; no runtime, Svelte, channel, audio, transport, TX, retry, teardown or session-finalization dependency; no side effects at module load.
- Public surface is exactly `configure, acquire, release, takeOperations, completeStart, completeStop, snapshot` — no `retry`, no `teardown`. No overclaim of later-child scope.
- **No integration conflict with `origin/main`:** `git ls-tree -r origin/main` has no `resource-demand*` file, and `frontend/src/lib/runtime/` on main is untouched between the merge-base and `0c6c97ec`. `mergeable: MERGEABLE`.

**Adversarial probes against the committed blob (27 probes, 25 pass, 2 fail = Finding 1):**
- **A1** drain *between* stop and late stale completion → no dispose. **PASS (prior Finding 1 closed)**
- **A2** stop drained *and* `completeStop`-acknowledged before the stale → no dispose. **PASS (closed)**
- **A3** three generations, two late stales with the stable handle → zero disposes. **PASS (closed)**
- **A4** stale *before* adoption of the same handle, no drain between → **FAIL — Finding 1**
- **A5** stale before adoption, drains only while the start is pending → **FAIL — Finding 1**
- **B1** two stale completions, same handle, same batch, no stop → exactly one dispose. **PASS (prior Finding 2 closed)**
- **B2** two stale completions, same handle, separate drains → exactly one. **PASS (closed)**
- **B3** five stale completions of the same handle across mixed drains → exactly one. **PASS**
- **B4** no over-suppression: three genuinely distinct orphan handles → one dispose each. **PASS**
- **C1** equal identity `SHARED` on `rx-audio` and `hardware-scope` → one independent dispose per resource. **PASS**
- **C2** adopted `SHARED` on `rx-audio` does not suppress orphan `SHARED` on `hardware-scope`; the live handle is untouched. **PASS**
- **D1** `Object.is` identity is real: structurally-equal-but-distinct objects both dispose, the same reference dedupes. **PASS**
- **D2** `NaN` handles dedupe (`Object.is`, not `===`). **PASS**
- **E1** generation replacement: stale stop rejected, new handle adopted, one stop per generation. **PASS**
- **E2** stop-before-start ordering preserved within one batch. **PASS**
- **E3** stop-before-start: releasing while a start is pending emits nothing, then orphans correctly. **PASS**
- **E4** duplicate `completeStart`/`completeStop` rejected; live handle never disposed or replaced. **PASS**
- **E5** forged (`{...op}`) and cross-instance same-epoch operations rejected; no dispose of the foreign handle. **PASS**
- **E6** failed start latches closed; re-acquire does not restart (documented deferral). **PASS**
- **F1** lease forgery (frozen clone) / cross-instance same-epoch / double release all rejected; demand never negative. **PASS**
- **F2** first/shared/last determinism over 5 leases. **PASS**
- **F3** `configure` whitelist holds against a prototype-backed wider object carrying `demand`/`health`/`activeHandle`/`pending` plus a getter on `selected`: only `available`/`selected` copied, getter evaluated exactly once, snapshot key set unchanged. **PASS**
- **F4** `snapshot()` is a copy — mutating it does not leak back. **PASS**
- **F5** unavailable/deselected mid-stream stops the live handle exactly once, no duplicate stop. **PASS**
- **F6** repeated/empty `takeOperations()` idempotent. **PASS**
- **F7** dispose deferred behind a never-resolving start stays queued across repeated drains, emits exactly once on resolve. **PASS**
- **F8** characterization (not a defect, see below): a recycled handle identity after a completed stop is not re-disposed. **PASS**

**Gates run locally at exact head** (`npm ci` in this worktree; probe files removed and tree re-verified clean before running):
- focused `npx vitest run src/lib/runtime/__tests__/resource-demand.test.ts` → **5/5 passed**
- `npm run check` → **4266 files, 0 errors, 0 warnings**
- `npm run lint` → clean
- `npm run build` → built in 6.63s (only the pre-existing chunk-size advisory)

**Exact-head CI** (`gh api repos/rigplane/rigplane-core/commits/73dd5104/check-runs`): `quick` success 1m35s (run 30303942536, includes the full frontend vitest suite), `grep-gate` success 5s (run 30303942796). `Agent Review Gate` fails solely for the missing PASS comment on this head.

## Non-blocking (please address or explicitly defer)

1. **Session-bounded ledger leaks a recycled handle identity** (`resource-demand.ts:99-103`). Probe F8: after `POOLED` is adopted and its stop acknowledged, a later *orphan* carrying the same identity gets **no** dispose — the ledger cannot distinguish "already cleaned" from "a new instance that reuses this identity". Re-adoption via a real start still works and still gets its own stop, so this is benign for object handles (identity is never recycled) but real for pooled/string handles. This is a direct, documented consequence of the contract the PR states, so I am not blocking on it — please just confirm it is intended and keep it visible for MOR-1128.
2. **Operations are mutable and alias private state** — `resource-demand.ts:123-127`. `acquire` freezes the lease (line 37) but `operation()` does not freeze the operation, and the object in `state.pending` is the same reference returned by `takeOperations()` and exposed via `snapshot().pending`. Fails closed, but `Object.freeze` would make it symmetric with the lease at no cost. Carried over from both prior reviews, still open.
3. **A `dispose` deferred behind a start that never resolves is retained indefinitely** (`resource-demand.ts:57`) — acknowledged MOR-1128 scope; flagging so it is not lost.

## Required to unblock

1. Fix Finding 1 so cleanup suppression is keyed on adoption ownership, making it order- and drain-independent in **both** directions.
2. Add the two regression orderings under **Test gap**.
3. Push, let `quick` + `grep-gate` go green on the new head, and request a fresh review.

Retry, terminal teardown, session finalization, failure/unavailable hardening, and consumer integration correctly remain with MOR-1128 and later children of MOR-1056 — not asked for here.

Leaving the PR in Draft.

morozsm:
Remediation pushed for the same-batch singleton alias cleanup ordering.

**Exact head:** `d1de5d794672ed40dbe7eaa01a0b7928a04cabfe`

### Change
- track adoption ownership separately from the first cleanup claim;
- suppress a queued orphan `dispose` by adopted ownership even after `reconcile` clears `activeHandle` to issue `stop`;
- preserve independent cleanup identity by `(resource, Object.is(handle))`.

### Regression evidence
The focused suite now covers both required stale-before-adoption orders:
1. stale completion → live adoption → final release → one drain;
2. stale completion → repeated drains while live start is pending → live adoption → final release.

Both emit only `stop` for the adopted singleton handle.

### Validation
- focused Vitest: 6/6 passed;
- `npm run check`: 0 errors, 0 warnings;
- `npm run lint`: clean;
- `npm run build`: passed (existing chunk-size advisory only);
- `git diff --check`: clean;
- no production consumers/imports found;
- cumulative scope against the PR merge-base: exactly 2 files, 200 net LOC (129 model + 71 focused tests);
- exact-head `grep-gate`: green;
- exact-head `quick`: green (run 30307166845).

PR remains Draft. Please run a fresh independent Claude Opus 5 review at effort high against exact head `d1de5d794672ed40dbe7eaa01a0b7928a04cabfe`. This author comment makes no review verdict.

morozsm:
Agent Review: PASS

**Head reviewed:** `d1de5d794672ed40dbe7eaa01a0b7928a04cabfe` — verified live (`gh pr view 2004 --json headRefOid`) against local `git rev-parse HEAD` both before and after the review; clean detached worktree, `git status --short` empty at start and end. Merge-base with `origin/main` = `53b8e5506132b5e3113927779f498a6dc22b3017`.

**Reviewer:** Claude Opus 5, effort high. Fresh and independent — I did not author this change, either remediation, or any prior review on this PR. I re-derived the invariants from the source rather than relying on the builder's regression evidence.

---

## Verdict

The `adoptedLedger` remediation genuinely closes the blocking defect from `issuecomment-5096777546`. I reproduced the fix independently rather than trusting the claim: the emit-time suppression is now a property of **adoption ownership** (`resource-demand.ts:56`, backed by the push at `:74`) instead of the transient `state.activeHandle`, which `reconcile` nullifies at `:110` the instant it issues the `stop`. That was the exact hole. It is closed in both directions and it no longer depends on drain placement.

The two ledgers are complementary, not a second authority: `cleanupLedger` (`:99-103`) is the *claim-time* dedupe ("this identity already had a cleanup claim"), `adoptedLedger` is the *emit-time* ownership check ("a `stop` owns this identity"). Every adopted handle is necessarily also in `cleanupLedger` (the adopt branch at `:73` claims first), so `cleanupLedger` is a strict superset and the two can never disagree. Completion authority is single-sourced per operation kind: reference identity (`issuedStarts` WeakSet + `state.pending !== op` for starts, `state.pending !== op` for stops), with `sessionEpoch` only as a cheap secondary filter — never as the authority. Cross-instance forgery is rejected on all three entry points.

No blocking findings.

## Scope enforcement

| Guard | Required | Actual | |
|---|---|---|---|
| Files | ≤ 3 | **2** | ✅ |
| Net LOC | ≤ 200 | **+200 / -0** | ✅ (exactly at the limit) |
| Production consumers | none | **0** | ✅ |
| Second authority | none | none | ✅ |
| Scope | Foundation-only | Foundation-only | ✅ |

`git diff --numstat 53b8e550...d1de5d79` → `129 0 frontend/src/lib/runtime/resource-demand.ts`, `71 0 frontend/src/lib/runtime/__tests__/resource-demand.test.ts`.

Consumer guard: `grep -rn "resource-demand\|ResourceDemand" src frontend/src docs .github --exclude-dir=node_modules`, excluding the two PR files → **no matches**. Nothing imports the model; no runtime, Svelte, channel, audio, transport, TX, retry, teardown, or session-finalization consumer is introduced.

## Commands and results

Run in the detached worktree at the exact head (`frontend/`):

| Command | Result |
|---|---|
| `npx vitest run src/lib/runtime/__tests__/resource-demand.test.ts` | **6/6 passed** (431 ms) |
| `npm run check` | **0 errors, 0 warnings**, 4266 files |
| `npm run lint` | **clean** |
| `npm run build` | **passed** (pre-existing chunk-size advisory only) |
| `git diff --check origin/main...HEAD` | **clean** |
| exact-head CI `quick` | **success**, `sha=d1de5d79…` (run 30307166845) |
| exact-head CI `grep-gate` | **success**, `sha=d1de5d79…` (run 30307166846) |

No test regenerated any fixture; the worktree needed no restoration and is clean.

## Bounded adversarial probes — 56/56 passed

Probes ran **outside tracked source**, against the committed blob (`git show HEAD:…/resource-demand.ts`, SHA-256 verified byte-identical to the worktree file: `22e7747a6635…`), transpiled with `esbuild` and executed as plain Node. Each run was bounded twice: an in-process 20 s watchdog (`setTimeout` → `process.exit(99)`) and an external `perl -e 'alarm 30; exec @ARGV' node probe.mjs`. No unbounded probe pattern was used.

The harness captures the pending start op **by reference** via `snapshot(r).pending`, so it can construct genuinely *no-drain* interleavings — the previous reviews' drain-discipline concern cannot hide here.

**Group A — an adopted handle is cleaned only by `stop`, with no earlier or later `dispose`**
- **A1** stale completion → live adoption of the *same singleton*, **no drain between** → single batch is `stop(rx-audio,S)` only, zero `dispose`. This is the previously-blocking repro; it now passes.
- **A2** repeated drains *while the live start is pending* → the `dispose` is deferred across drains 1/2/3, then adoption → final batch is `stop` only.
- **A3** pure no-drain (never drained until the end) → `start, start, stop`, zero `dispose`.
- **A4** late stale completion *after* adoption of the same handle → no `dispose`.
- **A5** orphan completion arriving *after* the `stop` was drained **and acknowledged** → emits nothing; adoption ownership persists for the whole session.
- **A6** no `dispose` anywhere across 6 drains straddling the release.

**Group B — orphan `dispose` exactly once**
- **B1/B2** duplicate orphan identity, with and without an intermediate drain → exactly one `dispose`.
- **B3** three orphans of the same identity across **three separate drains** → one `dispose` total.
- **B4** distinct orphan identities → one `dispose` each.
- **B5** equal handle under `rx-audio` and `hardware-scope` → independent, one `dispose` per resource.
- **B6** adoption under resource X does **not** suppress an orphan `dispose` under resource Y holding an equal handle.
- **B7** deferred `dispose` is not lost when the pending start **fails** — released on the next drain, `health = 'failed'`.
- **B8** deferred `dispose` released exactly once when the live start adopts a **different** handle (verified across three consecutive drains).

**Group C — operation / lease identity (15 probes)**
Forged start copy, forged stop copy, structural lease copy, cross-instance same-epoch lease / start / stop, wrong-epoch op, double `completeStart`, double `completeStop`, duplicate `release` — **all rejected**. Genuine op/lease accepted; leases frozen and distinct per `acquire`.

**Group D — reconcile / demand semantics**
Unavailable+selected issues no start; available+unselected issues no start; replacement overlap emits nothing and keeps the handle active (no transient stop); `configure` reads only `available`/`selected` and ignores injected `demand`/`activeHandle`/`health`/`pending`; a stale `stop` superseded by a re-acquire is no longer acknowledgeable and leaks no `dispose`.

**Group E — `Object.is` edge cases**
`NaN`/`NaN` dedupe to one `dispose` (where `===` would not); `+0`/`-0` stay **distinct** identities (where `===` would conflate them); structurally-equal distinct objects dispose separately; the same object reference dedupes.

## Full-suite regression check

The local full frontend run is **not** a regression signal on this machine, and I verified that rather than asserting it:

- at the PR head: `7 failed files | 81 failed / 2277 tests`
- at `origin/main` **without the PR files** (materialized non-mutatingly via `git archive origin/main frontend` into `/tmp` with `node_modules` symlinked): `9 failed files | 106 failed / 2271 tests`

The baseline fails **more**. Every failure is `localStorage` being `undefined` under this local jsdom install (`i18n/store`, `stores/layout`, `skins/registry`, `LanguageSelector`, `mod-input-auto`, …) — a pre-existing local environment issue, reproduced identically on a tree that does not contain this PR. The delta of exactly `2277 − 2271 = 6` tests is precisely the 6 new focused tests. Exact-head CI `quick` is green, which is the authoritative signal.

## Non-blocking observations

None of these block; all are consistent with the explicitly deferred hardening scope of #2001 ("excludes retry, terminal teardown/ended state, queued dispose behavior, advanced completion invalidation").

1. **`undefined` as a handle value is silently dropped** — `resource-demand.ts:108`/`:116` use `state.activeHandle !== undefined` as the "has active handle" predicate, and `State.activeHandle?: H` is an optional property. Probe F1: `completeStart(op, { handle: undefined })` returns `true` and adopts, but `reconcile` then never issues a `stop`, so the handle is never cleaned. Only reachable if a consumer instantiates `ResourceDemand<T | undefined>`. Worth constraining `H` (e.g. `H extends object`) or documenting the precondition when the first consumer lands.
2. **`failed` is sticky for the session** — probe F2: after a start error, `configure(false,false)` then `configure(true,true)` leaves `health = 'failed'` and issues no new start (`:116`). Correct given retry is out of scope, but it means `configure` cannot recover a resource; recovery belongs with the deferred retry work.
3. **`snapshot()` shares the live `pending` operation by reference** and `takeOperations()` hands out unfrozen operation objects (probe F3). The PR body already defers operation-freezing explicitly, and reference-identity checks remain load-bearing, so mutation is caller self-harm only. Fine to leave to Hardening.
4. **A deferred `dispose` is retained indefinitely** if a pending start never completes at all. Bounded by the documented session lifetime.
5. **Both ledgers are O(n) linear scans** that grow one entry per distinct adopted/orphan-cleaned identity per session — intentional and documented in the PR body.
6. **Stale PR-body reference:** the "Cleanup ledger contract" section still reads `Exact head: 73dd51041a4638693dd72923be0fb8022406d8ef`, which predates this remediation. Cosmetic; worth correcting for the record.

## MOR-1129 / #2001 acceptance

Every acceptance criterion in #2001 is met and independently verified:

- opaque unique leases — branded `ResourceLease`, frozen, distinct per `acquire` (C13–C15) ✅
- first / shared / last demand declarations (D3) ✅
- duplicate and unknown release no-op (C2, C3, C12) ✅
- unavailable selection does not start (D1, D2) ✅
- replacement overlap avoids a transient stop (D3) ✅
- focused test, `check`, `lint`, `build`, file/LOC guards, exact-head CI all pass ✅
- PR references #2001 and does **not** close #1994 — body states `Closes #2001` with "umbrella #1994 remains open" ✅

**MOR-1129 / #2001 can close on merge.** Umbrella #1994 must stay open, and the deferred items above belong to the Hardening issue that #2001 blocks.

**One process note, not a review objection:** this PR is still Draft, so per `AGENTS.md` it must not merge as-is. The author should run `gh pr ready` and let the `Agent Review Gate` pick up this PASS for the current head. I did not mark it ready, approve through the GitHub review API, push, merge, retarget, or touch #2006. No hardware evidence is claimed anywhere in this review — all findings are static analysis plus bounded in-process probes.

Issue #1953, opened 2026-07-27
Bind managed provider generation and authoritative PTT read lifecycle


morozsm:
Linear: https://linear.app/morozsm/issue/MOR-1044/core-bind-managed-provider-generation-and-explicit-authoritative-ptt
Parent prerequisite: MOR-1043
Prerequisites merged: #1945 / PR #1949 and #1947 / PR #1950.

## Summary

Add the private provider lifecycle seam needed between `ManagedRadioRuntime` and the authoritative IC-7610 PTT adapter. The host remains the sole provider-generation owner while later effect execution can bind observations, request a fresh PTT read, and invalidate a stuck generation without mutating the supervisor directly.

## Scope

Exactly these files:

- `src/rigplane/runtime/managed_radio_runtime.py`
- `src/rigplane/runtime/radio.py`
- new `tests/test_managed_provider_tx_binding.py`

Maximum 200 cumulative net lines. No effect executor, SDK/Web/rigctld routing, frontend, public raw-provider API, or unrelated refactor.

## Acceptance

- [ ] Every replacement, including an effectless first replacement, binds the exact host-owned generation before readiness can admit ON.
- [ ] Observer bind/unbind stays private and generation-bound; stale callbacks after replacement cannot affect current authority.
- [ ] A managed explicit fresh PTT read sends the actual CI-V PTT query through the existing serialized request mechanism; cache/write/ACK truth is not used.
- [ ] Current-generation invalidation/reset is coordinated by the host and advances host/supervisor generation together; stale invalidation is rejected.
- [ ] Replacement, readiness, request, and shutdown behavior from #1945 remains compatible.
- [ ] No raw provider, observer callback, or bare `set_ptt` is exposed to managed ingress.
- [ ] Focused tests cover effectless binding, generation continuity, fresh read, stale callback/read, and invalidation consistency.

## Compatibility

- Public `Radio`, CLI, config, rigctld wire, and docs: unchanged.
- Existing bare-provider `set_ptt(bool)`: unchanged and remains unmanaged/internal-facing.
- `ManagedRadioRuntime` is an internal target host; existing constructor behavior remains compatible when no lifecycle binding is supplied.

## Verification

Focused tests, existing managed-runtime and Icom authority tests, full non-integration pytest, Ruff/format, changed-source mypy, import-linter, and diff guard.

morozsm:
[...]

PR #2005, opened 2026-07-27
MOR-1121: serialize fresh PTT reads and terminal shutdown

### LINKED ISSUE #2003: MOR-1121: serialize fresh PTT reads and make shutdown terminal
Linear: MOR-1121
Parent: MOR-1104 / #1953
Depends on merged #1997.

## Outcome

Serialize authoritative fresh PTT reads against managed provider lifecycle mutation and make shutdown permanently terminal while preserving bounded de-key/provider release behavior.

## Acceptance

- [ ] `request_fresh_ptt` accepts only the exact current bound generation and passes the exact generation/observer through the typed CoreRadio port.
- [ ] Fresh-read await serializes with replacement, invalidation, and shutdown; cancellation releases the lane without publishing current-generation truth.
- [ ] Pre-bind, post-invalidate, and post-shutdown reads return not-ready.
- [ ] Shutdown unbinds, advances host generation, and permanently rejects later replace/read/readiness/ON and old callbacks.
- [ ] Transport/source epoch loss cannot leave the managed host bound/ready or publish stale PTT.
- [ ] Bind, unbind, read exceptions and cancellation fail closed.
- [ ] Shutdown bounds de-key service and always releases/unbinds the provider, with exactly-once physical release attempts when possible and explicit behavior when the provider is dead.
- [ ] Red-before adversarial terminal/A→B→A/in-flight cancellation tests, focused/full non-integration tests, Ruff/format, strict changed mypy, import-linter, diff guard, exact-head CI, and independent review pass.

## Scope

Exactly:

Maximum 200 net LOC relative to merge `53b8e5506132b5e3113927779f498a6dc22b3017`.

No CI-V token/observer changes, executor lane/deadlines, SDK, Web/frontend, public API, or raw provider exposure.

morozsm:
## Summary

- serialize authoritative fresh PTT reads with provider replacement, invalidation, and shutdown
- invalidate binding/readiness and advance host generation on read failure or cancellation
- make shutdown permanently terminal and idempotent
- unbind, invalidate, and release the provider after one bounded de-key attempt

## Why

MOR-1120 established one managed provider-generation authority but intentionally left two lifecycle boundaries open: in-flight authoritative reads could race lifecycle mutation, and shutdown did not invalidate or permanently close the binding. This closes that bounded remainder without changing CI-V token/observer internals or introducing an executor.

Within the allowed scope, authoritative-read failure/cancellation is the host-visible transport/source-epoch loss signal. It now fails closed before the lifecycle lane is released. A dead provider is never reported as physically de-keyed: service/release errors still surface after terminal cleanup.

## Verification

- red-before: 14 existing tests passed; 3 new tests failed on missing fresh-read and terminal behavior
- `uv run pytest tests/test_managed_radio_runtime.py tests/test_tx_safety.py tests/test_radio.py -q -n auto --tb=short` — 316 passed
- `uv run pytest tests/ --ignore=tests/integration -n auto --tb=short --timeout=300 --timeout-method=thread -q` — 8438 passed, 12 skipped, 11 xfailed, 1 xpassed
- `uv run ruff check src/ tests/`
- `uv run ruff format --check src/ tests/`
- strict mypy for both changed files
- `git diff --check`

One unrelated load-sensitive `TestCW.test_send_cw_long_text` CI-V timeout occurred in the first full run. It passed three times in isolation and in the clean full rerun.

## Scope

- 2 files
- 199 net LOC
- no CI-V token/observer changes, executor lane/deadlines, SDK, Web/frontend, or public API changes

Linear: MOR-1121

Closes #2003

### DIFF
--- a/src/rigplane/runtime/managed_radio_runtime.py
+++ b/src/rigplane/runtime/managed_radio_runtime.py
@@ -54,6 +54,10 @@ class ManagedRadioRuntime:
         self._provider_generation = 0
         self._bound_generation: int | None = None
         self._provider_state: _ProviderLifecycleState = "unbound"
+        self._lifecycle_lock = asyncio.Lock()
+        self._observation_version = 0
+        self._terminal = False
+        self._shutdown_transition: TxTransition | None = None
         self._shutdown_timeout = shutdown_timeout_seconds
 
     @property
@@ -69,12 +73,15 @@ class ManagedRadioRuntime:
 
     def _observe_ptt(self, observation: ProviderPttObservation) -> None:
         if (
-            self._provider_state == "bound"
+            not self._terminal
+            and self._provider_state == "bound"
             and observation.provider_generation
             == self._bound_generation
             == self._provider_generation
         ):
-            self._tx_safety.observe_ptt(observation)
+            transition = self._tx_safety.observe_ptt(observation)
+            if transition.outcome is TxOutcome.APPLIED:
+                self._observation_version += 1
 
     def _unbind(self) -> BaseException | None:
         lifecycle = self._provider_lifecycle
@@ -88,46 +95,78 @@ class ManagedRadioRuntime:
             return exc
         return None
 
-    async def replace_provider(self, *, ready: bool) -> TxTransition:
-        unbind_error = self._unbind()
-        self._provider_generation += 1
-        generation = self._provider_generation
-        transition = self._tx_safety.replace_provider(generation, ready=False)
-        self._provider_state = "unbound"
-        if unbind_error is not None:
-            raise unbind_error
-        lifecycle = self._provider_lifecycle
-        if lifecycle is None:
-            return transition
-        try:
-            lifecycle._bind_authoritative_ptt_observer(
-                provider_generation=generation, observer=self._observe_ptt
-            )
-        except BaseException:
-            try:
-                lifecycle._unbind_authoritative_ptt_observer()
-            except BaseException:
-                pass
-            raise
-        self._bound_generation = generation
-        self._provider_state = "bound"
-        if ready:
-            transition = self._tx_safety.set_provider_ready(generation, ready=True)
-        await self._service_effects(transition)
-        return transition
-
-    async def invalidate_provider(self, provider_generation: int) -> TxTransition:
-        if provider_generation != self._provider_generation:
-            return TxTransition(TxOutcome.STALE, self.tx_snapshot)
+    def _advance_provider(self) -> tuple[TxTransition, BaseException | None]:
         unbind_error = self._unbind()
         self._provider_generation += 1
         transition = self._tx_safety.replace_provider(
             self._provider_generation, ready=False
         )
         self._provider_state = "unbound"
-        if unbind_error is not None:
-            raise unbind_error
-        return transition
+        return transition, unbind_error
+
+    async def replace_provider(self, *, ready: bool) -> TxTransition:
+        async with self._lifecycle_lock:
+            if self._terminal:
+                return self._not_ready()
+            transition, unbind_error = self._advance_provider()
+            if unbind_error is not None:
+                raise unbind_error
+            lifecycle = self._provider_lifecycle
+            if lifecycle is None:
+                return transition
+            try:
+                lifecycle._bind_authoritative_ptt_observer(
+                    provider_generation=self._provider_generation,
+                    observer=self._observe_ptt,
+                )
+            except BaseException:
+                try:
+                    lifecycle._unbind_authoritative_ptt_observer()
+                except BaseException:
+                    pass
+                raise
+            self._bound_generation = self._provider_generation
+            self._provider_state = "bound"
+            if ready:
+                transition = self._tx_safety.set_provider_ready(
+                    self._provider_generation, ready=True
+                )
+            await self._service_effects(transition)
+            return transition
+
+    async def invalidate_provider(self, provider_generation: int) -> TxTransition:
+        async with self._lifecycle_lock:
+            if self._terminal:
+                return self._not_ready()
+            if provider_generation != self._provider_generation:
+                return TxTransition(TxOutcome.STALE, self.tx_snapshot)
+            transition, unbind_error = self._advance_provider()
+            if unbind_error is not None:
+                raise unbind_error
+            return transition
+
+    async def request_fresh_ptt(self) -> TxTransition:
+        async with self._lifecycle_lock:
+            lifecycle = self._provider_lifecycle
+            if (
+                self._terminal
+                or lifecycle is None
+                or self._provider_state != "bound"
+                or self._bound_generation != self._provider_generation
+            ):
+                return self._not_ready()
+            version = self._observation_version
+            try:
+                await lifecycle._request_authoritative_ptt_read(
+                    provider_generation=self._provider_generation,
+                    observer=self._observe_ptt,
+                )
+                if self._observation_version == version:
+                    raise RuntimeError("provider returned no authoritative PTT")
+            except BaseException:
+                self._advance_provider()
+                raise
+            return TxTransition(TxOutcome.APPLIED, self.tx_snapshot)
 
     async def set_provider_ready(self, *, ready: bool) -> TxTransition:
         if ready and (
@@ -170,17 +209,31 @@ class ManagedRadioRuntime:
         return transition
 
     async def shutdown(self, *, release_provider: ProviderRelease) -> TxTransition:
-        transition = self._tx_safety.emergency_release(
-            reason=TxReleaseReason.SERVER_SHUTDOWN
-        )
-        try:
-            if transition.outcome is not TxOutcome.NOOP:
-                await asyncio.wait_for(
-                    self._service_effects(transition),
-                    timeout=self._shutdown_timeout,
-                )
-        except TimeoutError:
-            pass
-        finally:
-            await release_provider()
-        return transition
+        async with self._lifecycle_lock:
+            if self._shutdown_transition is not None:
+                return self._shutdown_transition
+            self._terminal = True
+            transition = self._tx_safety.emergency_release(
+                reason=TxReleaseReason.SERVER_SHUTDOWN
+            )
+            self._shutdown_transition = transition
+            error: BaseException | None = None
+            try:
+                if transition.outcome is not TxOutcome.NOOP:
+                    await asyncio.wait_for(
+                        self._service_effects(transition),
+                        timeout=self._shutdown_timeout,
+                    )
+            except TimeoutError:
+                pass
+            except BaseException as exc:
+                error = exc
+            _, unbind_error = self._advance_provider()
+            try:
+                await release_provider()
+            except BaseException as exc:
+                error = error or exc
+            error = error or unbind_error
+            if error is not None:
+                raise error
+            return transition
--- a/src/rigplane/runtime/managed_radio_runtime.py
+++ b/src/rigplane/runtime/managed_radio_runtime.py
@@ -54,6 +54,7 @@ class ManagedRadioRuntime:
         self._provider_generation = 0
         self._bound_generation: int | None = None
         self._provider_state: _ProviderLifecycleState = "unbound"
+        self._provider_observer = self._observe_ptt
         self._lifecycle_lock = asyncio.Lock()
         self._observation_version = 0
         self._terminal = False
@@ -114,10 +115,11 @@ class ManagedRadioRuntime:
             lifecycle = self._provider_lifecycle
             if lifecycle is None:
                 return transition
+            self._provider_observer = self._observe_ptt
             try:
                 lifecycle._bind_authoritative_ptt_observer(
                     provider_generation=self._provider_generation,
-                    observer=self._observe_ptt,
+                    observer=self._provider_observer,
                 )
             except BaseException:
                 try:
@@ -159,7 +161,7 @@ class ManagedRadioRuntime:
             try:
                 await lifecycle._request_authoritative_ptt_read(
                     provider_generation=self._provider_generation,
-                    observer=self._observe_ptt,
+                    observer=self._provider_observer,
                 )
                 if self._observation_version == version:
                     raise RuntimeError("provider returned no authoritative PTT")
@@ -212,7 +214,6 @@ class ManagedRadioRuntime:
         async with self._lifecycle_lock:
             if self._shutdown_transition is not None:
                 return self._shutdown_transition
-            self._terminal = True
             transition = self._tx_safety.emergency_release(
                 reason=TxReleaseReason.SERVER_SHUTDOWN
             )
@@ -228,6 +229,7 @@ class ManagedRadioRuntime:
                 pass
             except BaseException as exc:
                 error = exc
+            self._terminal = True
             _, unbind_error = self._advance_provider()
             try:
                 await release_provider()
--- a/tests/test_managed_radio_runtime.py
+++ b/tests/test_managed_radio_runtime.py
@@ -30,6 +30,12 @@ class ProviderLifecycle:
         self.bindings: list[tuple[int, Callable[[ProviderPttObservation], None]]] = []
         self.current: tuple[int, Callable[[ProviderPttObservation], None]] | None = None
         self.bind_error = self.unbind_error = False
+        self.unbinds = 0
+        self.reads: list[tuple[int, Callable[[ProviderPttObservation], None]]] = []
+        self.read_started = asyncio.Event()
+        self.read_release = asyncio.Event()
+        self.read_blocked = False
+        self.read_error: BaseException | None = None
 
     def _bind_authoritative_ptt_observer(
         self,
@@ -43,6 +49,7 @@ class ProviderLifecycle:
             raise RuntimeError("bind failed")
 
     def _unbind_authoritative_ptt_observer(self) -> None:
+        self.unbinds += 1
         if self.unbind_error:
             raise RuntimeError("unbind failed")
         self.current = None
@@ -54,6 +61,12 @@ class ProviderLifecycle:
         observer: Callable[[ProviderPttObservation], None],
     ) -> None:
         assert self.current == (provider_generation, observer)
+        self.reads.append((provider_generation, observer))
+        self.read_started.set()
+        if self.read_blocked:
+            await self.read_release.wait()
+        if self.read_error is not None:
+            raise self.read_error
         observer(ProviderPttObservation(RadioTx.OFF, provider_generation, 1, 10.0))
 
 
@@ -67,6 +80,10 @@ def ids() -> Iterator[str]:
 async def no_effects(
     _supervisor: TxSafetySupervisor, _transition: TxTransition
 ) -> None:
+    pass
+
+
+async def no_release() -> None:
     pass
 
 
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: BLOCKED

**Head reviewed:** `8a1a79584b8763c1e0e9fcb20dce391c0c00d492` (base `53b8e5506132b5e3113927779f498a6dc22b3017`)
**Reviewer:** independent, did not author this branch.

Scope, static checks and the whole suite are clean and every claim in the PR body reproduced exactly (evidence at the bottom). Two defects survive that the suite cannot see, because the new fake's authority check is weaker than the production one.

---

## 1. BLOCKER — `request_fresh_ptt()` can never succeed against a real `CoreRadio`, and destroys a healthy provider binding on every call

**Severity:** high (feature inoperative + self-inflicted loss of TX authority)
**Where:** `src/rigplane/runtime/managed_radio_runtime.py:162` vs `:120`, against `src/rigplane/runtime/_civ_rx.py:644`

`replace_provider()` binds with `observer=self._observe_ptt` (line 120) and `request_fresh_ptt()` reads with `observer=self._observe_ptt` (line 162). Attribute access on a method produces a **new bound-method object every time**: the two are `==` but never `is`.

The CI-V authority check is identity-based:

```python
# src/rigplane/runtime/_civ_rx.py:644
and self._ptt_observer is read.observer
```

So `_ptt_read_is_current()` is always `False`, `request_authoritative_ptt_read()` returns `False` at `_civ_rx.py:619-620` **before any CI-V frame is dispatched**, and `CoreRadio._request_authoritative_ptt_read` raises `CommandError("PTT response was not authoritative for this request")` (`radio.py:2979`). `request_fresh_ptt`'s `except BaseException` at line 166 then calls `_advance_provider()` — tearing down a provider that was perfectly healthy.

This is not a hypothetical reading of the contract: **MOR-1120's own test asserts this exact rejection** — `tests/test_radio.py:890-904` (`test_fresh_ptt_read_rejects_rebind_and_equal_distinct_observer`) constructs an `EqualObserver` whose `__eq__` returns `True` and requires `CommandError`. The production composition added here is precisely that case.

**Repro** (real `ManagedRadioRuntime` + real `IcomRadio`, healthy radio with a queued PTT response):

```python
rt = ManagedRadioRuntime("radio-a", service=no_effects, provider_lifecycle=radio)
await rt.replace_provider(ready=True)
await _route_ptt(radio, b"\x00")           # seed authoritative OFF
transport.queue_response(_ptt_response(False))   # radio WILL answer
await rt.request_fresh_ptt()
```

Observed:

```
error   = CommandError: PTT response was not authoritative for this request
civ frames sent by the read = 0
provider_generation : 1 -> 2      (healthy provider invalidated)
provider_ready      : True -> False
radio_tx            : off -> unknown
civ observer bound  : False
```

Control, identical flow with one pinned observer object (`stable = rt._observe_ptt; rt._observe_ptt = stable`): `outcome=applied`, generation stays `1`. That isolates observer identity as the sole cause.

Consequence beyond "feature does not work": every call costs a full re-bind, and if a lease is live the runtime drops into `replace_provider(gen+1, ready=False)`, where `_service_release(force=False)` returns no effects because `_ready` is now `False` — so the pending release cannot be serviced until a new provider becomes ready.

**Why the suite is green:** the fake asserts with `==`:

```python
# tests/test_managed_radio_runtime.py:63
assert self.current == (provider_generation, observer)
```

Bound methods compare equal, so the fake accepts exactly what `_civ_rx.py:644` rejects. This is the "no permissive fake / tests must not mirror the implementation" criterion failing in substance even though no `MagicMock` is used.

**Required fix:** hold one stable observer reference for the lifetime of a binding — e.g. bind `self._observer = self._observe_ptt` once in `__init__` (or per binding) and pass `self._observer` at both line 120 and line 162. Then tighten the fake to `is` (`assert self.current is not None and self.current[0] == provider_generation and self.current[1] is observer`) so the test can no longer pass while production fails.

---

## 2. BLOCKER — `shutdown()` can never confirm a de-key it actually performed

**Severity:** medium (fails safe, but every clean shutdown reports a FAULTED unconfirmed release and burns the full timeout)
**Where:** `src/rigplane/runtime/managed_radio_runtime.py:215` interacting with `:76`

`shutdown()` sets `self._terminal = True` at line 215 **before** awaiting the bounded de-key at lines 221-226. But `_terminal` is the same flag the new guard at line 76 uses to reject observations. So during the de-key the only channel that can cross the supervisor's release boundary — `_observe_ptt` → `TxSafetySupervisor.observe_ptt` → `_clear_managed()` — is already severed. `settle_attempt` then records `read_ptt_unconfirmed`.

**Repro** — identical de-key driven two ways through the same executor stand-in (WRITE_OFF, radio OFF readback via the bound observer, settle):

| arm | phase | lease_id | radio_tx | radio actually keyed |
|---|---|---|---|---|
| `request_off()` (control, no `_terminal`) | `idle` | `None` | `off` | `False` |
| `shutdown()` | `faulted` | held | `unknown` | `False` |

`release_last_error='read_ptt_unconfirmed'`. The radio *was* de-keyed; shutdown just cannot see it. Against the acceptance wording, shutdown never falsely claims de-key (good), but it also cannot service "bounded de-key/release exactly once **when possible**" — confirmation is impossible by construction, not merely unlucky.

**Required fix:** the `not self._terminal` clause at line 76 is redundant for its stated purpose. Verified post-shutdown state:

```
_provider_state    = 'unbound'
_bound_generation  = None
_provider_generation = 2
```

`_advance_provider()` (line 231) already sets `_provider_state = "unbound"` and clears `_bound_generation`, so the pre-existing `self._provider_state == "bound"` guard rejects every late callback on its own — `test_shutdown_is_terminal_and_cleans_up_exactly_once` still holds without the `_terminal` clause. Either drop `not self._terminal` from `_observe_ptt`, or move `self._terminal = True` to just before `_advance_provider()` so the bounded de-key can still confirm. Add a test that drives a real WRITE_OFF + OFF readback through `shutdown()` and asserts the lease clears.

---

## Minor (non-blocking, listed for the author's call)

- `managed_radio_runtime.py:167` — `self._advance_provider()` discards the returned `unbind_error`. Defensible (the read error is more relevant, and state still fails closed) but silent.
- `managed_radio_runtime.py:213` — `_shutdown_transition` is memoized before the work runs, so a `shutdown()` that raised will, on a retry, return that transition without re-raising. Idempotent as specified; just note a retrying caller sees the second call as error-free.

## Closure assessment

**The merge would not complete the MOR-1104 / MOR-1044 roll-up.** MOR-1120 deliberately left two boundaries open; this PR closes the *serialization* half correctly — the single `asyncio.Lock` genuinely serializes read/replace/invalidate/shutdown, generations are strictly monotonic so A→B→A and old-completion callbacks cannot publish, cancellation is re-raised honestly after synchronous invalidation, and repeated shutdown unbinds/releases exactly once. But the *fresh authoritative read* half is inoperative against the only production provider (#1), so the transport/source-epoch-loss signal the design leans on is unreachable, and terminal shutdown cannot confirm its own de-key (#2). The lifecycle gap remains unowned.

## What I ran (all at `8a1a7958`, worktree clean)

- `uv run pytest tests/test_managed_radio_runtime.py tests/test_tx_safety.py tests/test_radio.py -q` — **316 passed**
- `uv run pytest tests/ --ignore=tests/integration -q -x -n auto --timeout=300` — **8438 passed, 12 skipped, 11 xfailed, 1 xpassed** (no flake seen; the `TestCW` timeout noted in the PR body did not recur)
- `uv run ruff check src/ tests/` — All checks passed
- `uv run ruff format --check src/ tests/` — 627 files already formatted
- `uv run mypy --strict src/rigplane/runtime/managed_radio_runtime.py` — Success
- `git diff --check` — clean
- CI at this exact head: `Tests (quick)` success, `Rebrand grep gate` success
- Scope verified: 2 files, +248/−49 = **199 net LOC**; no `_civ_rx` token/observer changes, no executor/SDK/Web/public-API changes — all within guardrails
- Adversarial probes (outside the worktree, tree untouched): production-composition fresh read, pinned-observer control, de-key confirmation under `shutdown()` vs `request_off()`, post-shutdown guard redundancy

## To re-run after the fix

1. `uv run pytest tests/test_managed_radio_runtime.py tests/test_radio.py tests/test_tx_safety.py -q`
2. A new test binding a real `IcomRadio` as `provider_lifecycle` and asserting `request_fresh_ptt()` returns `APPLIED` with the generation unchanged
3. A new test asserting `shutdown()` clears the lease when the executor confirms the de-key
4. `uv run pytest tests/ --ignore=tests/integration -q -n auto` + ruff + mypy --strict + lint-imports

Leaving the PR in Draft.

morozsm:
BLOCKED review remediation pushed at `186aca586dc0c7c4ec7943e3571fd9596a385580`.

## Fixed

1. `ManagedRadioRuntime` now pins one observer object per provider generation and passes that exact object to both bind and authoritative fresh read. The lifecycle fake checks observer identity with `is`.
2. Shutdown now keeps the bound observation channel open during the bounded de-key/readback service. The terminal latch is set immediately before atomic unbind/generation invalidation/provider release.

## New production-faithful evidence

- real `CoreRadio` + CI-V mock transport sends exactly one fresh-read frame
- decoded OFF is applied without changing generation 1 or clearing readiness
- shutdown WRITE_OFF + OFF/READ_PTT confirmation clears the lease before terminal invalidation
- repeated shutdown remains exact-once and late callbacks remain rejected

## Verification

- focused runtime/radio/TX: 315 passed
- full non-integration: 8437 passed, 12 skipped, 11 xfailed, 1 xpassed
- Ruff/format: PASS
- strict mypy for both changed files: PASS
- import-linter: 5/5 contracts kept
- cumulative scope from base: 2 files, 197 net LOC

Requesting a fresh independent Opus 5 max review for this exact head after CI completes. This is not a self-review or PASS claim.

morozsm:
Agent Review: PASS

**Head reviewed:** `186aca586dc0c7c4ec7943e3571fd9596a385580` (base `53b8e5506132b5e3113927779f498a6dc22b3017`)
**Reviewer:** independent replacement reviewer; did not author this branch. Prior review at `8a1a7958` inspected, then every claim independently re-derived rather than trusted.

Both blockers from the `8a1a7958` review are closed, and I confirmed each fix against the **real** `CoreRadio`/`CivRuntime` rather than the test fake. All custom probes were run under an explicit ≤30 s wrapper with `faulthandler.dump_traceback_later` armed and a per-probe `asyncio.wait_for`; nothing timed out.

---

## Prior reviewer's probe hang — diagnosed, not a product defect

`/tmp/probe_2005/probe_b.py` hung in `c3_replace_provider_cancelled_in_service_effects`, always after C2 (both `b.out` and `c.out` stop at the same point).

Root cause is in the probe, not the runtime: C3 called `replace_provider(ready=True)` on a runtime that **held no TX lease**, then blocked on `await entered.wait()`, where `entered` is set from inside the injected `service`. `_service_effects()` only calls the service `if transition.effects:` — and a lease-less `replace_provider` produces none, so the gate event was never set and the probe waited forever on its own Event.

Measured (`D1`):

```
leaseless_effects=()  service_calls_leaseless=0
with_lease_effects=1  with_lease_service_calls=1
```

Rewritten with a real lease acquired first, the same probe passes immediately (`C3` below). No production hang exists on this path.

---

## 1. Pinned observer identity — CLOSED, verified against real CI-V

`_provider_observer` is pinned once per binding (`managed_radio_runtime.py:57`, re-pinned at `:118`) and the *same object* is passed to both `_bind_authoritative_ptt_observer` (`:122`) and `_request_authoritative_ptt_read` (`:164`). That satisfies the identity authority check at `_civ_rx.py:644` (`self._ptt_observer is read.observer`).

Verified with a real `CoreRadio` + `MockTransport` (no fake lifecycle), plus the counterfactual that isolates the pin as the sole cause:

```
P1 real CoreRadio fresh read APPLIED with pinned observer
   outcome=APPLIED tx=OFF gen=1 ready=True bound_is_pinned=True sent=1
P2 counterfactual: un-pinned (== but not is) observer fails closed
   equal=True identical=False
   raised=CommandError: PTT response was not authoritative for this request
   gen=2 ready=False
P4 rebind re-pins observer and fresh read still authoritative
   repinned=True bound_is_second=True outcome=APPLIED gen=2
```

P2 re-creates the exact pre-fix defect on this head by assigning a fresh bound method (`== True`, `is False`) — it still fails closed, which proves the pin is load-bearing and not incidentally green. The fake was correctly tightened to `is` at `tests/test_managed_radio_runtime.py:67`, so the test can no longer pass while production fails, and `test_real_core_radio_fresh_read_preserves_healthy_generation` (`:280`) is the real-provider test the prior review asked for.

## 2. Terminal latch timing — CLOSED

`self._terminal = True` moved from before the bounded de-key to `:232`, after `asyncio.wait_for(self._service_effects(...))` and immediately before `_advance_provider()`. The de-key can now be confirmed through `_observe_ptt` → `observe_ptt` → `_clear_managed()`:

```
V1 confirmed physical-OFF de-key clears the lease at shutdown
   lease=None tx=UNKNOWN serviced=1 releases=1 outcome=ACCEPTED
```

I initially flagged a residual `lease_id` after shutdown and then disproved it as my own over-strict assertion: the lease is retained only when the de-key is never confirmed OFF, which is correct fail-closed behaviour, and every TX entry point still refuses:

```
V2 unconfirmed de-key retains the lease but every TX entry point refuses
   lease_retained=id-1 request_on=NOT_READY set_ready=NOT_READY
   replace=NOT_READY fresh=NOT_READY invalidate=NOT_READY tx=UNKNOWN
```

## Adversarial matrix — all green

| Probe | Result |
|---|---|
| `C3` cancel during `replace_provider` service effects | CancelledError raised, lock released, runtime still operable |
| `C3b` cancelled replace cannot permit a new key-up | `request_on=RELEASE_PENDING` (refused) |
| `C4` cancel `request_fresh_ptt` mid-read | invalidates synchronously, gen 1→2, ready=False, lock free, observer detached |
| `C5` cancel one lock waiter | next waiter is not stranded, lock free |
| `C6` cancel during shutdown de-key | terminal latched, unbind=1, release=1, CancelledError re-raised honestly |
| `S1` eight concurrent `shutdown()` | 1 de-key, 1 unbind, 1 release, all callers get the identical transition object |
| `S3`/`V3` late PTT=ON callback after shutdown | fully inert: gen unchanged, `_observation_version` unchanged, observer detached |
| `S2` PTT=ON inside the pre-terminal de-key window | cannot re-arm TX; ends `NOT_READY`/`UNKNOWN`/terminal |
| `S4` `release_provider` raises | error surfaces once; repeat shutdown returns cached transition, does not retry |
| `S5` unbind raises during shutdown | terminal latched, release still performed |
| `S6` de-key service raises | terminal latched, unbind + release still performed |
| `P3` real CoreRadio read with no response | fail closed, gen advanced, lane freed, subsequent read `NOT_READY` |
| `V4` two consecutive fresh reads of an unchanged value | both `APPLIED` (`_observation_version` 1→2) — the version gate does not mistake "no change" for "no answer" |
| `V5` provider answers with a stale generation | `RuntimeError: provider returned no authoritative PTT`, fails closed |

Serialization is structural: one `asyncio.Lock` covers `replace_provider` / `invalidate_provider` / `request_fresh_ptt` / `shutdown`, and no lock-order inversion exists (`_civ_rx._ptt_read_lock` is only ever acquired *under* `_lifecycle_lock`; `bind`/`unbind` are synchronous and take no lock).

## Non-blocking observations

- **`request_fresh_ptt` has zero production callers.** `grep -rn request_fresh_ptt src/ frontend/src` matches only the definition. The primitive is now correct but dormant — nothing in `web/`, `rigctld/`, or the effect executor calls it. This is consistent with the issue scope (executor/SDK/Web changes are explicitly excluded), so it is not a defect here, but it is the substantive reason the roll-up is not yet realised.
- **`shutdown()` lock acquisition is not bounded by `shutdown_timeout_seconds`.** Measured in `S7`: with a stuck provider read holding the lane, shutdown stayed blocked for the full 1.0 s I let it run despite `shutdown_timeout=0.05`, and completed only once the read was cancelled. In production this is bounded by `CoreRadio._civ_get_timeout = min(timeout, 2.0)` (`radio.py:857`), so worst case is ≈2 s + de-key timeout, and it is unreachable today given no callers. Worth a follow-up if a non-CoreRadio `ProviderTxLifecycle` ever appears.
- **`request_fresh_ptt:169` discards the `unbind_error`** returned by `_advance_provider()`. Fails closed and the read error is the more relevant one, but the unbind failure is silent. Carried over from the prior review; still accurate.
- **`_shutdown_transition` is memoised before the work runs** (`:220`), so a `shutdown()` that raised returns error-free on retry (`S4`: `first=RuntimeError: release failed`, `second_outcome=ACCEPTED`, `releases=1`). Exactly-once as specified; noting it for a retrying caller.
- **Two tests added in `8a1a7958` were removed in `186aca58`** (`test_fresh_read_serializes_terminal_shutdown`, `test_shutdown_unbind_error_still_releases_and_stays_terminal`). I checked the base: all 12 pre-existing tests survive and the PR net-adds 4, so **no merged coverage was lost** — the author revised their own unmerged tests, plausibly to stay inside the 200 net LOC cap (197 used). The consequence is that the *shutdown* and *invalidation* arms of acceptance bullet 2 now have no suite coverage; I verified both behaviours by probe (`S7`, `S5`), and the replacement arm is still covered at `:353`. Cheap follow-up, not a blocker.
- Worktree carries one unrelated dirty file, `frontend/src/lib/state/__tests__/fixtures/empty-store-field-status.json` (last touched by `bf1736e1`, MOR-558). Not part of this PR — `changedFiles` is 2 — and untouched by me.

## Closure assessment

**MOR-1121 / #2003 — closed.** All eight acceptance bullets are met. Fresh reads accept only the exact bound generation and pass the exact generation/observer through the typed CoreRadio port (P1/P4/V5); the await serialises with replacement, invalidation and shutdown behind one lock (S7 + structural); pre-bind/post-invalidate/post-shutdown reads return not-ready (V2); shutdown unbinds, advances the host generation and permanently rejects later replace/read/readiness/ON and old callbacks (V2/V3/S3); transport/source-epoch loss cannot leave the host bound or ready (P3/V5); bind/unbind/read exceptions and cancellation all fail closed (C4/C6/S5/S6); and the bounded de-key + exactly-once release holds under eight-way concurrency (S1). Residual gap is test coverage only, noted above.

**MOR-1104 / MOR-1044 roll-up (#1953) — not closed; keep open.** The two lifecycle boundaries MOR-1120 deliberately left open are both genuinely closed by this PR, which is the remainder the roll-up was waiting on. But #1953's own acceptance spans the `radio.py` seam and `tests/test_managed_provider_tx_binding.py`, and the capability delivered here is still dormant — no production path calls `request_fresh_ptt`. The roll-up should not be closed until a consumer exists; that wiring is out of scope for this issue by construction.

## What I ran (all at `186aca58`)

- `uv run pytest tests/test_managed_radio_runtime.py -q` — **18 passed**
- `uv run pytest tests/test_managed_radio_runtime.py tests/test_radio.py tests/test_tx_safety.py -q` — **315 passed**
- `uv run ruff check src/ tests/` — All checks passed
- `uv run ruff format --check src/ tests/` — 627 files already formatted
- `uv run mypy --strict src/rigplane/runtime/managed_radio_runtime.py` — Success
- `git diff --check` base→head — clean
- Scope: **2 files, +247/−50 = 197 net LOC** — within ≤3 files / ≤200 net LOC; no `_civ_rx` token/observer changes, no executor/SDK/Web/public-API changes
- CI at this exact head: `quick` success, `grep-gate` success, `Update Agent Review Gate status` success (`Agent Review Gate` red only for the missing PASS comment this review supplies)
- Adversarial probes in `/tmp/rv2005/` (outside the worktree; tree untouched): 4 real-CoreRadio identity probes incl. counterfactual, 6 cancellation probes + hang diagnosis, 7 shutdown/terminal/error-path probes, 5 adjudication probes — 22 checks, all green

Full non-integration suite was already green on this exact head; I did not re-run it.

Issue #2, opened 2026-02-26
feat(audio): RX high-level API (start_audio_rx_pcm)


morozsm:
## Why
Users need RX audio as PCM directly from the library (no manual Opus decode plumbing).

## Depends on
- #1

## Scope
- Add high-level API:
  - `start_audio_rx_pcm(callback, sample_rate=..., channels=...)`
- Callback receives PCM bytes (+ metadata)
- Graceful stop/cancel and robust task lifecycle

## Acceptance Criteria
- [ ] Callback receives PCM (not raw Opus)
- [ ] start/stop cycles are stable
- [ ] Callback exceptions do not break transport loop
- [ ] Tests cover normal stream + jitter + disconnect

PR #2014, opened 2026-07-28
Cut audio FFT over to presentation demand

### LINKED ISSUE #2013: Cut audio FFT over to App-owned demand with lifetime-neutral subscribers
Linear: MOR-1123

## Scope
Atomically move /api/v1/audio-scope lifetime from ScopeController frame subscriptions to the single App-session presentation resource host merged by #2008.

## Exact files
1. frontend/src/lib/runtime/scope-controller.svelte.ts
2. frontend/src/lib/runtime/__tests__/scope-controller.test.ts
3. frontend/src/components-v2/panels/audio-scope/AudioSpectrumPanel.svelte

Maximum 3 files and 200 net LOC. No App.svelte, registry, layout shell, hardware scope, or RX-audio files.

## Acceptance
- ScopeController remains the sole audio-FFT parser, broadcaster, and exact channel driver.
- subscribe/unsubscribe becomes lifetime-neutral.
- Panel owns one opaque audio-fft demand lease plus one frame subscription, released exactly once.
- First/shared/last demand opens once, shares the exact handle, and closes once.
- Late start, stale A-B-A completion, duplicate release, unavailable, failure, and final teardown are fail-closed and exact-handle safe.
- No hidden hardware-scope fallback and no transport import in the panel.
- Focused tests, check, lint, build, exact-head CI, and independent review pass.

Closes through Linear MOR-1123.

morozsm:
## Summary
- keep ScopeController as the sole `/api/v1/audio-scope` channel, parser, and broadcaster driver
- use a unique opaque token for every start while retaining the one memoized named `WsChannel`
- make frame subscribe/unsubscribe lifetime-neutral and give AudioSpectrumPanel one opaque `audio-fft` lease plus one subscription

## Control plane
- Linear: MOR-1123
- GitHub: Closes #2013

## Exact scope
- Base: `58bbf21eae6c9f7d1917e9fa96023b0d4329aa55`
- Head: `6b47a8c9c33e0b6df21d534786a359cc1bc71b06`
- 3 files, +186/-78, net +108 LOC
- no App.svelte, registry, layout, hardware scope, RX audio, resource-demand, resource-host, or adapter-factory changes

## Remediation of independent review BLOCKED
- Red-before on a production-shaped memoized channel: 4 failures/9 tests reproduced channel-handle aliasing, orphan leak, duplicate callback remount, and teardown-race publication
- ScopeController now maps each unique per-start token to the exact memoized channel callback/unsubscribe binding
- stop/dispose removes only the captured token binding and disconnects the shared physical channel only after its last binding is removed
- stale/orphan callbacks cannot publish because only the current token may broadcast frames

## Evidence
- Focused ScopeController: 10/10 passed
- Related ScopeController + PresentationResourceHost + ResourceDemand: 34/34 passed
- Full frontend fresh bounded run: 132 files, 2309/2309 passed
- One prior bounded aggregate run exposed the known `mod-input-tx-guard` order flake; its exact test passes 13/13 alone, and the final fresh full run is all green
- `npm run check`: 0 errors, 0 warnings
- `npm run lint`: passed
- `npm run build`: passed
- Diff guard: exactly 3 owned files, net +108 LOC; panel has no transport import or hardware-scope fallback

## Compatibility and hardware boundary
No public API, CLI, config, backend, or wire change. No real-radio or hardware acceptance was run or claimed; FTX-1 hardware acceptance remains outside this slice and remains a release gate.

Closes #2013

### DIFF
--- a/frontend/src/components-v2/panels/audio-scope/AudioSpectrumPanel.svelte
+++ b/frontend/src/components-v2/panels/audio-scope/AudioSpectrumPanel.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { runtime } from '$lib/runtime';
+  import { presentationResources, runtime } from '$lib/runtime/frontend-runtime';
   import { deriveAudioSpectrumProps } from '$lib/runtime/adapters/panel-adapters';
   import AudioSpectrumCanvas from './AudioSpectrumCanvas.svelte';
 
@@ -13,15 +13,24 @@
   let fftBandwidth = $state(48000);
   let fftPush: ((data: Uint8Array) => void) | null = null;
 
-  // Scope subscription — delegates lifecycle to ScopeController (ADR INV-2, INV-5)
+  // Scope frames and resource demand have separate lifetimes (ADR INV-2, INV-5).
   $effect(() => {
-    return runtime.scope.subscribe((frame) => {
+    runtime.scope.registerPresentationDriver(presentationResources);
+    const lease = presentationResources.acquire('audio-fft', 'AudioSpectrumPanel');
+    const unsubscribe = runtime.scope.subscribe((frame) => {
       fftPixels = frame.pixels;
       if (frame.endFreq > frame.startFreq) {
         fftBandwidth = frame.endFreq - frame.startFreq;
       }
       fftPush?.(frame.pixels);
     });
+    let released = false;
+    return () => {
+      if (released) return;
+      released = true;
+      unsubscribe();
+      presentationResources.release(lease);
+    };
   });
 </script>
 
--- a/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
+++ b/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
@@ -6,7 +6,10 @@
  */
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
 import { ScopeController } from '../scope-controller.svelte';
+import { PresentationResourceHost } from '../resource-host';
 import type { ScopeFrame } from '../scope-controller.svelte';
 
 // ── Helpers ──
@@ -50,40 +53,40 @@ describe('ScopeController', () => {
     ctrl = new ScopeController(() => channel as any);
   });
 
-  it('subscribe() opens the WS channel on first subscriber', () => {
+  it('keeps frame subscriptions lifetime-neutral', () => {
     expect(channel.connect).not.toHaveBeenCalled();
 
-    ctrl.subscribe(vi.fn());
+    const unsubscribe = ctrl.subscribe(vi.fn());
+    unsubscribe();
+    unsubscribe();
 
-    expect(channel.connect).toHaveBeenCalledTimes(1);
-    expect(channel.connect).toHaveBeenCalledWith('/api/v1/audio-scope');
-  });
-
-  it('second subscribe() reuses the existing channel without reconnecting', () => {
-    ctrl.subscribe(vi.fn());
-    ctrl.subscribe(vi.fn());
-
-    expect(channel.connect).toHaveBeenCalledTimes(1);
-    expect(channel.onBinary).toHaveBeenCalledTimes(1);
-  });
-
-  it('last unsubscribe() closes the channel', () => {
-    const unsub1 = ctrl.subscribe(vi.fn());
-    const unsub2 = ctrl.subscribe(vi.fn());
-
-    unsub1();
+    expect(channel.connect).not.toHaveBeenCalled();
     expect(channel.disconnect).not.toHaveBeenCalled();
-
-    unsub2();
-    expect(channel.disconnect).toHaveBeenCalledTimes(1);
   });
 
-  it('both subscribers receive parsed frames', () => {
+  it('registers one exact-handle driver for first/shared/last demand', async () => {
+    const host = new PresentationResourceHost<unknown>('session');
+    ctrl.registerPresentationDriver(host);
+    const first = host.acquire('audio-fft', 'first');
+    const shared = host.acquire('audio-fft', 'shared');
+    await vi.waitFor(() => expect(channel.connect).toHaveBeenCalledTimes(1));
+    expect(channel.connect).toHaveBeenCalledWith('/api/v1/audio-scope');
+    expect(host.snapshot('audio-fft').activeHandle).toBe(channel);
+
+    expect(host.release(first)).toBe(true);
+    expect(channel.disconnect).not.toHaveBeenCalled();
+    expect(host.release(shared)).toBe(true);
+    await vi.waitFor(() => expect(channel.disconnect).toHaveBeenCalledTimes(1));
+    expect(host.release(shared)).toBe(false);
+  });
+
+  it('both subscribers receive parsed frames', async () => {
     const h1 = vi.fn<(frame: ScopeFrame) => void>();
     const h2 = vi.fn<(frame: ScopeFrame) => void>();
 
     ctrl.subscribe(h1);
     ctrl.subscribe(h2);
+    await ctrl.audioFftDriver.start();
 
     channel._fire(makeScopeFrameBuffer());
 
@@ -95,12 +98,13 @@ describe('ScopeController', () => {
     expect(frame.endFreq).toBe(14_200_000);
   });
 
-  it('unsubscribed handler no longer receives frames', () => {
+  it('unsubscribed handler no longer receives frames', async () => {
     const h1 = vi.fn();
     const h2 = vi.fn();
 
     const unsub1 = ctrl.subscribe(h1);
     ctrl.subscribe(h2);
+    await ctrl.audioFftDriver.start();
 
     unsub1();
     channel._fire(makeScopeFrameBuffer());
@@ -109,9 +113,10 @@ describe('ScopeController', () => {
     expect(h2).toHaveBeenCalledTimes(1);
   });
 
-  it('ignores malformed frames (wrong magic byte)', () => {
+  it('ignores malformed frames (wrong magic byte)', async () => {
     const handler = vi.fn();
     ctrl.subscribe(handler);
+    await ctrl.audioFftDriver.start();
 
     // Bad magic — parseScopeFrame returns null
     const bad = new ArrayBuffer(20);
@@ -120,35 +125,30 @@ describe('ScopeController', () => {
     expect(handler).not.toHaveBeenCalled();
   });
 
-  it('reopens the channel after all subscribers leave and a new one joins', () => {
-    const unsub = ctrl.subscribe(vi.fn());
-    unsub();
+  it('does not let a stale exact-handle stop close the current channel', async () => {
+    const next = makeMockChannel();
+    ctrl = new ScopeController(vi.fn()
+      .mockReturnValueOnce(channel as any)
+      .mockReturnValueOnce(next as any));
+    const first = await ctrl.audioFftDriver.start();
+    await ctrl.audioFftDriver.stop(first);
+    const current = await ctrl.audioFftDriver.start();
+    await ctrl.audioFftDriver.stop(first);
 
     expect(channel.disconnect).toHaveBeenCalledTimes(1);
-
-    ctrl.subscribe(vi.fn());
-    expect(channel.connect).toHaveBeenCalledTimes(2);
+    expect(next.disconnect).not.toHaveBeenCalled();
+    await ctrl.audioFftDriver.stop(current);
+    expect(next.disconnect).toHaveBeenCalledTimes(1);
   });
 
-  it('counts subscriptions: same handler reference subscribed twice requires two unsubscribes', () => {
-    const handler = vi.fn<(frame: ScopeFrame) => void>();
-
-    const unsub1 = ctrl.subscribe(handler);
-    const unsub2 = ctrl.subscribe(handler);
-
-    expect(channel.connect).toHaveBeenCalledTimes(1);
-    expect(channel.disconnect).not.toHaveBeenCalled();
-
-    // Fire a frame — both subscriptions should receive it
-    channel._fire(makeScopeFrameBuffer());
-    expect(handler).toHaveBeenCalledTimes(2);
-
-    // First unsubscribe — channel stays open
-    unsub1();
-    expect(channel.disconnect).not.toHaveBeenCalled();
-
-    // Second unsubscribe — now channel closes
-    unsub2();
-    expect(channel.disconnect).toHaveBeenCalledTimes(1);
+  it('keeps the panel on one demand lease without transport or hardware fallback', () => {
+    const source = readFileSync(
+      resolve('src/components-v2/panels/audio-scope/AudioSpectrumPanel.svelte'),
+      'utf8',
+    );
+    expect(source).toContain("acquire('audio-fft'");
+    expect(source).toContain('release(lease)');
+    expect(source).toContain('registerPresentationDriver');
+    expect(source).not.toMatch(/\$lib\/transport|hardware-scope/);
   });
 });
--- a/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
+++ b/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
@@ -27,6 +27,9 @@ function makeMockChannel() {
     _fire(buf: ArrayBuffer) {
       for (const h of binaryHandlers) h(buf);
     },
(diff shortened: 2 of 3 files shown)


morozsm:
**Agent Review: BLOCKED**

Exact head: `63ee3722aad14bac26b4651631aa93ba1aba42d6`
Exact base: `58bbf21eae6c9f7d1917e9fa96023b0d4329aa55`
Independent fresh review (no implementation authorship). No hardware was used and no hardware evidence is claimed.

---

## Blocking: the resource handle is a session-memoized channel, so exact-handle authority does not hold

`ws-client.ts:753-760` — `getChannel(name)` **memoizes per name**:

```ts
export function getChannel(name: string): WsChannel {
  let ch = _channels.get(name);
  if (!ch) { ch = new WsChannel(); _channels.set(name, ch); }
  return ch;
}
```

`scope-controller.svelte.ts:73` calls `this._getChannel('audio-scope')` on every `start()` and returns that channel as the driver handle, and `:39` keys the binding registry by it (`_bindings = new Map<WsChannel, () => void>()`). In production every start therefore yields **the same object**, so (a) the demand model cannot distinguish an orphaned start's handle from the current one, and (b) `_bindings` can only ever hold one binding per channel.

Reproduced with the real `ScopeController` + real `PresentationResourceHost` (only the WS channel mocked, memoized like production; probes run and deleted, tree clean):

**1. Orphaned start is not disposed → channel left open.** After one adopted open/close cycle, an `acquire()` whose demand drops to 0 before the start's completion microtask leaves the socket live:

```
[P2] after orphan: connect=2 disconnect=1 liveHandlers=1 health=inactive activeHandle=undefined
```

The model believes `audio-fft` is inactive with no handle; the transport is still connected with a bound frame callback and `_activeChannel` still pointing at it. Cause: `ResourceDemand.completeStart` (`resource-demand.ts:78`) only pushes the `dispose` op when `claimCleanup` succeeds, and `claimCleanup` (`resource-demand.ts:153-157`) already holds that handle from the earlier adopted start — the ledger is keyed by handle identity and never pruned. With per-start handles this is correct; with one memoized handle the dispose is silently skipped for every orphan after the first.

**2. Next mount double-binds; the previous callback becomes unreachable.** `scope-controller.svelte.ts:87` does `this._bindings.set(ch, unsubscribe)` unconditionally, so the earlier unsubscribe is dropped on the floor. Both handlers close over the same `ch`, so the `_activeChannel !== ch` guard at `:77` does not filter either one:

```
[P3] after remount:       liveHandlers=2 subscriberCalls=2   <- every frame delivered twice
[P3] after final release: disconnect=2   liveHandlers=1      <- one handler bound forever
```

Every frame is published twice to every subscriber and `markScopeFrame()` fires twice, for the rest of the session. This is not self-healing: the final release unregisters only the newest callback.

**3. Final teardown does not prevent later frame publication.** Teardown racing an in-flight start (same one-microtask window):

```
[P4] after teardown: disconnect=1 liveHandlers=1 framesPublished=1 activeFrame=true
```

`ch._fire(frame)` after `await host.teardown()` still reaches a subscriber and still writes `ctrl.audioScopeFrame`. This is a direct miss on "final teardown unregisters the exact current callback and prevents later frame publication".

**Test fidelity.** The one shipped test that targets this property — `does not let a stale exact-handle stop close the current channel` — injects two *distinct* channel objects (`vi.fn().mockReturnValueOnce(channel).mockReturnValueOnce(next)`), a shape `getChannel` cannot produce. It passes while the production shape (one memoized channel, orphaned start) is uncovered.

**Reachability, stated honestly.** The window is one microtask: demand must reach 0 between `acquire()` (which invokes `driver.start()` → `_connect()` synchronously) and `completeStart`. That means a same-flush mount/unmount, or `presentationResources.teardown()` (`frontend-runtime.ts:191`, invoked from `App.svelte:90`) landing in the same flush as a mount. Narrow — but it is exactly the "late start after demand disappears" case this slice is meant to make fail closed, and the damage is permanent for the session rather than transient. I did not reproduce an end-to-end user gesture that opens the window; the reproduction is at the controller+host level.

**Required fix (validated).** Give the driver a unique opaque handle per start (a token wrapping the channel) and key `_bindings` by that token, so `_disconnect(token)` unregisters the exact callback and a stale handle can never alias the current one. Same memoized channel, per-start tokens:

```
[P6a] orphan:  connect=2 disconnect=2 liveHandlers=0
[P6a] remount: liveHandlers=1 framesPerFire=1
[P6b] after teardown: disconnect=2 liveHandlers=0 published=0
```

All three failures clear, entirely inside the owned file. A narrower alternative — make `_connect()` refuse to re-bind a channel already present in `_bindings` — fixes #2 only; #1 and #3 need per-start handle identity (or a matching change in `resource-demand.ts`, which is outside this slice's 3-file scope).

---

## Non-blocking

- `scope-controller.svelte.ts:33-34` — `audioScopeAvailable` / `activeScope` are hardcoded constants, so the resource's `available`/`selected` never reflect capability. Fail-closed today only because `RightSidebar.svelte:38` gates the mount on `hasAudioFft()`; the invariant is enforced by the layout, not the demand model.
- A failed start latches `health: 'failed'` and nothing calls `host.retry('audio-fft')`, so a transient WS-open failure disables the audio scope until reload. Base was not better (the throw escaped into the `$effect`) — follow-up, not a regression.
- The deleted base test `counts subscriptions: same handler reference subscribed twice requires two unsubscribes` was the only coverage of the claim still documented at `scope-controller.svelte.ts:56`.

---

## Verified correct

- Scope: exactly 3 owned files, `+113/-82` = **net +31 LOC**. No App, registry, layout, RX-audio, adapter-factory, or `$lib/transport` import in the panel (panel-boundary lint rule green).
- Sole driver: only `scope-controller.svelte.ts` opens `/api/v1/audio-scope`. `createAudioScopeConnection` has **0 call sites** (not revived); legacy `SpectrumPanel` uses `/api/v1/scope` (hardware endpoint), not a second audio-FFT channel.
- Frame subscribe/unsubscribe are lifetime-neutral — no `connect`/`disconnect` on either path; double-unsubscribe inert.
- Single App-session host (`frontend-runtime.ts:38`), teardown wired at `:191` ← `App.svelte:90`.
- Panel holds exactly one lease + one subscription; cleanup is idempotent twice over (`released` latch, plus `ResourceDemand.release` returning `false` for an already-released lease). Stale release inert.
- **First/shared/last, production handle identity:** open once, second `acquire` shares the exact handle (`snapshot().activeHandle === ch`), first release does not close, final release closes once and clears `audioScopeFrame`, zero handlers left.
- **Driver failure fails closed:** `health='failed'`, channel disconnected, no handler leaked, no `activeHandle`.
- A stale `stop` cannot be issued out of order by the host — `execute` invokes `driver.stop` synchronously at op creation, and the binding lookup disambiguates on `adopted`.

## Checks run at exact head

| Check | Result |
|---|---|
| `npx vitest run src/lib/runtime/__tests__/scope-controller.test.ts` | **7/7 pass** |
| `npx vitest run src/lib/runtime/__tests__/` | **31/31 shipped pass** |
| `npm run check` | **4268 files, 0 errors, 0 warnings** |
| `npm run lint` | **clean** |
| `npm run build` | **built in 2.99s** |
| `npm test` (full) | 65 failures / 6 files |

**Exact-base comparison for the pre-existing failures.** Base tree materialized via `git archive 58bbf21e`, same `node_modules`:

- head `63ee3722`: **65 failures / 6 files** — `Toast.component`, `i18n/locale-contract`, `i18n/pseudo-locale.smoke`, `i18n/store`, `runtime/adapters/mod-input-auto`, `skins/registry`
- base `58bbf21e`: **97 failures / 8 files** — the same six **plus** `i18n/index` and `lib/stores/layout`

Head's failing set is a strict subset of base's. The mod-input failure is `mod-input-auto.test.ts` (25), present at **both** head and base with an identical signature — `TypeError: Cannot read properties of undefined (reading 'clear')` (`localStorage` env class, not order-dependent on this diff). Pre-existing, not attributable to this PR. `scope-controller.test.ts` is green in both full runs.

---

Also note the PR is still **draft**, so it cannot merge regardless of this verdict.

**To clear this review:** apply the per-start handle fix in `scope-controller.svelte.ts`, add a regression test that drives `PresentationResourceHost` through a memoized single-channel factory (orphaned start after an adopted cycle → asserts `connect == disconnect`, one handler per frame, nothing published after `teardown()`), then re-run `npx vitest run src/lib/runtime/__tests__/`, `npm run check`, `npm run lint`, `npm run build`, and request a fresh review for the new head.

morozsm:
Author remediation evidence (not an independent review verdict)

Exact head: `6b47a8c9c33e0b6df21d534786a359cc1bc71b06`
Exact base: `58bbf21eae6c9f7d1917e9fa96023b0d4329aa55`

Addressed the memoized `WsChannel` alias blocker without changing `resource-demand.ts` or `resource-host.ts`: each driver start now returns a unique frozen opaque token, and ScopeController maps that token to the exact callback/unsubscribe binding on the one memoized named channel. Captured token cleanup cannot erase another binding; the physical channel disconnects only after its last binding is removed.

Red-before on the same-channel production shape: 4 failures/9 tests reproduced all review scenarios. Green-after durable coverage proves:
- adopted cycle followed by orphaned repeated start: balanced 2 connects / 2 disconnects, zero handlers, no active handle;
- remount after orphan: exactly one callback/frame and zero handlers after final release;
- teardown racing a repeated start: balanced cleanup, zero handlers, zero later publication;
- first/shared/last, duplicate/stale release, repeated same-handler subscriptions, and panel lease/boundary contracts remain intact.

Checks at this exact head:
- focused ScopeController: 10/10 PASS
- related ScopeController + host + demand: 34/34 PASS
- full frontend fresh bounded run: 132 files, 2309/2309 PASS
- check: 0 errors / 0 warnings
- lint: PASS
- build: PASS
- cumulative scope: 3 files, +186/-78, net +108 LOC

A prior bounded aggregate attempt hit the already-observed `mod-input-tx-guard` order flake; that exact file passes 13/13 alone and the final fresh full run passed 2309/2309. No hardware run or hardware evidence is claimed. PR remains Draft; no Ready, merge, Linear transition, or reviewer invocation was performed.

morozsm:
**Agent Review: PASS**

**Exact head: 6b47a8c9c33e0b6df21d534786a359cc1bc71b06**

Fresh independent review (MOR-1123) after remediation of the prior BLOCKED finding on `63ee3722`. I did not rely on author-supplied evidence; every claim below was re-derived locally at the exact head.

---

### Verification context

| Item | Result |
|---|---|
| Head reviewed | `6b47a8c9c33e0b6df21d534786a359cc1bc71b06` (matches `gh pr view` `headRefOid`) |
| Base | `58bbf21e`; `merge-base(origin/main, HEAD)` = `58bbf21e` |
| Current `origin/main` | `7237fec0` — diverges only by `frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts` (test-only PR #2015). **Disjoint from all 3 files here** → merge-compatible; `mergeable: MERGEABLE` |
| Scope | 3 files, +186/−78 = **+108 net** (within ≤3 files / ≤200 LOC) |

---

### Prior BLOCKER — resolved

The old shape used the memoized `WsChannel` itself as the resource handle. Production `getChannel` (`frontend/src/lib/transport/ws-client.ts:751-760`) returns **the same object** for every `getChannel('audio-scope')`, so repeated starts aliased handle identity — `claimCleanup` (`resource-demand.ts:153-157`) deduped by `Object.is` and swallowed orphan disposal, and the single `_unsubBinary` slot was overwritten.

The fix keys lifecycle on a unique opaque per-start token instead:

- `scope-controller.svelte.ts:76` — `Object.freeze({ token: Symbol('audio-fft') })`, minted per `_connect()`, never the channel.
- `scope-controller.svelte.ts:41,90` — `_bindings: Map<AudioFftHandle, {channel, unsubscribe}>` maps each token to its **own** `onBinary` remover, so no unsubscribe is ever overwritten.
- `scope-controller.svelte.ts:105-112` — `shared` check disconnects the physical channel only when no other binding still references it.
- `scope-controller.svelte.ts:80` — `if (this._activeHandle !== handle) return;` gates publication to the current start.

No second physical channel or controller is created: `_getChannel` is still called once per start and returns the one memoized instance (probe P1 asserts exactly one distinct factory arg).

Relevant transport facts I confirmed rather than assumed:
- `WsChannel.connect()` (`ws-client.ts:80-86`) early-returns while `OPEN`/`CONNECTING` → repeated `connect()` on the shared channel is idempotent, no socket leak.
- `WsChannel.disconnect()` (`ws-client.ts:171-185`) does **not** clear `binaryHandlers` → a leaked handler would survive reconnects and duplicate frames. This is exactly why per-start unsubscribe tracking is load-bearing, and why the handler-count assertions are meaningful.

---

### Independent probes (9/9 pass; temp file, removed afterwards; tree clean)

Mock models production reality exactly: one memoized channel object returned for every call, `Set` of binary handlers, real per-handler removers.

| Probe | Proves | Result |
|---|---|---|
| **P0 control** | Old shape (handle === channel) *is* genuinely defective — two orphaned starts yield **<2** cleanup calls | reproduced |
| P1 | 3 repeated starts → 3 distinct frozen symbol tokens, none `=== channel`; 1 physical channel; last stop disconnects **once** | pass |
| P2 | Stopping the earlier start leaves the later `unsubscribe` intact (handlers 2→1), channel stays open under shared demand, exactly **one** frame delivery (no dup) | pass |
| P3 | 5 mount/unmount cycles → handler count never exceeds 1, exactly one frame per cycle (no 2^n growth) | pass |
| P4 | A-B-A double orphan → both disposed, 0 leaked handlers, channel closed, `activeHandle` undefined, late frame reaches nobody | pass |
| P5 | `teardown()` racing an in-flight start → no post-teardown publication, `audioScopeFrame` null, 0 handlers, closed | pass |
| P6 | `connect()` throwing → health `failed`, 0 handlers, its own attempt cleaned up, **no** silent retry/revival | pass |
| P7 | `available:false` → transport never started | pass |
| P8 | Double `stop` + `dispose` of a stale token → idempotent, exactly one disconnect | pass |

On the race window: `driver.start` is synchronous, so the `_connect` → `completeStart` → deferred-dispose → `_disconnect` chain contains no macrotask boundary. A WS `message` event cannot interleave, so the transient `_activeHandle` assignment on an orphaned start cannot publish. P5 confirms this empirically.

### Recheck matrix

Lifetime-neutral frame subscribers ✅ · first/shared/last demand ✅ (P2) · stale A-B-A ✅ (P4) · unavailable/failure ✅ (P6/P7) · final teardown ✅ (P5) · panel holds exactly one lease + one subscription behind a `released` re-entry guard (`AudioSpectrumPanel.svelte:19-33`) ✅ · no adapter revival/fallback and no `$lib/transport` import from the panel ✅.

### Gates

- Focused `scope-controller.test.ts`: **10/10 pass**.
- `npm run lint`: clean. `npm run check` (svelte-check + tsc): **4269 files, 0 errors, 0 warnings**.
- Exact-head CI: `quick` **pass**, `grep-gate` **pass**. `Agent Review Gate` fails only for the missing PASS comment on this head — which this comment supplies.
- Full local suite: **not** green here — 76–106 failures across 9 files, varying between runs. All are `localStorage`/`window` env artifacts thrown at import time (`layout.svelte.ts:52`, `mod-input-auto.test.ts:164`) in i18n/layout/skins/Toast files that are **byte-identical at base** (outside the 3-file diff), so they are base-reproduced by construction and unrelated to this change. I could not independently confirm the author's "full 2309/2309 green" locally; exact-head CI `quick` is the authoritative green signal.

---

### Non-blocking follow-up (recommend a separate ticket; out of guardrail here)

`AmberScope.svelte:178` and `AmberCockpit.svelte:221` call `runtime.scope.subscribe()` **without** acquiring a lease. Now that `subscribe()` no longer opens the channel, those LCD panels render FFT only while `AudioSpectrumPanel` — the sole lease holder — is mounted. In shipped layouts that holds (`LcdLayout.svelte:94` mounts `RightSidebar`, `mode` defaults to `all`, defaults include `audio-scope`, and `CollapsiblePanel` collapse is CSS-only so children stay mounted).

It is, however, user-reachable to break: `drag-reorder.svelte.ts:259` calls `_removePanel(dragPanelId)` on a cross-sidebar drop, and `AudioSpectrumPanel` is imported only by `RightSidebar`. Dragging AUDIO SCOPE to the peer sidebar unmounts it permanently, releasing the only `audio-fft` lease — on the LCD skins that silently kills the main scope, with no remaining consumer able to re-establish demand. Pre-PR these panels were self-sufficient. Wiring them to the demand model exceeds this PR's 3-file/200-LOC scope, so it should not block the cutover.

No hardware was used and no hardware evidence is claimed.

Issue #2012, opened 2026-07-28
Cut RX-audio LIVE intent over to the App-session resource host


morozsm:
Linear: MOR-1125

## Scope
Atomically move browser RX-audio LIVE start/stop lifetime behind the single App-session presentation resource host. Preserve AudioManager as the only WebSocket and playback execution authority.

## Exact files
1. frontend/src/lib/runtime/frontend-runtime.ts
2. frontend/src/lib/runtime/commands/panel-commands.ts
3. frontend/src/lib/runtime/__tests__/frontend-runtime.test.ts

Maximum 3 files and 200 net LOC. No App.svelte, AudioManager, component, layout, skin, loader, TX, or scope files.

## Acceptance
- Canonical makeRxAudioHandlers dispatches semantic LIVE intent to FrontendRuntime and has no direct AudioManager startRx or stopRx call.
- First/repeated/shared LIVE and final exit have exact start/stop counts and stable AudioManager session identity.
- Handler replacement cannot stop a still-LIVE session.
- Duplicate or stale release and A-B-A completion cannot stop or overwrite the current session.
- Unavailable and start-failure paths fail closed without hidden retry or fallback.
- Final App-session teardown stops current RX exactly once before control teardown.
- Existing AudioManager reconnect/coalescing tests remain green.
- Focused tests, check, lint, build, exact-head CI, and independent lifecycle review pass.

Closes through Linear MOR-1125.

## Ordered compatibility re-plan

Slice A must merge first: one-file test-only update to frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts. Replace the obsolete same-stack LIVE-to-radio stop assertion with bounded settlement while preserving exact stop count and all canonical authority/saved-AF/volume assertions. No production code; do not close this issue.

Slice B remains the original three implementation files and closes this issue. A direct synchronous AudioManager stop bypass, a second lifetime owner, a fourth file in either PR, and resource-host changes are forbidden.

### Slice A2 — semantic runtime mock compatibility

Before production Slice B, merge a second one-file test-only prerequisite in frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts. Add a complete frontendRuntime semantic seam mock that delegates to the existing AudioManager spies. It is unused on current production, preserves all exact assertions, prevents the future panel-commands import from loading unrelated full-runtime mocks, and adds a post-settlement exact stop-count re-assert. No production code and no closing keyword.

[...]

PR #2015, opened 2026-07-28
MOR-1125 Slice A: await RX authority settlement

morozsm:
## Summary

- Behavior-neutral prerequisite Slice A for Linear MOR-1125 and GitHub issue #2012.
- Replaces the obsolete same-stack RX stop assertion with a bounded 100 ms settlement assertion.
- Preserves exact start/stop counts, factory identity, shared saved-AF state, browser volume, and duplicate-authority checks.

## Ordering

This test-only slice must merge before the MOR-1125 RX LIVE implementation slice. This prerequisite is non-closing; the production Slice B closes the linked implementation issue.

## Exact evidence

- Base: `58bbf21eae6c9f7d1917e9fa96023b0d4329aa55`
- Head: `9f71797a75fa48de85e53f33e5f4097a8647435e`
- Scope: 1 test file, +5 / -2 lines (net +3); no production code.

## Verification

- Focused RX authority test: 4/4 passed.
- Related wiring tests: 103/103 passed.
- `pnpm check`: 0 errors, 0 warnings.
- `pnpm lint`: passed.
- `pnpm build`: passed (existing chunk-size warnings only).
- `git diff --check` and exact-path scope guard: passed.

No hardware behavior changed or hardware evidence claimed.

### DIFF
--- a/frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts
+++ b/frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts
@@ -69,7 +69,7 @@ describe('RX-audio presentation command authority (MOR-1124)', () => {
     expect(setMuted).toHaveBeenNthCalledWith(2, false);
   });
 
-  it('keeps LIVE start, exit stop, and browser-volume semantics on the shared authority', () => {
+  it('keeps LIVE start, settled exit stop, and browser-volume semantics on the shared authority', async () => {
     const desktop = makeWiringRxAudioHandlers();
     const essentials = makeRuntimeRxAudioHandlers();
 
@@ -79,7 +79,10 @@ describe('RX-audio presentation command authority (MOR-1124)', () => {
     desktop.onMonitorModeChange('radio');
 
     expect(audioManager.startRx).toHaveBeenCalledTimes(1);
-    expect(audioManager.stopRx).toHaveBeenCalledTimes(1);
+    await vi.waitFor(
+      () => expect(audioManager.stopRx).toHaveBeenCalledTimes(1),
+      { timeout: 100, interval: 1 },
+    );
     expect(audioManager.setRxVolume).toHaveBeenCalledWith(0.37);
     expect(setVolume).toHaveBeenCalledWith(37);
   });


morozsm:
## Agent Review: PASS

**Exact head:** `9f71797a75fa48de85e53f33e5f4097a8647435e`
**Base:** `58bbf21eae6c9f7d1917e9fa96023b0d4329aa55`
Independent fresh review (not the implementation agent). No hardware was used and no hardware behavior is claimed.

### 1. Head / base / scope — verified

```
$ git rev-parse HEAD
9f71797a75fa48de85e53f33e5f4097a8647435e
$ git diff --numstat 58bbf21e 9f71797a
5  2  frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts
$ git diff --check 58bbf21e 9f71797a   # exit 0
```

Exactly **1 file, +5 / -2 (net +3)**, well inside the 200-LOC bound. `git diff --name-only` lists that path and nothing else, so there are **zero production changes** — `panel-commands.ts` and `command-bus.ts` are byte-identical to base. Slice A is behavior-neutral by construction.

### 2. The assertion is bounded, and it fails on 0 and on 2 — proven, not assumed

Bound is explicit and tight: `{ timeout: 100, interval: 1 }`, two orders of magnitude below vitest's 5000 ms default test timeout, so a regression fails with the count assertion message rather than an ambiguous test timeout.

I read the installed `vi.waitFor` implementation (vitest 4.1.0, `node_modules/vitest/dist/chunks/test.CTcmp4Su.js:3688`). It calls `checkCallback()` **synchronously and immediately**, and `if (checkCallback() === true) return;` — on first-check success it returns without ever installing `setTimeout`/`setInterval`.

Three throwaway probes (created untracked, run, deleted; tree verified clean afterwards, each well under 30 s):

| Probe | Setup | Result |
|---|---|---|
| A | `stopRx` never called | **FAILS** at 101 ms — `expected "vi.fn()" to be called 1 times, but got 0 times` |
| B | `onMonitorModeChange('radio')` twice → two synchronous stops | **FAILS** at 103 ms — `expected "vi.fn()" to be called 1 times, but got 2 times` |
| C | one sync stop + one stop deferred to `setTimeout(…, 5)` | assertion **passes**; settled count logged as **2** |

Probes A and B discharge the requirement directly: on the current production path the assertion fails if `stopRx` is never called *and* if it is called more than once.

### 3. Current production direct path still passes

`makeRxAudioHandlers().onMonitorModeChange` calls `audioManager.stopRx()` directly and synchronously (`src/lib/runtime/commands/panel-commands.ts:632`). Consequence: by the time `waitFor` runs its first synchronous check, the stop count is already final, so the exact-count semantics of the baseline are fully preserved today.

Confirmed empirically — the reworded LIVE test reports **0 ms**, i.e. it resolved on the first synchronous check with no polling and no timers installed:

```
✓ …rx-audio-authority.test.ts > keeps LIVE start, settled exit stop, and browser-volume semantics… 0ms
Test Files 1 passed (1) · Tests 4 passed (4)
```

I also ran the **base version of the file verbatim** against current production as a probe: **4/4 pass**. So this PR is not papering over a currently-red test — it is purely forward-looking, which is exactly what "behavior-neutral prerequisite" should look like.

### 4. Could the await mask a wrong extra stop or an unrelated timer?

- **Unrelated timer — no.** In the passing path `waitFor` installs no timers at all (§2). Additionally, `grep -rln "useFakeTimers" src/` returns **zero files** across the whole frontend suite, so `vi.isFakeTimers()` is unconditionally false and `waitFor`'s `advanceTimersByTime(interval)` branch is never taken. Nothing can be advanced or leaked into sibling files, which matters because this file runs in the non-isolated `fast` pool.
- **Extra stop — not on today's path; a bounded prospective gap.** `vi.waitFor` is first-success-wins, so it cannot observe a duplicate that arrives *after* the first successful check (probe C). That shape is unreachable while production is synchronous — the reachable duplicate shape is probe B, and it is caught. This is a property of the primitive, not a defect introduced here.

**Non-blocking follow-up for Slice B:** once the host defers the stop across an await, probe C's shape becomes reachable. Add a post-settlement quiescence step — e.g. drain microtasks after the `waitFor` and re-assert `toHaveBeenCalledTimes(1)` — so a late duplicate stop cannot hide behind the first successful poll.

### 5. Surviving assertions — all intact

| Contract | Location | Status |
|---|---|---|
| Canonical factory identity (`toBe`) | test 1 | unchanged |
| No duplicate authority (`savedAfLevel`, direct `audioManager.startRx/stopRx(` in command-bus) | test 2 | unchanged |
| Shared saved AF across both public paths (`sendCommand` nth 1/2, `patchActiveReceiver`, `setMuted` nth 1/2) | test 3 | unchanged |
| LIVE behavior + exact start count `startRx` × 1 | test 4, line 81 | unchanged, still asserted **synchronously before** the await |
| Exact stop count `stopRx` × 1 | test 4, lines 82-85 | preserved, now settlement-bounded |
| Browser volume (`setRxVolume(0.37)`, `setVolume(37)`) | test 4 | unchanged (`toHaveBeenCalledWith`, order-insensitive) |

Only the timing contract of the stop assertion moved; no coverage was dropped and no assertion was weakened to an "at least once" form. The renamed test title ("settled exit stop") accurately describes the new contract.

### 6. Gates run locally at exact head

- Focused `rx-audio-authority.test.ts` — **4/4 passed**
- Related wiring suite `src/components-v2/wiring/` — **9 files, 103/103 passed**
- `npm run check` (svelte-check + tsc) — **4268 files, 0 errors, 0 warnings**
- `npm run lint` (eslint) — **clean**
- `npm run build` — **passed** (pre-existing chunk-size and dynamic-import warnings only)
- Full `npx vitest run` — 2207 passed / 85 failed in 9 files. Every failure is `TypeError: Cannot read properties of undefined (reading 'clear'|'getItem')` on `localStorage`, the known local-only Node `--localstorage-file` environment issue. The affected files (i18n, skins registry, layout store, LanguageSelector, Toast, mod-input-auto) have no import path to the changed leaf test file, and none of them is a wiring test. I did not run a separate base baseline of the full suite; the discriminating evidence is that the diff contains zero production code and CI runs this suite green in a correctly configured environment.

### 7. Exact-head CI

Check runs on `9f71797a75fa48de85e53f33e5f4097a8647435e`:

- `quick` (Tests (quick), incl. frontend `npm ci` → `check` → `npx vitest run` → `build`) — **success**
- `grep-gate` (Rebrand grep gate) — **success**
- `Agent Review Gate` — red solely because no fresh PASS comment existed for this head; this comment addresses that.

### 8. Author claims — independently reproduced

Every claim in the PR description was re-derived from scratch rather than taken on trust: one file / +3 net ✔, focused 4/4 ✔, wiring 103/103 ✔, check + lint + build ✔, exact-head CI ✔.

### Merge preconditions (not actions I took)

This PR is currently a **draft**. Per the repo workflow, drafts must not merge — the author needs `gh pr ready` before merge, once required checks are green. I did not merge, mark ready, push, edit any code, or update Linear; the only mutation from this review is this comment.

**Verdict: Agent Review: PASS.**

PR #2016, opened 2026-07-28
MOR-1125 Slice A2: add RX runtime mock contract

morozsm:
## Summary

- Test-only prerequisite Slice A2 for Linear MOR-1125; related implementation issue: #2012.
- Adds the future FrontendRuntime RX semantic mock contract while preserving current direct-production assertions.
- Adds a fixed post-settlement checkpoint that reasserts the exact single stop count.

## Ordering

Merge this prerequisite before the MOR-1125 Slice B implementation. Issue #2012 remains for Slice B.

## Exact evidence

- Base: `7237fec0224cb5f1c0cf050dcd0d20b3961b3d9f`
- Head: `8b2213299320f44a439a75354c7cab1e22801590`
- Scope: 1 test file, +20 / -0 lines; no production code.

## Verification

- Focused RX authority baseline: 4/4 passed.
- Related wiring tests: 103/103 passed.
- `pnpm check`: 0 errors, 0 warnings.
- `pnpm lint`: passed.
- `pnpm build`: passed (existing chunk-size warnings only).
- Exact-path scope and diff checks: passed.

## Bounded mutation evidence

- A temporary throwing runtime-mock factory still produced 4/4, proving current production does not load the new mock.
- A temporary semantic-runtime panel routing mutation produced 4/4, proving the mock supports Slice B without loading unrelated runtime dependencies.
- Both temporary mutations were fully restored; final diff contains only the test contract.

No hardware behavior changed or hardware evidence claimed.

### DIFF
--- a/frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts
+++ b/frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts
@@ -28,6 +28,23 @@ vi.mock('$lib/stores/audio.svelte', () => ({
   setVolume: vi.fn(),
 }));
 
+vi.mock('$lib/runtime/frontend-runtime', async () => {
+  const { audioManager } = await import('$lib/audio/audio-manager');
+  const { setMuted, setVolume } = await import('$lib/stores/audio.svelte');
+  return {
+    runtime: {
+      setRxLive: vi.fn((live: boolean) => {
+        if (live) audioManager.startRx();
+        else audioManager.stopRx();
+      }),
+      get rxEnabled() { return audioManager.rxEnabled; },
+      setRxVolume: vi.fn((level: number) => { audioManager.setRxVolume(level); }),
+      setVolume: vi.fn((level: number) => { setVolume(level); }),
+      setMuted: vi.fn((muted: boolean) => { setMuted(muted); }),
+    },
+  };
+});
+
 import { audioManager } from '$lib/audio/audio-manager';
 import { setMuted, setVolume } from '$lib/stores/audio.svelte';
 import { patchActiveReceiver } from '$lib/stores/radio.svelte';
@@ -83,6 +100,9 @@ describe('RX-audio presentation command authority (MOR-1124)', () => {
       () => expect(audioManager.stopRx).toHaveBeenCalledTimes(1),
       { timeout: 100, interval: 1 },
     );
+    await Promise.resolve();
+    await Promise.resolve();
+    expect(audioManager.stopRx).toHaveBeenCalledTimes(1);
     expect(audioManager.setRxVolume).toHaveBeenCalledWith(0.37);
     expect(setVolume).toHaveBeenCalledWith(37);
   });


morozsm:
**Agent Review: PASS**

Exact head: `8b2213299320f44a439a75354c7cab1e22801590`
Base: `7237fec0224cb5f1c0cf050dcd0d20b3961b3d9f` · origin/main at review time: `02955e106410bed285f5f424b9744830456bbbb2`

Fresh independent review (no self-review). All author claims were re-derived from scratch; two were corrected/tightened below. No hardware was used and no hardware evidence is claimed.

---

## 1. Scope and compatibility — verified

| Requirement | Result |
|---|---|
| Exact head reviewed | `8b221329` == PR head ✓ |
| Exactly one file | `frontend/src/components-v2/wiring/__tests__/rx-audio-authority.test.ts` ✓ |
| Net LOC ≤ 200 | `+20 / -0` (numstat) ✓ |
| No production change | only a file under `__tests__/` ✓ |
| No closing keyword | none in PR body or commit message ✓ |
| Disjoint from `#2014` | `git merge-base 02955e10 8b221329` == `7237fec0` (the PR base); `#2014` touched `AudioSpectrumPanel.svelte` / `scope-controller*` only — zero overlap ✓ |

`pnpm-lock.yaml` is out of sync with `frontend/package.json` (`json-schema-to-typescript`) — **pre-existing on both base and `origin/main`**, not introduced here.

## 2. Behaviour neutrality on current production — proven

Replaced the mock factory body with an unconditional `throw`:

```
vi.mock('$lib/runtime/frontend-runtime', async () => {
  throw new Error('PROBE_P1_MOCK_FACTORY_EXECUTED');
```
→ **4/4 still pass.** The factory never executes, so nothing in today's `command-bus` → `panel-commands` graph resolves `$lib/runtime/frontend-runtime`. Confirmed statically too: `panel-commands.ts` imports `audioManager` / `setMuted` / `setVolume` directly and does not reference the runtime. The mock is inert on current production. Tree restored.

## 3. Planned Slice B routing — proven supported

Temporary mutation of `panel-commands.ts` `makeRxAudioHandlers` only, leaving **zero** direct `audioManager.*` and zero direct store calls in the RX section (i.e. the shape acceptance criterion #1 of #2012 demands):

`startRx()`→`runtime.setRxLive(true)`, `stopRx()`→`runtime.setRxLive(false)`, `rxEnabled`→`runtime.rxEnabled`, `setRxVolume`→`runtime.setRxVolume`, `setVolume`→`runtime.setVolume`, `setMuted`→`runtime.setMuted`.

→ **4/4 pass.** Every required semantic is preserved through the seam: `startRx` ×1, `stopRx` ×1, `setRxVolume(0.37)`, `setVolume(37)`, `setMuted` nth-1 `true` / nth-2 `false`, shared saved-AF (`set_af_level` `0` then `0.42` + `patchActiveReceiver({afLevel:0.42}, true)`), canonical factory identity (`makeWiringRxAudioHandlers === makeRuntimeRxAudioHandlers`), and the `command-bus` source guard. Tree restored.

## 4. No second lifetime owner / no hidden direct calls — proven

Because every mock method delegates to the *same* spies the assertions read, a stray production call and a routed call increment one counter:

| Mutation (on top of full Slice B routing) | Result |
|---|---|
| routed only (clean Slice B) | PASS |
| routed **+ leftover** `audioManager.stopRx()` | **FAIL — caught** |
| routed **+ leftover** `audioManager.startRx()` | **FAIL — caught** |

The seam cannot mask a duplicate RX lifetime owner. This is the property that matters most here, and it holds.

## 5. Bounded settlement + post-settlement re-assert — measured

| Injected defect | Caught? |
|---|---|
| zero stop (never stops) | **yes** (waitFor times out) |
| immediate duplicate (sync ×2) | **yes** |
| late duplicate, `queueMicrotask` | **yes** |
| late duplicate, 3-deep nested promise | **yes** |
| zero stop **+ unrelated timer churn** | **yes** — cannot win on an unrelated timer |
| late duplicate, `setTimeout(…, 0)` | **no** |

Microtask depth sweep of the re-assert window: caught at depth ≤3, missed from depth ≥4.

So the checkpoint is a genuine strengthening (base had `waitFor` only), and it is exactly what the PR body claims — "reasserts the exact single stop count". It is **not** a general duplicate-stop guard: a duplicate landing ≥4 microtask ticks or any macrotask later is missed. `PresentationResourceHost` is entirely promise-based (no timers), but its `execute()`/`teardown()` chains `await` well past 3 ticks. **Not blocking** — duplicate/stale-release and A-B-A coverage are explicitly Slice B acceptance criteria and belong in `frontend-runtime.test.ts` (Slice B's own file, isolated pool). Flagging so Slice B does not lean on this file for that guarantee.

## 6. Risk checked and cleared: `isolate: false` mock pinning

This file is **not** in the `isolated` pool in `vite.config.ts`, and it does module-scope `vi.mock` — which the config's own `#771` comment says can be silently no-op'd when a sibling pins the real module first. Static reachability over all 132 test files found exactly one fast-pool file that loads the real `frontend-runtime`: `src/skins/__tests__/entrypoints.test.ts` (via `lib/runtime/index.ts`).

Reproduced the worst case directly — a padded probe test importing the real module, forced to run **first** in the **same** worker (`--no-file-parallelism`, real module confirmed loaded via `typeof runtime.stopRx === 'function'`), then the authority test under full Slice B routing: **4/4 pass**. On vitest 4.0.18 the per-file mock stays authoritative. Hypothesis disproved; no action needed. Probe file deleted.

## 7. Gates (run at exact head)

- Focused `rx-audio-authority`: **4/4** (`|fast|` pool)
- `src/components-v2/wiring/`: **103/103**, 9 files
- `pnpm check`: 4265 files, **0 errors, 0 warnings**
- `pnpm lint`: **clean**
- `pnpm build`: **success** (pre-existing chunk-size warning only)
- Exact-head CI `8b221329`: `quick` success, `grep-gate` success. `Agent Review Gate` was failing only for "Missing Agent Review PASS comment" — this comment addresses it.

Working tree restored to clean after every probe (`git status` empty, verified before posting).

---

## Required for Slice B (this file pins the contract)

Slice A2 **defines** the seam, and #2012 forbids a 4th file in Slice B — so Slice B must conform to this mock exactly, not the other way round:

1. `FrontendRuntime` must gain **`setRxLive(live: boolean)`** and a **`rxEnabled`** getter with these exact names. Neither exists today. A different name, or routing through the existing `runtime.startRx()`/`runtime.stopRx()` passthroughs (deliberately absent from the mock), fails with `TypeError` and would force an edit to this file.
2. `pnpm check` gives **no** protection here — `vi.mock` factories are not structurally typed against the real module, so the mock/real divergence is invisible to the type checker. Confirmed: check is green while the mock exposes `setRxLive`/`rxEnabled` and omits `startRx`/`stopRx`.
3. `beforeEach` clears the `audioManager`/store spies but **not** `runtime.*` — those `vi.fn` counters accumulate across tests. Harmless today; do not assert on `runtime.setRxLive` call counts in this file without clearing them.

## Minor, non-blocking, pre-existing

- The browser-volume assertions use `toHaveBeenCalledWith(...)` rather than count/nth, so an *extra* `setVolume` write is not caught (verified: a leftover direct `setVolume(1)` alongside the routed call still passes). Unchanged by this PR — lines 106–107 are context, not additions.
- There is a source guard for direct `audioManager.startRx|stopRx(` in `command-bus.ts` but no equivalent guard for `panel-commands.ts`. Count-based detection covers the duplicate case (§4); a source guard would additionally catch "direct call left in, routing never added". Slice B's call.

## Merge note

PR is currently **draft** — per CLAUDE.md it must not merge while draft. This PASS is valid for head `8b221329`; run `gh pr ready` and let required checks settle, then merge. Ordering stands: Slice A (merged) → **Slice A2 (this)** → Slice B.

PR #2048, opened 2026-07-28
MOR-1141: expose source-qualified scope health facts

### LINKED ISSUE #2044: MOR-1141: expose source-qualified ScopeController health facts
Linear: https://linear.app/morozsm/issue/MOR-1141
Parent program: MOR-1140

## Exact scope

1. `frontend/src/lib/runtime/scope-controller.svelte.ts`
2. `frontend/src/lib/runtime/__tests__/scope-controller.test.ts`

Maximum 2 files and 200 net LOC.

## Acceptance

- One existing ScopeController exposes handle-qualified `{ demanded, transport, frameSeen }` facts for hardware and audio FFT.
- Audio gains exact channel state subscription/teardown and stale-handle protection symmetric with hardware.
- `frameSeen` is accepted only for the exact active handle while latest observed transport is connected; reset on start, stop and every non-connected state.
- Exact idempotent health subscriber teardown; stale A→B→A callbacks cannot publish, clear or revive current facts.
- Repeated panel driver registration is once-per-host and cannot overwrite later central capability-derived configuration.
- No ResourceHost, FrontendRuntime, panel, transport API, protocol, capabilities, TX or hardware behavior change.
- Focused/full frontend/check/lint/build, exact-head CI and fresh independent Claude Opus 5 max review pass.

No connection-epoch, continuous-freshness or real-hardware acceptance claim.

morozsm:
## Summary

- expose immutable, source-qualified `{ demanded, transport, frameSeen }` snapshots and exact health subscriptions from the existing `ScopeController`
- qualify audio and hardware state/frame callbacks by the exact active handle and current observed connected interval
- add symmetric audio binary/state teardown, reconnect resets, and stale A→B→A resistance
- preserve hardware compatibility getters from the same canonical health facts
- make audio-FFT driver registration once-per-host while preserving the first legacy configuration or a future central capability-derived configuration

## Scope

- exactly 2 files changed
- 234 additions, 36 deletions (198 net LOC)
- no ResourceHost, FrontendRuntime, panel, transport API, server, protocol, capabilities, TX, App, StatusBar, SpectrumPanel, profile, or hardware behavior changes
- no fallback and no continuous-freshness or WebSocket-epoch claim

## Validation

RED characterization before implementation:

- focused ScopeController suite: 4 failed / 14 passed (missing health API, audio state handling, stale callback protection, and once-per-host registration)

PASS on Node 26 with `NODE_OPTIONS=--no-experimental-webstorage`:

- `npx vitest run src/lib/runtime/__tests__/scope-controller.test.ts` — 18 passed
- `npx vitest run src/components-v2/panels/lcd/__tests__/audio-fft-demand.test.ts` — 3 passed
- `npm test` — 134 files, 2387 tests passed
- `npm run check` — 0 errors, 0 warnings
- `npm run lint` — passed
- `npm run lint:boundaries` — passed
- `npm run i18n:check` — passed
- `npm run build` — passed
- `git diff --check` — passed

No real-hardware validation was performed or claimed.

Closes #2044
Linear: MOR-1141

### DIFF
--- a/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
+++ b/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
@@ -99,6 +99,74 @@ describe('ScopeController', () => {
     expect(host.release(shared)).toBe(false);
   });
 
+  it('registers audio FFT once per host without overwriting central config', () => {
+    const host = new PresentationResourceHost<unknown>('session');
+    const configure = vi.spyOn(host, 'configure');
+
+    ctrl.registerPresentationDriver(host, { available: false, selected: false });
+    ctrl.registerPresentationDriver(host);
+
+    expect(configure).toHaveBeenCalledTimes(1);
+    expect(host.snapshot('audio-fft')).toMatchObject({ available: false, selected: false });
+    host.acquire('audio-fft', 'panel');
+    expect(channel.connect).not.toHaveBeenCalled();
+  });
+
+  it('publishes immutable audio health for the observed connected interval', async () => {
+    const listener = vi.fn();
+    const unsubscribe = ctrl.subscribeHealth(listener);
+    expect(ctrl.snapshotHealth('audio_fft')).toEqual({
+      demanded: false, transport: 'disconnected', frameSeen: false,
+    });
+    expect(Object.isFrozen(ctrl.snapshotHealth('audio_fft'))).toBe(true);
+
+    const handle = await ctrl.audioFftDriver.start();
+    expect(ctrl.snapshotHealth('audio_fft')).toEqual({
+      demanded: true, transport: 'disconnected', frameSeen: false,
+    });
+    for (const state of ['connecting', 'reconnecting', 'disconnected'] as const) {
+      channel._setState(state);
+      channel._fire(makeScopeFrameBuffer());
+      expect(ctrl.snapshotHealth('audio_fft')).toEqual({
+        demanded: true, transport: state, frameSeen: false,
+      });
+    }
+    channel._setState('connected');
+    channel._fire(makeScopeFrameBuffer());
+    expect(ctrl.snapshotHealth('audio_fft')).toEqual({
+      demanded: true, transport: 'connected', frameSeen: true,
+    });
+    channel._setState('reconnecting');
+    expect(ctrl.snapshotHealth('audio_fft').frameSeen).toBe(false);
+
+    unsubscribe();
+    unsubscribe();
+    const calls = listener.mock.calls.length;
+    await ctrl.audioFftDriver.stop(handle);
+    expect(ctrl.snapshotHealth('audio_fft')).toEqual({
+      demanded: false, transport: 'disconnected', frameSeen: false,
+    });
+    expect(listener).toHaveBeenCalledTimes(calls);
+    expect(channel._handlerCount()).toBe(0);
+    expect(channel._stateHandlerCount()).toBe(0);
+  });
+
+  it('keeps repeated health subscriptions exact and independently removable', async () => {
+    const listener = vi.fn();
+    const first = ctrl.subscribeHealth(listener);
+    const second = ctrl.subscribeHealth(listener);
+    const handle = await ctrl.audioFftDriver.start();
+    expect(listener).toHaveBeenCalledTimes(2);
+    first();
+    first();
+    channel._setState('connecting');
+    expect(listener).toHaveBeenCalledTimes(3);
+    second();
+    channel._setState('connected');
+    expect(listener).toHaveBeenCalledTimes(3);
+    await ctrl.audioFftDriver.stop(handle);
+  });
+
   it('both subscribers receive parsed frames', async () => {
     const h1 = vi.fn<(frame: ScopeFrame) => void>();
     const h2 = vi.fn<(frame: ScopeFrame) => void>();
@@ -106,6 +174,7 @@ describe('ScopeController', () => {
     ctrl.subscribe(h1);
     ctrl.subscribe(h2);
     await ctrl.audioFftDriver.start();
+    channel._setState('connected');
 
     channel._fire(makeScopeFrameBuffer());
 
@@ -124,6 +193,7 @@ describe('ScopeController', () => {
     const unsub1 = ctrl.subscribe(h1);
     ctrl.subscribe(h2);
     await ctrl.audioFftDriver.start();
+    channel._setState('connected');
 
     unsub1();
     channel._fire(makeScopeFrameBuffer());
@@ -137,6 +207,7 @@ describe('ScopeController', () => {
     const first = ctrl.subscribe(handler);
     const second = ctrl.subscribe(handler);
     const handle = await ctrl.audioFftDriver.start();
+    channel._setState('connected');
     first();
     channel._fire(makeScopeFrameBuffer());
     expect(handler).toHaveBeenCalledTimes(1);
@@ -150,6 +221,7 @@ describe('ScopeController', () => {
     const handler = vi.fn();
     ctrl.subscribe(handler);
     await ctrl.audioFftDriver.start();
+    channel._setState('connected');
 
     // Bad magic — parseScopeFrame returns null
     const bad = new ArrayBuffer(20);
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent lifecycle/safety review of PR #2048 (MOR-1141), reviewed at exact head **58cb43d1e944e7b10875ed99d146894ccdef9dd6** against base **351cef24f7ea9842a1c8073fe69a114dc6159d41**. Head re-confirmed against `gh pr view` immediately before posting; worktree clean at that SHA.

## Scope

Exactly two files, +234/-36 (net +198, within the 200 LOC guardrail):

- `frontend/src/lib/runtime/scope-controller.svelte.ts`
- `frontend/src/lib/runtime/__tests__/scope-controller.test.ts`

No changes to `resource-host.ts`, `resource-demand.ts`, `frontend-runtime.ts`, panels, transport API, capabilities, protocol, TX, profiles, fallback, or hardware behavior. Verified by `git diff --stat` on the exact SHA range.

## Probes run

All bounded, local, `NODE_OPTIONS=--no-experimental-webstorage` (Node 26 local; CI uses Node 20):

| Probe | Result |
|---|---|
| `vitest run src/lib/runtime/__tests__/scope-controller.test.ts src/components-v2/panels/lcd/__tests__/audio-fft-demand.test.ts` | 2 files / 21 tests passed |
| `vitest run src/lib/runtime src/components-v2` | 77 files / 1523 tests passed |
| `vitest run` (full frontend suite) | 134 files / 2387 tests passed |
| `npm run check` (svelte-check + tsc) | 4271 files, 0 errors, 0 warnings |
| `npm run lint` (eslint src/) | clean |

Static reachability probes: `grep` sweeps for every changed/added API across `src/` to enumerate real callers; read-through of `WsChannel` (`connect`/`setState`/`onBinary`/`onStateChange`), `PresentationResourceHost.execute/pump/configure`, and `ResourceDemand.reconcile/teardown`.

## Lifecycle reasoning

**Single source of truth.** `_health: Record<ScopeSource, ScopeHealth>` is the only fact store. `snapshotHealth` returns `Object.freeze({...})` of a copy — the stored value is never frozen and never aliased out. The three compatibility getters (`hardwareScopeDemanded/Connected/FrameLive`, lines 47-49) are pure derivations of `_health.hardware`; no second store or profile is introduced.

**Handle qualification.** Audio handles are `Object.freeze({token: Symbol()})`, hardware handles `Object.freeze({generation: ++n})` — identity is unique per start, so A→B→A cannot alias. Every binary and state callback guards on exact `_activeHandle`/`_activeHardwareHandle` identity, and binary additionally rejects unless `_readHealth(...).transport === 'connected'`. Verified against the adversarial tests that retain stale callbacks out of `_binaryHistory`/`_stateHistory` and invoke them directly after teardown (lines 436-474 and 383-410): stale callbacks neither clear, publish, close, nor revive current facts.

**frameSeen is per-observed-connected-interval.** Set only inside the fully-guarded binary path; forced `false` on start (lines 128-130, 202-204), on every non-connected state, and on stop. Reconnect keeps it false because the `connected` transition recomputes `state === 'connected' && current.frameSeen` from an already-cleared value — so `connected → reconnecting → connected` yields `frameSeen: false` until a new frame lands.

**Ordering / callback-before-assignment.** The seed `_setHealth(..., transport: ch.state, ...)` before subscribing is required and correct, not redundant: `WsChannel.onStateChange` does **not** replay current state, and `connect()` early-returns when `readyState` is `OPEN`/`CONNECTING` without emitting an event — so on an already-open shared channel the seed is the only source of truth. Because both handlers are now registered *before* `ch.connect(...)`, the synchronous `setState('connecting')` inside `_open()` is captured; base ordering (audio subscribed after `connect()`) could drop it.

**Symmetric audio teardown.** `ChannelBinding` now carries both `unsubscribeBinary` and `unsubscribeState`; `_disconnect` releases both in nested `finally` and only calls `channel.disconnect()` when the channel is unshared (computed after the binding is removed from the map). Test asserts `_handlerCount() === 0 && _stateHandlerCount() === 0`. Failure paths in `_connect`/`_connectHardware` delete the binding, unsubscribe both, reset facts when the failed handle is active, and disconnect only if unshared.

**Duplicate stop / stop-after-restart.** Both `_disconnect` and `_disconnectHardware` early-return on missing binding, so duplicate `stop`, `dispose`-after-`stop`, and stop of a superseded handle are inert. `PresentationResourceHost.execute` matches bindings by `Object.is` on the handle and `ResourceDemand.takeOperations` filters aliasing disposes, so no double release reaches the driver.

**subscribeHealth idempotence.** Each call takes a fresh `_nextId`, so two subscriptions of the same function reference are independent and each unsubscribe removes exactly one; repeated unsubscribe is a no-op `Map.delete`. Covered explicitly at lines 154-170. `_setHealth` dedupes on all three fields, so no spurious publishes.

**registerPresentationDriver once-per-host.** WeakSet keyed on host identity; first call wins, later calls are fully inert (`configure` not re-entered). Verified this preserves current behavior: all three production callers — `AmberCockpit.svelte:221`, `AmberScope.svelte:178`, `AudioSpectrumPanel.svelte:18` — pass no `initialConfig` and share the module singleton `presentationResources` (`frontend-runtime.ts:41`), so the first mount configures `{available: true, selected: true, driver}` exactly as before and later mounts/`$effect` re-runs no-op. Nothing else configures `'audio-fft'` in production, so the skip cannot drop the driver or the demand config, and `acquire()` still drives `refresh()` → `pump()`. The post-`teardown()` case is also safe: `ResourceDemand.teardown()` latches `ended = true` with no revival path (`acquire` throws, `reconcile` short-circuits on `!this.ended`), so re-`configure` after teardown was already inert before this PR.

**No overclaiming.** `transport` is the observed `WsChannel` state and `frameSeen` is an accepted-frame latch — the code claims neither a true WebSocket epoch, nor continuous freshness, nor verified real transport, nor hardware validation. That matches the facts it can actually establish.

## Remaining limitations (none blocking this slice)

1. **Health listeners run inside teardown, before resources are released.** `_disconnect` (lines 178-184) and `_disconnectHardware` (lines 249-255) invoke `_setHealth` — which synchronously calls arbitrary `subscribeHealth` listeners — *before* `unsubscribeBinary()/unsubscribeState()/channel.disconnect()`. A throwing listener would abort teardown: WS handlers stay registered and the socket is never closed, while `PresentationResourceHost.execute`'s `finally` still calls `completeStop` and `pump`'s `.catch(() => {})` swallows the error — leaving demand facts saying "stopped" against a still-open channel. Symmetrically, a throw from the pre-`try` `_setHealth` in `_connect`/`_connectHardware` (lines 128, 202) would strand `demanded: true` because `_disconnect` early-returns on the not-yet-registered binding. **Not reachable today** — grep confirms zero production `subscribeHealth` callers — and the unguarded subscriber loop matches the existing `_subscribers`/`_hardwareSubscribers` convention and `WsChannel`'s own `forEach` dispatch. **Required before the first production `subscribeHealth` consumer lands:** isolate listener invocation (try/catch per listener) or publish after the cleanup `finally`.
2. **Failed start demotes a previously active handle.** `_activeHandle`/`_activeHardwareHandle` is assigned before the `try` (lines 126, 200); if subscription or `connect()` throws while a *different* handle was active and still bound, the catch nulls the pointer and the older handle is permanently muted with its socket still open. Base assigned the pointer only after success. Unreachable through the only production path — `ResourceDemand` issues at most one `start` per resource at a time (serialized by `state.pending`), confirmed by the first/shared/last-demand test where two acquires produce exactly one `connect`. Worth revisiting only if concurrent handles per resource ever become legal.
3. **`snapshotHealth` is deliberately non-reactive** (`_readHealth` wraps the read in `untrack`), so a `$derived` built on it will not update; consumers must use `subscribeHealth` or the reactive compatibility getters. Worth a doc comment on the method.
4. **`markScopeFrame()` for audio now fires only for valid frames on the active handle while connected** (moved inside `if (frame)`), aligning audio with the hardware path. No observable effect: `scopeLastFrame` in `connection.svelte.ts` is written but read nowhere in `src/`.
5. The three compatibility getters currently have **no production consumers** (test-only), so the `$state`-getter conversion carries no live blast radius this slice.
6. **Verification is unit-level only.** No live IC-7610/hardware validation and no real-WebSocket run were performed; all evidence is the mock-channel suite plus static analysis of `WsChannel`/`ResourceDemand`.

## Merge preconditions (not actioned by this review)

PR is currently **draft** and `Agent Review Gate` shows FAILURE pending a fresh PASS for this head; `quick` and `grep-gate` are green. Per CLAUDE.md this PASS means the PR may merge once required checks are green for head `58cb43d1`, the gate re-evaluates this comment, and the PR is marked ready. I did not merge, did not change draft state, and did not use the GitHub review API.

PR #2050, opened 2026-07-28
MOR-1144: isolate ScopeController health listener failures

### LINKED ISSUE #2049: MOR-1144: isolate ScopeController health listener failures
Linear: https://linear.app/morozsm/issue/MOR-1144/isolate-scopecontroller-health-listener-failures-before-runtime

## Scope

Exact owned files:

1. `frontend/src/lib/runtime/scope-controller.svelte.ts`
2. `frontend/src/lib/runtime/__tests__/scope-controller.test.ts`

Maximum 2 files and 200 net LOC.

## Contract

- Keep the existing canonical health state and public `snapshotHealth` / `subscribeHealth` API.
- Notify a stable listener snapshot synchronously and isolate failures per listener.
- A throwing observer must not prevent later observers or abort audio/hardware start, state/frame handling, stop, failed-start cleanup, or teardown.
- Preserve exact handler cleanup, disconnect ownership, final health facts, idempotent unsubscribe, and deterministic subscribe/unsubscribe-during-notify behavior.
- Use existing frontend logging; no new UI state, retry policy, store/profile/controller, async lifecycle work, fallback, protocol/TX/hardware change, or hardware claim.

## Acceptance

Adversarial focused tests cover throwing and healthy listeners, listener mutation during notify, audio/hardware stop and failed-start cleanup, exact handler/disconnect counts and final snapshots. Full frontend/check/lint/boundaries/i18n/build, exact-head CI, and fresh independent Claude Opus 5 max review are required.

Blocks MOR-1142, the first production health subscriber. Closes only through a reviewed PR.

morozsm:
## Summary

- notify a stable snapshot of health listeners synchronously
- isolate each listener exception and report it through the existing frontend `console.warn` convention
- prove deterministic listener mutation plus audio/hardware start, state/frame, stop, failed-start cleanup, and teardown behavior

## Root cause

Health listeners were invoked directly from the live `Map`. A thrown exception escaped the lifecycle transition and prevented later listeners from receiving it, while listeners added during iteration could receive the in-progress transition.

## RED

Before the production change:

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run src/lib/runtime/__tests__/scope-controller.test.ts`
- result: **4 failed, 18 passed**
- failures proved the observer exception escaped audio/hardware start, later listeners were skipped, and subscribe-during-notify received the current transition

## GREEN

- focused ScopeController: **22 passed**
- MOR-1143 LCD audio fixture: **3 passed**
- full frontend suite on local Node 26.4.0 with `NODE_OPTIONS=--no-experimental-webstorage`: **134 files, 2,391 tests passed**

## Gates

- `npm run check`: 0 errors, 0 warnings
- `npm run lint`: passed
- `npm run lint:boundaries`: passed
- `npm run i18n:check`: passed (3 locale files, 166 source keys)
- `npm run build`: passed (existing chunk/dynamic-import warnings only)
- `git diff --check origin/main`: passed
- exact scope: 2 files, +103/-2 (105 touched lines, net +101), base `468ac4522ae0e41d5ab9f31c7ea2b737dc87766b`

## Hardware

No hardware validation was performed and no hardware claim is made; lifecycle coverage uses injected fake WebSocket channels.

Closes #2049

Linear: MOR-1144

### DIFF
--- a/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
+++ b/frontend/src/lib/runtime/__tests__/scope-controller.test.ts
@@ -5,7 +5,7 @@
  * safe to run in the fast (non-isolated) vitest pool.
  */
 
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { readFileSync } from 'node:fs';
 import { resolve } from 'node:path';
 import { ScopeController } from '../scope-controller.svelte';
@@ -69,6 +69,10 @@ describe('ScopeController', () => {
   beforeEach(() => {
     channel = makeMockChannel();
     ctrl = new ScopeController(() => channel as any);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
   });
 
   it('keeps frame subscriptions lifetime-neutral', () => {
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

**PR #2050 · MOR-1144 · head `806009e5f8c7e2e4c8be122469686f4d3cecba5c` · base `468ac4522ae0e41d5ab9f31c7ea2b737dc87766b`**

Independent lifecycle/safety review. Verified against the exact head SHA in a clean worktree (confirmed clean immediately before posting; `headRefOid` from the GitHub API matches the reviewed SHA).

---

## Scope compliance

| Check | Result |
|---|---|
| Files | 2 — exactly the permitted `scope-controller.svelte.ts` + `__tests__/scope-controller.test.ts` |
| Net LOC | +101 (`103` added / `2` deleted) — within 200 |
| Public API / health meaning | Unchanged: `subscribeHealth`, `snapshotHealth`, `ScopeHealth`, `HealthHandler` signatures untouched |
| New store/profile/controller | None |
| Demand / runtime / panel / transport / protocol / TX / hardware | Untouched |
| Retry / async lifecycle / fallback | None — dispatch remains fully synchronous |
| Hardware claim | None made |

Production change is 6 lines in `_setHealth` (`scope-controller.svelte.ts:276-282`): materialize the subscriber list into an array, then wrap each `listener(...)` in `try/catch` + `console.warn`.

## Why this is a real defect fix, not cosmetic

`_setHealth` is invoked from six lifecycle sites that are **not** protected by the surrounding `try`/`finally`:

- `_connect:128` and `_connectHardware:202` — *before* the `try` block, after `_activeHandle`/`_activeHardwareHandle` is already set but **before** binding registration and `ch.connect()`.
- `_disconnect:181` and `_disconnectHardware:252` — *after* `this._bindings.delete(handle)` but **before** the `try/finally` that unsubscribes handlers and disconnects the channel.
- `_connect:160` / `_connectHardware:233` — inside failed-start cleanup.
- `_connect:146` / `_connectHardware:215` — inside the binary frame path, before frame subscribers are fanned out.

On base, one throwing observer therefore aborted the lifecycle. I reproduced this directly against a reconstructed base build:

- **Start path leak** — base: `ch.connect()` never called, `[binary, state]` handlers `[0, 0]`, yet `_activeHandle` left non-null and health stuck at `{demanded: true, transport: 'disconnected', frameSeen: false}`. Unrecoverable stuck-demanded state.
- **Stop path leak** — base: `stop()` rejects, `[binary, state]` handlers remain `[1, 1]`, `disconnect()` never called. Binding already removed from the map, so the handlers are permanently orphaned.

Head fixes both; probes confirm handler counts return to `[0, 0]`, `disconnect` called exactly once, handles null, and honest final facts.

## Contract verification

1. **Stable synchronous snapshot** — `[...values()]` materialized before iteration; loop body has no `await`.
2. **Per-listener isolation** — `try` wraps only `listener(...)`; no lifecycle statement is inside the guarded region, so nothing is swallowed.
3. **Deterministic mutation semantics** — both directions corrected versus base:
   - Listener added during dispatch does **not** see the in-progress transition (base's live `Map` iterator *did* deliver it — appended entries are picked up mid-iteration).
   - Listener removing a *later* snapshot peer does **not** skip that peer (base's live iterator *did* skip it).
4. **Idempotent / self unsubscribe** — `Map.delete(id)` by identity; repeated and self-directed unsubscribe are inert and do not skip snapshot peers.
5. **Cleanup exactness preserved** — untouched: exact handler-closure unsubscribe, `some(c => c.channel === binding.channel)` shared-channel guard (disconnects only unshared/owned channels), and `_activeHandle === handle` exact-identity clearing.
6. **Dedup** — three-field early-return unchanged; identical health does not re-notify.
7. **Stale handle callbacks** — after stop, stale state/binary callbacks stay inert via the identity/generation guards; no health publication, no throw, even with a throwing observer registered.
8. **Logging convention** — `console.warn('Scope health subscriber failed', error)` mirrors the existing precedent `radio.svelte.ts:79` (`'Local extension radio state subscriber failed'`), one of 13 non-test `console.warn` call sites under `src/lib`.

## Probes run

15 temporary adversarial probes (multiple throwing listeners; throwing-listener isolation across start/stop/frame/failed-start paths; peer removal mid-dispatch; late subscribe during dispatch; self- and repeated-unsubscribe; nested/reentrant transition; stale handle callbacks; dedup; base-vs-head differential). All temp files removed — worktree confirmed clean.

Checks:

```
npx vitest run src/lib/runtime/__tests__/scope-controller.test.ts   → 22/22 pass
npx vitest run src/lib/runtime                                      → 9 files / 184 tests pass
npx vitest run                                    (full frontend)   → 134 files / 2391 tests pass
npx eslint src/lib/runtime/scope-controller.svelte.ts \
           src/lib/runtime/__tests__/scope-controller.test.ts       → exit 0
npx svelte-check --tsconfig ./tsconfig.app.json                     → 4271 files, 0 errors, 0 warnings
```

The four added tests are genuine regression tests — each fails on base (throw escapes `start()`; late subscriber wrongly notified). The added `afterEach(vi.restoreAllMocks)` correctly prevents `console.warn` spy leakage between cases.

## Limitations and notes (non-blocking)

- **Reentrancy ordering is unchanged by this diff.** A listener triggering a nested transition yields, on *both* head and base, byte-identical ordering: `["L1:true/connected", "L1:false/disconnected", "L2:false/disconnected", "L2:true/connected"]`. A peer later in the snapshot receives the outer (now-stale) transition after the nested one already completed, so its last-seen payload can be `demanded:true/connected` while actual state is `demanded:false/disconnected`. Pre-existing, neither introduced nor worsened here; `snapshotHealth()` stays authoritative and honest in both builds. Fixing it would require queued/async dispatch, which is explicitly out of scope. **Recommendation for the first consumer:** treat the callback as an invalidation signal and re-read `snapshotHealth()` rather than trusting the payload as latest.
- **Pathological `console.warn` that itself throws** escapes `_setHealth` and reproduces the base failure mode. Not production-reachable (browser/Node `console.warn` does not throw); reachable only from a harness that mocks it to throw. Not treated as a blocker.
- **No production consumer exists yet** — `subscribeHealth`/`snapshotHealth` have zero call sites outside the controller. This is pre-consumer hardening, so first-consumer safety was assessed by contract and probe rather than live usage.
- **No hardware validation** — no radio involved; this change is transport-agnostic. Local runs on Node 26 with `NODE_OPTIONS=--no-experimental-webstorage`; CI uses Node 20. The logic is plain `Map`/array/`try-catch`, so no Node-version sensitivity.
- **Process note (not a code defect):** this PR is currently **Draft**. Per repo policy a draft PR must not merge — the implementation agent should `gh pr ready` and let required checks go green. I did not change Draft/Ready state.

No blocking defects found. This PASS covers the code at `806009e5f8c7e2e4c8be122469686f4d3cecba5c` only and is void if the head moves.

PR #2056, opened 2026-07-28
MOR-1146: implement TX reducer lease and authority foundation

### LINKED ISSUE #2054: MOR-1146: implement the TX reducer lease and authority foundation
## Summary

Implement the first atomic child of the MOR-1052 reducer program: the pure TX
lease/source/authority/dispatch/release-correlation foundation. It emits
declarative effects and performs no I/O.

Linear: https://linear.app/morozsm/issue/MOR-1146/tx-reducer-foundation-lease-authority-dispatch-and-release-correlation
Program: https://linear.app/morozsm/issue/MOR-1052/program-deliver-the-pure-tx-reducer-and-adversarial-proof
Next production slice: #2057 / MOR-1148

Two exact-head Claude Opus 5 max reviews and a fresh transition audit proved
that correlated timer/command/fault completion cannot fit safely inside this
issue's 200-total-net-LOC cap. The accepted MOR-982 contract is preserved by
dependency-ordered #2057; it is not delegated to the effect runner.

## Exact scope

- `frontend/src/lib/runtime/tx-controller/model.ts`
- `frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts`
- maximum 2 files and **200 total net LOC**, including tests

## Foundation acceptance

- [ ] Phase and intent remain orthogonal; source-qualified leases/intent cannot be stolen or resumed after release/failure.
- [ ] Start requires exact eligibility, known target, allowed permit and fresh authoritative current-epoch PTT OFF newer than the latest accepted PTT marker.
- [ ] Start never regresses authoritative RF truth.
- [ ] MOD restore obligation is recorded before audio preparation.
- [ ] Audio readiness atomically records possible ON ownership, uncertain risk, one ON dispatch and an ON-timeout obligation.
- [ ] External ON during pre-dispatch preparation cancels without OFF; accepted ON while may-own escalates confirmed risk.
- [ ] Target/permit loss routes by ownership; external-only TX receives no OFF.
- [ ] Qualified release emits one ordered OFF, immediate OFF-timeout obligation and local-audio stop; duplicate release is inert.
- [ ] Pending OFF identity and one-time actual-delivery/cross-epoch barrier binding are preserved.
- [ ] Backend de-key clears local ownership without OFF; later re-key is external.
- [ ] Terminal cleanup permits at least two complete consecutive leases without retained per-lease markers.
- [ ] Strict-forward epoch/baseline behavior and queued-OFF barriers are covered.
- [ ] Exactly 2 files / ≤200 total net LOC; focused/full frontend/check/lint/boundaries/i18n/build pass.
- [ ] Exact-head CI and fresh independent Claude Opus 5 max review PASS.

## Explicit follow-up boundary

#2057 / MOR-1148 adds correlated timer-fired and command-result/error events,
complete timeout/fault/no-retry transitions, failed-release
preservation/discharge and normalized cleanup-effect guard semantics. MOR-1147
then adds production-zero adversarial proof. MOR-1053 may execute only the
final merged reducer contract.

## Compatibility

No public API, CLI, config, rigctld wire, protocol, backend, presentation,
provider, audio implementation or hardware change. This foundation remains
unused until the dependency chain lands. No hardware evidence is claimed.

morozsm:
## Summary

- add the pure MOR-1146 TX reducer foundation for source-qualified lease
  ownership, monotonic PTT authority, ON/OFF dispatch obligations,
  target/permit loss, backend de-key, and queued-OFF release correlation
- keep the model unused and I/O-free until the dependency-ordered reducer
  completion and effect-runner slices land
- preserve the full accepted MOR-982 contract through #2057 / MOR-1148 rather
  than compressing correlated timer/command/fault behavior past the 200-line
  safety cap

## Current remediation required

Formal Claude Opus 5 review on the current head identified foundation defects
that must be fixed before PASS: consecutive-lease marker cleanup, latest-PTT
start monotonicity, release risk escalation, and ON dispatch timeout arming.
The updated Linear/GitHub acceptance also requires source-qualified intent and
external-ON preparation handling.

## Scope

- 2 files
- maximum 200 total net lines
- no I/O, timer implementation, transport/audio implementation,
  Svelte/App/component, vendor/provider/profile, protocol, or hardware change

## Validation required

- focused reducer tests including two complete consecutive leases and authority races (20/20 passed)
- `npm run check`
- `npm run lint`
- `git diff --check`
- exact-head CI
- fresh independent Claude Opus 5 max `Agent Review: PASS`

Prior BLOCKED comments apply only to their recorded exact heads; no PASS is
claimed for the current head.

Linear: [MOR-1146](https://linear.app/morozsm/issue/MOR-1146/tx-reducer-foundation-lease-authority-dispatch-and-release-correlation)
Program: [MOR-1052](https://linear.app/morozsm/issue/MOR-1052/program-deliver-the-pure-tx-reducer-and-adversarial-proof)
Next production slice: [MOR-1148](https://linear.app/morozsm/issue/MOR-1148/tx-reducer-add-correlated-command-timer-and-fault-completion) / #2057

No hardware evidence is claimed.

Closes #2054

### DIFF
--- a/frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts
+++ b/frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from 'vitest';
+import { initialTxState, transition, type Eligibility, type PttObservation, type TxEvent, type TxState } from '../model';
+const marker = (seq: number, authorityEpoch = 1) => ({ authorityEpoch, pttObservationSeq: seq, pttLastObservedMonotonic: seq });
+const target = { receiver: 'MAIN', slot: 'A', frequencyHz: 14_074_000 } as const;
+const eligible: Eligibility = { catPtt: true, browserTxAudio: true, controlLive: true, permit: 'allowed', target };
+const ptt = (value: boolean, seq: number, epoch = 1): PttObservation => ({ value, observed: true, fresh: true, source: 'radio-readback', marker: marker(seq, epoch) });
+const start = (state = initialTxState(1, marker(1)), overrides = {}) => transition(state, { type: 'start', sourceId: 'desktop', leaseId: 'lease', intent: 'momentary', eligibility: eligible, ptt: ptt(false, 2), ...overrides });
+const types = (result: ReturnType<typeof transition>) => result.effects.map((effect) => effect.type);
+function active(): TxState {
+  let result = start(); const guard = result.state.guard!; result = transition(result.state, { type: 'intent', sourceId: 'desktop', guard, intent: 'latched' }); expect(result.state.intent).toBe('latched');
+  result = transition(result.state, { type: 'audio-ready', guard, commandId: 'on' }); result = transition(result.state, { type: 'on-sent', guard, commandId: 'on', barrier: marker(2) }); expect(result.state.phase).toBe('key-confirm-pending'); return transition(result.state, { type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility: eligible, offCommandId: 'off' }).state;
+}
+function keyPending(): TxState { const pending = start(); const guard = pending.state.guard!; const dispatched = transition(pending.state, { type: 'audio-ready', guard, commandId: 'on' }); return transition(dispatched.state, { type: 'on-sent', guard, commandId: 'on', barrier: marker(2) }).state; }
+function complete(state: TxState, seq: number): TxState {
+  const begun = start(state, { leaseId: `lease-${seq}`, ptt: ptt(false, seq) }); const guard = begun.state.guard!; const dispatched = transition(begun.state, { type: 'audio-ready', guard, commandId: `on-${seq}` }); expect(types(dispatched)).toContain('dispatch-on');
+  const sent = transition(dispatched.state, { type: 'on-sent', guard, commandId: `on-${seq}`, barrier: marker(seq) }); const keyed = transition(sent.state, { type: 'authority', epoch: 1, ptt: ptt(true, seq + 1), eligibility: eligible, offCommandId: 'off' });
+  const released = transition(keyed.state, { type: 'release', guard, commandId: `off-${seq}` }); const off = released.state.pendingOff!; const delivered = transition(released.state, { type: 'off-sent', ...off, eventEpoch: 1, barrier: marker(seq + 2) }); return transition(delivered.state, { type: 'authority', epoch: 1, ptt: ptt(false, seq + 3), eligibility: eligible, offCommandId: 'off' }).state;
+}
+describe('TX reducer', () => {
+  it.each([['CAT PTT', { ...eligible, catPtt: false }, ptt(false, 2)], ['browser audio', { ...eligible, browserTxAudio: false }, ptt(false, 2)],
+    ['live control', { ...eligible, controlLive: false }, ptt(false, 2)], ['known permit', { ...eligible, permit: 'unknown' }, ptt(false, 2)],
+    ['known target', { ...eligible, target: null }, ptt(false, 2)], ['fresh OFF', eligible, ptt(false, 1)]] as const)('fails closed without %s', (_name, eligibility, observation) => {
+    const result = start(initialTxState(1, marker(1)), { eligibility, ptt: observation }); expect(result).toMatchObject({ state: { phase: 'failed', fault: 'not-eligible', txRisk: 'none', leaseId: null }, effects: [] });
+  });
+  it('models the happy path without optimistic RF truth', () => {
+    const result = start(); expect(result.state).toMatchObject({ phase: 'audio-start-pending', intent: 'momentary', sourceId: 'desktop', leaseId: 'lease', generation: 1,
+      authorityEpoch: 1, leaseTarget: target, startPttBaseline: marker(2), modBarrier: marker(2), radioTx: 'off', mayOwnKey: false, pendingOff: null });
+    expect(types(result)).toEqual(['start-audio', 'arm-audio-timeout']);
+    expect(start(result.state, { sourceId: 'mobile', leaseId: 'other' })).toEqual({ state: result.state, effects: [] });
+    const guard = result.state.guard!; expect(transition(result.state, { type: 'intent', sourceId: 'mobile', guard, intent: 'latched' })).toEqual({ state: result.state, effects: [] }); expect(transition(result.state, { type: 'release', sourceId: 'mobile', guard, commandId: 'off' })).toEqual({ state: result.state, effects: [] });
+    const released = transition(result.state, { type: 'release', guard, commandId: 'off' }); expect(transition(released.state, { type: 'intent', sourceId: 'desktop', guard: released.state.guard!, intent: 'latched' })).toEqual({ state: released.state, effects: [] });
+    expect(active()).toMatchObject({ phase: 'active', intent: 'latched', radioTx: 'on', onConfirmed: marker(3) });
+  });
+  it('rejects stale start evidence without regressing authoritative truth', () => { const external = transition(initialTxState(1, marker(1)), { type: 'authority', epoch: 1, ptt: ptt(true, 10), eligibility: eligible, offCommandId: 'off' }); const stale = start(external.state, { ptt: ptt(false, 2) }); expect(stale.state).toMatchObject({ phase: 'failed', fault: 'not-eligible', pttMarker: marker(10), radioTx: 'on' }); });
+  it('supports consecutive successful leases without retained markers', () => { const once = complete(initialTxState(1, marker(1)), 2); const twice = complete(once, 6); expect(twice).toMatchObject({ phase: 'idle', generation: 4, leaseTarget: null, startPttBaseline: null, modBarrier: null, onCommandId: null, onDispatch: null, onConfirmed: null, fault: null }); });
+  it('cancels pre-key without OFF and coalesces qualified release', () => {
+    const pending = start(); const cancelled = transition(pending.state, { type: 'release', sourceId: 'desktop', guard: pending.state.guard!, commandId: 'off-pre' });
+    expect(types(cancelled)).toEqual(['cancel-timers', 'stop-local-audio']);
+    const keyed = active(); const released = transition(keyed, { type: 'release', sourceId: 'desktop', guard: keyed.guard!, commandId: 'off' });
+    expect(types(released)).toEqual(['cancel-timers', 'dispatch-off', 'arm-off-timeout', 'stop-local-audio']); expect(released.state.phase).toBe('releasing');
+    expect(released.state).toMatchObject({ phase: 'releasing', generation: 2, pendingOff: { commandId: 'off', leaseId: 'lease', generation: 2,
+      originalEpoch: 1, deliveryEpoch: null, deliveryPttBarrier: null, deliveryRebound: false } });
+    const opened = transition(released.state, { type: 'epoch', epoch: 2, baseline: marker(1, 2), offCommandId: 'other' }); const premature = transition(opened.state, { type: 'authority', epoch: 2, ptt: ptt(false, 2, 2), eligibility: eligible, offCommandId: 'other' }); expect(types(premature)).toEqual([]); const off = opened.state.pendingOff!; const sent = { type: 'off-sent' as const, ...off, eventEpoch: 2, barrier: marker(2, 2) }; const delivered = transition(opened.state, sent); expect(types(delivered)).toEqual(['arm-off-timeout']); expect(delivered.state.pendingOff?.deliveryRebound).toBe(true); expect(transition(delivered.state, sent)).toEqual({ state: delivered.state, effects: [] });
+    expect(transition(released.state, { type: 'release', sourceId: 'desktop', guard: released.state.guard!, commandId: 'other' })).toEqual({ state: released.state, effects: [] });
+  });
+  it('retains exactly one OFF obligation from ON dispatch through teardown', () => {
+    const pending = start(); const guard = pending.state.guard!; const dispatched = transition(pending.state, { type: 'audio-ready', guard, commandId: 'on' });
+    expect(dispatched.state).toMatchObject({ mayOwnKey: true, txRisk: 'uncertain', onDispatch: marker(2) }); expect(types(dispatched)).toEqual(['cancel-timers', 'dispatch-on', 'arm-on-timeout']); expect(transition(dispatched.state, { type: 'audio-ready', guard, commandId: 'duplicate' })).toEqual({ state: dispatched.state, effects: [] });
+    const routes: TxEvent[] = [{ type: 'release', sourceId: 'desktop', guard, commandId: 'off' }, { type: 'fail', guard, fault: 'ptt-on-rejected', offCommandId: 'off' }, { type: 'fail', guard, fault: 'audio-failed', offCommandId: 'off' },
+      { type: 'epoch', epoch: 2, baseline: marker(1, 2), offCommandId: 'off' }, { type: 'authority', epoch: 1, ptt: ptt(false, 2), eligibility: { ...eligible, controlLive: false }, offCommandId: 'off' }];
+    for (const event of routes) {
+      const released = transition(dispatched.state, event); expect(types(released)).toEqual(['cancel-timers', 'dispatch-off', 'arm-off-timeout', 'stop-local-audio']); expect(released.state.pendingOff).toMatchObject({ commandId: 'off', deliveryPttBarrier: null });
+      expect(transition(released.state, { type: 'on-sent', guard, commandId: 'on', barrier: marker(3) })).toEqual({ state: released.state, effects: [] });
+    }
+  });
+  it('handles external and locally-correlated authoritative ON races', () => {
+    const pending = start(); const external = transition(pending.state, { type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility: eligible, offCommandId: 'off' });
+    expect(types(external)).toEqual(['cancel-timers', 'stop-local-audio']); expect(external.state).toMatchObject({ phase: 'releasing', radioTx: 'on', mayOwnKey: false, modRestorePending: true });
+    const settled = transition(external.state, { type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off' }); expect(types(settled)).toContain('restore-mod');
+    const begun = start(); const guard = begun.state.guard!; const dispatched = transition(begun.state, { type: 'audio-ready', guard, commandId: 'on' }); const confirmed = transition(dispatched.state, { type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility: eligible, offCommandId: 'off' });
+    expect(confirmed.state).toMatchObject({ phase: 'audio-start-pending', txRisk: 'confirmed-on', onConfirmed: marker(3) });
+    const rebound = transition(confirmed.state, { type: 'on-sent', guard, commandId: 'on', barrier: marker(3) }); expect(rebound.state.phase).toBe('active'); expect(types(rebound)).toEqual(['arm-on-timeout']); expect(transition(rebound.state, { type: 'on-sent', guard, commandId: 'on', barrier: marker(3) })).toEqual({ state: rebound.state, effects: [] });
+    const dekeyed = transition(rebound.state, { type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off' }); expect(dekeyed.state.fault).toBe('backend-dekeyed'); const released = transition(dispatched.state, { type: 'release', guard, commandId: 'off' }); const raced = transition(released.state, { type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility: eligible, offCommandId: 'off' });
+    expect(raced.state).toMatchObject({ phase: 'releasing', radioTx: 'on', txRisk: 'confirmed-on', onConfirmed: marker(3) });
+  });
+  it('qualifies post-ON failures and rejects stale epochs while rebasing pre-key MOD cleanup', () => {
+    for (const fault of ['ptt-on-rejected', 'audio-failed'] as const) {
+      const keyed = keyPending(); const rejected = transition(keyed, { type: 'fail', guard: keyed.guard!, fault, offCommandId: 'off' });
+      expect(rejected.state).toMatchObject({ phase: 'releasing', fault, mayOwnKey: true, pendingOff: { commandId: 'off' } }); expect(types(rejected)).toContain('dispatch-off'); const off = rejected.state.pendingOff!; const sent = transition(rejected.state, { type: 'off-sent', ...off, eventEpoch: 1, barrier: marker(3) }); const idle = transition(sent.state, { type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off' }); expect(idle.state).toMatchObject({ phase: 'idle', fault: null, onCommandId: null, modBarrier: null });
+    }
+    const pending = start(); const opened = transition(pending.state, { type: 'epoch', epoch: 2, baseline: marker(1, 2), offCommandId: 'off' });
+    expect(opened.state.modBarrier).toEqual(marker(1, 2)); for (const [epoch, baselineEpoch] of [[2, 2], [1, 1], [3, 2]] as const) expect(transition(opened.state, { type: 'epoch', epoch, baseline: marker(9, baselineEpoch), offCommandId: 'stale' })).toEqual({ state: opened.state, effects: [] });
+    expect(types(transition(opened.state, { type: 'authority', epoch: 2, ptt: ptt(false, 2, 2), eligibility: eligible, offCommandId: 'off' }))).toContain('restore-mod');
+  });
+  it.each([['receiver', { ...eligible, target: { ...target, receiver: 'SUB' as const } }], ['slot', { ...eligible, target: { ...target, slot: 'B' as const } }], ['frequency', { ...eligible, target: { ...target, frequencyHz: target.frequencyHz + 1 } }], ['unknown', { ...eligible, target: null }], ['permit', { ...eligible, permit: 'unknown' }]] as const)('routes %s loss by key ownership', (_name, eligibility) => {
+    const pending = start(); expect(types(transition(pending.state, { type: 'authority', epoch: 1, ptt: ptt(false, 2), eligibility, offCommandId: 'off' }))).not.toContain('dispatch-off');
+    expect(types(transition(active(), { type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility, offCommandId: 'off' }))).toContain('dispatch-off');
+  });
+  it('treats qualifying backend de-key and later re-key as external', () => { const dekeyed = transition(active(), { type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off' }); expect(dekeyed.state).toMatchObject({ phase: 'failed', fault: 'backend-dekeyed', txRisk: 'none', leaseId: null, radioTx: 'off', modRestorePending: false }); expect(types(dekeyed)).toEqual(['cancel-timers', 'stop-local-audio', 'restore-mod']); const rekeyed = transition(dekeyed.state, { type: 'authority', epoch: 1, ptt: ptt(true, 5), eligibility: eligible, offCommandId: 'off' }); expect(rekeyed).toMatchObject({ state: { radioTx: 'on', leaseId: null }, effects: [] }); });
+  it('resets only discharged faults and preserves authority truth', () => {
+    const rejected = start(initialTxState(1, marker(1)), { eligibility: { ...eligible, permit: 'unknown' } }); const reset = transition(rejected.state, { type: 'reset-fault' }); expect(reset.state).toMatchObject({ phase: 'idle', intent: null, fault: null, generation: rejected.state.generation, authorityEpoch: 1, pttMarker: rejected.state.pttMarker, radioTx: 'unknown' }); expect(start(reset.state).state.phase).toBe('audio-start-pending');
+    const dekeyed = transition(active(), { type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off' }); const recovered = transition(dekeyed.state, { type: 'reset-fault' }); expect(recovered.state).toMatchObject({ phase: 'idle', fault: null, radioTx: 'off', onCommandId: null }); expect(start(recovered.state, { ptt: ptt(false, 5) }).state.phase).toBe('audio-start-pending');
+    const pending = start(); const failed = transition(pending.state, { type: 'fail', guard: pending.state.guard!, fault: 'audio-failed', offCommandId: 'off' }); expect(transition(failed.state, { type: 'reset-fault' })).toEqual({ state: failed.state, effects: [] });
+    const keyed = active(); const released = transition(keyed, { type: 'release', guard: keyed.guard!, commandId: 'off' }); expect(transition(released.state, { type: 'reset-fault' })).toEqual({ state: released.state, effects: [] });
+  });
+});
--- a/frontend/src/lib/runtime/tx-controller/model.ts
+++ b/frontend/src/lib/runtime/tx-controller/model.ts
@@ -1,0 +1,114 @@
+export type TxPhase = 'idle' | 'audio-start-pending' | 'key-confirm-pending' | 'active' | 'releasing' | 'failed';
+export type TxIntent = 'momentary' | 'latched' | null;
+export type TxTarget = { receiver: 'MAIN' | 'SUB'; slot: 'A' | 'B' | null; frequencyHz: number } | null;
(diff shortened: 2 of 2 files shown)


morozsm:
Agent Review: BLOCKED

**Reviewed head:** `545d910076705dc50609ecb124a70e27dbd07e8f` (verified before review and again immediately before posting; PR OPEN, draft, base `main`)
**Base:** `175ab682accb52d31d46bb00aa1bd857b7f3b66d`
**Contract:** MOR-1146 under accepted parent MOR-982 (App-owned TX controller)

**Scope: satisfied.** Exactly 2 files, `+200 / -0` = 200 net LOC (at the limit):
`frontend/src/lib/runtime/tx-controller/model.ts`, `frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts`.
`model.ts` has zero imports; no I/O, timers, Svelte/App/component, transport/audio implementation, manufacturer policy, capability profile, or hardware claim. Nothing in the repo imports the module yet.

All mechanical gates pass. This is blocked on three defects in the public transition contract, each reproduced by driving the exact reviewed `model.ts` directly (no repository file was modified; probes were run from `/tmp`).

---

## F1 — blocking — an ON that may have left the tab carries no de-key obligation

`audio-ready` (`model.ts:77-80`) records `onCommandId` and emits `dispatch-on`, but leaves `mayOwnKey: false`, `txRisk: 'none'`, and arms no timer. `mayOwnKey` is set only by `on-sent` (`model.ts:86-88`), and `release()` dispatches OFF only when `mayOwnKey` is true (`model.ts:56-60`). So in the window between the ON being handed to the transport and `on-sent` being fed back, **every teardown path tears down with no OFF and no `pendingOff`**:

| event dispatched after `audio-ready`, before `on-sent` | effects | `pendingOff` |
|---|---|---|
| `release` (momentary tap) — `model.ts:76` | `['cancel-timers','stop-local-audio']` | `null` |
| `fail` / `'ptt-on-rejected'` — `model.ts:82-84` | `['cancel-timers','stop-local-audio']` | `null` |
| `epoch` bump (reconnect) — `model.ts:96-98` | `['cancel-timers','stop-local-audio']` | `null` |
| eligibility loss (`controlLive:false`) — `model.ts:111`/`114` | `['cancel-timers','stop-local-audio']` | `null` |

The generation bump then makes the real `on-sent` an exact no-op (verified: returns the identical state object, `[]` effects), so the ON correlation is discarded permanently. For contrast, the same `fail` one event later yields the correct `['cancel-timers','dispatch-off','stop-local-audio']`.

Failure scenario: operator taps a momentary PTT; `dispatch-on` is executed and the frame reaches the radio; the release lands before the App feeds `on-sent` back. The radio is keyed, the reducer holds no OFF obligation, no timer is armed in that window, and no later event can create one. Same outcome for a reconnect or a control-loss during ON dispatch — the cross-epoch `pendingOff` machinery (`model.ts:90-98`, contract point 7) is unreachable from this window.

`'ptt-on-rejected'` is the sharpest case: it names the outcome of the ON dispatch, so its realistic arrival time is precisely when `mayOwnKey` is false. A rejected or timed-out correlated command is not proof of non-delivery — a command-response timeout occurs after the frame is on the wire. The model treats it as "nothing happened". `model.test.ts:57-62` exercises `'ptt-on-rejected'` only from `active()` (after PTT-ON was confirmed by readback), an ordering that cannot occur in practice, which is what masks this.

This also breaches contract point 10: while an ON is outstanding the state asserts `txRisk: 'none'`, an optimistic claim that no RF is up.

Risk: **stuck transmitter with no de-key attempt** — the exact hazard this reducer exists to prevent, and a violation of the accepted policy in `docs/plans/2026-07-25-ui-composition-discovery-and-parity.md` (P0 gap 1; R8 "prioritized de-key attempt, no stale replay").

**Required fix:** treat `onCommandId !== null` as key-may-be-owned. From `audio-ready`, set `txRisk: 'uncertain'` and arm the OFF obligation so that `release`, `fail`, `epoch`, and eligibility loss in that window all emit the same ordered qualified `dispatch-off` → `stop-local-audio` correlated to the lease/generation/epoch. A redundant OFF when the ON genuinely never left is safe and idempotent, and is the accepted fail-safe direction.

## F2 — blocking — `failed` is an absorbing state; a routine fail-closed start permanently disables TX

`start` is accepted only from `idle` (`model.ts:67`); a rejected start parks the machine in `failed` (`model.ts:69`); the only edge back to `idle` is the confirmed-release path, gated on `state.phase === 'releasing'` (`model.ts:109`). There is no `reset`/`clear-fault` member of `TxEvent`, and `epoch` does not clear the phase (`model.ts:96-98`). Verified: from `failed`, a fully eligible `start` returns the identical state object with `[]` effects, before and after a new authority epoch, and likewise after a backend de-key.

Failure scenario (routine, not adversarial): control reconnects, `epoch` sets `epochBaseline` to the newest marker. Contract point 2 requires a **strictly newer** fresh authoritative PTT-false, so an operator press arriving before the first fresh PTT poll of the new epoch is correctly rejected — and the controller is then dead for the lifetime of the state object. Three subsequent good polls and two further epoch bumps do not revive it; every later press is a silent no-op. Same terminal outcome after any backend de-key (`model.ts:102-104`), so an operator can never re-key after one.

Risk: silent, permanent loss of TX availability after a common transient. Fails safe with respect to RF, but it is a functional defect in the lease model this ticket delivers.

**Required fix:** provide an explicit recovery edge — a `reset`/`clear-fault` event, or return `failed` → `idle` once its obligations are discharged (`modRestorePending === false` and no `pendingOff`). If recovery is instead delegated to the App re-creating state via `initialTxState`, state that in the module contract and address generation monotonicity: `initialTxState` restarts `generation` at 0 while `sameGuard` (`model.ts:33-34`) compares only leaseId + generation + epoch, so a reused `leaseId` would let a previous life's in-flight callback match a new lease.

## F3 — required — the OFF dispatch→sent window is unarmed and can park the machine forever

`release()` emits `dispatch-off` but arms no timer (`model.ts:50-62`); `arm-off-timeout` is armed only once `off-sent` binds delivery (`model.ts:92-94`). If `off-sent` never arrives (queued OFF dropped on reconnect, transport never confirms), `pendingOff.deliveryPttBarrier` stays `null`, so the MOD-restore/return-to-idle branch can never fire — correctly, per contract point 7 — and nothing else can either. Verified: after four authoritative PTT-false polls the state remains `phase: 'releasing'`, `txRisk: 'confirmed-on'`, `modRestorePending: true`, `deliveryPttBarrier: null`, and a new press is a no-op.

The model owns timer arming everywhere else (`arm-audio-timeout`, `arm-on-timeout`, `arm-off-timeout`), so leaving both dispatch→sent windows unarmed is an asymmetry, not an obvious delegation.

**Required fix:** arm an escalation for the dispatch→sent windows (or document explicitly that the App owns it and how the reducer is told the dispatch failed).

## Non-blocking notes

- `newer()` (`model.ts:37-42`) fails closed when the barrier and the observation disagree on which marker field is populated (e.g. a `pttObservationSeq`-less baseline against seq-bearing observations). Safe, but it would silently and permanently block start, OFF confirmation and MOD restore. Worth a documented invariant that the App emits one marker vocabulary per epoch.
- `PttObservation` carries no receiver/target qualification, so rejecting "global-only" PTT evidence (contract point 8) rests entirely on the App normalising such observations to `source: 'other'` or `observed: false`. The reducer cannot enforce it; worth stating in the module contract.
- `effect()` (`model.ts:47-48`) drops a falsy `commandId`, so an empty-string command id would be silently omitted.

---

## Verified as correct

- **(1)** Phase/intent orthogonal; `intent` only mutates intent under `sameGuard`; `start` from a non-idle phase is a no-op and a foreign-`sourceId` release is inert — a lease/source cannot be stolen.
- **(2)** Start fail-closed on all six axes (CAT PTT, browser TX audio, live control, `permit === 'allowed'`, non-null target, strictly-newer fresh authoritative current-epoch PTT-false), with redundant `marker.authorityEpoch === state.authorityEpoch` tagging.
- **(3)** `modRestorePending` and `modBarrier` are set in the same transition that emits `start-audio`, and survive pre-key cancel, pre-key `fail`, and epoch change — including the correct rebasing of `modBarrier` onto the new epoch baseline when there is no `pendingOff` (`model.ts:98`), without which the old-epoch barrier could never be satisfied.
- **(5)** Receiver/slot/frequency change, known→unknown target, and permit loss all route through `mayOwnKey` (pre-key cancel vs. qualified release), on both the fresh-observation and stale-observation paths. Stale-epoch authority is an exact no-op. External-only TX produces no OFF.
- **(6)** `epoch` accepts only strictly forward epochs whose baseline is tagged with that same epoch; duplicate, older, and mismatched-baseline events return the identical state object with no effects.
- **(7)** `pendingOff` retains command/lease/generation/original epoch across epoch changes; only the first exact matching `off-sent` binds delivery epoch and barrier (duplicates inert); cross-epoch queued OFF binds with `deliveryRebound: true` once; a pre-delivery PTT-false restores nothing and holds the lease.
- **(8)** OFF confirmation requires a strictly newer fresh authoritative observation at the bound delivery barrier and epoch; there is no ACK/command-response event in the vocabulary, and `source: 'other'`, `observed: false`, `fresh: false` are all rejected.
- **(9)** Backend de-key clears ownership without emitting OFF, and a later re-key produces no effects.
- **(10)** Holds everywhere except the ON-in-flight `txRisk: 'none'` window described in F1.

## Commands and results

| command | result |
|---|---|
| `gh pr view 2056 --json headRefOid,state,isDraft,baseRefName` | `head=545d9100… state=OPEN draft=true base=main` (twice: before review, and immediately before posting) |
| `git diff --stat 175ab682..545d9100` | 2 files changed, 200 insertions(+), 0 deletions(-) |
| `git diff --check 175ab682..545d9100` | clean (exit 0) |
| `npx vitest run src/lib/runtime/tx-controller/__tests__/model.test.ts` | 1 file / **12 tests passed** |
| `npm run check` | `4273 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS` |
| `npm run lint` | clean, no output |
| reducer probes (F1/F2/F3), run from `/tmp` against the reviewed `model.ts` | reproduced as described above |

## Checks to rerun after fixes

1. `npx vitest run src/lib/runtime/tx-controller/__tests__/model.test.ts`
2. `npm run check` and `npm run lint`
3. `git diff --check` and the 2-file / ≤200-net-LOC scope check
4. New cases: `release`, `fail('ptt-on-rejected')`, `epoch` bump, and eligibility loss each dispatched after `audio-ready` and before `on-sent` — each must emit exactly one ordered `dispatch-off` → `stop-local-audio` and create a correlated `pendingOff`; the late `on-sent` must stay inert
5. New case: `fail('ptt-on-rejected')` driven from `key-confirm-pending` (not only from `active()`)
6. New case: recovery from `failed` back to a usable `idle` (or an explicit, tested reset contract preserving generation monotonicity)
7. New case: `dispatch-off` with no `off-sent` must escalate rather than park in `releasing`

**No hardware evidence was reviewed or claimed.** This review is static analysis plus bounded pure-reducer probes; no hardware, browser, network, or full-repository probes were run, and no repository file was modified.

morozsm:
Agent Review: BLOCKED

**Reviewed head:** `a3df1df30335145a11c7b3835d159633e1a72544` (verified exact, OPEN, draft, base `main` @ `175ab682accb52d31d46bb00aa1bd857b7f3b66d`) — re-verified immediately before posting.
**Contract:** MOR-1146, under accepted parent MOR-982.

## Scope — compliant

- Exactly 2 files: `frontend/src/lib/runtime/tx-controller/model.ts` (+114), `frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts` (+86).
- 200 total net LOC, 0 deletions — at the cap, not over it.
- `model.ts` has zero imports: no I/O, timers, Svelte/App/component, transport/audio implementation, manufacturer policy, capability profile, or hardware claim. Timers are modelled as declarative `arm-*` effects only. Pure-reducer constraint is met.

## Prior findings — all three remediated on this head

Re-derived independently against the new head, not inherited from `issuecomment-5104402709`:

1. **ON-in-flight teardown before `on-sent`** — `mayOwnKey` now flips at `audio-ready` atomically with `dispatch-on` (`model.ts:76`). All five teardown routes (`release`, `fail`/`audio-failed`, `fail`/`ptt-on-rejected`, `epoch`, `authority` eligibility loss) emit exactly `cancel-timers, dispatch-off, arm-off-timeout, stop-local-audio`. **Fixed.**
2. **Guarded `reset-fault`** — `model.ts:72` now requires `!pendingOff && !modRestorePending && !mayOwnKey`. Verified inert while a release is outstanding. **Fixed.**
3. **OFF timeout arming** — `release` arms at dispatch (`model.ts:54`) and exact `off-sent` re-arms from the bound delivery barrier (`model.ts:92`); a pre-delivery PTT-false does not restore MOD (`model.ts:106`, `deliveryPttBarrier === null` ⇒ no restore). **Fixed.**

Blocking on four new defects found on this head.

---

## Blocker 1 — TX keys exactly once, then silently never keys again

`audio-ready` is gated on `state.onCommandId === null` (`model.ts:75`), but `onCommandId` is **only** cleared by `reset-fault` (`model.ts:73`). Neither `start` (`model.ts:67`) nor the successful `releasing → idle` path (`model.ts:107`) resets it — `onCommandId`, `onDispatch` and `onConfirmed` all leak across leases.

So after one entirely successful TX cycle, every later `audio-ready` falls through every branch and returns `{ state, effects: [] }`.

```
cycle 1: start ok, audio-ready effects=["cancel-timers","dispatch-on"] -> KEYS
         settled phase=idle onCommandId="on-1"
cycle 2: start ok, audio-ready effects=[] -> NEVER KEYS (inert)
cycle 3: start ok, audio-ready effects=[] -> NEVER KEYS (inert)
```

Failure scenario: operator keys, unkeys cleanly (OFF dispatched, delivered, RF-confirmed off, state back to `idle`). On the next PTT press the lease opens, `start-audio` fires, TX audio starts streaming — and the ON command is never dispatched. The machine sits in `audio-start-pending` with `localAudio: 'starting'`, `mayOwnKey: false`, no key, and no error surfaced anywhere: the event is a total no-op, so there is no fault, no effect, and no state delta to observe.

There is no clean escape. A subsequent `release` settles back to `idle` with `onCommandId` still poisoned:

```
stuck: phase=audio-start-pending mayOwnKey=false onCommandId="on-1"
after release + PTT-false: phase=idle onCommandId="on-1"
retry after clean release: STILL NEVER KEYS
after forced audio-failed + MOD restore + reset-fault: phase=idle onCommandId=null
retry after fault recovery: KEYS
```

Only an induced fault path (`fail` → MOD discharge → `reset-fault`) restores the ability to transmit.

This is masked by the test file because every scenario builds from a fresh `initialTxState` — `active()` (`model.test.ts:9`) and `keyPending()` (`model.test.ts:16`) both start clean, and no test drives two consecutive leases through the success path.

**Required fix:** clear `onCommandId` / `onDispatch` / `onConfirmed` when a lease is established (`start`, `model.ts:67`) and when the machine returns to `idle` (`model.ts:107`), matching what `reset-fault` already does. Add a test that runs two full successful lease cycles from one state and asserts the second emits `dispatch-on`.

## Blocker 2 — start freshness is gated on `epochBaseline`, not the latest authoritative observation

`model.ts:63` computes `newer(state.epochBaseline, event.ptt)`. `epochBaseline` is only rewritten by an `epoch` event (`model.ts:96`); `pttMarker` is the field that tracks the latest authoritative observation. A PTT-false that post-dates the epoch baseline but pre-dates the last observed truth therefore passes the start gate.

```
after external key: radioTx=on pttMarker.seq=10 epochBaseline.seq=1
start w/ stale seq=2: phase=audio-start-pending radioTx=off pttMarker.seq=2 effects=["start-audio","arm-audio-timeout"]
```

Failure scenario: the radio is authoritatively observed keyed at seq 10 (external/second client TX). A start carrying a cached seq-2 PTT-false is accepted; `model.ts:67` then unconditionally writes `radioTx: 'off'` and regresses `pttMarker` from seq 10 back to seq 2, destroying the newer RF truth. The reducer proceeds to prepare audio and key against a transceiver its own last authoritative evidence said was already transmitting.

This is a fail-open against contract item 2 ("a strictly newer fresh authoritative current-epoch PTT false") and an optimistic RF claim against item 10. `pttMarker` is monotonic within an epoch and initialised to the baseline, so gating on it is strictly stronger and never rejects a genuinely fresh start.

**Required fix:** gate on `newer(state.pttMarker, event.ptt)` at `model.ts:63`, and do not let `start` move `pttMarker` backwards at `model.ts:67`. Add a test: authoritative PTT-true at a high seq, then a start with a lower-seq PTT-false ⇒ `not-eligible`, `pttMarker` unchanged.

## Blocker 3 — `txRisk` never escalates when RF is observed keyed after a qualified OFF

`txRisk` is set to `confirmed-on` only at `release` time from the then-current `radioTx` (`model.ts:53`) and on the `key-confirm-pending → active` edge (`model.ts:104`). While `phase` is `releasing` or `failed`, an authoritative PTT-true updates `radioTx` (`model.ts:99`) but leaves `txRisk` alone.

```
after release: phase=releasing radioTx=off txRisk=uncertain
in-flight ON wins race: radioTx=on txRisk=uncertain phase=releasing
```

Failure scenario: teardown between `dispatch-on` and `on-sent` — the qualified OFF is dispatched while ON is still in flight, the ON reaches the radio last, and the radio is now keyed. A readback confirms `radioTx: 'on'`, but the reducer's risk field still reports `uncertain`. Any consumer built on `txRisk` (the field that exists specifically to carry RF risk) will under-report a confirmed live carrier as merely uncertain. This is the same class of optimism item 10 forbids, in the risk field rather than the phase field.

**Required fix:** escalate `txRisk` to `confirmed-on` whenever an authoritative PTT-true is accepted while `mayOwnKey` is true, regardless of phase. Add a test for the OFF/ON race asserting `txRisk === 'confirmed-on'`.

## Blocker 4 — `dispatch-on` arms no timeout

`audio-ready` emits `cancel-timers` immediately followed by `dispatch-on` (`model.ts:77`), cancelling the audio timeout and arming nothing. `arm-on-timeout` is emitted only at `on-sent` (`model.ts:86`).

```
dispatch-on  effects = ["cancel-timers","dispatch-on"]
dispatch-off effects = ["cancel-timers","dispatch-off","arm-off-timeout","stop-local-audio"]
```

Failure scenario: ON is dispatched, `mayOwnKey` is true, and the transport never reports `on-sent` (lost command response on a still-live control link, so eligibility never drops and the `authority` release path at `model.ts:109` never fires). The machine sits in `audio-start-pending` indefinitely with TX audio streaming and the key possibly live, with no armed timer to force teardown. In this phase an authoritative PTT-true is not even promoted — `model.ts:104` requires `key-confirm-pending` — so the machine can watch `radioTx` go `on` and still never act.

This is the exact asymmetry the accepted OFF remediation fixed: arm at dispatch, re-arm at delivery. The ON path only does the second half.

**Required fix:** emit `arm-on-timeout` alongside `dispatch-on` at `model.ts:77`, letting `on-sent` re-arm from the real delivery barrier. Add a test asserting `audio-ready` effects include an arm.

---

## Non-blocking observations

- **`arm-off-timeout` has no consuming event.** `TxEvent` has no timeout variant, and `fail` is inert once `phase === 'releasing'` (`release` short-circuits). An OFF that is dispatched but never RF-confirmed therefore parks the machine terminally in `releasing` with `mayOwnKey: true` — fail-safe, and correctly never claims de-key, but there is no re-assert path and TX is locked out permanently. Worth an explicit decision record on whether re-assertion belongs to the controller ticket.
- Same terminal shape when an `epoch` advances after `deliveryPttBarrier` is bound: the barrier stays tagged with the old epoch, so `newer()` can never match again and MOD is never restored. Conservative, but stuck.
- **`startPttBaseline`** is written at `model.ts:67` and cleared at `model.ts:73`, but never read by any guard — dead state.
- **Effect guard inconsistency:** `release` attaches the post-bump guard to `cancel-timers`/`stop-local-audio` (`model.ts:50,56`), while `fail` and the backend-dekey path attach the pre-bump guard (`model.ts:82,102`). Not provably wrong without the executor, but the correlation semantics should be pinned before the controller consumes them.

## Verified-correct behaviour (for the record)

Phase/intent orthogonality and lease non-stealability (`start` inert unless `idle`; `intent`/`release` guarded by `sameGuard` + `sourceId`); `authoritative()` correctly excludes ACK/command-response/`other` sources; `epoch` accepts only strict-forward epochs with a self-tagged baseline, duplicate/older/mismatched are exact no-ops; `off-sent` correlation is complete (command, lease, generation, original epoch, current-epoch delivery, self-tagged barrier) and one-time via `deliveryEpoch !== null`, including the queued cross-epoch case setting `deliveryRebound`; pre-key eligibility/target loss cancels without OFF while post-`audio-ready` loss releases with OFF; `sameTarget` treats known→unknown as loss; stale-epoch `authority` is inert; external-only TX never receives OFF (`release` requires a guard); backend de-key clears ownership without OFF and restores MOD only against `onConfirmed`; `modRestorePending` is set before `start-audio` and survives every pre-key path, with `modBarrier` rebased on epoch change only when no OFF is outstanding.

## Checks run

| Check | Result |
|---|---|
| `npx vitest run src/lib/runtime/tx-controller/__tests__/model.test.ts` | 1 file / **14 passed** |
| `npm run check` (svelte-check + tsc) | **0 errors, 0 warnings**, 4273 files |
| `npm run lint` (eslint src/) | **clean** |
| `git diff --check 175ab682..a3df1df3` | **clean** |
| Bounded out-of-repo reducer probes | **5 expectations failed** → Blockers 1–4 |

The PR's own validation claims all reproduce. The blockers are behavioural gaps the current tests do not cover, not check failures. No repository file was created or modified during this review; probes ran outside the checkout against the module under review.

## Re-run after fixes

`npx vitest run src/lib/runtime/tx-controller/__tests__/model.test.ts`, `npm run check`, `npm run lint`, `git diff --check`, plus new coverage for: two consecutive successful lease cycles; stale-PTT start rejection with `pttMarker` preserved; OFF/ON race `txRisk` escalation; `arm-on-timeout` at `dispatch-on`.

**No hardware evidence was reviewed or claimed.** This review is limited to the pure reducer contract at the exact head above.

morozsm:
Agent Review: PASS

**Reviewed head:** `4adb63062303ddfc7adf1a90a87a0acbcdf702a8` (verified exact before review and again immediately before posting: OPEN, draft, base `main` @ `175ab682accb52d31d46bb00aa1bd857b7f3b66d`). Both prior formal reviews targeted superseded heads (`545d9100`, `a3df1df3`); every finding was re-derived from scratch against this head rather than inherited.
**Contract:** MOR-1146 under accepted parent MOR-982.

## Live split boundary

Reviewed against the live scope split, not an assumed one. MOR-1146 owns the safe, reusable pure foundation: source-qualified lease/intent, monotonic PTT authority, ON/OFF dispatch obligations and barriers, target/permit loss, backend de-key, terminal/multi-cycle cleanup, strict epochs, queued-OFF correlation.

`#2057` (MOR-1148) is recorded OPEN and explicitly *"blocked by merged MOR-1146 / PR #2056"*, and exclusively owns: lease/generation/authority-qualified **timer-fired** events; generic **command-result/error** vocabulary (ACK/response never establishing RF truth); complete audio/ON/OFF **timeout and fault** transitions with no retry; `failed/release-not-confirmed` **preservation and authoritative discharge**; and **normalized cleanup-effect guard semantics** for MOR-1053. MOR-1147 then adds the production-zero adversarial matrix.

Per that boundary, the absence of the MOR-1148 surface is **not** treated as a defect. What I did test is the three conditions that *would* block: the foundation does not make MOR-1148 impossible, does not leak unsafe authority, and does not violate its own acceptance. All three hold — see *Residual liveness* below.

## Scope — compliant

- Exactly 2 files: `frontend/src/lib/runtime/tx-controller/model.ts` (+114), `frontend/src/lib/runtime/tx-controller/__tests__/model.test.ts` (+86).
- **200 total net LOC**, 0 deletions — at the cap, not over.
- `model.ts` has **zero imports**. Static audit confirms no I/O, real timers, `Date`/`Math.random`, Svelte/App/component, transport/audio implementation, provider/vendor policy, capability profile, protocol, or hardware claim. Timers exist only as declarative `arm-*` / `cancel-timers` effects. Exported surface is types plus `initialTxState` and `transition`.
- The module is **dormant** — nothing in the repo imports it yet, so this head carries no production behavior change.
- Verified pure by execution: across the whole event alphabet, `transition` never mutates the input state or event and is deterministic on re-run; unknown event types are exact no-ops.

## Prior findings — all six fixed on this head

Re-derived independently, not inherited.

| Prior finding | Status on `4adb6306` |
|---|---|
| F1 — ON in flight before `on-sent` carried no de-key obligation | **Fixed.** `audio-ready` (`model.ts:75-76`) sets `mayOwnKey`/`txRisk: 'uncertain'`/`onCommandId` atomically with `dispatch-on`. All teardown routes in that window now emit the qualified OFF. |
| F2 — `failed` was absorbing; one fail-closed start killed TX permanently | **Fixed.** Guarded `reset-fault` (`model.ts:73`) recovers `failed → idle`; the recovered state starts again. |
| F3 — OFF dispatch→sent window unarmed | **Fixed.** `arm-off-timeout` is emitted at dispatch (`model.ts:55`) and re-armed at delivery binding (`model.ts:91`). |
| B1 — `onCommandId` leaked across leases; TX keyed exactly once then silently never again | **Fixed.** `start` (`model.ts:68`) and the successful `releasing → idle` path (`model.ts:107`) both route through `clearLease`. Drove **three** consecutive successful leases from one state object: all three dispatch ON. |
| B2 — start freshness gated on `epochBaseline` instead of the latest observation | **Fixed.** `model.ts:64` now gates on `newer(state.pttMarker, event.ptt)`. |
| B3 — `txRisk` never escalated when RF was observed keyed after a qualified OFF | **Fixed.** `model.ts:100` escalates on any accepted true while `mayOwnKey`, with no phase gate. |
| B4 — `dispatch-on` armed no timeout | **Fixed.** `arm-on-timeout` is emitted alongside `dispatch-on` (`model.ts:76`). |

## Safety reasoning — 12 required properties, adversarially proven

**316 assertions across five bounded out-of-repo probes, all passing.** Highlights beyond the PR's own suite:

1. **Phase/intent orthogonality.** `intent` applies only under exact `sameGuard` *and* exact current `sourceId`, in a legal pre-release phase, leaving `phase` untouched. Foreign-source `intent`/`release`, and wrong lease/generation/epoch guards, are exact no-ops. `start` is inert from **every** non-idle phase — no source can steal or resume a lease.
2. **Start fail-closed.** Rejected on all 13 probed axes — CAT PTT, browser TX audio, live control, `permit: 'denied'` *and* `'unknown'`, null target, `ptt=true`, `observed:false`, `fresh:false`, `source:'other'`, equal-seq, older-seq, wrong epoch — each landing `failed`/`not-eligible` with no effects and no key claim. A stale false after a newer true cannot regress `pttMarker` or `radioTx`: on the `start` path the marker holds at seq 10 with `radioTx: 'on'`; on the `authority` path the stale and equal-seq observations return the identical state object.
3. **MOD obligation precedes start-audio.** `modRestorePending` and `modBarrier` are set in the same transition that emits `start-audio`, and survive pre-key cancel.
4. **audio-ready atomicity.** One transition marks that ON may leave, `txRisk: 'uncertain'`, ON correlation, dispatch barrier, `dispatch-on` and immediate `arm-on-timeout`. Duplicate, different-commandId, wrong-guard and wrong-phase `audio-ready` are inert; wrong-commandId, wrong-epoch-barrier, stale-guard and duplicate `on-sent` are inert; the exact `on-sent` rebinds the barrier and re-arms **without** a second ON.
5. **Teardown.** Nine teardown routes × two windows (ON-in-flight and fully active) = 18 cases, each emitting exactly `cancel-timers, dispatch-off, arm-off-timeout, stop-local-audio` — one qualified OFF, the timeout obligation immediately, then local audio stop. Probed six follow-on events per case (release/fail on old *and* new guard, epoch, eligibility loss): **no second `dispatch-off` in any of the 108 combinations**, and `pendingOff` is never disturbed. True pre-dispatch cancel and external-only TX emit no OFF. Duplicate release retains the original OFF identity.
6. **Authority escalation.** External authoritative ON during audio preparation cancels locally with no OFF and holds MOD restore pending until a qualifying false (which then discharges exactly once). Accepted true escalates to `confirmed-on` and records `onConfirmed` whenever this lease may own the key — verified **before `on-sent`** (and `on-sent` then promotes straight to `active`) and **during `releasing`**.
7. **Routing by key ownership.** Receiver, slot, slot→null, frequency, known→unknown target, `permit: 'denied'` and `'unknown'` each cancel without OFF pre-key and emit a qualified OFF post-key. Stale/foreign-epoch `authority` (0, 2, 99) is inert; non-authoritative sources cannot drive routing or truth.
8. **Epochs.** Strict-forward with self-tagged baseline; duplicate, older, epoch 0, and both mismatched-baseline directions are exact no-ops. The pre-key MOD barrier rebases onto the new baseline only when there is no pending OFF — confirmed it is *not* rebased while an OFF is outstanding.
9. **Queued-OFF correlation.** `pendingOff` retains command, lease, generation and original epoch. All six wrong-field `off-sent` variants are inert. The first exact delivery binds epoch and barrier once and re-arms; second delivery and a different-barrier delivery are inert; the cross-epoch case rebinds once with `deliveryRebound: true` while preserving `originalEpoch`. Premature, cached/stale, wrong-source, unobserved and not-fresh falses cannot confirm; a false *at* the bound barrier cannot confirm, only strictly past it.
10. **Backend de-key.** Requires a strictly newer authoritative same-epoch false after ON confirmation — stale, equal-seq, non-authoritative and unobserved falses do not trigger it. It stops local audio and restores MOD once (carrying the ON-confirmation barrier), clears ownership, and emits **no OFF**. It does not repeat, and a later re-key records radio truth only, with no ownership and no effects.
11. **Multi-cycle cleanup.** Drove **three** complete consecutive successful leases from one state object. Each start is accepted, each **dispatches ON**, each reaches `active`, each emits the qualified OFF, and each settles cleanly to `idle` with `restore-mod`. After every cycle all 11 per-lease fields (`leaseTarget`, `startPttBaseline`, `modBarrier`, `onCommandId`, `onDispatch`, `onConfirmed`, `pendingOff`, `guard`, `leaseId`, `sourceId`, `intent`) are null, while `generation` and `pttMarker` stay strictly monotonic and authority epoch and radio truth are preserved. Guarded `reset-fault` preserves the same invariants and is correctly blocked while an OFF, a MOD obligation, or `mayOwnKey` is outstanding.
12. **No optimism.** `radioTx: 'off'` at start is derived from the authoritative false that gated the start, never asserted. No state reaches `idle` with `mayOwnKey` or a surviving `pendingOff`. No effect asserts hardware state.

## Residual liveness — fail-safe, and squarely MOR-1148's

If an epoch advances *after* `deliveryPttBarrier` is bound, the barrier stays tagged to the old epoch, so `newer()` can never match again and the machine parks in `releasing`. I drove this to the stuck state and probed it: it never returns to `idle`, never claims de-key, keeps `modRestorePending: true` and honest `txRisk: 'confirmed-on'`, emits no further OFF, and makes `start` inert — it **locks TX out rather than leaking authority**. It is therefore a liveness gap, not a safety one, and it is exactly the acceptance criterion MOR-1148 records (*"OFF error/timeout records `failed/release-not-confirmed`, preserves pending ownership/risk/MOD obligations, performs no retry, and never claims de-key"*).

MOR-1148 is confirmed reachable from this foundation: the stuck state retains `guard`, and `pendingOff` retains `commandId`, `leaseId`, `generation`, `originalEpoch` and `deliveryEpoch`, while the `arm-off-timeout` effect carries both `guard` and `commandId` — everything a correlated timer-fired transition needs.

**Non-blocking notes** (no fix required to merge):
- `fail` arriving while `phase === 'releasing'` with `mayOwnKey` overwrites only the `fault` label (`model.ts:79`): effects are empty and `pendingOff` is byte-identical, but the returned object is a new reference, so it is not identity-inert. Cosmetic/diagnostic; falls under MOR-1148's fault-completion work.
- `startPttBaseline` (`model.ts:68`) is written and cleared but never read by any guard — dead state.
- Cleanup-effect guard attachment is still pre-bump in `fail`/backend-de-key (`model.ts:81`, `103`) vs post-bump in `release` (`model.ts:51-57`) — explicitly MOR-1148's scope, noted only for continuity.
- `newer()` fails closed when barrier and observation disagree on which marker field is populated; safe, but worth a documented invariant that the App emits one marker vocabulary per epoch.

## Commands and results

| Command | Result |
|---|---|
| `gh pr view 2056 --json headRefOid,state,isDraft,baseRefName` | `head=4adb6306… state=OPEN draft=true base=main` (before review and immediately before posting) |
| `git diff --stat 175ab682..4adb6306` | 2 files changed, **200 insertions(+), 0 deletions(-)** |
| `git diff --numstat` net LOC | **net=200** (cap 200) |
| `git diff --check 175ab682..4adb6306` | **clean** |
| `npx vitest run src/lib/runtime/tx-controller/__tests__/model.test.ts` | 1 file / **20 tests passed** |
| `npm run check` (svelte-check + tsc) | **4273 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS** |
| `npm run lint` (eslint src/) | **clean**, exit 0 |
| Bounded out-of-repo reducer probes A–E | **316/316 assertions passed** |
| `git status --porcelain` | **empty** — no repository file was created or modified |

Every probe ran outside the checkout against the exact-head `model.ts` blob. No command reached its time bound. No hardware, browser, or network probe and no unbounded full-repository probe was run.

**No hardware evidence was reviewed or claimed.** This review is static analysis of the exact-head diff plus bounded pure-reducer execution.

PR #2061, opened 2026-07-28
MOR-1053: add the TX reducer effect controller

### LINKED ISSUE #2060: MOR-1053: coordinate TX audio, PTT effects, timers, and stale completion
## Summary

Build the dependency-injected controller/effect runner around the merged pure TX reducer. The runner executes reducer obligations and translates external callbacks into exact reducer events; it must not reinterpret state, ownership, failure policy, or RF truth.

Linear: https://linear.app/morozsm/issue/MOR-1053/core-tx-controller-coordinate-audio-ptt-effects-timers-and-stale
Parent: MOR-1006

## Exact scope

Only these files may change:

1. `frontend/src/lib/runtime/tx-controller/controller.ts`
2. `frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts`

Maximum 2 files and 200 total net LOC, including tests. Do not edit `model.ts` or `model.test.ts`; any reducer defect requires a separate bounded remediation.

## Acceptance criteria

- [ ] Audio preparation completes before any PTT ON attempt.
- [ ] PTT ON is attempted once and never retried, queued, replayed, or reconstructed after interruption.
- [ ] Qualified OFF is attempted/coalesced once before local audio cleanup.
- [ ] Synchronous OFF dispatch failure still runs idempotent local stop in a `finally` path.
- [ ] Timer scheduling, audio completions, command events, and authority events dispatch exact lease/generation/authority/command-qualified reducer events.
- [ ] ACK/response/error callbacks cannot manufacture RF truth.
- [ ] Conservative MOD ownership remains in the existing adapter; restoration executes only for the reducer restore obligation.
- [ ] Exact fake-clock tests prove effect order, timeout delivery, release coalescing, no-ON replay, OFF-send failure cleanup, and stale effect suppression.
- [ ] Focused controller and reducer suites, full frontend, check, lint, boundaries, i18n, and build pass.
- [ ] Exact-head CI and a fresh independent Claude Opus 5 max review pass before normal merge.

## Non-goals and compatibility

No App/component/source migration, transport implementation, audio adapter, capability policy, reducer-model, protocol, CLI, config, docs, or hardware change. No public compatibility change is intended. No hardware evidence is claimed.

morozsm:
## Summary

Add the dependency-injected effect controller around the merged pure TX reducer.

Linear: https://linear.app/morozsm/issue/MOR-1053/core-tx-controller-coordinate-audio-ptt-effects-timers-and-stale
Closes #2060

## Scope

- `frontend/src/lib/runtime/tx-controller/controller.ts`
- `frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts`
- exactly 2 files, +200/-0 = net +200 LOC
- reducer `model.ts` and `model.test.ts` unchanged

## Behavior

- executes reducer effects in exact order through a reentrancy-safe event queue
- prepares audio before one ON; stale completion cannot replay ON
- preserves reducer-owned OFF coalescing and OFF-before-local-stop ordering
- normalizes synchronous PTT send failures into exact qualified command results
- delivers exact lease/generation/epoch/command/revision-qualified timer and transport callbacks
- isolates throwing subscribers and timer cancellation so they cannot suppress OFF/stop obligations
- converts scheduling failure into the same one-shot qualified timer event for audio, ON, and OFF deadlines
- cancels timer records by exact guard or cleanup lineage without crossing a new lease generation
- executes MOD restoration only for the reducer restore obligation caused by authoritative evidence

## Verification

- focused controller + reducer: 51/51
- full frontend: 136 files, 2467/2467
- independent mutation verification: all 6 B1-B3 mutant families killed non-vacuously
- prior independent P0 adversarial probes: 8/8 pass
- check: 0 errors, 0 warnings
- lint, boundaries, i18n, build, and local diff guard: pass
- latest remediation adds the proof rows and removes one production blank line; controller behavior is unchanged from `4a90aef39398be348b22a1cf5bc4bbe7b7b4c161`

The B1-B3 remediation proves audio resolve/reject/synchronous throw, truly synchronous/reentrant transport callbacks and ON throw, fired one-shot behavior, and direct exact-guard cancellation. `commandId`, `stopLocalAudio`, and `restoreMod` remain total synchronous DI operations; MOR-1053 does not add adapter failure policy for those hooks. No App/component/source migration, production transport/audio adapter, capability policy, reducer policy, protocol, or hardware change. No hardware evidence is claimed.

Fresh exact-head Claude Opus 5 max review: PASS for `c5b8a0b46714f68b59115b0a83c1f3103fb5b54e`. All required CI must remain green before merge.

### DIFF
--- a/frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts
+++ b/frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts
@@ -1,0 +1,89 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { TxController, type TxControllerDependencies } from '../controller';
+import type { Eligibility, PttMarker, PttObservation, TxEvent } from '../model';
+const marker = (seq: number, epoch = 1): PttMarker => ({ authorityEpoch: epoch, pttObservationSeq: seq, pttLastObservedMonotonic: seq });
+const ptt = (value: boolean, seq: number): PttObservation => ({ value, observed: true, fresh: true, source: 'radio-readback', marker: marker(seq) });
+const eligible: Eligibility = { catPtt: true, browserTxAudio: true, controlLive: true, permit: 'allowed', target: { receiver: 'MAIN', slot: 'A', frequencyHz: 14_074_000 } };
+const start = (): TxEvent => ({ type: 'start', sourceId: 'desktop', leaseId: 'lease', intent: 'momentary', eligibility: eligible, ptt: ptt(false, 1) });
+
+function harness(audio: Promise<string | null> = Promise.resolve(null), failure: 'off' | 'cancel' | 10 | 20 | 30 | null = null) {
+  const log: string[] = []; const reports: Array<{ command: 'on' | 'off'; report: Parameters<TxControllerDependencies['sendPtt']>[3] }> = [];
+  let id = 0;
+  const dependencies: TxControllerDependencies = {
+    startAudio: vi.fn(() => { log.push('audio'); return audio; }),
+    sendPtt: vi.fn((command, _id, _correlation, report) => { log.push(command); reports.push({ command, report }); if (command === 'off' && failure === 'off') throw new Error('closed'); }),
+    stopLocalAudio: vi.fn(() => { log.push('stop'); }), restoreMod: vi.fn(() => { log.push('restore'); }),
+    commandId: vi.fn((command) => `${command}-${++id}`), schedule: vi.fn((callback, delay) => { if (delay === failure) throw new Error('clock'); return setTimeout(callback, delay); }),
+    cancel: vi.fn((handle) => { clearTimeout(handle as ReturnType<typeof setTimeout>); if (failure === 'cancel') throw new Error('cancel'); }),
+    timeoutMs: { 'audio-start': 10, 'on-confirmation': 20, 'off-confirmation': 30 },
+  };
+  return { controller: new TxController(1, marker(0), dependencies), dependencies, log, reports };
+}
+const flush = async () => { await Promise.resolve(); await Promise.resolve(); };
+
+describe('TxController', () => {
+  afterEach(() => vi.useRealTimers());
+
+  it('runs audio before one ON and command acknowledgements never create RF truth', async () => {
+    vi.useFakeTimers(); const h = harness(); h.controller.dispatch(start()); expect(h.log).toEqual(['audio']);
+    await flush(); expect(h.log).toEqual(['audio', 'on']); expect(h.reports.filter((item) => item.command === 'on')).toHaveLength(1);
+    const sent = h.reports[0].report; sent({ outcome: 'sent', eventEpoch: 1, barrier: marker(2) }); sent({ outcome: 'ack', eventEpoch: 1, barrier: null });
+    expect(h.controller.snapshot()).toMatchObject({ phase: 'key-confirm-pending', radioTx: 'off', mayOwnKey: true });
+  });
+
+  it('delivers the exact audio timeout and suppresses its late async completion', async () => {
+    vi.useFakeTimers(); let resolve!: (value: string | null) => void; const h = harness(new Promise((done) => { resolve = done; }));
+    h.controller.dispatch(start()); vi.advanceTimersByTime(10);
+    expect(h.controller.snapshot()).toMatchObject({ phase: 'failed', fault: 'audio-timeout' }); expect(h.log).toEqual(['audio', 'stop']);
+    resolve(null); await flush(); expect(h.log).toEqual(['audio', 'stop']); expect(h.reports).toHaveLength(0);
+  });
+
+  it('delivers correlated ON and OFF deadlines at their exact fake-clock durations', async () => {
+    vi.useFakeTimers(); const h = harness(); h.controller.dispatch(start()); await flush();
+    vi.advanceTimersByTime(20);
+    expect(h.controller.snapshot()).toMatchObject({ phase: 'releasing', fault: 'on-timeout' }); expect(h.log).toEqual(['audio', 'on', 'off', 'stop']);
+    vi.advanceTimersByTime(30);
+    expect(h.controller.snapshot()).toMatchObject({ phase: 'failed', fault: 'release-not-confirmed' });
+  });
+
+  it('coalesces release, attempts OFF before stop, and cleans up after a synchronous send failure', async () => {
+    vi.useFakeTimers(); const h = harness(Promise.resolve(null), 'off'); h.controller.dispatch(start()); await flush();
+    const guard = h.controller.snapshot().guard!; h.controller.dispatch({ type: 'release', guard, commandId: 'off-release' });
+    expect(h.log).toEqual(['audio', 'on', 'off', 'stop']); expect(h.controller.snapshot()).toMatchObject({ phase: 'failed', fault: 'release-not-confirmed' });
+    h.controller.dispatch({ type: 'release', guard: h.controller.snapshot().guard!, commandId: 'off-duplicate' });
+    h.reports.find((item) => item.command === 'on')!.report({ outcome: 'sent', eventEpoch: 1, barrier: marker(2) });
+    expect(h.log.filter((item) => item === 'off')).toHaveLength(1); expect(h.log.filter((item) => item === 'stop')).toHaveLength(1);
+  });
+
+  it('isolates throwing observers and timer cancellation so release still performs OFF then stop', async () => {
+    vi.useFakeTimers(); const cases = [harness(), harness(Promise.resolve(null), 'cancel')]; cases[0].controller.subscribe(() => { throw new Error('observer'); });
+    for (const h of cases) { h.controller.dispatch(start()); await flush(); h.controller.dispatch({ type: 'release', guard: h.controller.snapshot().guard!, commandId: 'off' }); expect(h.log).toEqual(['audio', 'on', 'off', 'stop']); }
+    expect(vi.mocked(cases[1].dependencies.cancel)).toHaveBeenCalled();
+  });
+
+  it('fails closed with exact qualified audio, ON, and OFF events when scheduling throws', async () => {
+    vi.useFakeTimers(); const audio = harness(Promise.resolve(null), 10); audio.controller.dispatch(start()); await flush(); expect(audio.controller.snapshot()).toMatchObject({ phase: 'failed', fault: 'audio-timeout' }); expect(audio.log).toEqual(['audio', 'stop']);
+    const on = harness(Promise.resolve(null), 20); on.controller.dispatch(start()); await flush(); expect(on.controller.snapshot()).toMatchObject({ phase: 'releasing', fault: 'on-timeout' }); expect(on.log).toEqual(['audio', 'on', 'off', 'stop']);
+    on.reports[0].report({ outcome: 'sent', eventEpoch: 1, barrier: marker(2) }); expect(on.log).toEqual(['audio', 'on', 'off', 'stop']);
+    const off = harness(Promise.resolve(null), 30); off.controller.dispatch(start()); await flush(); off.controller.dispatch({ type: 'release', guard: off.controller.snapshot().guard!, commandId: 'off' });
+    expect(off.controller.snapshot()).toMatchObject({ phase: 'failed', fault: 'release-not-confirmed' }); expect(off.log).toEqual(['audio', 'on', 'off', 'stop']);
+  });
+
+  it('never replays ON when a pre-key lease is released before audio completes', async () => {
+    vi.useFakeTimers(); let resolve!: (value: string | null) => void; const h = harness(new Promise((done) => { resolve = done; }));
+    h.controller.dispatch(start()); h.controller.dispatch({ type: 'release', guard: h.controller.snapshot().guard!, commandId: 'off-cancel' });
+    resolve(null); await flush(); vi.runAllTimers();
+    expect(h.log).toEqual(['audio', 'stop']); expect(h.reports).toHaveLength(0); expect(h.controller.snapshot().mayOwnKey).toBe(false);
+  });
+
+  it('restores MOD only from the reducer obligation after post-OFF authority', async () => {
+    vi.useFakeTimers(); const h = harness(); h.controller.dispatch(start()); await flush();
+    h.reports[0].report({ outcome: 'sent', eventEpoch: 1, barrier: marker(2) });
+    h.controller.dispatch({ type: 'authority', epoch: 1, ptt: ptt(true, 3), eligibility: eligible, offCommandId: 'off-a' });
+    h.controller.dispatch({ type: 'release', guard: h.controller.snapshot().guard!, commandId: 'off-a' });
+    h.reports.find((item) => item.command === 'off')!.report({ outcome: 'sent', eventEpoch: 1, barrier: marker(3) });
+    const cancelled = vi.mocked(h.dependencies.cancel).mock.calls.length; h.controller.dispatch({ type: 'authority', epoch: 1, ptt: ptt(false, 4), eligibility: eligible, offCommandId: 'off-a' }); expect(vi.mocked(h.dependencies.cancel).mock.calls).toHaveLength(cancelled + 2); expect(vi.getTimerCount()).toBe(0);
+    h.controller.dispatch({ type: 'authority', epoch: 1, ptt: ptt(false, 5), eligibility: eligible, offCommandId: 'off-a' });
+    expect(h.log.filter((item) => item === 'restore')).toHaveLength(1); expect(h.controller.snapshot().phase).toBe('idle');
+  });
+});
--- a/frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts
+++ b/frontend/src/lib/runtime/tx-controller/__tests__/controller.test.ts
@@ -5,47 +5,39 @@ const marker = (seq: number, epoch = 1): PttMarker => ({ authorityEpoch: epoch, 
 const ptt = (value: boolean, seq: number): PttObservation => ({ value, observed: true, fresh: true, source: 'radio-readback', marker: marker(seq) });
 const eligible: Eligibility = { catPtt: true, browserTxAudio: true, controlLive: true, permit: 'allowed', target: { receiver: 'MAIN', slot: 'A', frequencyHz: 14_074_000 } };
 const start = (): TxEvent => ({ type: 'start', sourceId: 'desktop', leaseId: 'lease', intent: 'momentary', eligibility: eligible, ptt: ptt(false, 1) });
-
-function harness(audio: Promise<string | null> = Promise.resolve(null), failure: 'off' | 'cancel' | 10 | 20 | 30 | null = null) {
+function harness(audio: Promise<string | null> = Promise.resolve(null), failure: 'off' | 'on' | 'cancel' | 'audio' | 'reentrant' | 'callback-throw' | 10 | 20 | 30 | null = null) {
   const log: string[] = []; const reports: Array<{ command: 'on' | 'off'; report: Parameters<TxControllerDependencies['sendPtt']>[3] }> = [];
   let id = 0;
   const dependencies: TxControllerDependencies = {
-    startAudio: vi.fn(() => { log.push('audio'); return audio; }),
-    sendPtt: vi.fn((command, _id, _correlation, report) => { log.push(command); reports.push({ command, report }); if (command === 'off' && failure === 'off') throw new Error('closed'); }),
+    startAudio: vi.fn(() => { log.push('audio'); if (failure === 'audio') throw new Error('audio'); return audio; }),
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: BLOCKED

Fresh independent TX-safety/lifecycle review of exact head `4a90aef39398be348b22a1cf5bc4bbe7b7b4c161` (base and merge-base both `3ffde9d74d67fffab69da9cd9e999eef52de8eb9`, PR is Draft). No repository file or branch was modified by this review; all probes ran in an isolated mirror outside the repo.

**No production defect was found.** Every safety-critical path I probed behaves exactly as MOR-1053 requires. The block is on **missing coverage**: three surviving mutants of `controller.ts` — one of which keys the radio after audio preparation has failed — pass the entire focused suite (48/48) and the entire frontend suite (2464/2464). Acceptance criterion 1 and the PR body's reentrancy claim are therefore asserted but not proven.

## Scope and gates (all verified, all green)

| Item | Result |
|---|---|
| Files | exactly 2: `controller.ts`, `__tests__/controller.test.ts` |
| Net LOC | +200 / -0 = **exactly the 200 cap** |
| `model.ts` / `model.test.ts` | unchanged — blobs identical to base (`3848f295…`, `95641ddb…`) |
| Focused controller + reducer | 48/48 pass |
| Full frontend (`NODE_OPTIONS=--no-experimental-webstorage npx vitest run --maxWorkers=4`) | 136 files, 2464/2464 pass |
| `npm run check` | 4275 files, 0 errors, 0 warnings |
| `npm run lint` / `lint:boundaries` / `i18n:check` / `build` | exit 0 (i18n: 3 locales, 166 keys) |
| Exact-head CI | `quick` success, `grep-gate` success; `Agent Review Gate` fails only for a missing fresh PASS |

Every numeric claim in the PR body reproduces. Note: there is no CI job named `diff-check`; the actual head checks are `quick` and `grep-gate`. The body's "independent adversarial probes: 8/8" is not verifiable from the diff (no probe artifact is committed) — not counted against the PR, just not evidence I can confirm.

## Invariant audit (production code — all correct)

Audited against the full MOR-1053 contract with 10 independent probes (P1–P10) driving a deterministic virtual clock through the injected `schedule`/`cancel` hooks. All 10 pass on the unmutated head:

- audio strictly before one ON, and ON never retried/replayed/reconstructed — `controller.ts:84-91`, `#run` order at `:52-62`
- qualified OFF once/coalesced, always before local stop; synchronous OFF send failure caught at `:57-58`, reported after the batch at `:66-68`, reaching `failRelease` with the exact identity
- exact qualification on every callback: timer events carry `commandId`/`armRevision`/`leaseId`/`generation`/`originalEpoch`/`eventEpoch`/`offCommandId` (`:97-99`), command results reuse the **effect** guard rather than live state (`:76-81`) — correct
- `eventEpoch` captured at arm time from post-transition state (`:94`) is right: on an `epoch` event the reducer computes `release` on the old state and the controller then arms with the new epoch, so `matchesOff` still holds
- ACK/`response-ok` cannot manufacture RF truth (P9: `radioTx` stays `off`, `onConfirmed` stays null)
- stale callbacks inert: cancel failure leaves a live host timer, and firing it after a **reused `leaseId` with a new generation** is a no-op (P8)
- MOD restored only on the reducer obligation with exact authority evidence (`:109`); the reducer emits `restore-mod` only inside the `authority` branch, so the `cause.type` gate drops nothing

The three prior P0 blockers are genuinely remediated and non-vacuously covered — each mutant is killed by the PR's own tests:

- throwing subscriber suppressing OFF/stop → per-listener `try/catch` at `:43`; removing it **fails 1 test**
- cancel exception suppressing the remaining batch → per-cancel `try/catch` at `:106`; removing it **fails 1 test**
- schedule exception leaving a deadline unarmed → `catch { fire(); }` at `:102`; removing it **fails 1 test**, and all three timer families are covered independently by the row at `controller.test.ts:64-70`

Mutation sweep: 30 mutants of `controller.ts`, **15 killed** by the PR tests (incl. eventEpoch, armRevision, commandId qualification, OFF correlation on both the dispatch and arm paths, cleanup-lineage cancellation, double/absent `stopLocalAudio`, restore-mod duplication).

## Blocking findings

### B1 — Acceptance criterion 1 is only half-proven: no test ever fails audio preparation

`startAudio(): Promise<string | null>` — a **non-null resolved string is the declared audio-failure signal** (`controller.ts:88`). The harness parameterises it (`controller.test.ts:9`) but every call site passes `Promise.resolve(null)` or a never-settling promise. No test supplies an error string, a rejection, or a synchronous throw. The suite proves `['audio']` precedes `['audio','on']` on the happy path only — which any implementation that calls `startAudio` first and then keys **regardless of outcome** also satisfies.

Reproduce (survives 48/48 focused *and* 2464/2464 full frontend):

```
# controller.ts:88-91 — treat an audio error as ready
-  void pending.then((error) => this.dispatch(error
-    ? { type: 'fail', guard, fault: 'audio-failed', offCommandId: ... }
-    : { type: 'audio-ready', guard, commandId: ... }),
+  void pending.then(() => this.dispatch({ type: 'audio-ready', guard, commandId: ... }),
```

That mutant dispatches PTT ON after audio preparation has failed — the exact hazard criterion 1 exists to prevent — and no test notices. Same family, also surviving 48/48:

- delete the rejection handler at `controller.ts:91` → rejection becomes an unhandled rejection, no `fail` event
- replace the `startAudio` synchronous-throw catch at `controller.ts:87` with `catch { return; }`

Required: rows asserting that for (a) a resolved non-null error string, (b) a rejected promise, (c) a synchronous throw — `sendPtt` is never called, state reaches `phase: 'failed' / fault: 'audio-failed'`, local stop runs, and no deadline remains live. My probes P1/P2/P3 kill all three; the production code passes them unchanged, so this is purely a test addition.

### B2 — the "reentrancy-safe event queue" claim has zero coverage; no synchronous or reentrant report is exercised

`harness.sendPtt` (`controller.test.ts:14`) only stores `report`; every report is invoked manually after `dispatch` has fully unwound. The queue at `controller.ts:34-47` is therefore never entered reentrantly.

Reproduce (survives 48/48):

```
# controller.ts:36
-  if (this.#processing) return;
+  if (false) return;
```

Under that mutant three of my probes fail: a synchronous OFF `transport-error` report processes `failRelease`'s `cancel-timers` **before** `arm-off-timeout` runs, leaving an orphan off-confirmation deadline live (P6 expects 0 live timers, observes 1); a subscriber that re-enters `dispatch` reorders the effect log (P7); the `startAudio` sync-throw path reorders (P3).

Related gap in the same row: `sendPtt` synchronous failure is only exercised for OFF (`command === 'off' && failure === 'off'`, `controller.test.ts:14`). The **ON** synchronous send failure — `controller.ts:57-58` → `release` with `on-command-failed` → OFF then stop — has no test at all.

Required: one synchronous-report row (ON `sent` reported inside `sendPtt`; OFF `transport-error` reported inside `sendPtt`, asserting OFF-once, stop-once and zero live timers) and one ON synchronous-send-failure row.

### B3 — two guard branches uncovered (cheap to close)

Both survive 48/48; both are killed by my probes:

- one-shot `fired` guard, `controller.ts:97`: deleting `if (fired) return;` survives. A `schedule` hook that invokes the callback and *then* throws produces two `timer-fired` dispatches (probe P10 asserts exactly one at the dispatch boundary).
- direct-guard branch of the cancel predicate, `controller.ts:104`: reducing it to `sameGuard(timer.cancelGuard, guard)` survives — no test needs the `timer.guard` branch. The cleanup-lineage branch **is** covered (dropping it fails a test), so only this half is unproven. Probe P6 kills it.

## Residual assumption — adjudicated, not blocking

The PR states `commandId`, `stopLocalAudio` and `restoreMod` are total synchronous DI hooks. Neither issue #2060 nor MOR-1053 requires adapter failure policy for them, and the enumerated failure modes the contract *does* require (OFF dispatch sync failure, scheduling, audio, stale callbacks) are all handled. I accept the assumption. Measured blast radius, for the record:

- **`commandId` throwing** while arming the ON deadline (`controller.ts:95`, the 4th mint on the ON path) leaves the observed state `phase: audio-start-pending`, `mayOwnKey: true`, effect log `['audio','on']`, **0 live timers** — ON already on the wire with no confirmation deadline, and the exception escaping the `.then` at `:88` as an unhandled rejection. Not reachable for a counter/UUID minter (which fails deterministically on the first mint, long before any ON), so this stays a residual limit. Non-blocking hardening: minting the `off` id lazily inside `fire`, or arming before sending, removes the window for free.
- **`stopLocalAudio` throwing** during the backend-dekeyed obligation aborts the effect loop at `controller.ts:108`, so the following `restore-mod` is skipped (`restoreMod` calls = 0) and the throw escapes `dispatch`. Not RF-unsafe — the backend has already de-keyed — and MOD ownership stays conservative in the adapter per the contract.
- **`restoreMod` throwing** loses nothing: `restore-mod` is last in both reducer emissions, and `#processing` is reset in `finally` so queued events drain on the next dispatch.

Also noted, non-blocking: the `finally` compensator at `controller.ts:64` is unreachable under the declared assumptions (every throwing effect on an OFF batch is individually caught, and `stopped` is already set when `stopLocalAudio` itself throws) — deleting it survives 48/48. The criterion is satisfied by the inline catch plus normal continuation, so this is dead-code observation only. `dispatch` also notifies subscribers on no-op transitions where `transition` returns the same state object; benign, but worth knowing before wiring presentation.

Mutants that survive even my probes and that I judge unreachable/equivalent under current reducer semantics — hardening only, **not** required for this PR: cross-lease `cancel-timers`, `sameGuard` dropping `generation` or `authorityEpoch`, `originalEpoch`/`generation` sourced from live state in `#send`/`#commandResult`, and the fired-record `delete` at `:97`.

## Budget note

The PR sits at exactly 200/200 net LOC. Closing B1–B3 in the existing dense style is a handful of lines but will exceed the cap. Please either compress within the two owned files or get an explicit budget decision from the ticket owner — do not silently exceed 200.

```
NODE_OPTIONS=--no-experimental-webstorage npx vitest run src/lib/runtime/tx-controller
NODE_OPTIONS=--no-experimental-webstorage npx vitest run --maxWorkers=4
npm run check && npm run lint && npm run lint:boundaries && npm run i18n:check && npm run build
git diff --numstat main...HEAD    # still 2 files, <=200 net, model.ts/model.test.ts untouched
```
Then confirm each B1/B2/B3 mutant above now **fails** at least one test, and re-request an independent exact-head review.

## Limits of this review

Verified only from the two owned files, the merged reducer, and repo-local gates at the exact head above. No claim is made about App integration, browser or cross-surface parity, production transport/audio adapter behavior, capability policy, protocol, or hardware. No hardware evidence.

morozsm:
Agent Review: PASS

Fresh independent final TX-safety/lifecycle review of exact head `c5b8a0b46714f68b59115b0a83c1f3103fb5b54e` against exact base `3ffde9d74d67fffab69da9cd9e999eef52de8eb9` (merge-base confirmed identical to base; `gh pr view 2061` reports exactly this head/base pair). All prior B1–B3 findings from [the earlier BLOCKED review](https://github.com/rigplane/rigplane-core/pull/2061#issuecomment-5110037883) are **non-vacuously closed**, and I found no production defect. No repository file, branch, or ref was modified by this review; `git status --porcelain` is empty before and after all probes, and every mutation ran in an isolated mirror at `/tmp/mor1053-mirror` (rsync of `frontend/` with `node_modules` symlinked, hard 180 s per-probe timeout).

## Scope and gates (all verified at the exact head)

| Item | Result |
|---|---|
| Files changed | exactly 2: `controller.ts`, `__tests__/controller.test.ts` |
| Net LOC | +200 / −0 = **exactly the 200 cap** (`controller.ts` +110, test +90) |
| `model.ts` / `model.test.ts` | unchanged — blobs identical to base (`3848f295…`, `95641ddb…`) |
| Focused controller + reducer | **51/51** pass |
| Full frontend (`--maxWorkers=4`) | **136 files, 2467/2467** pass |
| `npm run check` | 4275 files, **0 errors, 0 warnings** |
| `npm run lint` / `lint:boundaries` / `i18n:check` / `build` | exit 0 (i18n: 3 locales, 166 keys; build ✓ in 3.36 s) |
| CI at exact head | `quick` **success**, `grep-gate` **success**; `Agent Review Gate` fails only for the missing fresh PASS this comment supplies |

Every numeric claim in the PR body reproduces. One correction to the body for the record: "*latest remediation changes tests only; production controller behavior is unchanged from `4a90aef3`*" — the production file **did** change by one line, but it is the removal of a single blank line (`controller.ts` 111 → 110). Behavior is genuinely identical; the deletion is a budget maneuver that, together with 10 blank lines compressed out of the test file, funds the three new test rows while holding the total at exactly 200. The prior review's budget note was respected — the cap was not exceeded.

## B1–B3 closure — differential mutation experiment

Because production code is byte-identical to the previously reviewed head modulo that blank line, closure is decidable by a clean controlled experiment: run the **same mutants** against the prior head's 48-test file and against this head's 51-test file. Attribution to the new rows is then unambiguous.

| Mutant of `controller.ts` | prior tests (48) | this head (51) |
|---|---|---|
| **B1a** audio error string treated as ready (keys the radio *after* audio prep failed) — `:87-89` | SURVIVED | **KILLED** |
| **B1b** drop the `.then` rejection handler — `:90` | SURVIVED | **KILLED** |
| **B1c** `startAudio` sync-throw catch → `catch { return; }` — `:86` | SURVIVED | **KILLED** |
| **B1d** `startAudio` sync-throw catch → dispatch `audio-ready` — `:86` | SURVIVED | **KILLED** |
| **B2a** remove reentrancy guard `if (this.#processing) return;` — `:35` | SURVIVED | **KILLED** |
| **B2c** swallow **ON** synchronous send failure — `:57` | SURVIVED | **KILLED** |
| **B3a** drop the one-shot `if (fired) return;` — `:96` | SURVIVED | **KILLED** |
| **B3b** cancel predicate → `sameGuard(timer.cancelGuard, guard)` only (drops the direct `timer.guard` branch) — `:103` | SURVIVED | **KILLED** |
| **B3c** cancel predicate → `sameGuard(timer.guard, guard)` only (drops cleanup lineage) — `:103` | KILLED | **KILLED** |

The prior-head column reproduces the earlier BLOCKED findings exactly, including the detail that only the *direct-guard* half of the cancel predicate was unproven while the cleanup-lineage half was already covered. Both halves are now covered.

Kills are attributable to the three added rows, each of which is the failing test:

- **B1** → `controller.test.ts:61-64` drives all three audio-failure shapes (resolved non-null error string, rejected promise, synchronous throw) and asserts `sendPtt` **not called**, `phase: 'failed' / fault: 'audio-failed'`, `log === ['audio','stop']`, and `vi.getTimerCount() === 0`. Kills B1a–B1d.
- **B2** → `controller.test.ts:65-69`. The `'reentrant'` harness invokes `report(...)` **synchronously inside `sendPtt`**, i.e. while `#processing` is true — a genuinely reentrant transport report, not a manually replayed one. Same row adds the ON synchronous-send-failure case (`fault: 'on-command-failed'`, then OFF, then stop, `sendPtt` twice). Kills B2a and B2c.
- **B3** → `controller.test.ts:70-73`. A `schedule` hook that invokes the callback **and then throws** exercises the fired-one-shot against the `catch { fire(); }` compensator, asserting exactly one state notification pair at the dispatch boundary. Kills B3a. B3b is killed by the B2 row; B3c by the MOD row at `:80-89`.

I also ran a stronger reentrancy mutant not in the original finding — replacing the queue with immediate recursive processing — which is **KILLED** by three rows. The queue at `:33-46` is genuinely exercised.

## Independent production re-review (31-mutant sweep, 23 killed)

Beyond B1–B3 I re-derived every axis from `model.ts` and `controller.ts` rather than trusting the PR narrative.

- **Ordered effects** — `#run` (`:47-68`) iterates the reducer's effect array in order; reversing it fails 6 tests.
- **No ON replay** — `dispatch-on` is emitted only by the reducer's `audio-ready` branch gated on `state.onCommandId === null` (`model.ts:103`); the controller never retries, queues, or reconstructs it. Proven by the pre-key-release row (`:74-79`), the late-audio-completion row (`:30-35`), and now the B1 row.
- **OFF priority / OFF-before-local-stop** — preserved from the reducer's ordering; asserted as `['audio','on','off','stop']` in six rows. Setting the `stopped` flag on the wrong effect fails 9 tests.
- **Subscriber / cancel / schedule / send failures** — per-listener `try/catch` (`:42`), per-cancel `try/catch` (`:105`), `catch { fire(); }` (`:100-101`), and the deferred send-failure normalization (`:56-57` → `:65-67`) are each independently killed when removed. Scheduling failure is covered for all three timer families (`:54-60`).
- **Exact qualification** — `commandId`, `armRevision`, `leaseId`, `generation`, `originalEpoch`, `eventEpoch`, `offCommandId` on timer events (`:96-98`) and command results (`:77-79`). Corrupting `eventEpoch`, `armRevision`, or either `offCommandId` mint kills 2–4 tests each. `eventEpoch` captured from post-transition state at arm time (`:93`) is correct across an `epoch` event: the reducer computes `release` on the old state, the controller then arms with the new epoch, so `matchesOff` (`model.ts:52`) still holds.
- **Stale callback rejection** — command results reuse the **effect** guard, not live state (`:70`, `:76`), so a late report carries the identity of the command that was actually sent.
- **Timer lineage** — `cancelGuard = this.#state.cleanupGuard ?? guard` (`:95`); forcing it to the effect guard kills a test. Records are registered (`:100`) and removed on cancel (`:104`) — dropping either kills tests.
- **ACK cannot manufacture RF truth** — `:24-29` asserts `radioTx: 'off'` after `sent` + `ack`.
- **Cleanup / MOD authority** — `restore-mod` executes only for the reducer obligation and only under an `authority` cause (`:108`); the MOD row asserts exactly one `restore`, `cancel` call count, and zero live timers.
- **Null-assertion safety** — I traced every `effect.guard!` / `effect.commandId!` / `effect.armRevision!` back to `model.ts`. No reducer path emits `cancel-timers`, `dispatch-*`, or `arm-*` with a missing guard/commandId: `release` requires a non-null guard (`model.ts:74`), `failLocal`/`failRelease`/the two `authority` restore branches all resolve `cleanupGuard ?? guard` from states where one is non-null, and `modRestorePending` can never outlive a null `cleanupGuard` (`clearLease` sites all clear it together). `arm-audio-timeout` correctly carries `commandId: null`, which the reducer requires (`model.ts:108`).
- **Fail-safe on an escaping effect** — if an effect throws out of `#run`, `finally` resets `#processing` (`:45`) and queued events drain on the next dispatch, where exact qualification rejects any that went stale.

## Survivors — all adjudicated equivalent or reducer-defended (8 of 31)

None is a defect in the production code; each is either a semantically equivalent rewrite or is blocked downstream by the merged reducer. This matches the prior review's own adjudication of the same families.

- **Report sync send failure inline instead of after the batch** — equivalent under the queue: `#commandResult` → `dispatch` only enqueues while `#processing` is true, and the epoch is captured at throw time either way.
- **Drop the `cause.type === 'authority'` gate** (`:108`) — the reducer emits `restore-mod` only inside its two `authority` branches (`model.ts:147`, `:154`), so the gate is a redundant fail-safe that currently drops nothing.
- **`#send` correlation from live state** (`:71`) — `#state` is fixed for the duration of a batch and is exactly the post-transition state the effect derives from.
- **`#commandResult` guard from live state** (`:76`) — the only genuinely late path, but a stale report is rejected downstream by `event.commandId === state.onCommandId` / `state.pendingOff.commandId` plus the phase gates, so no observable difference. Effect-scoped identity is still the correct design and is what production does.
- **Fired record not deleted from `#timers`** (`:96`) — leaves a dead record a later matching `cancel-timers` would `cancel()` again; the `fired` flag still prevents a second dispatch. Hygiene, not RF safety.
- **Drop the `finally` stop compensator** (`:63`) — confirmed dead code under the declared assumptions: `stopped` is set when the effect is *encountered*, before it runs, and no effect that can throw precedes a `stop-local-audio` in any reducer batch. Harmless belt-and-braces; I verified this by enumerating the batches rather than assuming it.
- **`sameGuard` ignoring `generation` / `authorityEpoch`** (`:17-18`) — only widens cancellation; older-generation timers are already cancelled by the time a newer-generation `cancel-timers` runs, so it is not distinguishable. Production keeps the full triple comparison.

## Residual assumptions — accepted, non-blocking

`commandId`, `stopLocalAudio`, and `restoreMod` remain total synchronous DI hooks; neither issue #2060 nor MOR-1053 asks for adapter failure policy on them, and every failure mode the contract *does* enumerate is handled. Re-measured at this head: the `off` id for an ON dispatch is minted at `:55` **before** `#send`, so a throwing minter cannot leave an unkeyed ON on the wire — the only window is `:94`, arming the ON deadline after the ON was already sent, exactly as previously measured. Also noted, non-blocking: `dispatch` notifies subscribers on no-op transitions (`:41-42`) — benign, but worth knowing before wiring presentation.

## Merge preconditions

PASS does not authorize merging as-is: **PR #2061 is still a Draft**. It needs `gh pr ready` plus green required checks (`Agent Review Gate` should flip once this comment registers for head `c5b8a0b4`). I am reviewer-only and have not marked it ready, merged, edited, or touched Linear or issue #2060.

## Limits of this review

Verified only from the two owned files, the merged reducer, and repo-local gates at the exact head above. No claim is made about App integration, browser or cross-surface parity, production transport/audio adapter behavior, capability policy, protocol, or hardware. **No hardware evidence is claimed or required for this software/mock slice.**

PR #2066, opened 2026-07-29
Publish synchronous control-session epoch transitions

### LINKED ISSUE #2065: MOR-1155: publish synchronous control-session epoch transitions
Linear: https://linear.app/morozsm/issue/MOR-1155/core-tx-app-publish-synchronous-control-session-epoch-transitions

Part of MOR-1008. Add a synchronous ws-client control-session transition subscription carrying state and epoch.

Acceptance:
- Connected publishes only after transport epoch increments.
- Connected transition is observable before pinned OFF commands drain.
- Reconnect publishes a new epoch in deterministic order.
- Unsubscribe is leak-free.
- Seam remains dormant in this slice; no presentation or controller activation.

Scope:
- frontend/src/lib/transport/ws-client.ts
- frontend/src/lib/transport/__tests__/ws-client.test.ts
- at most 165 net LOC

Verification: focused ws-client tests, frontend lint/type/boundary checks, fresh exact-head independent review.

morozsm:
## Summary

Publishes synchronous control-session transition notifications from the canonical WebSocket control channel. Each distinct connection-state transition carries the current transport epoch.

## Why

Lifecycle consumers can now correlate their own control-session state with queued command delivery. The transition is emitted synchronously after the state handler notification and before pinned `ptt_off` delivery is drained on reconnection, preserving the de-key safety ordering.

## Scope

- Adds the public transition type and subscription seam in `ws-client.ts`.
- Covers connect, reconnect, disconnect, epoch progression, OFF delivery ordering, and idempotent unsubscribe.
- Leaves consumers dormant: this atomic slice exposes the canonical seam without changing UI behavior.

## Verification

Fresh independent verifier passed the exact uncommitted patch before publication:

- Scope: 2 files, +86 net LOC.
- Stable patch-id: `ea227a6dc234bde57d48e422e5b3cf04f05b521d`.
- Targeted frontend test suite passed; no source/test edits after verification.
- `git diff --cached --check` passed before commit.

Closes #2065

Linear: [MOR-1155](https://linear.app/morozsm/issue/MOR-1155)

### DIFF
--- a/frontend/src/lib/transport/__tests__/ws-client.test.ts
+++ b/frontend/src/lib/transport/__tests__/ws-client.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type { WsCommand, WsMessage } from '../../types/protocol';
 import type { ReceiverState, ServerState } from '../../types/state';
-import type { CommandDeliveryEvent } from '../ws-client';
+import type { CommandDeliveryEvent, ControlSessionTransition } from '../ws-client';
 
 type ServerStateWithObservation = ServerState & {
   observationSeq?: number;
@@ -510,6 +510,72 @@ describe('control channel singleton', () => {
     globalThis.WebSocket = originalWebSocket;
     vi.useRealTimers();
     vi.resetModules();
+  });
+
+  it('publishes each control session epoch before draining a pinned OFF', async () => {
+    const {
+      connect,
+      disconnect,
+      onCommandDelivery,
+      onControlSessionTransition,
+      sendCommand,
+    } = await import('../ws-client');
+    const transitions: ControlSessionTransition[] = [];
+    const order: string[] = [];
+    onControlSessionTransition((transition) => {
+      transitions.push(transition);
+      order.push(`session:${transition.state}:${transition.epoch}`);
+    });
+    onCommandDelivery((event) => {
+      order.push(`delivery:${event.commandId}:${event.eventEpoch}`);
+    });
+
+    expect(sendCommand('ptt_off', {}, 'off-1')).toBe(false);
+    connect('ws://test/api/v1/ws');
+    instances[0].simulateOpen();
+    instances[0].simulateClose();
+    expect(sendCommand('ptt_off', {}, 'off-2')).toBe(false);
+    vi.advanceTimersByTime(1300);
+    instances[1].simulateOpen();
+    disconnect();
+
+    expect(transitions).toEqual([
+      { state: 'connecting', epoch: 0 },
+      { state: 'connected', epoch: 1 },
+      { state: 'disconnected', epoch: 1 },
+      { state: 'reconnecting', epoch: 1 },
+      { state: 'connected', epoch: 2 },
+      { state: 'disconnected', epoch: 2 },
+    ]);
+    expect(order).toEqual([
+      'session:connecting:0',
+      'session:connected:1',
+      'delivery:off-1:1',
+      'session:disconnected:1',
+      'session:reconnecting:1',
+      'session:connected:2',
+      'delivery:off-2:2',
+      'session:disconnected:2',
+    ]);
+  });
+
+  it('unsubscribes the control session transition handler idempotently', async () => {
+    const { connect, disconnect, onControlSessionTransition } = await import('../ws-client');
+    const transitions: ControlSessionTransition[] = [];
+    const unsubscribe = onControlSessionTransition((transition) => transitions.push(transition));
+
+    connect('ws://test/api/v1/ws');
+    instances[0].simulateOpen();
+    unsubscribe();
+    unsubscribe();
+    disconnect();
+    connect('ws://test/api/v1/ws');
+    instances[1].simulateOpen();
+
+    expect(transitions).toEqual([
+      { state: 'connecting', epoch: 0 },
+      { state: 'connected', epoch: 1 },
+    ]);
   });
 
   it('applies optimistic data mode updates before sending', async () => {
--- a/frontend/src/lib/transport/__tests__/ws-client.test.ts
+++ b/frontend/src/lib/transport/__tests__/ws-client.test.ts
@@ -440,6 +440,66 @@ describe('WsChannel', () => {
     expect(ch.state).toBe('disconnected');
   });
 
+  it('snapshots control session subscribers for each transition', async () => {
+    const { WsChannel } = await import('../ws-client');
+    const ch = new WsChannel();
+    const order: string[] = [];
+    let added = false;
+    ch.onSessionTransition((transition) => {
+      order.push(`first:${transition.state}`);
+      if (!added) {
+        added = true;
+        ch.onSessionTransition((next) => order.push(`late:${next.state}`));
+      }
+    });
+    ch.onSessionTransition((transition) => order.push(`second:${transition.state}`));
+
+    ch.connect('ws://test');
+    expect(order).toEqual(['first:connecting', 'second:connecting']);
+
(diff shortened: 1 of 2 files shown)


morozsm:
Agent Review: PASS

Independent formal review at the exact head passed. The epoch increment precedes the synchronous connected transition, which precedes the pinned OFF drain; the regression test pins that order across reconnect N+1. Unsubscribe double-call and disconnect deduplication behavior are covered. The dormant seam has no presentation/controller consumer and no second epoch owner.

Scope: 2 files, net +86; within the approved bounds.

Checks reviewed: targeted ws-client test 39/39; lint and boundary lint clean; type check 0 errors/warnings; PR quick and grep-gate green. The broader runtime test artifact is diff-independent because its failing test wholly mocks ws-client.

Non-blocking activation-slice follow-ups: isolate throwing transition subscribers before a consumer is wired, and consider a stale onclose identity guard.

Reviewed head: 935863a642f3f294fb104eb14ec8a5a8ea1e4304

morozsm:
Agent Review: PASS

Sole formal independent review at the exact head. Verdict is based on re-derived evidence, not on the earlier PASS comment — that one was written for the parent commit `935863a6` and is stale for this head (it predates the `fix frontend reconnect reentrancy` commit, which is the substantive half of the safety story).

## Scope

2 files, +168/-1 (`frontend/src/lib/transport/ws-client.ts` +41/-0, `ws-client.test.ts` +127/-1). Within the ≤3 files / ≤200 LOC guardrail. Purely additive to the transport layer; no other layer touched.

## Contracts validated

Each contract below was verified by reading the code path *and* by an isolated mutation probe (sandbox copy at `/tmp`, repo worktree never modified — confirmed clean and byte-identical afterward). A contract only counts as pinned if breaking it turns a test red.

| # | Contract | Probe | Result |
|---|---|---|---|
| 1 | Serialized transition queue | replace queue+flag with direct dispatch | KILLED — `serializes reentrant transitions…` |
| 2 | Per-event subscriber snapshot | iterate live `Set` instead of `[...set]` | KILLED — `snapshots control session subscribers…` |
| 3 | Handlers added during dispatch excluded from the in-flight event | same as #2 | KILLED |
| 4 | Post-callback socket/state/epoch revalidation | delete the `onopen` guard | KILLED — `serializes reentrant transitions…` |
| 5 | Epoch published before pinned-OFF drain | move OFF delivery ahead of `setState('connected')` | KILLED (2 tests) |
| 6 | Reconnect N+1 epoch progression | emit `connected` before `++transportEpoch` | KILLED (2 tests) |
| 7 | Disconnect dedupe | drop the `changed` guard in `setState` | KILLED (2 tests) |
| 8 | No second owner after intentional close | force `if (true)` on the reconnect branch | KILLED (2 tests) |
| 9 | Idempotent unsubscribe | make the returned disposer a no-op | KILLED — `unsubscribes … idempotently` |

Detail on the two that carry the safety weight:

**Serialization + total order.** `_emitSessionTransition` (`ws-client.ts:92-104`) pushes to `sessionTransitionQueue` and returns early when `dispatchingSessionTransition` is set, so a reentrant `disconnect()` from inside a `connected` subscriber cannot interleave. Traced against the mock (whose `close()` fires `onclose` synchronously, so the nested path is genuinely exercised): the inner `onclose` → `setState('disconnected')` is queued, `second:connected` completes, then `first:disconnected`/`second:disconnected` drain. The test pins the full six-element order, and the snapshot is taken per event inside the `while` loop, so the `late:` subscriber added during `connecting` is absent from `connecting` and present from `connected` onward.

**Revalidation before side effects.** The `onopen` guard (`ws-client.ts:125-130`) re-checks all four of `this.ws !== ws`, `readyState !== OPEN`, `_state !== 'connected'`, `socketEpoch !== transportEpoch` *after* subscriber callbacks and *before* `_resetHeartbeat`, `_startKeepalive`, the pinned-OFF drain, the send-queue drain, and the subscribe re-send. Worth calling out explicitly: because the guard returns before line 134-135, `pendingPttRelease` is **not** cleared on the aborted path — the OFF stays pinned for the next session rather than being consumed against a dead socket. That is the correct de-key semantics, and the test asserts it indirectly (`instances[0].sent` empty, `deliveries` empty, timer count 0, one socket only).

**Dedupe.** `disconnect()` sets `this.ws = null` then calls `ws.close()`; the resulting `onclose` performs the state change, so `disconnect()`'s own `setState('disconnected')` sees `changed === false` and emits nothing. This holds under both the synchronous-mock and real-browser (async `close()`) orderings, since the second caller in either sequence is the no-change one.

## Dormant seam / single epoch owner

- `grep` over `frontend/src` for `onControlSessionTransition|onSessionTransition|ControlSessionTransition`: matches only in `ws-client.ts` and its own test. Zero production consumers — the slice is genuinely dormant as claimed.
- `transportEpoch` is mutated at exactly one site, `ws-client.ts:122` (`++this.transportEpoch` in `onopen`). No second owner; the published epoch and the `CommandDeliveryEvent` epoch cannot diverge.
- Existing `stateHandlers` behavior is unchanged: the `forEach` at line 86 still fires unconditionally, and the new emit is strictly appended behind the `changed` guard.

## Checks

- Targeted suite: **41/41** pass at this head.
- `npm run check` (svelte-check): 4276 files, **0 errors, 0 warnings**; `tsc --noEmit`: exit 0.
- `npm run lint` and `npm run lint:boundaries`: clean.
- CI at `cc1c304f`: `quick` **success**, `grep-gate` **success**. The `quick` frontend block ran (paths-filter matched exactly the 2 changed files) and reported **`Test Files 137 passed (137)`**, with `ws-client.test.ts (41 tests)` green.
- Local full-suite shows 76 failures across 7 files — these are the known local-env `localStorage` artifact (`ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`), not a regression. Diff-independent on two axes: none of the 7 files is `ws-client.test.ts`, and the only one that references the transport at all (`mod-input-auto.test.ts`) wholly `vi.mock`s `$lib/transport/ws-client`. CI's 137/137 is the authoritative baseline.

## Non-blocking follow-ups for the activation slice

None of these block merge — they are unreachable while the seam has no subscriber — but each should be closed before the first consumer is wired.

1. **Isolate throwing subscribers.** In `_emitSessionTransition`, a handler that throws escapes both loops. The `finally` clears the flag, but the remaining handlers for that event are skipped, the still-queued transitions linger in `sessionTransitionQueue` and are delivered late on the *next* transition, and the exception propagates out of `setState` into the caller. From `onclose` that means the reconnect timer at lines 186-189 is never scheduled and the channel wedges permanently. Note the pre-existing `stateHandlers.forEach` at line 86 has the identical exposure, so this is a shared hardening, not a new defect.
2. **Mirror the guard on `onclose`.** `onclose` (lines 182-190) has no socket-identity check: it unconditionally nulls `this.ws` and, when `!intentionalClose`, schedules `_open()`. That is pre-existing and untouched here, but this PR adds a second synchronous callback surface *inside* `onclose` ahead of that scheduling — a future `disconnected` subscriber calling `connect()` would get a second `_open()` and two sockets. Same asymmetry the `onopen` guard just closed.
3. **Document the epoch phase.** `epoch` on `connecting`/`reconnecting` carries the *previous* established generation (0 and 1 in the pinned expectations), not the upcoming one. That is currently pinned only by test expectation; a doc comment on the exported `ControlSessionTransition` interface would prevent a consumer from reading it as the incoming epoch.

## Merge conditions

PASS is subject to the standard gate: this PR is currently a **draft**, so it must not merge as-is. Mark it ready, then merge once required checks are green with this PASS fresh for the head.

Reviewed head: cc1c304f95cdb8a08e73559df1317fcb2ae5502b

PR #2899, opened 2026-08-31
test: mark wall-clock profiling as benchmarks instead of gates

morozsm:
## Why

`tests/test_civ_command_profiling.py::TestOperationThroughput::test_bcd_encoding_throughput` failed a full-suite run at `48935 ops/sec` against `assert ops_per_sec > 50_000`, while the host sat at load average ~47 on 10 cores. Idle, the same loop measures **922,000-1,054,000 ops/sec** — roughly 20x the floor. It did not catch a slow `bcd_encode`; it caught a busy machine, which is exactly what `pytest -n auto` produces.

## It is a class, not a test

Measured on idle hardware, threshold against measurement:

| Module | asserts | headroom |
|---|---|---|
| `tests/test_civ_command_profiling.py` | 9 | 16x - 400x |
| `tests/test_performance_regressions.py` | 5 | **7x** - 50,000x |
| `tests/test_web_audio_streaming_profile.py` | 5 | 7.8x - 588x |

Sample measurements behind those: BCD latency 1.27 µs against a 20 µs threshold; frame parsing 0.04 µs against 10 µs; frame creation 3,991,778 fps against a 10,000 fps floor; `p50` 0.21 µs against 10 µs; and in the SLO file, 0.002 ms against 100 ms, 7.019 ms against 50 ms, 2.542 ms against 20 ms.

Nothing in that range can gate. `bcd_encode` could grow fifteen times slower and still clear its floor, while a host twenty times busier fails it. Note that the tightest margin is **not** in the file that went red: `test_performance_regressions.py` clears `< 20 ms` at 2.54 ms, so the next failure of this kind was not going to be the same test.

Two of the three modules already say what they are, in their own docstrings — "profiles key radio operations to establish baseline latencies", "Measures latency and throughput". They `print` their numbers, and `-q` hides those, so the default suite runs them for their assertions alone: the half that cannot discriminate.

## What this does

- Declares a `benchmark` marker and deselects it by default, alongside the `slow` marker the repo already had: `addopts = "-v --tb=short -m 'not slow and not benchmark'"`.
- Marks the wall-clock tests: `tests/test_civ_command_profiling.py` and `tests/test_performance_regressions.py` carry a module-level `pytestmark = pytest.mark.benchmark`; `tests/test_web_audio_streaming_profile.py` marks its profiling tests individually, leaving `test_client_queue_saturation` (a correctness assertion) in the default run.
- Records the reason in each module docstring, in terms that stay true: a threshold tight enough to catch a regression is also tight enough to catch a busy neighbour, so tightening one means revisiting the marker.

Verified in both directions: **23 tests** leave the default suite and exactly those 23 collect under `-m benchmark` (8 + 7 + 8 per module, measured 2026-09-15 at the current head with `uv run pytest -m benchmark --collect-only -q`).

## What this gives up

Stated plainly, because it is a real loss: a catastrophic regression — seven times slower or worse — now ships unnoticed by the default suite. These assertions did cover that case, badly.

The alternative that keeps the guard is to measure each operation against a reference workload timed in the same process, so the ratio cancels host speed. That is a rewrite of those 23 tests rather than a marker, and worth doing separately if the coverage is wanted back.

The most contestable part of this change is including `tests/test_performance_regressions.py`, which is named for regression testing and speaks of SLOs. It is here because at 7x-50,000x headroom it does not do that, and its tightest assertion is the most exposed of the three files.

## Verification

- `uv run pytest tests/ -n auto -q --tb=short --timeout=300 --timeout-method=thread` — `11878 passed, 182 skipped, 48 xfailed`
- `uv run pytest -m benchmark --collect-only -q` over the three modules — `23/24 tests collected (1 deselected)`
- `uv run ruff check src/ tests/`, `uv run ruff format --check src/ tests/` — clean
- `uv run mypy --strict src/rigplane/web` — clean

## Guardrails

4 files, 66 insertions + 9 deletions = 75 changed lines against current `main` — under the 6 files / 600 lines soft threshold.

**Refreshed 2026-09-15:** this branch was two weeks and 566 commits behind `main`, so `origin/main` was merged into it (merge commit, no rebase). `git diff origin/main...HEAD` is still exactly the four files above. The counts in this body were corrected at the same time after an independent review found them stale.

## Review

Needs an independent agent review before merge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,12 +117,13 @@ markers = [
     "e2e: end-to-end tests for PCM API and CLI audio workflows",
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "validation_hardware: real-radio validation runs; requires explicit env opt-in",
+    "benchmark: wall-clock measurements, reported rather than gated on; deselected by default (run with -m benchmark)",
 ]
 testpaths = ["tests"]
 python_files = ["test_*.py"]
 python_classes = ["Test*"]
 python_functions = ["test_*"]
-addopts = "-v --tb=short -m 'not slow'"
+addopts = "-v --tb=short -m 'not slow and not benchmark'"
 
 [tool.ruff]
 target-version = "py311"
--- a/tests/test_civ_command_profiling.py
+++ b/tests/test_civ_command_profiling.py
@@ -10,6 +10,17 @@ Operations profiled:
 - Command queue processing
 - Response parsing latency
 - Full roundtrip (send + receive + parse)
+
+Marked ``benchmark`` and so deselected from the default suite: every
+assertion below compares a wall-clock measurement against a fixed
+threshold that idle hardware clears by more than an order of magnitude.
+That makes them measurements rather than gates -- a slowdown small enough
+to matter passes, while a host busy enough to slow the process by that
+same order fails, which is what ``pytest -n auto`` on a loaded machine
+produces. Run them deliberately with ``uv run pytest -m benchmark
+tests/`` and read the printed numbers; tightening a threshold here means
+revisiting that marker, because a threshold tight enough to catch a
+regression is also tight enough to catch a busy neighbour.
 """
 
 from __future__ import annotations
@@ -30,6 +41,8 @@ from rigplane.commands import (
     build_civ_frame,
 )
 from rigplane.types import Mode, bcd_encode
+
+pytestmark = pytest.mark.benchmark
 
 
 class TestCIVCommandProfiling:
--- a/tests/test_civ_command_profiling.py
+++ b/tests/test_civ_command_profiling.py
@@ -11,16 +11,19 @@ Operations profiled:
 - Response parsing latency
 - Full roundtrip (send + receive + parse)
 
-Marked ``benchmark`` and so deselected from the default suite: every
-assertion below compares a wall-clock measurement against a fixed
-threshold that idle hardware clears by more than an order of magnitude.
-That makes them measurements rather than gates -- a slowdown small enough
-to matter passes, while a host busy enough to slow the process by that
-same order fails, which is what ``pytest -n auto`` on a loaded machine
-produces. Run them deliberately with ``uv run pytest -m benchmark
-tests/`` and read the printed numbers; tightening a threshold here means
-revisiting that marker, because a threshold tight enough to catch a
-regression is also tight enough to catch a busy neighbour.
+Marked ``benchmark`` and so deselected from the default suite. These
+assertions compare wall-clock measurements against fixed thresholds that idle
+hardware clears many times over, which makes them measurements rather than
+gates: a slowdown small enough to matter passes, while a host busy enough to
+slow the process by that same factor fails.
+``TestOperationThroughput.test_bcd_encoding_throughput`` failed a full-suite
+run at 48935 ops/sec against its 50000 floor with the host at load average
+~47, having measured over twenty times that floor idle;
+``TestLatencyDistribution`` below records the same effect on p99 from the
+other direction. Run them deliberately with ``uv run pytest -m benchmark
+tests/ -s`` and read the printed numbers. Tightening a threshold here means revisiting the marker, because a threshold
+tight enough to catch a regression is also tight enough to catch a busy
+neighbour.
 """
 
 from __future__ import annotations
@@ -265,10 +268,11 @@ class TestLatencyDistribution:
     def test_frame_creation_latency_distribution(self):
         """Measure frame creation latency percentiles.
 
-        Gates the MOR-416 frame-building performance SLO on p50/p95, which
-        both carry a >20x margin over baseline (~0.17µs / ~0.38µs vs the
-        10µs / 30µs budgets below) and stay stable even on a contended
-        CI runner.
+        Reports the MOR-416 frame-building percentiles. It no longer gates
+        them: this module is marked ``benchmark`` and the default suite does
+        not run it (see the module docstring). p50/p95 carry a >20x margin
+        over baseline (~0.17µs / ~0.38µs vs the 10µs / 30µs budgets below),
+        which is why they held on the contended CI runner where p99 did not.
 
         p99 with only 100 samples is not a statistical percentile -- index
         99 of a 100-sample sort is literally the single worst sample, so a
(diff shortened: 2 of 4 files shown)


morozsm:
Agent Review: PASS 26c510d70f5d9b03460c0d31937a98df3bca11b2

Independent exact-head review of the merge head `26c510d7` (worktree detached at it, clean; `gh pr view 2899 --json headRefOid` returned the same SHA). Evidence pasted below.

## 1. Merge integrity — clean

`git diff origin/main...HEAD --stat` shows exactly the PR's own 4 files:

```
 pyproject.toml                            |  3 ++-
 tests/test_civ_command_profiling.py       | 25 +++++++++++++++++++++----
 tests/test_performance_regressions.py     | 17 +++++++++++++++++
 tests/test_web_audio_streaming_profile.py | 30 ++++++++++++++++++++++++++----
 4 files changed, 66 insertions(+), 9 deletions(-)
```

The two PR commits alone produce the identical stat (`git diff cf6d8855^ 365f17b3 --stat` — same 4 files, same 66+/9-), so the merge commit contributed no content: no silent reverts of `main`, no re-introductions. `git log --oneline origin/main..HEAD` is just `26c510d7` (merge) + the two PR commits. Note: `origin/main` has since advanced one commit (`f50aaf5a`, #3500); that is normal main movement and does not affect this head.

## 2. Marker behaves as claimed

Default (`addopts = "-v --tb=short -m 'not slow and not benchmark'"`):

```
tests/test_civ_command_profiling.py       → collected 8 items / 8 deselected / 0 selected
tests/test_performance_regressions.py     → collected 7 items / 7 deselected / 0 selected
tests/test_web_audio_streaming_profile.py → 1 passed, 8 deselected   (only the unmarked test_client_queue_saturation runs)
tests/test_command_union.py (unrelated)   → 1 passed                 (0 deselected — selection otherwise unchanged)
```

With `-m benchmark`:

```
civ  → 8 passed          perf → 7 passed          web audio → 8 passed, 1 deselected
```

Globally: `pytest tests/ --collect-only -q` → `16822/16849 tests collected (27 deselected)`; with `-m benchmark` → `23/16849 tests collected (16826 deselected)`. The 27 = 23 benchmark (8+7+8, grep confirms the marker exists only in the three intended files) + 4 pre-existing `slow` marks (test_audio_stress, test_mock_icom_session, test_scope_stress, test_session_lifecycle_e2e). Nothing else was silently deselected.

## 3. No correctness check was disabled

Every assert in the two `pytestmark` modules is a wall-clock/throughput threshold; all results are discarded (`_ =`):
- `test_civ_command_profiling.py` — 9 asserts, all timing gates: latency<100µs (L107), BCD<20µs (L130), parse<10µs (L163), queue<100µs (L196), roundtrip<50µs (L222), fps>10000 (L245), ops/sec>50000 (L262), p50<10µs (L305), p95<30µs (L306).
- `test_performance_regressions.py` — 7 asserts, all timing gates: <100ms ×2 (L79, L92 — the "parse" loops are no-ops), <50ms ×2 (L116, L142), <20ms ×2 (L155, L194), >1000 frames/ms (L226).
- `test_web_audio_streaming_profile.py` — marked asserts are all thresholds: p50<1ms (L70), >1M samples/sec (L107), p50<100µs (L131), >1M frames/sec (L154), p50<2ms (L282). Three marked tests assert nothing and only print (relay×2, frame_size_impact), as the docstring says. The one behavioral assert — `test_client_queue_saturation` L321 `assert full_count > 0`, deterministic bounded-queue backpressure — is deliberately NOT marked and still runs in the default suite. ✓

## 4. Mechanism proven live in both directions (mutated in place, then `git checkout --` restored)

- Removed `and not benchmark` from `addopts`: civ `8 passed` (was 8 deselected), web audio `9 passed` (was 1+8), unrelated module unchanged (`1 passed`).
- Added `@pytest.mark.benchmark` to the ordinary `tests/test_command_union.py` test: default run `1 deselected`, `-m benchmark` `1 passed`.

## 5. Prose at this head

Docstring claims checked by measurement on this host: BCD throughput 1,505,463 ops/sec vs 50,000 floor (>20x, claim "over twenty times" holds); `test_frame_overhead_acceptable` 6,279 frames/ms vs 1000 floor (6.3x, "about seven times" holds); `test_frame_encode_throughput` 8,799,269 frames/sec vs 1M (8.8x, "about eight times" holds); latency run p50=0.17µs matches the docstring's "~0.17µs"; p99=2.62µs vs p50=0.17µs demonstrates the claimed worst-sample effect; perf module indeed prints nothing; `-s` runs do print numbers. No docstring names a threshold or timing absent from the code.

The PR body, however, predates the pre-merge review commit `365f17b3` and is stale — see REQUIRED BEFORE MERGE. All threshold *values* it names still exist in the code.

## 6. CI as found (reported, not waited on)

```
Agent Review Gate  fail      Missing trusted standalone exact-head Agent Review PASS directive.
quick              skipping  https://github.com/rigplane/rigplane-core/actions/runs/35033233402/job/104596381345
```

`quick` skipped because PR #2899 is still a **draft** (quick.yml:98-102 requires `draft == false`) — by design for drafts. Its `classify` job ran at this exact head and correctly emitted `core: true, frontend: false, ci: false, docs: false` for the diff 7625924b→26c510d7, so marking Ready will start the required core run. `Agent Review Gate: fail` is expected until this directive lands.

## REQUIRED BEFORE MERGE (prose only — code verdict is PASS)

The PR body makes claims false at this head; whoever merges must correct the body (or state the corrections in the merge record):

1. "Marks all three modules with `pytestmark = pytest.mark.benchmark`" — false now: only two modules use `pytestmark`; `test_web_audio_streaming_profile.py` uses class/method marks, leaving `test_client_queue_saturation` unmarked on purpose.
2. "24 tests deselected ... 24 passed" — now 23 marked tests (default deselects 27 = 23 benchmark + 4 pre-existing slow).
3. "5 asserts" for `tests/test_performance_regressions.py` in the headroom table — the module has 7 asserts (at the original commit too; miscounted from the start).
4. "4 files, 43 insertions + 1 deletion = 44 changed lines" — now 4 files, 66 insertions + 9 deletions = 75 changed lines; still far under the 6-file/600-line soft threshold.

No docstring correction is needed: every claim added to code is true at this head.

PR #3321, opened 2026-09-07
fix(face): format scalar readings on semantic surfaces

morozsm:
## Problem

On the `sdr-test` face, a scalar reading whose declared domain is the 0..1
wire fraction was printed with `String(value)`. Observed on a live IC-7300 on
2026-09-06:

- `RFG 0.8196078431372549` on `VfoIndicatorRow` — 209/255. The same value is
  recorded in `frontend/src/lib/runtime/adapters/__tests__/fixtures/ic7300-state.json`
  at `main.rfGain`.
- `AF 0.2196078431372549` on `RxAudioSurface` — 56/255.
- `COMP ?` on `TxAuxSurface` for a toggle the radio never reported.

## Fix

Presentation only — no fact is rescaled. Which formatter applies is decided by
the field's **declared** domain, never by the value:

- RF gain and AF level text now go through `formatKnownLevel` in
  `frontend/src/semantic/format-level.ts`, the existing shared readout
  formatter (MOR-1447), which already takes the declared `[min, max]` tuple
  and only converts when that tuple is exactly `0..1`.
- `TxAuxSurface`'s unknown token `?` becomes the em dash `—`, matching the
  other semantic surfaces. `?` reads as a prompt, not as "the radio never
  reported this".

The AF slider still receives the unrescaled 0..1 fact; only the `<output>`
string changed.

| Field | Formatter | Metadata consulted | Renders |
|---|---|---|---|
| `rfGain` (`VfoIndicatorRow`) | `formatKnownLevel` | `RF_FRONT_END_LEVELS[0]` = `['rfGain', 'RF gain', 0, 1, 0.01]` — the same tuple `RfFrontEndInstrumentHost` reads for the RF-gain scalar; indices `[2]`,`[3]` give the 0..1 domain | `82%` |
| `afLevel` (`RxAudioSurface`) | `formatKnownLevel` | the 0..1 domain `RxAudioViewModel.afLevel` and the file's own rule (4) declare | `22%` |
| TX-aux toggles (`TxAuxSurface`) | none (no numeric domain) | `TX_AUX_TOGGLES` — ATU's enum plus three booleans | `on` / `off` / `—` |

Bandwidth (Hz), attenuator (dB), preamp ordinal, antenna index and RIT/XIT
offsets are left as plain numbers: their declared domains are not fractions.

## Tests

- `VfoIndicatorRow.test.ts` — the three existing current/stale/current cases
  now assert `RFG 75%` and the matching `aria-label`; two new cases pin a known
  0.8196078431372549 rendering as `82%` (and *not* containing `0.81`), and an
  unread RF gain with no display observation rendering as the em dash.
- `RxAudioSurface.test.ts` — the verbatim-fact case now asserts `42%` on the
  readout while the slider still reads 0.42; two new cases pin 0.2196078431372549
  as `22%` and an unread level as the em dash with no `%`. The file's
  import-allow-list test gains `./format-level`, and now also asserts that
  `format-level.ts` declares no imports of its own, so the closure that test
  guards stays closed.
- `TxAuxSurface.test.ts` — two new parametrised cases over `TX_AUX_TOGGLES`:
  an unobserved toggle reads `<label>: —` and contains no `?`; an observed one
  keeps its own word.
- `semantic-rx-audio-wiring.component.test.ts` — the adapter-seam case keeps
  its `aria-valuenow` 0.42 assertion and updates the readout to `42%`.

## Mutation evidence

Both mutations were applied to this head, run, and reverted; `git diff HEAD`
was empty before the final run.

- **Revert the formatters** (`rfGainText` → `String(level)`, `afText` →
  `String(f.reading.value)`, TxAux `'—'` → `'?'`): **11 failed / 299 passed**
  — 4 in `VfoIndicatorRow.test.ts`, 2 in `RxAudioSurface.test.ts`, 4 in
  `TxAuxSurface.test.ts`, 1 in `semantic-rx-audio-wiring.component.test.ts`.
- **Wrong declared domain** (`RF_GAIN_MAX` and `AF_LEVEL_MAX` set to 255, so
  `formatKnownLevel` falls through to `String`): **7 failed / 303 passed** —
  the same set minus the four `TxAuxSurface` cases, which that mutation does
  not touch.

The two new "unread → em dash" cases survive both mutations, as expected:
neither mutation changes the unknown branch. They pin that branch against a
future formatter that would render `NaN%` or `0%` for an unread field.

## Soft threshold

7 files / 120 changed lines (`git diff --numstat origin/main..HEAD`, measured
at this head) — over the 6-file soft threshold, under the hard ceiling.

This is one unit of work: a single defect class — a 0..1 declared domain
printed with `String` — on the three semantic surfaces that carry such a
reading, plus the four pinned test files that hold those surfaces' rendered
text. Splitting it per surface would leave the face inconsistent between
merges, and the wiring test asserts one of the same strings, so it cannot be
separated from `RxAudioSurface` anyway.

## Left unfixed, same class

Enumerated with `grep -rn "'?'" frontend/src --include='*.svelte' --include='*.ts'`
at this head, excluding test files.

**Same defect (0..1 fraction via `String`), deliberately not touched here:**

- `frontend/src/semantic/VfoSurface.svelte`, the `RFG` badge in the
  indicator-badge builder (lines 541–543 at this head) — the same `rfGain`
  fraction, still via `displayValue`. That file is owned by the in-flight
  receiver-deck PR; editing it here would conflict.

**Unknown token `'?'` — needs a ruling on whether `?` is a deliberate
convention before any sweep:**

- `frontend/src/semantic/RfFrontEndSurface.svelte` — `export const UNKNOWN_TEXT = '?'`
- `frontend/src/semantic/FilterSurface.svelte`
- `frontend/src/semantic/ScopeControlsSurface.svelte` (two sites)
- `frontend/src/semantic/DspSurface.svelte` (three sites)
- `frontend/src/semantic/RfFrontEndInstrumentHost.svelte`
- `frontend/src/semantic/TxAuxScalarHost.svelte`
- `frontend/src/semantic/bar-meter-projector.ts` (two sites)
- `frontend/src/components-v2/meters/smeter-scale.ts` — `'S ?'` / `'?'`
- `frontend/src/components-v2/panels/ModInputTxWarning.svelte` — `sourceLabel ?? '?'`
- `frontend/src/skins/segmentline/lcd-display-helpers.ts` (five sites) — this
  one looks like a deliberate convention rather than an oversight: it uses
  `'?'` for *unknown* and `'—'` for *unsupported* in the same expression, so a
  blanket sweep would erase a distinction. Flagging rather than changing.

Not in the class, listed so the enumeration is checkable:
`components-v2/layout/vfo-layout-tokens.ts:207` (query-string parsing),
`components-v2/layout/keyboard-map.ts:66` (a keybinding), and
`lib/runtime/commands/radio-intents.ts:92` (an optional-field name suffix).

## Gates

Run in `frontend/` at this head:

- `npm run test -- VfoIndicatorRow RxAudioSurface TxAuxSurface semantic-rx-audio-wiring` — 4 files, **310 passed, 0 failed**
- `npm run lint` — clean, no output
- `npm run check` — `4775 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS`

The full Python suite is not run locally; `quick.yml` on this PR is the record.

### DIFF
--- a/frontend/src/components-v2/wiring/__tests__/semantic-rx-audio-wiring.component.test.ts
+++ b/frontend/src/components-v2/wiring/__tests__/semantic-rx-audio-wiring.component.test.ts
@@ -289,7 +289,8 @@ describe('AF level: 0..100 becomes 0..1 exactly once, at the adapter seam', () =
     render();
     expect(Number(afSlider()!.getAttribute('aria-valuenow')))
       .toBeCloseTo(0.42, 10);
-    expect(text('af-value')).toBe('0.42');
+    // The fact stays 0.42 (the slider above); only the READOUT is formatted.
+    expect(text('af-value')).toBe('42%');
   });
 
   it.each([0, 7, 50, 100])('renders a browser volume of %i on the 0..1 scale', (volume) => {
--- a/frontend/src/semantic/RxAudioSurface.svelte
+++ b/frontend/src/semantic/RxAudioSurface.svelte
@@ -49,6 +49,7 @@
 -->
 <script module lang="ts">
   import { MOD_INPUT_SOURCES, modInputSourceLabel } from '$lib/radio/mod-input';
+  import { formatKnownLevel } from './format-level';
   import type {
     AudioFocus, ModInputReadiness, MonitorMode, RxAudioField,
   } from './radio-view-model';
@@ -78,6 +79,17 @@
   /** Honest text: an unread fact reads as unknown, never as a default. */
   export const textOf = (f: RxAudioField<unknown>): string =>
     f.reading.status === 'known' ? String(f.reading.value) : UNKNOWN_TEXT;
+  /** `afLevel`'s declared domain (rule 4 above, and `RxAudioViewModel.afLevel`
+   *  in the contract). `formatKnownLevel` reads that domain — not the value —
+   *  to decide the readout, so a 0..1 fraction shows as a percent instead of
+   *  the wire float `String()` printed. */
+  const AF_LEVEL_MIN = 0;
+  const AF_LEVEL_MAX = 1;
+  /** The AF readout. The fact stays 0..1 (rule 4); only the STRING changes. */
+  const afText = (f: RxAudioField<number>): string =>
+    f.reading.status === 'known'
+      ? formatKnownLevel(f.reading.value, AF_LEVEL_MIN, AF_LEVEL_MAX)
+      : UNKNOWN_TEXT;
 </script>
 
 <script lang="ts">
@@ -200,7 +212,7 @@
         <!-- 0..1 — the contract's OWN unit (rule 4). No rescale in either
              direction: the value in is the fact, the value out is the intent. -->
         {@render handles.afLevel()}
-        <output data-testid="rx-audio-af-value">{textOf(rx.afLevel)}</output>
+        <output data-testid="rx-audio-af-value">{afText(rx.afLevel)}</output>
       </label>
     {/if}
 
--- a/frontend/src/semantic/TxAuxFiniteHost.svelte
+++ b/frontend/src/semantic/TxAuxFiniteHost.svelte
@@ -15,8 +15,13 @@
       ? disabledReasonText(field.availability)
       : usable(field) ? undefined : disabledReasonText({ structural: true, operational: false });
 
+  /** An unobserved reading renders as the em dash, not '?': a question mark
+   *  reads as a prompt rather than as "the radio never reported this". Only
+   *  `TX_AUX_TOGGLES` reaches this — ATU's three-state enum and three
+   *  booleans — so no numeric domain is formatted here; the TX-aux levels
+   *  are rendered by `TxAuxScalarHost`. */
   const textOf = (field: TxAuxField<unknown>): string =>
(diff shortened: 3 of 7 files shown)


morozsm:
Closing: half of this landed another way while the PR sat, and the conflicts are exactly there. The branch `codex/face-format-readings` is kept and the remainder is recorded on MOR-1675.

Verified on current `main` before closing:

- Already done: `semantic/RxAudioSurface.svelte` and `semantic/VfoSurface.svelte` use `formatKnownLevel` (`semantic/format-level.ts`), which is the change this PR proposed for them.
- Still true today: `semantic/VfoIndicatorRow.svelte:165` renders `RFG {rfGainNumber(indicator.rfGain)}` from the raw 0..1 fraction, and `semantic/TxAuxFiniteHost.svelte:19,84` still shows `'?'` for an unobserved toggle.

`git merge-tree` reports conflicts in `RxAudioSurface.svelte`, `VfoIndicatorRow.svelte` and their two test files — i.e. in the files `main` already changed. The surviving defect is two sites and fits a fresh, tiny diff on current `main`, which is cheaper and safer than re-resolving this one.

PR #3480, opened 2026-09-15
fix: validate FTX-1 notch, IF shift and CW pitch against profile domains

morozsm:
## Behaviour

`YaesuCatRadio.set_manual_notch_freq`, `set_if_shift` and `set_cw_pitch` now
validate their value against the active profile's published normalized control
domain before any CAT write: the value must be an integer inside
`[raw_min, raw_max]` and on the `raw_origin + k*raw_step` lattice. Off-domain
values raise `ValueError` naming the control, its allowed range and its step,
with no wire traffic. A profile that publishes no normalized domain for a
control fails the same honest way instead of passing the value through.

- The hardcoded 300/1050 range check in `set_cw_pitch` is removed; the range,
  step and the Hz→index mapping `(freq - raw_origin) // raw_step` all come from
  the profile's `cw_pitch` domain. In-range but off-lattice values (e.g. 305 Hz)
  are now rejected instead of being silently floored.
- Existing pins keep passing: `test_yaesu_cat_poller.py` CW pitch 300→`KP00;`,
  700→`KP40;`, 1050→`KP75;`, and 299/1051 rejected with no CAT write (the new
  error message keeps the `300-1050` range form that test asserts).
- Only `rigs/ftx1.toml` targets the Yaesu CAT backend (TX-500/X6100/X6200 have
  no backend), so no other profile's setter path is affected. Icom backends are
  untouched.
- Stale "0–255" prose was deleted from the FTX-1 notch docstrings and the
  `rigs/ftx1.toml` header comment. No domain value in the TOML changed.

## Search before write

Searched `git grep -n -E "raw_step|raw_origin|on_control_lattice|quantiz" -- src/rigplane`
plus validator-shaped names (`lattice|clamp|quantiz|constrain|within|snap`) in
`web/radio_poller.py`, `web/server.py`, `core/command_service.py`, `profiles/`:
the only existing lattice check was the loader's private, load-time-only
`_on_control_lattice` (`src/rigplane/profiles/rig_loader.py`). No public runtime
validator existed, so it is reused as-is behind one new public wrapper,
`validate_control_raw_value` in the profiles layer, instead of a parallel
backend-side check.

## Files

- `src/rigplane/profiles/rig_loader.py` — public `validate_control_raw_value`
  reusing `_on_control_lattice`.
- `src/rigplane/backends/yaesu_cat/radio.py` — three setters validate first;
  stale docstring prose removed; CW pitch index derived from the domain.
- `rigs/ftx1.toml` — comment-only deletion of the stale "(0-255, …)" claim.
- `tests/test_ftx1_radio.py` — domain acceptance/rejection cases for all three
  setters plus the missing-domain failure; `test_set_if_shift_negative` moved
  from −150 (off the 20 Hz lattice) to −160.

## Checks

- RED first: 10 new rejection/missing-domain tests failed on the pre-change
  code (acceptance pins already passed).
- `uv run pytest tests/test_ftx1_radio.py tests/test_yaesu_cat_poller.py tests/test_ftx1_control_domains.py -q --tb=short`
  → 535 passed.
- `uv run pytest` on every other tests file matching
  `git grep -l -E "set_if_shift|set_manual_notch_freq|set_cw_pitch" -- tests`
  (16 files incl. `test_radio.py`, `test_rigctld_handler.py`, `test_web_server.py`,
  `test_mor1543_cmd29_gating.py`, validation suites) → 2525 passed, 2 skipped
  (pre-existing skips).
- `uv run ruff check` and `uv run ruff format --check` on changed files → clean.
- `uv run lint-imports` → 5 contracts kept, 0 broken.
- `uv run mypy src/rigplane/backends/yaesu_cat/radio.py src/rigplane/profiles/rig_loader.py`
  → 9 errors, all reproduced identical on a clean clone of origin/main
  (f27f1132) with a fresh `uv sync --all-extras`; zero new errors.

The frontend half of these tickets and the FTX-1 hardware rerun are separate
deliveries and are not part of this PR.

Part of MOR-1680
Part of MOR-1681
Part of MOR-1682


morozsm:
Agent Review: PASS 08f3de1fa22c539837306103f5ba278f80529062

## Findings

**1. `validate_control_raw_value` vs `_on_control_lattice` — CONFIRMED correct.**
Traced `RadioProfile.controls` construction (`profiles/rig_loader.py: _RigTomlConfig.to_profile`):
`raw_min/raw_max/raw_step/raw_origin` are produced via `int(_control_number(..., integer=True))`
(`rig_loader.py:418-423`), which raises `RigLoadError` for any non-`int` TOML value — so these
fields are real Python `int`s at runtime, never `Decimal`/`str`/`float`/`bool`. The
`isinstance(..., int)` guard in `validate_control_raw_value` cannot spuriously reject a valid
domain (negative `if_shift`, non-zero `cw_pitch`/`notch` origin included — verified against
`rigs/ftx1.toml`'s `[controls.if_shift]` origin=0/step=20 and `[controls.cw_pitch]`
origin=300/step=10). New public helper is justified: `git grep -n -E "raw_step|raw_origin|lattice|quantiz" -- src/rigplane`
and a search for `clamp|quantiz|constrain|snap` in `web/radio_poller.py`, `web/server.py`,
`core/command_service.py` turn up no other public runtime validator; the only prior check was the
private, load-time-only `_on_control_lattice`, correctly reused rather than duplicated.
Layering: `uv run lint-imports` → 5 contracts kept, 0 broken.

**2. Production callers / exception path — CONFIRMED not a crash, but NARROWED: found a real
untested regression outside the diff.**
Traced the full path for a `ValueError` from the new validation: `command_service.execute`
re-raises after lifecycle bookkeeping (`core/command_service.py:217-225`) → `_execute_write`
only catches `CommandExecutionInvalidatedError` → `_cmd_set_level` has no catch → the top-level
`RigctldHandler.execute` dispatch catches it generically (`handler.py:1527-1529`,
`except ValueError: return _err(HamlibError.EINVAL)`). So a rejected write degrades to a
well-formed RPRT error, not a crashed connection — confirmed.

However, I reproduced (live, against the real `YaesuCatRadio`+ftx1 profile, `_write` stubbed)
that `rigctld/routing.py`'s `set_level` for **IFSHIFT does zero local validation** before calling
`radio.set_if_shift(round(value))`, and for **CWPITCH clamps only the Hz *range*
(300-1050), not the 10 Hz lattice**. Pre-PR, `set_if_shift` had no validation at all and
`set_cw_pitch` silently floored off-lattice values — both previously succeeded for any in-range
integer. Post-PR:
```
set_if_shift(15)   -> ValueError: if_shift must be within -1200-1200 on the 0 + k*20 lattice
set_if_shift(-5)   -> ValueError (same)
set_cw_pitch(301)  -> ValueError: cw_pitch must be within 300-1050 on the 300 + k*10 lattice
```
i.e. a hamlib client sending `L IFSHIFT 15` or `L CWPITCH 301` — previously accepted — now gets
EINVAL. This exact interaction is untested: `tests/test_rigctld_handler.py`'s `yaesu_radio`
fixture is `AsyncMock(spec=_FakeYaesuRadio)` (line 3374), so `set_if_shift`/`set_cw_pitch` are
fully mocked there and the real validation never runs through that suite — confirmed by reading
the fixture. `routing.py:353-355`'s comment ("`radio.set_cw_pitch` ... clamps to FTX-1 range
... internally") is now stale (it validates-and-rejects, it doesn't clamp).
This is a genuine risk but sits entirely in `rigctld/routing.py`, a file this PR does not touch,
and is outside its stated "backend half" scope (MOR-1680/1681/1682). Not blocking this PR — no
commit to this diff can fix it — but flagged for a prompt follow-up (align `routing.py`'s local
CWPITCH clamp to the profile lattice, add IFSHIFT bounds/lattice handling, and correct the stale
comment).

**3. Only `rigs/ftx1.toml` drives `YaesuCatRadio` — CONFIRMED.**
`backends/factory.py`'s two construction sites (`YaesuCatBackendConfig` case, and the
`_YAESU_MODELS`/`model.startswith("FT")` serial case) both call `YaesuCatRadio(device=...,
baudrate=..., rx_device=..., tx_device=..., audio_sample_rate=...)` with no `profile=` argument,
so the constructor's default `profile="ftx1"` always applies. No other profile can reach these
setters.

**4. Tests — CONFIRMED, including mutation kill.**
`uv run pytest tests/test_ftx1_radio.py tests/test_yaesu_cat_poller.py tests/test_ftx1_control_domains.py -q`
→ `535 passed`. Mutation test (on a `git archive` scratch copy, never touching the reviewed
worktree): patched `validate_control_raw_value` to a no-op returning `(0,0,1,0)` without raising,
ran the 10 new rejection/missing-domain tests via `PYTHONPATH=<scratch>/src <worktree venv>/bin/python3 -m pytest`
→ all 10 failed with `DID NOT RAISE <class 'ValueError'>`. Confirms these tests genuinely pin the
validate-before-write behavior. `git status --short` in the reviewed worktree is clean throughout.

**5. Prose — CONFIRMED true for changed lines; no stale neighbours found.**
Checked all remaining "0-255" docstrings in `radio.py` (S-meter, AF, RF gain, squelch, COMP/ALC
meters) — these are genuinely unrelated 0-255-scale controls, not leftovers of the removed notch
range. `rigs/ftx1.toml`'s edited comment (line 35) is now accurate.

**6. Mypy — CONFIRMED pre-existing (line-based check, per dispatch constraint).**
`uv run mypy src/rigplane/backends/yaesu_cat/radio.py src/rigplane/profiles/rig_loader.py` → 9
errors, at lines 94/96/98/263/268/797/1647/2533 (radio.py) and 657 (rig_loader.py). None fall
inside any hunk this diff changed (radio.py: 33,1569,1587,1593,1601-1606,1610,1619,1740-1745,
2666,2668-2676; rig_loader.py: 9,185-218) — no new mypy error on a changed line.

**7. CI.** `gh pr checks 3480`: `classify` pass, `quick` pass (run 34974349616, matches
coordinator's reported 16323 passed/220 skipped/15 xfailed), `grep-gate` pass. Agent Review Gate
is red only for the missing PASS directive, expected before this comment posts.

No defect found inside this diff itself; its own stated scope (validate-before-write for the
three FTX-1 setters, data-driven from the profile, honest failure on a missing domain) is met and
test-confirmed. The one real regression risk (item 2) lives outside the diff and is recommended
as a prompt follow-up, not a merge blocker.

PR #3482, opened 2026-09-15
fix: quantize hamlib IF shift and CW pitch to the profile lattice

morozsm:
## Behaviour

`rigctld \set_level` IFSHIFT/CWPITCH used to forward `round(value)` (IFSHIFT) or a hardcoded 300–1050 clamp (CWPITCH) while the Yaesu backend rejects any value off the profile's raw lattice with `ValueError` → `RPRT EINVAL`. Now, when the radio's profile publishes a normalized control domain (`raw_min`/`raw_max`/`raw_step`/`raw_origin` all ints), `RigctldCommandRouting.set_level` quantizes the requested value to the nearest lattice point before calling the setter, and returns `RPRT EINVAL` **without calling the setter** only when the value is outside `[raw_min, raw_max]`. Radios whose profile publishes no normalized domain (IC-7300 legacy `cw_pitch` shape, `profile=None`, mock radios) keep exactly today's code path: IFSHIFT `round(value)`, CWPITCH the existing 300–1050 clamp.

New generic public helper `quantize_control_raw_value(controls, control, value) -> int | None` in `profiles/rig_loader.py`, next to `validate_control_raw_value`; domain extraction is shared via `_normalized_raw_domain` (no duplicated range checks).

## Tie rule

A value exactly halfway between two lattice points rounds **away from `raw_origin`** (ties go to the point with the larger `abs(value - raw_origin)`), documented in the docstring and pinned in tests: if_shift 10 → 20, −10 → −20; cw_pitch 305 → 310.

## Pre-write search

`git grep -n -E "quantiz|nearest|snap|lattice" -- src/rigplane` — no existing quantizer for profile control lattices; nearest were `_on_control_lattice` (exact-lattice predicate, reused) and handler.py's ad-hoc `min(key=abs(x-db))` attenuation snapping (not profile-driven); neither fits, so the new helper was written.

## Files

- `src/rigplane/profiles/rig_loader.py` — `_normalized_raw_domain` factored out of `validate_control_raw_value` (error messages byte-identical); new `quantize_control_raw_value`.
- `src/rigplane/rigctld/routing.py` — IFSHIFT/CWPITCH quantize-or-EINVAL with legacy fallback; false "clamps ... internally" comment deleted; `_profile_controls`/`_quantize_profile_level` helpers.
- `tests/test_rigctld_level_quantize.py` — NEW: 19 tests. Helper unit tests (on-lattice unchanged, nearest, tie, out-of-range `ValueError`, legacy/absent domain → `None`, negative domain + non-zero origin, float inputs, agreement with `validate_control_raw_value`) plus routing tests through `YaesuRouting.set_level` with a radio double carrying the REAL FTX-1 profile (loaded via `load_rig(rigs/ftx1.toml).to_profile()`) whose setters validate exactly like `YaesuCatRadio`; no-domain double and the real IC-7300 profile pin the legacy path.

NOTCHF, the backend setters, and profile data are untouched. The no-domain CW clamp is unchanged.

## Checks

- `uv run pytest tests/test_rigctld_level_quantize.py tests/test_ftx1_radio.py tests/fake_rigctld.py tests/test_command_service.py tests/test_managed_tx_assembly.py tests/test_proxy.py tests/test_rigctld_key_down_backstop.py tests/test_rigctld_managed_ingress_scheduling.py tests/test_rigctld_routable_extension.py tests/test_rigctld_tx_interlock.py tests/test_usb_audio_capture_open_timeout_mor1438.py tests/test_rigctld_handler.py tests/test_rigctld_protocol.py -q --tb=short` → **1204 passed**, 23 warnings.
- `uv run ruff check` (changed files) → passed. `uv run ruff format --check` (changed files) → 3 files already formatted.
- `uv run lint-imports` → 5 contracts kept, 0 broken (`rigctld` → `profiles` is layering-legal).
- `uv run mypy src/rigplane/rigctld/routing.py src/rigplane/profiles/rig_loader.py` → 2 errors, both pre-existing on lines outside every changed hunk, proven byte-identical to `origin/main` line content (`routing.py:504` ≡ main `:481`; `rig_loader.py:714` ≡ main `:657`).

Part of MOR-2469


morozsm:
Agent Review: PASS 9f0c493daed43dde32e39a49bf2a221a06d4b5c1

Independent verifier, exact head `9f0c493daed43dde32e39a49bf2a221a06d4b5c1` (detached worktree, `uv sync --all-extras`). Read-only except this comment and one restored in-place mutation (tree clean, HEAD unchanged at end).

## 1. `quantize_control_raw_value` (src/rigplane/profiles/rig_loader.py:243) — holds

- **Tie rule**: docstring says "exactly halfway rounds AWAY from `raw_origin`". Code: `k = floor(offset + 0.5)` for `offset >= 0`, `ceil(offset - 0.5)` otherwise. For a tie at `k + 0.5` this yields `k+1`; at `-(k + 0.5)` it yields `-k-1` — away from origin in both directions. Verified live: step-20 origin-0 domain: `15 → 20`, `-15/-10 → -20`; odd step-3 domain: `4.5 → 6`, `-4.5 → -6`, `1.5 → 3`, `-1.5 → -3`; non-zero-origin domain (min 5, max 95, step 10, origin 15): tie `20 → 25`, tie `10 → 5`, `9 → 5`, `91 → 95`. Matches the pinned tests (`test_quantize_tie_rounds_away_from_origin`).
- **Negative domains / nearest**: `-29 → -20`, `-31 → -40`, `29 → 20`, `31 → 40` (tests pass).
- **Floats near boundaries**: `1199.9 → 1200`, `1189.9 → 1180`, `20.4 → 20`, `300.9 → 300`; `-1200.4`, `1200.4`, `1050.4` raise `ValueError` (range check precedes rounding). `NaN`, `±inf`, `True`, `'5'` all raise — the inline comment "NaN fails every comparison, ±inf falls outside the range" is accurate.
- **Return type**: domain fields are ints (enforced by `_normalized_raw_domain`), so `raw_origin + k * raw_step` is int arithmetic → always `int`, as the setter requires.
- The final `max(raw_min, min(raw_max, ...))` clamp is unreachable for loader-validated profiles (endpoints on the lattice ⇒ nearest point to an in-range value is in range); defensive only, and the docstring says "clamped", so prose and code agree. Not a defect.
- **Refactor equivalence**: `_normalized_raw_domain` (rig_loader.py:189) is the old inline block of `validate_control_raw_value` moved verbatim — same field order `(raw_min, raw_max, raw_step, raw_origin)`, same `Mapping`/bool/int checks, `None` ↔ the old raise. #3480's tests in `tests/test_ftx1_radio.py` pass unchanged (371 passed combined with the new file).

## 2. Routing — holds

- The quantizing `set_level` is in `class YaesuRouting:` (src/rigplane/rigctld/routing.py:151; method at :332, IFSHIFT :366-372, CWPITCH :373-387). **Yaesu-only**: `create_routing` (routing.py:486) dispatches via `RigctldRoutable`; the only implementor of `rigctld_routing` is `YaesuCatRadio` (src/rigplane/backends/yaesu_cat/radio.py:3063).
- **Icom path untouched**: the diff does not modify handler.py; the Icom fallback (`handler.py:2721-2761`, CWPITCH `round(value)` at :2740) is byte-identical to `origin/main`. `rigs/ic7300.toml` publishes the legacy `cw_pitch` shape (raw_min/raw_max only, line 1577) and no `if_shift` control at all, so even a hypothetical routed Icom takes the legacy branch — pinned by `test_cwpitch_real_ic7300_profile_keeps_legacy_clamp` with the real profile.
- **Missing/None profile**: `_profile_controls` (routing.py:69) uses `getattr(..., None)` at both hops — absent attr, `profile=None`, `controls=None` all yield `None` → quantize returns `None` → legacy path (pinned by the two `_PassthroughRadio` tests).
- **Layering**: `uv run lint-imports` → `Contracts: 5 kept, 0 broken.` (rigctld → profiles is permitted; contrast `commands/` which may not import profiles).

## 3. Duplication — none to reuse

`git grep -E "quantiz|nearest|snap|lattice|min\(.*key=abs" -- src/rigplane`: the only nearest-value code in Python is handler.py:2746-2757 (PREAMP/ATT `min(hardcoded 3-4 element dB list, key=abs)`, Icom-specific, not lattice/profile-driven) and commands/scope.py:969 `_span_index_for_hz` (explicitly exact-match, "not a nearest-value one" per its own docstring). Frontend `control-domain.ts` implements `nearest_ties_*` but is TypeScript display-side, unreachable from Python routing. No reusable Python quantizer existed; adding one is correct.

## 4. Tests — hold, mutation-checked

Real FTX-1 profile via `load_rig(rigs/ftx1.toml).to_profile()`; `_ValidatingRadio` setters call the real `validate_control_raw_value` exactly like the production setters (`YaesuCatRadio.set_if_shift` radio.py:1745, `set_cw_pitch` radio.py:2673) — a validating double, not a bare AsyncMock. Routing is exercised through `YaesuRouting.set_level` itself.

**Mutation check** (in-place edit, only the named test file run, then `git checkout --`): reverting the IFSHIFT branch to `await radio.set_if_shift(round(value))` →

```
E   ValueError: if_shift must be within -1200-1200 on the 0 + k*20 lattice; got 1300
FAILED tests/test_rigctld_level_quantize.py::test_ifshift_quantizes_nearest
FAILED tests/test_rigctld_level_quantize.py::test_ifshift_out_of_range_einval_no_call[-1210]
FAILED tests/test_rigctld_level_quantize.py::test_ifshift_out_of_range_einval_no_call[1300]
========================= 3 failed, 16 passed in 0.21s =========================
```

Restored; `git status --short` empty, HEAD still `9f0c493d…`.

**Runs** (tails):
- `uv run pytest tests/test_rigctld_level_quantize.py tests/test_ftx1_radio.py -q --tb=short` → `371 passed in 2.57s`
- `uv run pytest tests/test_rigctld_handler.py tests/test_rigctld_key_down_backstop.py tests/test_hamlib_dump_caps.py -q --tb=short` (all remaining `set_level`-referencing rigctld test files) → `432 passed, 15 warnings in 12.31s`

## 5. Lint / types

- `uv run ruff check` on the 3 files: `All checks passed!`; `ruff format --check`: `3 files already formatted`.
- `uv run mypy src/rigplane/rigctld/routing.py src/rigplane/profiles/rig_loader.py`: 2 errors — rig_loader.py:714 (`redundant-cast`, from #2674) and routing.py:504 (`no-any-return` in `create_routing`). **Both pre-existing on lines this diff did not touch** (verified against the diff hunks; :714 predates via 4431f57e). Zero errors on added lines.

## 6. Prose — true and checkable

Every added docstring/comment verified: the NaN/±inf comment (REPL-verified), the tie-rule sentence (matches code exactly), `_profile_controls`/`_quantize_profile_level` docstrings, and the test module docstring's claims (EINVAL without setter call, legacy passthrough/clamp preserved) are each pinned by an assertion.

## 7. CI as found (not waited on)

```
Agent Review Gate  fail  Missing trusted standalone exact-head Agent Review PASS directive.
quick              pass  6m26s
```

## Ticket MOR-2469 compliance

Quantize to nearest legal point ✓; explicit tested tie rule ✓; out-of-range → EINVAL without calling the setter (asserted empty call lists) ✓; no normalized domain → today's path exactly (IFSHIFT `round` passthrough incl. float `15.7 → 16`, CWPITCH legacy 300-1050 clamp) ✓; data-driven, no radio constants added ✓; real routing path with a validating double ✓; Icom unchanged ✓; NOTCHF and legacy CW clamp constants left out of scope ✓.

morozsm:
Closing without merge (owner decision, 2026-09-15).

The independent review passed the code, but the change places `from ..profiles.rig_loader import quantize_control_raw_value` in `src/rigplane/rigctld/routing.py`, and `src/rigplane/rigctld/LAYER.md` allows rigctld to depend only on `core`, `commands` and `runtime` and lists `from rigplane.profiles` as a forbidden pattern. `.importlinter`'s layers contract does not enforce that line, which is why `lint-imports` stayed green.

A mechanism audit of control-domain conversion concluded that the conversion math belongs in `profiles`, is applied at the runtime Radio boundary, and that consumers such as rigctld must not convert themselves. MOR-2469 will be redone as step 4 of that plan: rigctld gets quantized/converted values through the Radio surface. The branch `codex/rigctld-lattice-quantize` is kept as reference for that work.

Part of MOR-2469

PR #3485, opened 2026-09-15
refactor: add profiles control-domain math and migrate Yaesu CW pitch onto it

morozsm:
## Module API and semantics

New `src/rigplane/profiles/control_domain.py` (MOR-2472, step 2 of MOR-2470) is the single Python mechanism for control-domain math, operating on the **published normalized domain mapping** (canonical decimal strings, exactly what the frontend receives):

- `decode_control_domain(domain, raw: int) -> str | None` — raw → canonical display string
- `quantize_control_domain(domain, display: str) -> str | None` — display lattice posture, all five modes (`nearest_ties_down`: `remainder*2 > step` rounds up; `nearest_ties_up`: `>=`; `floor`; `ceil`; `reject`: off-lattice → `None`), result outside the axis → `None`, **no clamping**
- `encode_control_domain(domain, display: str) -> int | None` — requires `restoration == "exact"`; lookup needs full raw-lattice coverage; linear/centered results guarded by a `decode(...) == quantized` round-trip
- `validate_control_raw_value(controls, control, value)` — moved unchanged (behaviour and error text)

All arithmetic is exact: decimals parse from canonical strings only (never floats) into integer `(coefficient, scale)` pairs; rendering goes through `_public_decimal`'s canonical rule. `Number.isSafeInteger` parity: |int| must fit 2^53−1, matching the TypeScript vectors. Invalid domain/input → `None`, mirroring TypeScript fail-closed posture. Data-driven: no radio constants in code.

## TypeScript parity

Semantics of all five mappings (`identity`, `linear`, `centered`, `lookup`, `encoded`-rejecting paths) and all five quantization modes match `frontend/src/lib/radio/control-domain.ts` exactly; `tests/test_control_domain_math.py` (38 cases) ports the full TS vector suite from `frontend/src/lib/radio/__tests__/control-domain.test.ts` — same inputs → same outputs. Ported TS case list (32 TS instances; `undefined`/`null` collapse into one Python `None` parametrization in two suites):

1. round-trips a valid identity domain
2. maps linear lattice points with negative values and nonzero raw origins
3. uses {nearest_ties_down,nearest_ties_up,floor,ceil,reject} at negative half-step boundaries (5)
4. keeps supported quantization {…} valid (5)
5. keeps tiny steps and huge coefficients exact (10^300 with 1e-20 step)
6. canonicalizes mixed-scale arithmetic without signed or fractional zeroes
7. uses centered anchors independently of nonzero lattice origins
8. inverts descending lookup values and rejects ambiguous entries
9. never claims reversible encoding for unavailable restoration
10. rejects values outside the axis before quantization or encoding
11. fails closed for malformed lookup {missing, empty-object, extra-key, duplicate-raw} (5→4)
12. rejects a centered domain whose center offsets do not align
13. fails closed for invalid quantization {none, zero, object, array, empty-string, unknown-word} (7→6)
14. fails closed for off-lattice raw, malformed decimals, bad mapping and overflow (raw_max = 2^53)

Plus real FTX-1 profile cases: linear `manual_notch_freq` raw 1→"10", 320→"3200" (encode back); identity `if_shift`/`cw_pitch`; half-step ties at notch display "15" for all five modes; reject off-lattice ("15"); out-of-axis ("5", "3210") and out-of-range raw (0, 321) → `None`.

## What moved

`validate_control_raw_value`, `_on_control_lattice`, and `_public_decimal` moved from `profiles/rig_loader.py` into `control_domain.py` unchanged; `rig_loader.py` re-imports them (with `validate_control_raw_value` re-exported via `__all__`) so every existing import path — including the `rigplane.rig_loader` sys.modules shim — keeps working. No second copy remains; the only external importer (`backends/yaesu_cat/radio.py`) now imports from the new module.

## Yaesu changes

- `read_cw_pitch` derives Hz via `decode_control_domain` on the profile `cw_pitch` domain (CAT index addresses the domain's raw lattice from its first point) — no `300`/`10` literals; a missing domain or out-of-domain index raises `ValueError`.
- `set_cw_pitch` keeps `validate_control_raw_value` first, so off-lattice/out-of-range rejection keeps the exact pinned `ValueError` text ("cw_pitch must be within 300-1050 …"); the Hz→index encoding uses the validator's `(raw_step, raw_origin)`.
- Docstrings/comments stating the hardcoded mapping were deleted in `radio.py` and `observations.py` (two stale comment lines; no test pins those comments).
- Public API units unchanged: CW pitch Hz, IF shift signed Hz, manual notch RAW index 1..320.

## Explicitly NOT done (later MOR-2470 steps)

- Manual notch API unit stays RAW index; notch not decoded on read paths.
- `src/rigplane/rigctld/` untouched (charter forbids importing profiles there; step 4).
- Icom not migrated.
- TypeScript mirror pinning with shared vectors is MOR-2477.

## Files (6)

`control_domain.py` (new), `rig_loader.py`, `yaesu_cat/radio.py`, `yaesu_cat/observations.py` (2 stale comment lines — required so the mandated `git grep -E "300 \+ idx|idx \* 10" -- src/rigplane/backends/yaesu_cat` prints nothing), `test_control_domain_math.py` (new), `test_ftx1_control_domains.py` (local `_decode`/`_encode` helpers replaced by the module). `tests/test_ftx1_radio.py` needed **no** changes — all pins hold.

## Checks (counts)

- `uv run pytest tests/test_control_domain_math.py tests/test_ftx1_control_domains.py tests/test_ftx1_radio.py tests/test_yaesu_cat_poller.py tests/test_rig_loader.py -q` → **945 passed, 1 skipped** (with `test_yaesu_cat_observation_adapter.py` added for the comment edit: 1157 passed)
- `uv run ruff check` / `ruff format --check` on all changed files → clean
- `uv run lint-imports` → 5 contracts kept, 0 broken (profiles imports only stdlib + core)
- `uv run mypy` on the three source files → 0 errors on changed lines; 9 pre-existing errors verified line-content-identical against `origin/main`
- `git grep -n -E "300 \+ idx|idx \* 10" -- src/rigplane/backends/yaesu_cat` → no output

Part of MOR-2472

**Correction after review (`d93867bc`):** `_CANONICAL_DECIMAL` now ends with `\Z` instead of `$`, so a trailing newline (e.g. `"0\n"`) is rejected exactly as the TypeScript mirror rejects it; pinned in `tests/test_control_domain_math.py`.


morozsm:
Agent Review: BLOCKED dce8b20fe62e93c3870d708ac48f10996f3800b8

Independent exact-head review of dce8b20fe62e93c3870d708ac48f10996f3800b8. One required fix (a demonstrated TS/Python parity divergence); everything else verified green.

## 1. TS parity — one input class diverges (the blocker)

Mappings (identity/linear/centered/lookup; unknown `mapping` → None), all five quantization modes, restoration gating, canonical decimal parsing/rendering, negative half-steps, floor semantics, 2^53 guards, lookup direction/ambiguity, cardinality — line-by-line equivalent, and the full TS vector suite passes on both sides (vitest 32 passed; Python port green).

Divergence: a canonical decimal string with one trailing newline.

- `frontend/src/lib/radio/control-domain.ts:6` — `/^-?(?:0|[1-9][0-9]*)(?:\.[0-9]*[1-9])?$/`; JS `$` without `m` matches only at end of string. Verified with node at this head: `DECIMAL.test('0\n') === false`.
- `src/rigplane/profiles/control_domain.py:35` — same pattern, but Python `$` also matches just before a trailing newline. Measured:
  - `_parse_decimal('0\n')` → `(0, 0)` (TS: `null`)
  - `_parse_decimal('-0\n')` → `(0, 0)` — a signed zero TS excludes by the `value === '-0'` + regex pair
  - `quantize_control_domain(LINEAR, '0\n')` → `'0'` (TS: `null`); the acceptance also flows into `encode_control_domain`

No in-repo caller passes a trailing newline today, but this module's sole deliverable is exact TS parity (docstring: "must stay semantically identical"), and MOR-2477 pins the two with shared vectors — the divergence would break pinning or get baked in.

Required fix: anchor with `\Z` (`re.compile(r"^-?(?:0|[1-9][0-9]*)(?:\.[0-9]*[1-9])?\Z")`) and pin with a test (`'0\n'` → `None`).

Exact-value spot checks (≥6, ties and negative half-steps) — Python results identical to TS vectors:
- quantize LINEAR `"-0.75"`: ties_down `"-1"`, ties_up `"-0.5"`, floor `"-1"`, ceil `"-0.5"`, reject `None` (tests/test_control_domain_math.py:75-90)
- encode LINEAR `"-0.75"` → `-4`; decode `-4` → `"-1"`; encode `"0.5"` → `2`
- huge 10^300 / 1e-20 step: decode 1 → `…00000000000000000001`, encode max → `2`
- lookup: decode `2` → `"-0.5"`, encode `"-0.5"` → `2`, ambiguous → `None`
- identity round-trip `-1`/`1`; `raw_max=2**53` → `None` (comment correctly marks `Number.MAX_SAFE_INTEGER + 1`)
- FTX-1: notch `1→"10"`, `320→"3200"`, ties at `"15"` for all five modes; `cw_pitch` `300/550/1050`

Case-count audit: TS 32 instances; Python 38 collected = 30 ported + 8 FTX-1. The 32→30 collapse is exactly `undefined`+`null` → one `None` parametrisation in two suites — semantically equivalent. Every TS case has a counterpart.

Non-divergence noted: for a non-Mapping `domain`, Python returns `None` (documented fail-closed) where TS would throw TypeError — unreachable under TS types; Python-side hardening, no result conflict.

## 2. Duplication — holds

- `git grep -n -E "def (_on_control_lattice|validate_control_raw_value|_public_decimal)" -- src` → exactly one definition each, all in `control_domain.py` (lines 46, 54, 64). `rig_loader.py:31-33` imports them; `tests/test_ftx1_control_domains.py`'s local `_decode`/`_encode` were deleted and replaced by the module. No second copy.
- `radio.py:2188 _cw_pitch_hz_from_index` re-implements only a two-key usability probe (bool-guarded int isinstance for `raw_min`/`raw_step`, radio.py:2197-2205). Calling `validate_control_raw_value` instead would impose the full lattice error text on an index that is deliberately allowed to be out of domain, and the new message must name the index. Partial and justified; not a second mechanism.
- Indexing from `raw_min` (radio.py:2211 `raw_min + idx * raw_step`) matches the FTX-1 KP contract: `rigs/ftx1.toml:700-705` has `raw_min = 300`, `raw_origin = 300` — coincident; `tests/test_ftx1_radio.py:2013-2043` pins idx 25 → 550 Hz plus the round-trip table, all pass. Note: `set_cw_pitch` (radio.py:2710) still indexes from `raw_origin`; identical today, would diverge on a hypothetical origin≠min profile — pre-existing line, unchanged here, observation only.
- `int(display)` (radio.py:2219): identity decode always yields an integer string (`_text((raw, 0))`); a hypothetical fractional domain would make `int()` raise ValueError, caught by the same callers. Fail-closed.

## 3. Failure path — traced, holds (with an intended behavior change)

- Poller: `poller.py:1609-1614` — `state.cw_pitch = await radio.get_cw_pitch()` sits in `try: … except NotImplementedError: pass except Exception: logger.debug("YaesuCatPoller: get_cw_pitch failed", exc_info=True)` → one bad reading is logged and dropped; the cycle continues.
- Observations: `observations.py:1331-1333` reads via `_safe_read`, whose `except (CatParseError, CatFormatError, ValueError, KeyError)` (observations.py:1394) catches the new ValueError; with `paths=(_CW_PITCH,)` declared it calls `_raise_declared_defect` (observations.py:1414-1447), which records a `DeclaredCommandDefect` on the acquisition scheduler and re-raises; `web_startup.py:159 _abort_on_startup_defect` then aborts startup. So yes — one bad reading can abort the observation pass, but through the pre-existing declared-defect channel that already treats malformed-frame ValueErrors identically (comment at observations.py:1398-1400). Net change: an out-of-domain idx previously became silently out-of-contract Hz (80 → 1100); now it fails loudly. Matches the PR body's stated intent.

## 4. Re-exports / layering — holds

- `uv run python -c "import rigplane.profiles.rig_loader, rigplane.profiles.control_domain"` → OK; `validate_control_raw_value` importable from `rigplane.profiles.rig_loader`, `__module__ == 'rigplane.profiles.control_domain'`; the `rigplane.rig_loader` shim re-exports via `__all__`.
- `uv run lint-imports` → `Contracts: 5 kept, 0 broken.`

## 5. Tests / mutation / static checks — all hold

`uv run pytest tests/test_control_domain_math.py tests/test_ftx1_control_domains.py tests/test_ftx1_radio.py tests/test_yaesu_cat_poller.py tests/test_rig_loader.py -q --tb=short`:
```
945 passed, 1 skipped, 1 warning in 13.71s
```
In-place mutation (`>` → `>=` in `nearest_ties_down`, control_domain.py:401), only `tests/test_control_domain_math.py`:
```
FAILED tests/test_control_domain_math.py::test_maps_linear_lattice_points_with_negative_values_and_nonzero_raw_origins
FAILED tests/test_control_domain_math.py::test_uses_quantization_at_negative_half_step_boundaries[nearest_ties_down--1]
FAILED tests/test_control_domain_math.py::test_ftx1_notch_half_step_ties[nearest_ties_down-10]
3 failed, 35 passed in 0.11s
```
Restored with `git checkout -- src/rigplane/profiles/control_domain.py`; `git status --short` → clean (empty).
`ruff check` → `All checks passed!`; `ruff format --check` → `6 files already formatted`; `uv run mypy src/rigplane/profiles/control_domain.py` → `Success: no issues found in 1 source file`. TS mirror: `npx vitest run src/lib/radio/__tests__/control-domain.test.ts` → `Tests 32 passed`.

## 6. Prose — holds

Docstrings match observed behavior; the `2**53` comment correctly names `Number.MAX_SAFE_INTEGER + 1`; the two deleted stale comment lines had no test pins; `git grep -n -E "300 \+ idx|idx \* 10" -- src/rigplane/backends/yaesu_cat` → empty, as the PR body claims; PR-body counts (32 TS instances / 38 Python, undefined-null collapse) audited correct; "no second copy" verified.

## 7. CI as found

`gh pr checks 3485 --required`: `quick` **pass** (1m35s); `Agent Review Gate` **fail** — "Missing trusted standalone exact-head Agent Review PASS directive" (this review).

## Required fixes

1. `src/rigplane/profiles/control_domain.py:35` — replace the `$` anchor with `\Z` in `_CANONICAL_DECIMAL` so a trailing newline is rejected exactly as TypeScript rejects it, and pin the case with a test (`'0\n'` → `None` through `_parse_decimal`/`quantize_control_domain`).

No other findings requiring code changes.

morozsm:
Agent Review: PASS d93867bc268a959820c194821b65653b1bb00596

Independent delta review, previous BLOCKED head `dce8b20f` → new head `d93867bc`. The prior finding (Python `$` anchor matched before a trailing newline, diverging from the TypeScript mirror) is closed.

**1. Delta scope — holds.** `git diff dce8b20f d93867bc` touches exactly `src/rigplane/profiles/control_domain.py` (+5/-1) and `tests/test_control_domain_math.py` (+38). The only source change is the anchor plus a comment:
`_CANONICAL_DECIMAL = re.compile(r"^-?(?:0|[1-9][0-9]*)(?:\.[0-9]*[1-9])?\Z")` (control_domain.py:38).

**2. Reproduction — holds.** Identity domain (raw -10..10 step 1 origin 0, display "-10".."10" step "1" origin "0", quantization "reject", restoration "exact"):
```
'0\n' -> None None
'-0\n' -> None None
'1\n' -> None None
' 0' -> None None
'0 ' -> None None
'0' -> '0' 0
```
`quantize_control_domain` / `encode_control_domain` both reject every newline/whitespace variant; `decode_control_domain(domain, 0) == '0'` also verified by the suite.

**3. Remaining `$` anchors — none.** `grep -n -E "re\.compile|re\.(match|fullmatch|search)"` over `control_domain.py` and `rig_loader.py` finds exactly one regex: control_domain.py:38 (now `\Z`). `rig_loader.py` contains no regex validation at all (only prose "match" strings).

**4. Mutation check — holds.** With `\Z` mutated back to `$` in place: `uv run pytest tests/test_control_domain_math.py -q --tb=line` → 3 failed, 43 passed; all three failures are the new `test_rejects_uncanonical_surrounding_whitespace_in_display_decimals[0\n|-0\n|1.5\n]` with `AssertionError: assert '0' is None`. The new tests would fail on the old anchor. File restored via `git checkout --`; `git status --short` clean afterwards.

**5. Focused suites — holds.** `uv run pytest tests/test_control_domain_math.py tests/test_ftx1_control_domains.py tests/test_ftx1_radio.py tests/test_rig_loader.py -q --tb=short` → **772 passed, 1 skipped** (7.11s). `uv run ruff check src/rigplane/profiles/control_domain.py tests/test_control_domain_math.py` → "All checks passed!".

**6. Comment and PR-body note — true.** The added comment correctly states `$` would accept `'0\n'` (confirmed by mutation) and that TS `$` is a true end-of-input anchor: `frontend/src/lib/radio/control-domain.ts:6` has `const DECIMAL = /^-?(?:0|[1-9][0-9]*)(?:\.[0-9]*[1-9])?$/;` with no `m` flag, so JS `$` does not match before `\n` — parity restored. The PR body's "Correction after review (d93867bc)" note accurately describes this exact change.

**7. CI as found (not waited on).** `gh pr checks 3485 --required`: `quick` **pass** (1m35s) at this head; `Agent Review Gate` fail — "Missing trusted standalone exact-head Agent Review PASS directive", which this comment supplies. Coordinator should confirm both are green at `d93867bc268a959820c194821b65653b1bb00596` immediately before merge.

PR #3489, opened 2026-09-15
fix: stop auto-tune from assuming a 600 Hz CW pitch

morozsm:
## Old behaviour

`ControlHandler._cw_auto_tune` detected the CW tone, then computed the VFO
shift against `state.cw_pitch if state.cw_pitch else 600`. With the pitch
never observed by the radio, the handler fabricated a 600 Hz default and
retuned the VFO by an offset derived from a value no radio reported. The
FTX-1 declares a CW pitch range of 300–1050 Hz, so a real pitch far from
600 Hz shifted the VFO by hundreds of Hz (MOR-2482).

## New response contract

When the CW pitch is unobserved (`state.cw_pitch` unset, zero, or otherwise
not a usable positive value), auto-tune queues **no** `SetFreq` and returns:

```json
{
  "detected": 650,
  "cw_pitch": null,
  "delta": null,
  "applied": false,
  "reason": "cw_pitch_unknown"
}
```

The existing keys stay present, `applied` is `false`, and the unknown case
is stated explicitly via `cw_pitch: null` plus the `reason` string — pinned
in `test_unobserved_pitch_returns_unknown_without_retune`. No substitute
constant is used and the profile is not consulted: an unobserved value
stays unobserved.

With an observed pitch, behaviour is byte-identical to `origin/main`
(pinned in `test_observed_pitch_retunes_exactly_as_before` and the
pre-existing tests, which are unchanged).

## Frontend check

`git grep -rn "cw_auto_tune\|auto_tune" -- frontend/src` shows the browser
side only *dispatches* the command fire-and-forget
(`frontend/src/lib/runtime/commands/panel-commands.ts:562`,
`dispatchRadioIntent({ name: 'cw_auto_tune', params: {} })`); no frontend
code reads the response keys (`cw_pitch`/`delta`/`applied`). **No frontend
change needed.**

## Files and counts

- `src/rigplane/web/handlers/control.py` — +13/−2
- `tests/test_cw_auto_tune_wiring.py` — +66
- Total: 2 files, +79/−2

## Checks

- RED first: `test_unobserved_pitch_returns_unknown_without_retune[unset|unusable]`
  failed against the old code (fabricated `cw_pitch: 600`, `applied: True`).
- `uv run pytest tests/test_cw_auto_tune_wiring.py -q --tb=short` — 24 passed.
- `uv run pytest tests -k "auto_tune or cw_pitch" -q --tb=short` — 110 passed, 16603 deselected.
- `uv run ruff check` / `uv run ruff format --check` on changed files — clean.
- `uv run mypy --strict src/rigplane/web` — Success: no issues found in 27 source files.
- `uv run lint-imports` — Contracts: 5 kept, 0 broken.

Part of MOR-2482


morozsm:
Agent Review: PASS 1ba040a80af91e0d01c5f03226daef08989c8e52

Independent verifier, frozen head `1ba040a80af91e0d01c5f03226daef08989c8e52` (matches `gh pr view 3489 --json headRefOid`). Diff: 2 files, +79/−2 (`git diff --numstat origin/main...HEAD`: `13 2 src/rigplane/web/handlers/control.py`, `66 0 tests/test_cw_auto_tune_wiring.py` — matches the PR body).

**1. Guard — HOLDS.** `src/rigplane/core/radio_state.py:279` — `cw_pitch: int = 0  # Hz`. Confirmed. The only runtime writer is `state.cw_pitch = await radio.get_cw_pitch()` (`src/rigplane/backends/yaesu_cat/poller.py:1610`) with `async def get_cw_pitch(self) -> int` (`src/rigplane/core/radio_protocol.py:1757`), so values are `int` end to end; `if cw_pitch <= 0:` cannot raise for any value that type allows. A negative pitch is handled exactly like zero — unknown, no retune — which is right: the pitch domain is 300–1050 Hz (profile `controls.cw_pitch`, cf. `frontend/src/lib/radio/filter-controls.test.ts:98`), a negative Hz value is physically meaningless, and refusing conservatively fabricates nothing. Note the old code was worse than its own doc: `state.cw_pitch if state.cw_pitch else 600` is truthy for `-1`, so a negative pitch was used as a real pitch (mutation run below shows `delta: 651` for pitch `-1` under the old logic).

**2. No fabricated constant — HOLDS.** `git grep -n "600" -- src/rigplane/web/handlers/control.py` → zero hits. Nothing else in the file re-introduces a substitute constant.

**3. Response contract — HOLDS.** Old code (`git show origin/main:src/rigplane/web/handlers/control.py`): `cw_pitch = state.cw_pitch if state.cw_pitch else 600`, then unconditional `{"detected": hz, "cw_pitch": cw_pitch, "delta": delta, "applied": abs(delta) > 5}`. New unknown case keeps all four pre-existing keys (`"cw_pitch": None, "delta": None, "applied": False`) and adds `"reason": "cw_pitch_unknown"`; no `SetFreq` is queued (return precedes the `q.put`). The observed path (cw_pitch > 0) is line-for-line identical to origin/main: same `delta = hz - cw_pitch`, same `abs(delta) > 5` gate, same `SetFreq`, same return dict. The pre-existing timeout/no-detection early returns (`{"detected": None, "applied": False}`) are unchanged.

**4. Frontend — HOLDS, fire-and-forget claim verified.** Only dispatch site: `frontend/src/lib/runtime/commands/panel-commands.ts:562` — `dispatchRadioIntent({ name: 'cw_auto_tune', params: {} });` — the returned `CommandLifecycle` (`radio-intents.ts:130`) is ignored at the call site. No non-test frontend code reads `cw_pitch`/`delta`/`applied` from an auto-tune response: the `cwPitch` hits are the sidetone control UI (`set_cw_pitch`, `CwPanel.svelte:168-178`), and `ControlButtonDemo.svelte:140`'s `600` is a local demo knob default, not retune math. Nothing reads the now-null fields; no frontend change needed.

**5. Tests — PASS; mutation killed.**
- `uv run pytest tests/test_cw_auto_tune_wiring.py -q --tb=short` → `24 passed in 0.77s`
- `uv run pytest tests -k "auto_tune or cw_pitch" -q --tb=short` → `110 passed, 16603 deselected, 2 warnings in 12.67s` (warnings are pre-existing `SyntaxWarning`s in `test_command_queue_execution.py`, untouched by this PR). Both match the PR body.
- Mutation (guard removed in place, old `state.cw_pitch if state.cw_pitch else 600` restored, only the named file run):
```
E     {'cw_pitch': -1} != {'cw_pitch': None}
E     {'applied': True} != {'applied': False}
E     {'delta': 651} != {'delta': None}
E     Right contains 1 more item:
E     {'reason': 'cw_pitch_unknown'}
FAILED tests/test_cw_auto_tune_wiring.py::TestCwAutoTuneWiring::test_unobserved_pitch_returns_unknown_without_retune[unset]
FAILED tests/test_cw_auto_tune_wiring.py::TestCwAutoTuneWiring::test_unobserved_pitch_returns_unknown_without_retune[unusable]
========================= 2 failed, 22 passed in 0.14s ==========================
```
Restored with `git checkout -- src/rigplane/web/handlers/control.py`; `git status --short` → clean (no output).

**6. Lint/type — PASS.** `uv run ruff check` on both changed files → `All checks passed!`; `uv run ruff format --check` → `2 files already formatted`; `uv run mypy --strict src/rigplane/web` → `Success: no issues found in 27 source files`. Matches the PR body.

**7. Prose — TRUE.** Handler comment (control.py:2214-2217) accurately describes the code below it. Test docstrings' claims ("no `SetFreq` is queued", "the profile is not consulted" — no profile read exists anywhere in this path; "the existing keys stay present") are pinned by assertions in the same test. The test comment "+50 Hz" is arithmetically true (650−600). PR body file counts and check results match observed values exactly.

**8. CI as found (not waited on).** `quick` — pass, 6m12s. `Agent Review Gate` — fail, "Missing trusted standalone exact-head Agent Review PASS directive": expected, that gate consumes this very comment and will re-evaluate on it.

No required fixes. No REQUIRED BEFORE MERGE corrections — every PR-body claim checked out against the head.

PR #3501, opened 2026-09-15
feat: decode the manual notch and NB depth through the control-domain contract

morozsm:
First of five PRs for the MOR-2475 remainder. It adds no behaviour to any screen: it extends the existing control-domain contract and fixes one readback gap, so the later PRs have a single path to consume.

## What changes

- `frontend/src/lib/radio/filter-controls.ts`: `resolveNrLevelContract`, which bridged exact and legacy domains for NR only, is generalized into `resolveControlContract(caps, control)`. The same two arms remain — an exact profile domain decodes through `control-domain.ts`, a legacy band keeps today's proportional math — now parameterized by control key, capability tag and legacy default. No second bridge is introduced; `resolveNrLevelContract` stays as the NR-shaped caller of the generalized one.
- `frontend/src/lib/runtime/adapters/radio-view-model-adapter.ts`: the manual-notch readback becomes capability-driven. Where the radio publishes a `manual_notch_freq` domain and the field's status is usable, `dsp.notchFreq` carries that field decoded to its display unit and the group publishes `dsp.notchFreqDomain`; otherwise it reads `notchFilter` exactly as before and publishes no domain. `nbDepth` now decodes through the same contract instead of the standalone `nbDepthRawToDisplay` call.
- `frontend/src/semantic/radio-view-model.ts`: `notchFreqDomain` added to the DSP group, optional, validated by the same validator `nrLevelProjection`'s domain uses.

## Why it matters

FTX-1 shows no manual-notch reading today: the Yaesu CAT poller stores `main.manual_notch_freq` while the adapter reads `notchFilter`, which that backend never fills. No Icom profile publishes `manual_notch_freq` at all, so the capability check is what keeps Icom untouched rather than a per-model branch.

## Measured behaviour

With FTX-1 capabilities (`manual_notch_freq` raw 1..320 → display 10..3200 Hz step 10, measured from the shipping profile): raw 160 → `notchFreq = 1600`, `notchFreqDomain = {min:10, max:3200, step:10, origin:10}`; raw 1 → 10; raw 320 → 3200; an out-of-domain raw reads `unknown`. With an Icom-shaped radio (no such domain, `notchFilter = 77`): `notchFreq = 77`, no `notchFreqDomain` key — byte-for-byte today's value. IC-7610 `nb_depth` raw 0 → 1 and raw 9 → 10; a radio without the control stays structurally absent.

## Tests

[...]

PR #3500, opened 2026-09-15
fix: validate the manual-notch position through the radio control domain

morozsm:
`web/handlers/control.py` refused any `set_notch_filter` position outside `0 <= level <= 255`. That literal is the web backend deciding a radio's range: FTX-1's published `manual_notch_freq` domain is raw 1..320, so the handler rejected 256–320 and accepted 0, neither of which the radio agrees with.

The handler now resolves the position through the Radio surface:

- A radio implementing `core/radio_protocol.py: ControlDomainCapable` that publishes a `manual_notch_freq` domain accepts exactly the raw positions `decode_control_raw` answers for. FTX-1: 320 is queued, 321 and 0 are refused before anything is enqueued.
- Every other radio keeps the documented raw 0-255 integer wire contract (`docs/api/command-catalog.md` declares `set_notch_filter` as `value: int`), enforced by the existing coercion `core/command_dispatch.py: _raw_int_level_from_param` — the same function `set_rf_gain` / `set_sql` / `set_squelch` use — rather than a literal in the handler. Icom behaviour is unchanged: 0/128/255 queue, 256 and −1 are refused synchronously with nothing enqueued, exactly as on `main`.

No web→`profiles` import is added; `lint-imports` stays at 5 contracts kept, 0 broken.

## Tests

- `tests/test_handlers_coverage.py`: a `_Ftx1NotchDomainRadio` double whose `ControlDomainCapable` methods delegate to a real `YaesuCatRadio` on the shipping `ftx1` profile and whose `set_notch_filter` validates like the backend — domain top accepted, off-domain refused on the WebSocket ingress path with an empty queue; a domain-less radio refused for −1 and 256, queued for 0/128/255.
- Mutations (coordinator-reviewed builder evidence): dropping the domain check fails 3 tests; dropping the wire-contract check fails 2.

## Checks

`uv run pytest -n auto tests -k "notch" -q` → 312 passed, 6 skipped. `ruff check` / `ruff format --check` clean on the changed files. `uv run mypy --strict src/rigplane/web` → no issues in 27 source files. `uv run lint-imports` → 5 kept, 0 broken.

## Scope

2 files, +135/−10 — within the 6-file soft threshold.

First review of the earlier head found the domain-less path had stopped refusing synchronously (the request answered `ok` and the failure surfaced later in `RadioPoller._execute`); `b69a3bfc` restores the `main` behaviour through the shared coercion.

Part of MOR-2474

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3502, opened 2026-09-16
feat: drive the DSP notch surfaces and command from the published domain

morozsm:
Second of five PRs for the MOR-2475 remainder, on top of #3501. The DSP notch surfaces and the notch command stop carrying a radio range of their own and take the published domain instead.

## What changes

- `lib/runtime/props/panel-props.ts: toDspProps` carries the manual-notch domain through to the surfaces, the way it already carries `nrLevelProjection`.
- `semantic/DspSurface.svelte` (its `DSP_LEVELS` row), `semantic/DspScalarHost.svelte` (`domain()`) and `components-v2/panels/DspPanel.svelte` take the notch slider's bounds and step from that domain where the radio publishes one. A radio that publishes none keeps its own `{0, 255, 1}` constants — the view model states the domain and never a fallback, so the fallback stays with the consumer.
- `lib/runtime/commands/panel-commands.ts: onNotchFreqChange` encodes display → raw through the same `resolveControlContract(caps, 'manual_notch_freq')` before dispatching. It previously sent the slider value verbatim, which was correct only while the slider carried raw codes.

No second bridge: everything goes through the contract #3501 generalized. The renderer and the notch marker are untouched — `AudioSpectrumPanel.svelte` reads the separate `toAudioSpectrumProps`, still `rx?.notchFilter ?? 0`, which is PR-3's scope.

## Measured behaviour

FTX-1, which publishes `manual_notch_freq` (raw 1..320 → display 10..3200 Hz step 10): the slider reads 10 / 3200 / 10, raw 160 shows as 1600 Hz, a change at 1600 Hz dispatches `{value: 160}`, and 9 / 1601 / 3201 are refused. An Icom-shaped radio, which publishes no such domain: bounds 0 / 255 / 1, `notchFreq` reads `notchFilter` verbatim (77 → 77), and a change at 96 dispatches `{value: 96}` — byte-for-byte today's behaviour, pinned by its own test.

The backend agrees with what the command now sends: `web/handlers/control.py: _notch_position_from_param` (merged in #3500) expects the raw code — any position `decode_control_raw('manual_notch_freq', …)` answers for, else the documented raw 0-255 wire integer.

## Tests

New `lib/runtime/props/__tests__/notch-freq-domain.isolated.test.ts`; appends to `panel-commands.intent.isolated.test.ts` and `semantic/__tests__/DspSurface.test.ts`. Mutations run by the builder and reported with output: a surface that ignores the published domain fails 4 DspSurface cases (`expected '0' to be '10'`); a command that dispatches unencoded fails on `value: 1600` where `160` is expected.

Not separately tested: the FTX-1 arms of `DspPanel` and `DspScalarHost` — the 10-file stop condition was reached first. Their no-domain fallback stays pinned by the existing suites.

## Checks

`npm --prefix frontend run test -- src/lib/runtime src/semantic src/components-v2/panels`: 185 files, 5637 passed. `check`: 0 errors. `lint`: clean.

## Scope

10 files, +345/−11 — over the 6-file soft threshold, under the hard ceiling. (Re-measured at the merge head `fc64e935` with `gh pr view 3502 --json changedFiles,additions,deletions`; the body previously said 8 files, +343/−9, which was true of an earlier push and not of this head. The two extra files and their +2/−2 are the debt-inventory re-key and the digest re-pin described below.) One unit of work: the surfaces cannot read a domain `toDspProps` does not carry, and a slider in Hz that dispatched unencoded values would send the radio a wrong position.

## Corrections applied at merge

Two false counts in this body were corrected before merge, re-measured at the head being merged (`fc64e935`) rather than carried over from an earlier push: the Scope line said 8 files / +343/−9 where the head has 10 / +345/−11, and the stop condition was stated as 8 files where it was 10. The difference is the two commits that answer the control-feedback debt guards — `4dfacbe5` re-keys one baseline entry after the `DspSurface.svelte` value expression changed, `fc64e935` re-pins the sha256 that entry's rename moved. Both were verified independently at this head: the baseline holds 39 entries, 39 unique, sorted and frozen, and recomputing the digest the way `control-feedback-debt-baseline.test.ts` computes it reproduces `04ba30ba57aa763fadd842e29b8d576ad7f1b078ca246275373f469681ebf38b`. The debt inventory did not grow — the re-key renames one existing site, it does not admit a new unlabelled control.

Part of MOR-2475


morozsm:
Agent Review: PASS fc64e935476560c40a82782880ff722eade80032

Independent exact-head review. Head confirmed with `git rev-parse HEAD`. CI as found: `quick` passed (5m19s), `visual` passed (1m41s); not rerun here. Focused checks I ran myself are listed below.

## Frozen-artifact audit (MOR-1713/MOR-1714) — verified clean

- Ran the scanner directly (`node`, `scanRepository()` + `auditSvelteSource()` against both the merge-base and head versions of `DspSurface.svelte`): the old file has exactly ONE audited site (`nr?.value ?? numberOf(dsp[field], min)`), the new file exactly ONE (the re-keyed `nr?.value ?? numberOf(dsp[field], notch?.min ?? min)`). This is a genuine rename of the same slider's value expression — no additional unlabelled control was introduced or hidden.
- Baseline recomputed: 39 entries, 39 unique, sorted (`new Set(baseline).size === 39`, sort check). Digest recomputed the way the test computes it (`sha256` over `join('\n')`) reproduces `04ba30ba…ebf38b` exactly. Scanner's radio-backed identities are a subset of the frozen set (no growth).
- `control-feedback-debt-inventory.test.ts` passes when given a real timeout (its default 5s shard timeout tripped on my machine only; 12/12 pass at 60s) and `control-feedback-debt-baseline.test.ts` passes.

## Command encode (panel-commands.ts onNotchFreqChange) — verified

- No domain: `displayDomain === null` → gate stays `knownReceiverField('notchFilter', …)` with `contract.receivers ?? undefined`, which resolves to the same receiver count the old omitted argument derived from caps — behavior-identical to the merge-base. Verbatim dispatch pinned by the Icom test and the pre-existing tests.
- Domain: `displayToRaw` requires a safe integer, `quantizeControlDomain` with `quantization: 'reject'` refuses off-lattice values (1601/9/3201 pinned), and `encodeControlDomain` verifies the decode round-trip before returning. Domain ends are exact (10↔1, 3200↔320 pinned). Backend admission (`_notch_position_from_param`, control.py:296-319) matches the test comment's claim.
- `min == max` cannot reach the new branches: `exactControlDisplayDomain` fails the cardinality/probe round-trip and `legacyControlRange` rejects `display_max <= display_min`, so a degenerate domain degrades to the raw fallback everywhere (DspSurface/DspScalarHost additionally re-validate `max > min`).

## Consumers / unit-mismatch sweep — no reachable mismatch introduced

- `toDspProps.notchFreq` consumers: DspPanel (binding value is feedback-based, not `p.notchFreq`), AmberCockpit's scope marker (`AmberAfScope`), MobileRadioLayout/AmberScope (mode only). The AmberCockpit marker now receives Hz where a domain exists while still normalizing raw — that is exactly the marker-geometry decision recorded on MOR-2475 as PR-3's work; not demanded here.
- DspPanel's `notchPositionBinding` and DspScalarHost's `notchFreq` binding take the Hz domain but keep a wire-raw `getDspControlFeedback('notchFilter')` lane. Not reachable today: the only domain publisher (rigs/ftx1.toml) observes `manual_notch_freq`, never `notch_filter`, so those bindings stay unavailable exactly as before this PR. Same for DspPanel's hbar fine-step (1 Hz quantum vs `reject` quantization). Recommend recording on MOR-2475 that PR-3/PR-4 must add the display-projection feedback lane before any profile publishes both fields.

## Test discrimination

- `notch-freq-domain.isolated.test.ts`: decodes + domain key FAIL under the old `toDspProps` (old impl yields 0 and no key). Discriminates. The Icom no-domain case is a regression pin by design.
- `panel-commands.intent.isolated.test.ts`: the FTX-1 encode case (1600→160) and the off-domain refusals FAIL under the old verbatim dispatch. Discriminates. The no-domain case is a regression pin.
- `DspSurface.test.ts`: the bounds cases (min 10 / max 3200 / step 10 / decoded value) FAIL under the old row constants (0/255/1). The emit-verbatim case pins the display-unit wire contract the encode depends on. Discriminates.

## Prose

Every added comment was checked against code, `control.py`, and `rigs/ftx1.toml` (fixture mirrors the profile exactly). No sentence found that could be false without a test failing. No REQUIRED BEFORE MERGE items.

## Checks I ran (this worktree, focused only)

- `npm ci` (fresh node_modules per repo rule), then: scanner/digest node audit above; `npx vitest run` on the two debt-guard tests (with `--testTimeout=60000` for the inventory one), the three new/changed test files (223 passed), and `DspPanel.component.test.ts` + `mor1560-dsp-family-conformance.isolated.test.ts` (60 passed).

Layer rules respected (no new imports across boundaries, no new abstractions — `resolveControlContract` reused); 10-file limit untouched.

PR #3503, opened 2026-09-16
fix(MOR-2228): open serial ports with DTR/RTS held inactive

morozsm:
## What

Every serial port in rigplane is now opened through one helper,
`rigplane.core.serial_open.open_serial_port`, which keeps the modem control
lines inactive.

pyserial 3.5 defaults to `dtr=True, rts=True` on open.

**This is a precaution, not a live hazard.** At the radio's default settings
neither line keys anything — Icom `SET -> Connectors -> USB SEND` ships as OFF
(owner, 2026-09-16). It matters only on a station that has deliberately pointed
USB SEND at DTR or RTS, which is the usual WSJT-X/fldigi arrangement and a
configuration we support. On such a station, asserting either line keys the
transmitter.

Before this change nothing in `src/` ever set or cleared those lines: `rg "\bdtr\b|\brts\b|dsrdtr|rtscts" src/ --type py -i`
returned exactly one hit, a comment in `hamlib_bridge.py`.

The helper applies both halves of the mitigation:

1. It builds the `Serial` instance with `serial_for_url(..., do_not_open=True)`,
   sets `.dtr = False` / `.rts = False` while the instance is still **closed**,
   then opens it and wraps it with `serial_asyncio.connection_for_serial`.
   pyserial stores those as `_dtr_state` / `_rts_state` and applies them in
   `Serial.open()` (verified by reading `serial/serialposix.py`: `open()` calls
   `_update_dtr_state()` / `_update_rts_state()` right after
   `_reconfigure_port`; `_reconfigure_port` itself does not).
2. It deasserts both lines again on the live instance after the open.

Which half is strictly necessary is **unestablished** and is checked on the
bench — see "Not covered" below. One reason to keep both: pyserial's own
`open()` swallows `EINVAL`/`ENOTTY` from `_update_*_state`.

### History of this PR: the first commit was wrong

`008f0a85` passed `dtr=False, rts=False` as **kwargs** to
`open_serial_connection`. Independent review blocked it: pyserial 3.5 does not
accept `dtr`/`rts` as constructor arguments — they are post-construction
properties. They fall into `**kwargs` and raise
`ValueError: unexpected keyword arguments` inside `Serial.__init__`, before any
device access. On real hardware **every** open would have failed: discovery
silently finding no radios, Yaesu never connecting, the CI-V link never
connecting.

Every test stayed green because every test injects a fake opener that accepts
any kwargs. `f1f7366c` replaces the mechanism and adds the test that closes that
gap (see Tests).

## Why one helper rather than four fixes

The rule is a property of opening a port, not of any one caller. A rule that
has to be remembered at each call site is a rule that will be missed at the
next one. All four open sites now delegate:

| Site | File |
|---|---|
| Icom CI-V link | `backends/icom7610/drivers/serial_civ_link.py` |
| Yaesu CAT transport (also serves the Yaesu probe) | `backends/yaesu_cat/transport.py` |
| CI-V discovery probe | `backends/discovery.py` |
| Xiegu model-ID probe | `backends/discovery.py` |

The helper lives in `core/` because all four call sites are in `backends/`,
which sits above `core/` in the import-linter matrix.

Note for reviewers: the ticket described the fourth site as the Kenwood probe.
That was wrong — `probe_serial_kenwood_cat` is a stub that returns `None` and
never opens a port. The fourth real site is `probe_xiegu_model_id`.

## Tests

Thirteen tests in `tests/test_serial_open.py` pin both halves independently,
per call site, plus the tolerance cases (a writer with no `.serial`; a `.serial`
whose setters raise).

The one that matters most is `TestRealPyserialConstruction::test_real_pyserial_construction_with_helper_kwargs`:
it runs the **real** pyserial construction path with no hardware
(`serial_for_url('loop://', do_not_open=True)`), which is the only kind of test
that can catch the class of bug the first commit shipped. A fake opener never
can. (A full end-to-end open over `loop://` is not possible: `SerialTransport`
calls `serial.fileno()`, which `loop://` does not support.)

Mutation results, each restored afterwards:

| Mutation | Reddens |
|---|---|
| drop `instance.rts = False` (pre-open) | `TestRealPyserialConstruction::test_real_pyserial_construction_with_helper_kwargs` |
| drop `instance.dtr = False` (pre-open) | same test |
| drop the post-open deassert | `TestOpenSerialPort::test_deasserts_lines_on_the_real_serial_object` + 5 others |
| drop the whole pre-open configuration | `test_real_pyserial_construction_with_helper_kwargs` |

The last mutation was reproduced independently by the coordinator: 1 failed,
12 passed, tree restored clean.

## Gates

Baseline on the five affected test files at the merge base `ca8ef0d3`: 224 passed.

- `uv run pytest tests/test_discovery.py tests/test_serial_civ_link.py tests/test_yaesu_cat_transport.py tests/test_serial_backend_smoke.py tests/test_icom7610_serial_radio.py tests/test_serial_open.py -q` -> **237 passed** (224 baseline + 13 new)
- `uv run ruff check src/ tests/` -> All checks passed!
- `uv run ruff format --check src/ tests/` -> 749 files already formatted
- `uv run lint-imports` -> Contracts: 5 kept, 0 broken

## Not covered by this PR

This is **defence in depth against a code gap, not a fix for a reproduced
failure**, and it is not reachable at all in a default radio configuration
(see above). MOR-2228's own correction records that the transmit events once
seen on the bench were the radio's VOX, not this mechanism.

The bench verification the ticket asks for — owner present, dummy load, USB
SEND set to RTS and then DTR, confirming zero TX on open, on start-up and on a
cable replug — is **not** done here and cannot be done in code. It is tracked
separately; that check is also what settles which half of the mitigation is
load-bearing.

## Size

7 files, +498/-32 (530 changed lines). Above the 6-file soft threshold by one
file — a two-line note in `core/LAYER.md` recording where the new transport
primitive lives — and under the 600-line soft threshold. One unit of work: one
rule, one helper, its four call sites and their tests.

Production code is 53 lines of the total; the rest is tests and docstrings.

## Also in this change: two exception swallows at the open sites

Owner instruction, 2026-09-16: exceptions must not simply be swallowed.

`discovery.py`'s CI-V and Xiegu probes caught `except Exception` and logged
`"cannot open %s @ %d"` at debug **with no exception detail**. That is precisely
what would have hidden the blocker above on real hardware. Both now separate
expected from unexpected: `OSError` (which pyserial's `SerialException`
subclasses — verified via MRO) stays quiet and returns the existing sentinel,
because a busy or absent port is normal during a scan; anything else is logged
at warning with the exception type, message and traceback, and is not reported
under a message naming a different cause.

`_deassert_control_lines` likewise moved from debug to warning: a control-line
write failing on a TX-safety path must be visible at default logging. It still
swallows rather than raises, because a virtual or Bluetooth port that refuses
the write is still a usable port.

One behaviour change worth review: an `ImportError` for a missing
`pyserial-asyncio` used to propagate out of the CI-V probe; it is now a warning
and a `None` return. No CLI caller caught it, and `enumerate_serial_ports` fails
earlier if pyserial itself is absent.

The other four blind swallows in `discovery.py` (probe response handling and USB
audio resolution) are **not** touched here — they are enumerated and ticketed
separately, per the rule that a review of one instance is a review of its class
but scope is not widened to fix it.

## Corrections this PR made to its own ticket and spec

- MOR-2228 prescribes the kwargs mechanism. That prescription is wrong (see
  History above).
- MOR-2228 describes the fourth open site as the Kenwood probe.
  `probe_serial_kenwood_cat` is a stub returning `None` that never opens a port;
  the fourth real site is `probe_xiegu_model_id`.

## Mechanism audit

Ran at `f1f7366c` (auditor role, method from `.claude/skills/mechanism-audit/SKILL.md`).

**Cleared.** There is exactly one way to obtain a serial connection in this tree:
the only construction of a pyserial object anywhere in `src/` is inside this
helper. `core/transport.py` is UDP/LAN; `rigctld_client`, `dx_cluster` and
`cli/_validate` use TCP; `hamlib_bridge` spawns `rigctld` with a TCP endpoint and
never opens a device node; `enumerate_serial_ports` only enumerates. Placement in
`core/` was judged correct on two independent grounds: `core` already owns the
other transport, and the four consumers span three unrelated backend families, so
any home under `backends/` would make one family the owner of the others' safety
rule.

**Acted on here** (`c9efb436`):

- The `core/LAYER.md` sentence claiming this helper is "the only way backends open
  a port" was false — `SerialCivLink(open_serial_connection=...)` bypasses it
  entirely. Narrowed, with the exception named. (The seam is test-only: the sole
  production construction, `_icom_serial_base.py` ~line 213, passes no opener.)
  Note the contrast: `discovery.py`'s two `opener=` seams do **not** bypass
  anything, because the injected opener is consumed *inside* `open_serial_port`.
- Deleted a stale comment in `discovery.py` naming `_default_open_serial`, removed
  earlier in this branch.
- Deleted the `SerialOpener` type alias: read nowhere outside its own module and
  character-for-character identical to `discovery.py`'s `_OpenSerial`. Safe
  despite being in `__all__` because `core/serial_open.py` does not exist on
  `origin/main`, so no out-of-repo consumer can exist.

**One audit verdict reversed.** The audit called
`_deassert_control_lines`'s `serial is None` branch dead in production, since the
installed `SerialTransport` always exposes `.serial`. But `pyproject.toml` pins
`pyserial-asyncio>=0.6` — a floor, not an exact version. A later release that
renames or drops that attribute would break every open; this branch is what keeps
the pre-open configuration working in that case. Kept, with the reason recorded in
the docstring so the next reader does not delete it.

**Ticketed, not fixed here:** MOR-2495 (retire the `SerialCivLink` seam, which
would let the stronger LAYER.md sentence come back), MOR-2496 (two test-injection
mechanisms for one job — the `opener=`/`_open_serial` chain has no production
caller at any level; plus `probe_serial_kenwood_cat`, a stub with four dead
parameters), MOR-2494 (the other four blind swallows in `discovery.py`).


morozsm:
Agent Review: BLOCKED 008f0a851a1d318d3b8620b065312c2ead01613e

## Blocker: the open-time kwargs half raises `ValueError` on every real port open (pyserial 3.5)

`rigplane.core.serial_open.open_serial_port` (src/rigplane/core/serial_open.py:49) forwards `dtr=False, rts=False` into `serial_asyncio.open_serial_connection` → `create_serial_connection` → `serial.serial_for_url` → `Serial(None, **kwargs)`. **pyserial 3.5's `SerialBase.__init__` does not accept `dtr`/`rts` keyword arguments** — they fall into `**kwargs` and hit `raise ValueError('unexpected keyword arguments: ...')` (only `do_not_open`, `writeTimeout`, `interCharTimeout` are popped; `dtr`/`rts` are post-construction properties, with `self._dtr_state = True; self._rts_state = True` defaults).

Reproduced end-to-end in this worktree's venv (uv.lock pins `pyserial==3.5`, `pyserial-asyncio==0.6`; pyproject says `pyserial>=3.5`), no hardware touched:

```
>>> await serial_asyncio.open_serial_connection(url="/dev/bogus-review-test", baudrate=19200, dtr=False, rts=False)
ValueError: unexpected keyword arguments: {'dtr': False, 'rts': False}
>>> await serial_asyncio.open_serial_connection(url="/dev/bogus-review-test", baudrate=19200)
SerialException: [Errno 2] could not open port ... No such file or directory
```

The ValueError fires inside `Serial.__init__`, **before any device access** — so on real hardware every open fails unconditionally, regardless of whether the port exists:

- `backends/discovery.py:192` (`_try_baud`) and `:335` (`probe_xiegu_model_id`): the existing `except Exception` swallows the ValueError into a debug log → discovery **silently finds no radios** (CI-V and Xiegu paths).
- `backends/yaesu_cat/transport.py:198`: `except Exception` → `CatTransportError` → Yaesu probe returns None. Yaesu CAT never connects.
- `backends/icom7610/drivers/serial_civ_link.py:548`: ValueError propagates from `_open()` into the connect loop. CI-V link never connects.

The post-open deassert half (`_deassert_control_lines`) is never reached, because the open itself now fails.

Why the test suite cannot see this: all 8 new tests (and the pre-existing transport tests) inject fake openers (`AsyncMock`, `_capturing_opener`) that accept `**kwargs` unconditionally. No test exercises the real `serial_for_url` path — there is no hardware in CI and none here either, but the kwargs rejection is a pure constructor behavior, no port needed. The mutation table below proves the tests watch the kwargs; it does not prove the kwargs are valid.

Consequence for the PR's prose: "it requests both lines inactive at open time (kwargs forwarded to pyserial's ``serial_for_url``)" — the forwarding is real, the *requesting* is not: pyserial rejects the request. "Which half is strictly necessary is unestablished" — the open-time-kwargs half is not merely unnecessary, it is fatal. The docstring "Open *url* via ``serial_asyncio`` with DTR and RTS held inactive" is likewise false for every real open.

Suggested fix directions (not prescriptions): build the `Serial` instance yourself with `serial_for_url(url, ..., do_not_open=True)`, set `.dtr = False` / `.rts = False` on the closed instance, `open()` it, then wrap it with `serial_asyncio.connection_for_serial(...)`; or drop the kwargs half and rely on the post-open deassert alone. Either way, add one test that runs the real `serial_for_url` kwargs path without hardware (e.g. asserting the constructor accepts the exact kwargs dict the helper builds, or using `do_not_open` semantics), because that is the exact seam this bug lives in.

## What I checked

1. **Completeness (four open sites).** Verified with independent searches, not the PR table: `open_serial_connection` (only remaining direct reference is the helper's own `_default_opener` and pass-through defaults), `serial_for_url|serial\.Serial|Serial(` (docstring mention only), `import serial|serial_asyncio` (dependency checks only), `\bdtr\b|\brts\b|dsrdtr|rtscts` (new file only). Production open paths found: `_try_baud`, `probe_xiegu_model_id`, `YaesuCatTransport.connect` (also serves `probe_serial_yaesu_cat` via the default transport factory), `SerialCivLink._resolve_opener` — all four route through the helper. `probe_serial_kenwood_cat` confirmed a stub returning None (discovery.py:542-560). The Hamlib path spawns `rigctld` as a subprocess (hamlib_bridge.py:187), no serial open. `discover_serial_radios`' `_open_serial` param is pass-through only. One structural note: `SerialCivLink(open_serial_connection=...)` (serial_civ_link.py:542) bypasses the helper when injected — currently test-only (sole production constructor is `_icom_serial_base.py:213`, no opener injected), so LAYER.md's "the only way backends open a port" is true of today's production paths but not enforced for that seam. Minor.
2. **Mutation discrimination** — all three reproduced at the exact head, each restored afterwards with `git checkout -- src/rigplane/core/serial_open.py` and `git status --short` verified empty (see below). Results match the PR table exactly. Note these mutations only prove the tests observe the helper's contract; they say nothing about the contract's validity against real pyserial (see blocker).
3. **The swallow in `_deassert_control_lines`.** Scoped per-line (`for line in ("dtr", "rts")`), each `setattr` individually wrapped, debug log carries the port name and exc_info. Worst case — serial object present, both writes raise — leaves the lines as pyserial's open set them (asserted on a real tty), with only a debug log. That case is currently unreachable-with-consequence because the kwargs blocker fails the open first; once the blocker is fixed, this residual is acceptable in shape (raising would convert usable virtual/BT ports into failed opens, and the discovery candidate filter already excludes non-USB ports) but the **debug** level makes a TX-safety-adjacent failure invisible at default logging config — consider `warning`. Non-blocking.
4. **The getattr chain.** Checked against installed pyserial-asyncio 0.6 source: `open_serial_connection` returns `asyncio.StreamWriter(transport, protocol, reader, loop)` where transport is `serial_asyncio.SerialTransport`, which exposes a public `.serial` property (set from the `serial_instance` created by `serial_for_url`). `asyncio.StreamWriter` always exposes `.transport`. So with the real opener the chain always resolves in production; a writer lacking `.transport.serial` occurs only for injected fakes (the tested tolerance case). No silent production skip.
5. **Regression risk per call site vs ca8ef0d3.** Yaesu `connect()`: dependency check moved from inline `import serial_asyncio` to `_require_pyserial_asyncio()` — still raises ImportError → wrapped into the same `CatTransportError(_DEPENDENCY_HINT)`; the `except Exception → CatTransportError` wrapper and `bytesize=8, parity="N", stopbits=1` are preserved (also pinned by test_yaesu_cat_transport.py:110). SerialCivLink `_resolve_opener`: `_ensure_serial_dependencies()` still raises the friendly ImportError eagerly; the redundant raw `importlib.import_module("serial_asyncio")` was correctly dropped (the helper re-resolves at call time). Discovery probes: opener resolution and `except Exception` handling unchanged. No lost behaviour at any site — the regression is the new shared one (the blocker), which sits in the helper.
6. **Layer.** `uv run lint-imports`: Contracts: 5 kept, 0 broken — `backends → core` is legal for all four sites.
7. **Prose honesty.** True and verified: four-sites table (incl. the Yaesu-probe-via-transport note and the Kenwood-is-a-stub correction); "exactly one hit, a comment in hamlib_bridge.py" at base — confirmed at ca8ef0d3 src/rigplane/hamlib_bridge.py:38 with word-boundary semantics; "eight new tests" (4 unit + 4 call-site); mutation table; gate outputs. False as written: every sentence asserting the open-time kwargs half works (see blocker) — PR body "requests both lines inactive at open time", docstring "with DTR and RTS held inactive", "applies both halves of the mitigation".

## Reproduced mutation results (tests/test_serial_open.py)

- M1 drop `rts=False` → **5 failed / 3 passed** (`test_passes_dtr_and_rts_false_to_opener` + 4 `TestCallSiteWiring`) — matches PR.
- M2 drop `dtr=False` → **5 failed / 3 passed** (same set) — matches PR.
- M3 drop `_deassert_control_lines(writer, url)` call → **4 failed / 4 passed** (`test_deasserts_lines_on_the_real_serial_object`, `test_refused_control_line_write_does_not_fail_open`, `test_yaesu_cat_transport`, `test_serial_civ_link_default_opener`) — matches PR ("+ 3 others").
- Tree restored clean after each (`git status --short` empty; HEAD re-verified as 008f0a851a1d318d3b8620b065312c2ead01613e).

## Gates (run at the exact head)

- `uv run pytest tests/test_discovery.py tests/test_serial_civ_link.py tests/test_yaesu_cat_transport.py tests/test_serial_backend_smoke.py tests/test_icom7610_serial_radio.py tests/test_serial_open.py -q --tb=short --timeout=300 --timeout-method=thread` → **232 passed, 9 warnings in 6.00s**
- `uv run ruff check src/ tests/` → **All checks passed!**
- `uv run ruff format --check src/ tests/` → **749 files already formatted**
- `uv run lint-imports` → **Contracts: 5 kept, 0 broken**

All four gates are green and the mutation table reproduces exactly — the blocker is invisible to all of them because every relevant test replaces the opener. CI state as found: not waited on; verdict is on the code, per the gate's role split.

## Required before this can pass

1. Replace the kwargs-at-open mechanism with one pyserial 3.5 actually accepts (pre-configured closed `Serial` + `connection_for_serial`, or post-open-only deassert), so real opens work again.
2. Add at least one test that would have caught this: assert the exact kwargs dict the helper builds is accepted by the real `serial.Serial` constructor path (e.g. `serial.serial_for_url` with `do_not_open=True`, or equivalent), or otherwise execute the real construction without hardware.
3. Correct the PR body and `serial_open.py` docstrings: the open-time-kwargs half does not "request the lines inactive" on pyserial 3.5.

morozsm:
Agent Review: PASS c9efb4361e737b1a9b51a7207db2539b2895a0df

Independent review of the replaced mechanism at the exact head (previous BLOCKED was against the kwargs construction; nothing from that review was assumed). Read `git diff ca8ef0d3..HEAD`, the PR body (treated as claims), and the **installed** libraries in this worktree's venv: pyserial 3.5 (Python 3.14.7), pyserial-asyncio 0.6.

## 1. Mechanism verified against installed pyserial — works

- `serial/serialutil.py:459-473` — the `rts`/`dtr` property setters store `_rts_state`/`_dtr_state` unconditionally and call `_update_*_state()` only `if self.is_open`. Setting them on a closed instance is a pure attribute store: accepted, no device access.
- `serial/__init__.py:59` — `serial_for_url` pops `do_not_open` before the class is instantiated, so it never reaches `Serial(**kwargs)`'s ValueError path.
- `serial/serialposix.py:311-374` — `Serial.open()` calls `_reconfigure_port(force_update=True)` (line 332), then `_update_dtr_state()`/`_update_rts_state()` (lines 336/338, guarded only by `not _dsrdtr`/`not _rtscts`, which no rigplane call site sets), and `_reconfigure_port` itself never touches TIOCM lines (verified lines 376-540: termios flags only). The EINVAL/ENOTTY swallow at 341 the PR body mentions is real. So the pre-open configuration is applied at open, and the PR's claim about *which* half pyserial's own `open()` implements is accurate.
- `serial_asyncio/__init__.py:481-509` read in full. `_open_with_idle_lines` reproduces `open_serial_connection` exactly where it matters: `connection_for_serial` (line 453) is literally what `create_serial_connection` calls after `serial_for_url`; `asyncio.StreamReader()` default limit is 65536 == `asyncio.streams._DEFAULT_LIMIT` (signature inspected, `_DEFAULT_LIMIT` printed); same `StreamReaderProtocol`, same `StreamWriter(transport, protocol, reader, loop)` construction. Only differences: the serial instance construction (the point of the PR) and `get_running_loop()` vs `get_event_loop()` (equivalent inside a coroutine).
- `SerialTransport.__init__` (`serial_asyncio/__init__.py:47-70`) never opens the port; it sets `timeout=0`/`write_timeout=0` (which triggers `_reconfigure_port` on the open port — line-tying safe, see above) and schedules `connection_made`/`_ensure_reader`. Passing it an already-open instance is what the old path did too.
- **Live end-to-end on a real device node** (macOS pty slave via `pty.openpty()`, `uv run` one-liner, no files written): `open_serial_port('/dev/ttys012', 19200)` returned a working pair — `writer.transport.serial` is a real `Serial`, `dtr`/`rts` read back `False`, `b'ping'` written through the writer arrived at the pty master, clean `wait_closed`. Control: stock `open_serial_connection` on the same pty opens with `dtr: True | rts: True` — the exact hazard this PR removes. (Both the helper and stock `open_serial_connection` stall if a naive harness blocks the event loop reading the master fd — inherent serial_asyncio poll behaviour, identical for old and new code, not a regression; with an executor-based read the helper completes fine.)
- `loop://` genuinely cannot go end-to-end: `protocol_loop.py` defines no `fileno` and `SerialTransport._ensure_reader` needs one (`serial_asyncio/__init__.py:322`) — the PR's statement about the test ceiling is true.
- All four call sites consume only the public StreamReader/StreamWriter surface (`write`/`drain`/`read`/`close`/`wait_closed`); grep confirms none touches `.transport` or `get_extra_info`.

## 2. Nothing lost vs `open_serial_connection`

Construction parity verified line by line (see above). Exception-on-failed-open parity: pyserial's posix `open()` closes its own fd and abort pipes on failure and sets `is_open=True` only on success (`serialposix.py:350-374`) — a failed `instance.open()` leaves a closed, non-leaked instance, same as a failed `serial_for_url` in the old path. If `SerialTransport(...)` construction failed *after* a successful open, the port would leak — but the old `open_serial_connection` has the identical ordering and hole. No regression.

## 3. Failure path

`serial_for_url` raise → no instance exists; `open()` raise → instance left closed by pyserial itself; either way the exception propagates. Call sites: Yaesu normalizes everything to `CatTransportError` (transport.py:208-211); CI-V link propagates to its existing session error handling; both probes now classify (below). Types each site sees are unchanged from the old path.

## 4. Exception split

`serial.SerialException.__mro__` verified: `SerialException → OSError → Exception`. `OSError` as the quiet set covers busy/absent/refusing ports plus every pyserial-wrapped termios error; nothing routine lands in the warning branch (the only realistic non-OSError from the open path is a programming error or the missing-dependency ImportError). The `ImportError` change (propagate → warning + `None`) is acceptable: `enumerate_serial_ports` still raises ImportError early when pyserial itself is absent, the Yaesu probe *already* quietly swallowed its ImportError at debug (discovery.py:470-471), and real connects still fail loudly (`CatTransportError`, `_ensure_serial_dependencies`). One awareness note: the warning fires once per baud — up to 4 per port on the CI-V probe on a broken install. Loud, not silent; fine.

## 5. Mutations (each applied, tested, restored; `git status --short` empty after every restore)

| Mutation | Result |
|---|---|
| drop `instance.dtr = False` | RED: `TestRealPyserialConstruction::test_real_pyserial_construction_with_helper_kwargs` (1 failed, 12 passed) |
| drop `instance.rts = False` | RED: same test |
| drop post-open `_deassert_control_lines` call | RED: 6 tests — `test_deasserts_lines_on_the_real_serial_object`, `test_refused_control_line_write_warns_but_does_not_fail_open`, all four `TestCallSiteWiring` |
| drop whole pre-open configuration | RED: `test_real_pyserial_construction_with_helper_kwargs` |
| CI-V unexpected branch `warning`→`debug` | RED: `TestOpenErrorClassification::test_civ_probe_unexpected_open_error_logs_warning` |
| Xiegu unexpected branch `warning`→`debug` | RED: `test_xiegu_probe_unexpected_open_error_logs_warning` |

Every mutation is caught by a named test. No uncaught mutation — no finding here. The exception-split mutations *are* discriminated, contrary to the question's worry.

## 6. Prose honesty

- `core/LAYER.md` claim checked exhaustively: grepped `src/` for `serial_for_url`, `Serial(`, `termios`, `os.open`, `posix_openpt` — the only pyserial construction anywhere is inside this helper; production `SerialCivLink(device=..., baudrate=...)` (`_icom_serial_base.py:213`) passes no opener; discovery's two `_open_serial` seams are consumed *inside* `open_serial_port` and cannot bypass the deassert. "Sole exception: SerialCivLink's injected test opener" is true.
- `_deassert_control_lines` docstring: the `serial is None` early-return is reached via `getattr(getattr(writer, "transport", None), "serial", None)` → a future transport without `.serial` makes the open still succeed (pre-open half already applied), only the second deassert is skipped. Claim is true as written.
- Module docstring and `_open_with_idle_lines` docstring ("wraps it the way open_serial_connection does", "Mirrors … except the pyserial instance comes from `_configure_serial_idle_lines`") — verified accurate against the installed source.
- PR body's pyserial claims (`_dtr_state`/`_rts_state`, applied in `open()` not `_reconfigure_port`, EINVAL/ENOTTY swallow, MRO) — all verified true.

**Minor findings (non-blocking):**

1. **Stale docstrings** — `discovery.py:155, 184, 330, 743`: `_open_serial` is still documented as "Override for ``serial_asyncio.open_serial_connection``". That function no longer participates in any production path; the override now replaces the helper's internal open step (and its result still passes through the post-open deassert). Should name `rigplane.core.serial_open`'s open step instead.
2. **Kwargs narrowing** — old `open_serial_connection` accepted `limit=`/`loop=` stream kwargs; in `open_serial_port` a hypothetical `limit=` would fall through to `Serial(**kwargs)` and raise ValueError. No caller passes either; informational only, but the `**kwargs` docstring ("Extra pyserial constructor parameters") is the correct reading and is honest.

## 7. Dead code

`SerialOpener`: zero references in `src/`, `tests/`, `docs/` (only a stale `__pycache__`/`.pyc`, untracked build artifact). `opener`'s annotation `Callable[..., Awaitable[tuple[Any, Any]]] | None` matches discovery's `_OpenSerial` (discovery.py:135) and every injected opener's shape. Clean.

## Gates (exact head c9efb436, this worktree, run in order before any mutation)

1. `uv run pytest tests/test_discovery.py tests/test_serial_civ_link.py tests/test_yaesu_cat_transport.py tests/test_serial_backend_smoke.py tests/test_icom7610_serial_radio.py tests/test_serial_open.py -q --tb=short --timeout=300 --timeout-method=thread` → `237 passed, 9 warnings in 6.22s` (warnings: pre-existing `asyncio.iscoroutinefunction` DeprecationWarnings from `command_service.py:325`)
2. `uv run ruff check src/ tests/` → `All checks passed!`
3. `uv run ruff format --check src/ tests/` → `749 files already formatted`
4. `uv run lint-imports` → `Contracts: 5 kept, 0 broken`

No failures observed, so no pre-existing-failure classification was needed; all gates ran at the exact head in this worktree. Tree verified clean (`git status --short` empty) after the final restore.

The two minor findings are real but do not warrant blocking a safety fix of this size; both are one-line prose/annotation cleanups that can ride any later touch of these files. PASS.

morozsm:
Agent Review: PASS e9317e253d17407c6f63739576ee8e0120df177a

Delta review of e9317e25 vs my PASS at c9efb436.

**Delta (exactly as stated):** a16aa0c6 (docstring fix in `src/rigplane/backends/discovery.py`, 4 sites) and e9317e25 (clean merge of origin/main 6ddbe5fd: docs-only #3504/#3506 plus frontend notch markers #3505).

**Checked:**
- Docstring fix is TRUE of the code. `_open_serial` is passed as `opener=` to `open_serial_port` (discovery.py:192, :340); `open_serial_port` runs `_deassert_control_lines` on the writer unconditionally after the opener returns (core/serial_open.py:56-57) — the "post-open control-line deassert still applies" clause is verified, not trusted. `rg "Override for .*open_serial_connection" src/` → no matches. Remaining `open_serial_connection` mentions are the DI parameter name in serial_civ_link.py (no stale path claim) and true "wraps/mirrors" comparisons in serial_open.py.
- Merge interaction: `git diff c9efb436..HEAD --stat` — nothing from main touched src/, core/, the serial/discovery path, or any of the six test files; the only src/tests delta is the branch's own docstring commit.
- No silent reversion: diff of all seven PR-owned files between c9efb436 and HEAD shows only the docstring change in discovery.py; the other six are byte-identical.

**Gates at e9317e25 (local, `uv run`):**
- pytest (6 files): `237 passed, 9 warnings in 6.00s`
- ruff check src/ tests/: `All checks passed!`
- ruff format --check: `749 files already formatted`
- lint-imports: `Contracts: 5 kept, 0 broken.`

c9efb436 conclusions (mechanism, exception split, layer placement, mutation coverage) carried forward unchanged — the delta does not touch any of them.

PR #3505, opened 2026-09-16
feat: place the notch markers by decoded frequency over the drawn passband

morozsm:
PR-3 of the MOR-2475 chain, on top of PR-2 (#3502). The notch markers stop placing themselves by the radio's internal code and take the decoded frequency where the radio publishes one.

## What was wrong

Both audio scopes placed the manual-notch marker as `notchFreq / 255` across the filter trapezoid. Once PR-2 made `notchFreq` a frequency in Hz for a radio that publishes `controls.manual_notch_freq` (the FTX-1: raw 1..320 → 10..3200 Hz step 10), that rule puts the marker nowhere in particular: a code divided by 255 has no relation to where that frequency falls on the drawn passband.

## What changes

- `toAudioSpectrumProps` gains the same manual-notch source selection `toDspProps` already has, extracted as a shared `manualNotchReading` helper inside `panel-props.ts` — one implementation, not a copy.
- `audio-spectrum-renderer.ts` and `AmberAfScope.svelte` place the marker at `notchFreq / filterHz` across `tl..tr` when a domain is published, clamped to the trapezoid edge, and keep `notchFreq / 255` byte-for-byte when none is. The domain is threaded through `AudioSpectrumPanel` → `AudioSpectrumCanvas` and `AmberCockpit` → `AmberAfScope` alongside the value that was already carried.

The placement expression is three lines duplicated across the two renderers. Rule of three: two call sites are a coincidence, so it is not abstracted.

## Why `filterHz` is the right span

Checked against the bar drawing rather than assumed. In `audio-spectrum-renderer.ts` the spectrum bars map `barFrac = i / numPoints` to `pixIdx = dcIdx + barFrac * passbandFrac * positiveLen` with `passbandFrac = filterHz / halfBw`: the left edge of the drawn spectrum is the DC bin, 0 Hz, and the right edge is `filterHz`. The drawn passband therefore runs 0..filterHz, which is what the new placement divides by. `AmberAfScope` draws its bins over the same 0 → passband span. `filterHz` is the animated width the trapezoid itself is built from, so marker and trapezoid move together while the filter animates instead of the marker sliding across a settling shape.

## Tests

Nine new cases across props, the renderer and the Amber scope — three per file, counted from the diff with `grep -cE '^\+\s*it\('`, no parameterised cases and none removed. Each was proved to discriminate by mutation, run and reverted, with the observed failures recorded: forcing the old props shape gives `expected +0 to be 1600` and `expected undefined to deeply equal {min:10,max:3200,…}`; forcing raw placement gives `expected 984.75… to be close to 222.65`; forcing Hz placement unconditionally gives `expected 137.71 to be close to 185.34`; removing the clamp gives `expected 313.25 to be close to 267.95`. Three cases pass against both implementations by design, and they are the unchanged-path pins: "keeps the raw 0-255 placement when no domain is published" in each of the two renderers, and "keeps the raw notchFilter value and emits no notchFreqDomain key" in the props. The reviewer confirmed by running the force-raw-placement mutation that those pins stay green under it, which is what an unchanged-path pin is for.

**Correction applied before merge.** This paragraph first said nine cases with one passing both ways; the both-ways count was wrong and is now three, found by the independent review. The review also reported twelve new cases rather than nine; re-counted from the diff, nine is correct and that part of the finding is not adopted. The builder's own report said six new props cases where the diff adds three. Three separate counts of the same thing disagreed, so the number above is the one the diff yields.

## Scope

9 files, +258/−24 — over the 6-file soft threshold, under the hard ceiling, measured at this head. One unit of work: the renderers cannot read a domain the props do not carry, and a marker left on `raw / 255` while the value is in Hz is the defect being removed.

`pbtRawToHz` is untouched everywhere, by the coordinator ruling recorded on MOR-2475: deleting the renderer's private copy is the same change as consolidating PBT onto the contract path, and it moves the number `audio-spectrum-renderer.test.ts` pins at 1191, so it belongs with PR-4 and its re-measurement. The DSP panel and the command path are untouched; PR-2 finished those.

## Known, not fixed here

The marker spans the trapezoid's top edge `tl..tr` while the spectrum bars span its bottom edge `bl..br`, which is wider by `slopeExtra` on each side. Marker and bar for the same frequency therefore coincide at mid-band and diverge toward the edges. This geometry pre-dates this PR, applies identically to the contour marker, and correcting it would change what the overlay means; recorded on MOR-2475 rather than changed here.

Where a domain is published but the reading is unusable, the props carry the raw `notchFilter` code while the renderer treats the value as Hz. This mirrors the slider semantics PR-2 settled and is unreachable on today's profiles; it is part of the `manualNotchFreq` / `notchFilter` feedback split already recorded for PR-3/PR-4 follow-up.

Part of MOR-2475


morozsm:
Agent Review: PASS 17ec4588d1d79a517b32972840df95e3accba0ea

Independent review of the exact head above. CI state as found, not waited on: `quick` pass (5m25s), `visual` pass, classifiers/observers green; the two Agent Review Gate rows fail only for the absence of this directive.

## Verified by running (worktree at this head)

- `npm --prefix frontend ci`, then the three touched test files: 60/60 pass.
- Re-ran three of the builder's mutations, red each time, restored, `git status` clean, HEAD unchanged:
  - clamp removed (renderer) → `expected 313.25 to be close to 267.95` — matches the reported text;
  - raw placement forced (renderer) → `expected 984.755882352941 to be close to 222.64999999999998`;
  - `toAudioSpectrumProps` reverted to the old shape → `expected +0 to be 1600` and `expected undefined to deeply equal { min: 10, max: 3200, step: 10, …(1) }`.
- `npm run check` (svelte-check + both tsc projects): 0 errors, 0 warnings. `eslint` on changed files and `lint:boundaries`: clean.

## Verified by reading

1. **Span claim holds.** `audio-spectrum-renderer.ts:308-314` maps `barFrac = i/numPoints` across `bl..br` to `pixIdx = dcIdx + barFrac·passbandFrac·positiveLen` (`passbandFrac = filterHz/halfBw`, line 290): leftmost bar reads the DC bin, i.e. the drawn passband's low edge **is** 0 Hz and the right edge is `filterHz`. `AmberAfScope.svelte:326-339` is the same construction over its own `bl..br`. Both markers divide by the same `filterHz` the trapezoid and bars are built from (`Math.round(rs.animFilterWidth)` / `Math.round(animatedFilterWidth)`), so marker, trapezoid and bars animate together. A marker-offset-by-low-edge defect does not exist.
2. **Top-edge vs bottom-edge mismatch is genuinely pre-existing.** The removed line already placed the marker on `tl..tr`; the contour markers use the identical span (renderer :382, Amber :242) and no geometry constant changed. Worst-case marker-vs-bar divergence is `slopeExtra` at each band edge, tapering linearly to 0 at mid-band: 44.1 px of 400 (top span 135.9 px) in the renderer; 42 px of 320 (top span 123.2 px) in Amber — visible at the edges, but unchanged in magnitude by this PR and shared with the contour. Recording it on MOR-2475 is right; fixing it here would break the byte-for-byte raw pin. The `filterRatio` `*0.75` asymmetry (renderer :164 vs Amber :181) is pre-existing, untouched, and each test mirrors its own renderer's constant — both pass, so neither is broken.
3. **Clamp correct at both ends.** `min(1,…)` pins 3200→`tr` (mutation-verified); `max(0,…)` pins ≤0→`tl`; `filterHz ≤ 0` short-circuits to 0 (no division); NaN `filterHz` cannot reach the expression (`showFilterOverlay` false); NaN `notchFreq` yields NaN coordinates and no path — the same silent skip the old `raw/255` had.
4. **Helper is one implementation; `toDspProps` unchanged.** `manualNotchReading` (panel-props.ts:631-645) reproduces the old inline block conjunct-for-conjunct and ternary-for-ternary; `activeRx` and `resolveControlContract` are pure, so the extra `activeRx` call is behavior-neutral; the return object is textually untouched and the six PR-2 `toDspProps` cases still pass. Duplicating the three-line placement expression under the rule of three is right — the two scopes are deliberately independent canvas implementations with different constants.
5. **Layer rules.** Type-only imports from `$lib/radio/filter-controls` in both panels; no transport/audio-manager imports; adapter pure; radio values come from published domains, FTX-1 numbers appear only as test fixtures; every new prop has a consumer.

## Findings

**REQUIRED BEFORE MERGE (PR-body correction, not fixable by a commit):** the Tests paragraph says "Nine new cases" and "One case — an Icom-shaped radio with no domain — passes against both implementations". Both counts are wrong. The diff adds **12** new `it` blocks (3 renderer + 3 Amber + 6 in `notch-freq-domain.isolated.test.ts`), and **three** cases pass against the old implementation by design: the Icom props pin **and** the two "keeps the raw 0-255 placement when no domain is published" renderer pins (verified: the raw pin stayed green under the forced-raw-placement mutation, and the no-domain path is byte-identical by the PR's own stated design). The discrimination substance is correct — every case that can discriminate does, per my own mutation runs — but whoever merges must correct these two counts in the body or squash message and note that it did.

Non-blocking notes: (a) the placement comments' operative rule is test-pinned, but the bars-alignment claim would not go red if the bar mapping were later changed — it is reading-verified only; (b) the domain-published-but-reading-unusable mixed-semantics case is disclosed in the body, unreachable on today's profiles, and already recorded as follow-up; (c) no renderer-level test pins the lower clamp (`notchFreq ≤ 0 → tl`) — the symmetric expression is covered by reading and the upper clamp's mutation evidence.

`pbtRawToHz`, the DSP panel, and the command path confirmed untouched, per the coordinator ruling.

PR #3508, opened 2026-09-16
refactor: route the audio-scope PBT overlay through the one shipped conversion

morozsm:
PR-4 of the MOR-2475 chain, on top of PR-3 (#3505). The audio spectrum renderer stops carrying its own copy of the PBT raw-to-Hz conversion and uses the one shipped implementation, with the range passed explicitly.

## What changes

The renderer's private `pbtRawToHz(raw, center = 128, maxHz = 1200)` is deleted. Its three call sites now use `$lib/radio/filter-controls: pbtRawToHz(raw, range)`. The range travels as `SpectrumState.pbtRange`, derived in `toAudioSpectrumProps` with `pbtRangeFromCaps(caps)` and threaded through `AudioSpectrumPanel` → `AudioSpectrumCanvas`, absent as a key rather than null when the radio publishes none — the same discipline PR-3 used for `notchFreqDomain`.

**The point is the dependency, not the arithmetic.** The shipped function falls back to a module-global capabilities store when called without a range. Dropping the private copy and calling it bare would have traded a hardcoded constant for a hidden global — worse, at identical numbers. The renderer never reaches that store.

## The numbers did not change, and that is the claim being made

`audio-spectrum-renderer.test.ts` pinned `pbtRawToHz(56) = -675`, `(255) = 1191`, `(0) = -1200` against the deleted private function. Re-measured through the new path rather than hand-edited: all three are unchanged, and so is `128 → 0`. That is expected — every profile publishing PBT declares `raw_center = 128` and `display_max = 1200`, and `PBT_DEFAULTS` carries the same pair — but it was measured, not assumed.

**Unchanged is not the same as correct.** Bench measurement on the IC-7610 the same day (MOR-2497) established that this Hz scale is wrong about the radio: the real step is 50 Hz, the number of steps is `filter_width / 50`, the full range of one edge is `±filter_width / 2`, and raw 0 and 255 are not reachable at all. At a 3600 Hz filter the top reachable position is +1800 Hz, not 1191. This PR neither fixes that nor makes it worse; it removes the second implementation so that when the scale is corrected there is one place to correct. MOR-2497 owns the correction.

## The no-range decision

Where no usable range is published, the renderer **refuses to draw the PBT overlay** rather than passing `undefined` and letting the shipped function consult the store. Reasons: the "must not reach the module-global store" requirement is unconditional; a stale store left by a previously connected radio would otherwise displace the current radio's trapezoid; `radio-view-model-adapter.ts` sets the precedent (MOR-1291: no scale means unknown, not a store lookup); and the profiles without PBT publish no PBT control at all, so nothing visible changes for them. There are **four** of those, not two: `ftx1`, `tx500`, `x6100` and `x6200` — counted with `grep -l '^\[controls.pbt_inner\]' rigs/*.toml`, which leaves `ic705`, `ic7300`, `ic7610` and `ic9700` as the publishers. An earlier version of this paragraph named only two; the independent review caught it and the coordinator confirmed the count against `rigs/` before correcting it.

## Tests

Six mutations, each broken, observed red, and restored, with the failure text recorded. The one that matters: making the renderer drop the range argument stays **green** in the fast pool and goes red only in the new isolated test, which installs a non-default `pbt_inner` range in the store so a store-consulting path produces visibly different geometry — `expected 240.5 to be close to 230.375`, where 240.5 is the store's position. A test checking only the arithmetic would not have caught it, which is why that test exists and why it lives in the `isolated` pool.

The others: disabling the overlay guard, dropping the no-range guards, removing the props spread, emitting the key unconditionally, and removing `Math.round` from the shipped function (`expected 1190.625 to be 1191`).

## Checks

`npm --prefix frontend run test -- audio-spectrum panel-props notch-freq-domain filter-controls mor1562-adapter-seams` → 8 files, 251 passed. `check` → 0 errors, 0 warnings across 4859 files. `lint` → clean.

## Scope

7 files, +276/−27, measured at this head — over the 6-file soft threshold, under the hard ceiling. One unit of work: the renderer cannot take a range the props do not carry, and deleting the copy without threading the range would substitute a global for a constant.

`pbtHzToRaw` and the legacy converter exports are untouched; unexporting what has no callers is PR-5, and the census recorded on MOR-2475 shows `pbtHzToRaw` still has five production call sites.

Part of MOR-2475


morozsm:
Agent Review: PASS 954a56d15001eb9d3e7e74ba20f8894762d9b824

Independent review of PR #3508 at the exact head above. CI state as found: not re-queried during this review; required checks are the coordinator's separate pre-merge step.

## Store reachability (the PR's purpose)

No path reaches the module-global capabilities store. The renderer has exactly three `pbtRawToHz` call sites (audio-spectrum-renderer.ts:162, 226, 227); all pass `pbtRange`, and each is guarded so the function is invoked only with a truthy range (`avgPbtHz` ternary; `if (pbtActive && pbtRange)`). In the shipped helper the store lookup `pbtRange()` sits behind `range ?? pbtRange()` and is evaluated only when `range` is null/undefined — a truthy-but-malformed range yields NaN, not a store read. The no-range branch computes literal `0` without calling anything. The only producer is `pbtRangeFromCaps(caps)` (validates finite/non-zero fields, cannot throw), spread into props only when `!== undefined`; Panel and Canvas thread it as a plain prop with no other PBT math, and Canvas adds it to the `$effect` dependency list so a late-arriving range repaints. Importing `filter-controls` has no store side effect. The bare `pbtRawToHz` calls at panel-props.ts:379-380 are pre-existing `toFilterProps` code, untouched by this diff (PR-5 scope). `toDspProps` and the PR-3 path are byte-for-byte untouched — the panel-props diff is 4 hunks, all imports/`AudioSpectrumProps`/`toAudioSpectrumProps`, additive.

## No-range decision

Correct, and invisible for shipped profiles. `pbt` capability + `[controls.pbt_inner]` (all `raw_center=128`, `display ±1200`, identical to the old hardcoded constants) exist only in ic705/ic7300/ic7610/ic9700 — exactly the radios whose state carries live PBT values. Profiles without PBT never poll it, so `pbtInner` stays at the `state_schema.py:125` default 128 → `pbtActive` false → no overlay before or after this PR. Refusing (not store-falling-back) also prevents a stale store from a previously connected radio displacing the current one's overlay.

## Mutations re-run (both restored; worktree clean after)

- M1, decisive — renderer drops the range argument at all three call sites: `audio-spectrum-pbt-range.isolated.test.ts` RED, `AssertionError: expected 240.5 to be close to 230.375, received difference is 10.125` (240.5 is the store-installed raw_center=100/±900 position). Full fast project under the same mutation: **264 files / 7436 tests GREEN** — the fast pool genuinely would have missed it; the isolated test is the only discriminator. Claim confirmed.
- M2 — remove the `pbtRange` spread from `toAudioSpectrumProps`: `panel-props.test.ts` RED, `AssertionError: expected undefined to deeply equal { rawCenter: 128, …(2) }`.

## Pins

Re-derived independently: `(200-128)·9.375=675`, `(56-128)·9.375=-675`, `(255-128)·9.375=1190.625→1191`, `(0-128)·9.375=-1200`; geometry `200+(675/2400)·180·0.6=230.375`, outer `169.625`. The shipped formula is term-for-term the deleted private one, and all 120 tests in the three touched files pass at the head. The pins are forced by arithmetic, not hand-fitted.

## Prose

The body correctly claims UNCHANGED and explicitly disclaims CORRECT ("Unchanged is not the same as correct", MOR-2497 owns the scale fix). No code comment claims the numbers are right. Minor: audio-spectrum-renderer.test.ts:241 has a mangled joined line (`() => {      const rs =`); eslint passes on it — cosmetic only.

## REQUIRED BEFORE MERGE

- PR body, "The no-range decision": "the only profiles without PBT — `ftx1`, `x6200`" is false. Four rig profiles publish no PBT: `ftx1`, `tx500`, `x6100`, `x6200` (verified: no `pbt` capability and no `[controls.pbt_inner]` in any of them; x6100/x6200 carry explicit comments that CI-V PBT probes time out). The sentence's conclusion ("nothing visible changes") still holds for all four — they publish no PBT state either — but the enumeration as written is a false claim that the squash body would carry onto `main`. Fix: edit the body (and the merge/squash body) to name all four before merging. Not fixable by a commit.

## Checks run

`git rev-parse HEAD` (head confirmed) · `npm --prefix frontend ci` (535 packages) · vitest run on the three touched test files (120 passed) · vitest run `--project fast` under M1 (7436 passed) · vitest on the isolated file under M1 (1 failed, text above) · vitest on panel-props under M2 (1 failed, text above) · eslint on both new/edited test files (clean).

Could not determine: whether the pins were produced by an actual measurement run versus derived — the numbers are arithmetically forced either way, so the distinction is not load-bearing here.

PR #3509, opened 2026-09-16
docs: narrow the "every serial open" claim to what both branches do

morozsm:
## What

Prose only; no behaviour change. 2 files, 24 changed lines.

`src/rigplane/core/serial_open.py:6` claimed:

> every serial open in rigplane routes through `open_serial_port`

`SerialCivLink`'s `open_serial_connection` seam does not.
`_resolve_opener` (`serial_civ_link.py:542-543`) returns the injected
callable as-is, so `_deassert_control_lines` never runs on that branch.

PR #3503 already narrowed the same claim in `src/rigplane/core/LAYER.md:26-29`
("Sole exception: `SerialCivLink`'s injected test opener (retiring it:
MOR-2495)") but left the module docstring at its original width, so the two
files disagreed about the same fact.

This change:

- narrows the `serial_open.py` sentence to production paths and names the
  exception;
- documents the distinction on `SerialCivLink`: `open_serial_connection`
  replaces the whole open step and takes no arguments, while
  `open_serial_port`'s own `opener` replaces only the open and still has its
  result deasserted;
- corrects "that whole step" to "that open step" in `open_serial_port`'s
  docstring, which described the narrower seam with the wider wording.

## Why not route the seam through `open_serial_port(opener=...)`

The seam is intentional and predates MOR-2228. `git log -L 160,162` on
`serial_civ_link.py` returns one commit — `28ee1611 feat(serial): implement
production SerialCivLink for IC-7610 USB CI-V` — i.e. the parameter arrived
with the class. #3503 only replaced the body of the non-injected branch.

The two seams have incompatible signatures: `open_serial_port` calls
`opener(url=..., baudrate=..., **kwargs)`, while `open_serial_connection` is
typed `Callable[[], Awaitable[tuple[Any, Any]]]` — zero arguments, device and
baud already bound in the test's closure. Routing one through the other would
`TypeError` and require reshaping the public constructor plus the five places
in `tests/test_serial_civ_link.py` that supply the seam — four constructor
calls (`:116`, `:650`, `:735`, `:772`) and one direct assignment to the
private attribute (`:853`). Counted with
`grep -n "open_serial_connection=" tests/test_serial_civ_link.py` (four hits)
and `grep -n "_open_serial_connection" tests/test_serial_civ_link.py` (one
further hit, line 853). That is MOR-2495's
scope, which #3503 already opened.

## How each claim was established

- "Nothing under `src/` passes `open_serial_connection`":
  `grep -rn "open_serial_connection" src` — the only non-prose hits are the
  parameter's own definition and use inside `serial_civ_link.py`;
  `grep -rn "SerialCivLink(" src` returns one construction,
  `_icom_serial_base.py:213`, without the kwarg.
- "the non-injected branch is covered by
  `tests/test_serial_open.py::test_serial_civ_link_default_opener`":
  `grep -rn "def test_serial_civ_link_default_opener" tests/test_serial_open.py`
  → line 293; it monkeypatches `serial_open._open_with_idle_lines` and asserts
  `("dtr", False)` and `("rts", False)` on the recording serial.
- Signature difference: read from `serial_open.py:58-61` and
  `serial_civ_link.py:161`.

## Checks

Prose-only change, but the paths are `src/rigplane/**/*.py`, so the quick
classifier routes it as backend/core — the documentation-only exception does
not apply and the normal backend checks run.

Locally, scoped to the change:

- `uv run ruff check` on both changed files — All checks passed
- `uv run ruff format --check` on both — 2 files already formatted
- `uv run pytest tests/test_serial_open.py tests/test_serial_civ_link.py -q`
  — 57 passed


morozsm:
Agent Review: PASS ecbab5139cb9eb18294d591cb09a1decbc840418

Independent review. I did not author this change. Prose-only; the prose is the
entire deliverable, so every added sentence was checked against code I read or
ran myself, not against the PR body.

## Claims checked

1. **`_resolve_opener` returns the injected callable as-is; the deassert does
   not run on that branch** — CONFIRMED. `serial_civ_link.py:555-563`
   (`SerialCivLink._resolve_opener`): `if self._open_serial_connection is not
   None: return self._open_serial_connection`. `connect()` at line 246-249
   awaits that callable directly. `git grep -n "_deassert" ecbab513 --
   src/rigplane/backends/` returns no hits, so nothing on that path reaches
   `_deassert_control_lines`. Executed against a `git archive ecbab513`
   extract: `_resolve_opener()` returned the injected object identically
   (`op is inj`), and the recording serial saw **`[]`** — no DTR/RTS writes.

2. **`open_serial_port`'s own `opener` still gets its result deasserted** —
   CONFIRMED by reading the body, not the docstring: `serial_open.py:59-63`
   does `reader, writer = await opener(...)` then
   `_deassert_control_lines(writer, url)` unconditionally, outside the
   `if opener is None` branch. Same scratch run with `opener=` gave
   **`[('dtr', False), ('rts', False)]`**. The same instrument produced `[]`
   in case 1 and a non-empty list here, so the comparison is demonstrably
   able to report inequality — it is not a silently-broken "both look alike".

3. **Signature difference** — CONFIRMED from annotation *and* call site.
   `inspect.signature(SerialCivLink.__init__)` at head reports
   `Callable[[], Awaitable[tuple[Any, Any]]] | None`; `connect()` invokes it
   as `opener()` with no arguments (`serial_civ_link.py:248`). `opener` is
   called `opener(url=url, baudrate=baudrate, **kwargs)`
   (`serial_open.py:61`). The docstring's "called with `url` and `baudrate`"
   is true and is not falsified by the additional `**kwargs`.

4. **"Nothing under `src/` passes `open_serial_connection`"** — CONFIRMED with
   independent vocabularies, not a repeat of the PR's grep:
   - `git grep -n "open_serial_connection" ecbab513 -- src/` → only the
     parameter's own definition/use in `serial_civ_link.py` plus prose.
   - `git grep -n "SerialCivLink(" ecbab513 -- src/` → one construction,
     `_icom_serial_base.py:213`, without the kwarg.
   - `git grep -n "opener" ecbab513 -- src/` → the only other injection seam
     in `src/` is `discovery.py:192` and `:340`, and both pass it as
     `open_serial_port(..., opener=_open_serial)`, i.e. through the helper, so
     the deassert still runs. This is what makes the totality word "One test
     seam" hold rather than being a NARROWED claim.
   - `git grep -nE "import serial|serial\.Serial|serial_for_url|create_serial_connection|aioserial|comports|Serial\("`
     over `src/` finds no serial open outside `core/serial_open.py`,
     `discovery.py`, `serial_civ_link.py`, `yaesu_cat/transport.py` — and the
     latter three all route through `open_serial_port`.

5. **The named test covers the non-injected branch and can actually fail** —
   CONFIRMED by mutation, not by reading.
   `tests/test_serial_open.py:293` `TestCallSiteWiring::test_serial_civ_link_default_opener`
   builds `SerialCivLink` *without* the kwarg, calls `_resolve_opener()`,
   awaits it, and asserts `("dtr", False)` / `("rts", False)`.
   - Baseline on an unmodified `git archive ecbab513` extract: 1 passed.
   - Mutation reinstating the pre-#3503 defect (`open_serial_port` patched at
     import time from a throwaway pytest plugin to skip
     `_deassert_control_lines`): **FAILED** —
     `assert ('dtr', False) in []`.
   - Control (legitimate nearby change: reorder the deassert loop to
     `("rts", "dtr")`): 13 passed, i.e. the check does not redden on correct
     work.
   The tree under review was never touched; all mutation happened in an
   extracted copy. `git status --short` on the worktree is clean at
   `ecbab5139cb9eb18294d591cb09a1decbc840418`.

6. **The seam predates MOR-2228** — CONFIRMED by my own archaeology, not the
   PR's. `git show bc5d1f78^:src/rigplane/backends/icom7610/drivers/serial_civ_link.py`
   (the parent of #3503) already contains
   `open_serial_connection: Callable[[], Awaitable[tuple[Any, Any]]] | None = None`
   at line 161 and the identical early return at 542-543; the only difference
   is that the non-injected branch then called
   `serial_asyncio.open_serial_connection` directly. #3503 replaced only that
   branch body. The PR body's `git log -L` pointing at `28ee1611` reproduces
   from head (the line has shifted to 175 after this PR's docstring).

7. **Consistency with `src/rigplane/core/LAYER.md:26-29`** — CONFIRMED. LAYER.md
   says "the normal way backends open a port. Sole exception: `SerialCivLink`'s
   injected test opener (retiring it: MOR-2495)". The two narrowed docstrings
   now assert the same fact with the same ticket. The disagreement #3509 set
   out to remove is gone, and no third site carries the old wide wording:
   `git grep -n "whole step"` over `src/ tests/ docs/` returns nothing, and
   `discovery.py:155,184,331,744` already said "the open step", so the
   corrected phrase matches its neighbours rather than diverging from them.

8. **Is any new sentence wider than the code?** — No sentence I could falsify.
   The totality word ("One test seam", "Nothing under `src/`") is the one that
   had to be swept, and item 4 sweeps the class rather than the instance.

## Neighbour / whole-file pass

`serial_open.py` entered review whole, not hunk-by-hunk. The untouched
neighbours of the edited lines are consistent with the correction: the `Args:`
entry `opener: Replacement for the real open step (used in tests)` already used
the narrower wording the diff moves the prose body to, and the MOR-2493
sentence about which half is load-bearing remains explicitly hedged as
unestablished. No stale sibling beside the corrected line.

## Guardrails and checks

2 files, 24 changed lines — under both the 6/600 soft threshold and the
10/1000 hard ceiling, so no PR-body justification or exception line is owed.
Paths are `src/rigplane/**/*.py`, so the documentation-only exception correctly
does **not** apply and the backend checks are the right scope; `quick` is green
at this exact head (1m32s), with `quick-v2-observe` still pending as the
observation shadow.

## Non-blocking finding

**PR body, case 3 (never enters the repository; editable in place, head SHA
unaffected). REQUIRED BEFORE MERGE — an edit, not a commit.**

The body says routing the seam through `open_serial_port(opener=...)` would
"require reshaping the public constructor plus six test call sites
(`tests/test_serial_civ_link.py:116,650,735,772,853`)". The count is wrong
twice over: five line numbers are listed, not six, and only four of them are
constructor kwarg sites. `git grep -n "open_serial_connection=" ecbab513 --
tests/` returns exactly four hits (116, 650, 735, 772); line 853 is
`link._open_serial_connection = _open`, a direct private-attribute assignment,
not a constructor call. The whole-repo sweep
`git grep -n "_open_serial_connection\|open_serial_connection" ecbab513 --
tests/` confirms no sixth site exists — the remaining hits are in
`tests/test_yaesu_cat_transport.py` and refer to
`serial_asyncio.open_serial_connection`, an unrelated symbol reached through
`opener=` at line 94.

Suggested replacement: "plus five test sites (`tests/test_serial_civ_link.py`:
four constructor kwargs at 116, 650, 735, 772 and one private-attribute
assignment at 853)".

This does not touch the diff and does not affect the conclusion it supports
(that the reshape is MOR-2495's scope), which is why it is not a block. Nothing
in the changed lines themselves carries this number.

## Observation, not a defect

"Nothing under `src/` passes `open_serial_connection`" is true today but is not
tied to anything that fails when it stops being true — a future production
caller would falsify it silently. I am not blocking on this: the equivalent
sentence already landed in `core/LAYER.md` via #3503 under the same exposure,
and MOR-2495 retires the parameter outright, which removes the claim rather
than pinning it. Worth a line in MOR-2495 rather than a commit here.

PR #3510, opened 2026-09-16
refactor: unexport the seven filter-conversion symbols nothing imports

morozsm:
PR-5, the last step of the MOR-2475 chain. Seven symbols in `lib/radio/filter-controls.ts` stop being exported. They are still used inside that file; only the export goes.

`controlRawToDisplay`, `controlDisplayToRaw`, `FILTER_BIPOLAR_MIN`, `FILTER_BIPOLAR_MAX`, `FILTER_WIDTH_MIN`, `FILTER_WIDTH_MAX`, `FILTER_WIDTH_STEP`.

## The census, and one correction to it

Counted at `6bf81ed7` across every `.ts` and `.svelte` under `frontend/src`, with comments stripped, excluding the declaring file: all seven have **zero** production references and **zero** test references. The module is imported by **24 files other than itself and its own unit test**, and none of them imports any of the seven. Counting rule stated because it changes the number: including `filter-controls.test.ts` gives 25, which is what the independent review counted. The claim that carries the change — that nothing outside the module imports these symbols — holds under either rule. Inside the declaring file they are all live — `controlRawToDisplay` 3 uses, `controlDisplayToRaw` 2, one each for the five constants — which is why this unexports rather than deletes.

An earlier census on MOR-2475 reported one external reference to `FILTER_WIDTH_STEP`. That was a raw grep counting a **comment** in `filter-controls.test.ts`; with comments stripped the count is zero, and the builder independently reached the same conclusion from the same line. Six symbols became seven.

## What is deliberately left alone

`pbtRawToHz` (7 production call sites) and `pbtHzToRaw` (5, all in `panel-commands.ts`) stay exported — the chain's original PR-5 line called for deleting them, which the census showed is not executable, and MOR-2475 records why: routing those twelve call sites onto the contract path would rescale PBT, and MOR-2497 has since measured that the Hz scale itself is wrong about the radio. `controlRangeFromCapsOrDefault` (1 caller), `nbDepthRawToDisplay` (2 callers) and `CONTROL_DEFAULTS.nb_depth` also stay; moving `nb_depth` onto the contract path is its own PR, and it is safe there precisely because that mapping is exact — raw 0..9 to display 1..10, ten values against ten.

## Checks

`npm --prefix frontend run check` → 4859 files, **0 errors, 0 warnings**. That is the check that matters here: it fails if any of the 24 importers still needs a now-private symbol. `lint` clean. `npm --prefix frontend run test -- filter-controls` → 2 files, 41 tests passed.

## Scope

1 file, +7/−7, measured at this head. Nothing but the `export` keyword changes — verified by filtering the diff for any line that is not one of the seven declarations.

No prose was deleted: no comment in the file described these as part of the public surface. The docblock mentions of `controlRawToDisplay` elsewhere in the file describe the `range` parameter, not the export, and remain true.

Part of MOR-2475


morozsm:
Agent Review: PASS 324c896c4f159185fded28acb788430d0c011142

Independent review of the exact head. I tried to refute the census and could not.

**Census (redone independently, comments stripped, whole repo).** Zero code references to any of the seven symbols outside `filter-controls.ts`. The only external hit is a `//` comment in `filter-controls.test.ts:89` mentioning `FILTER_WIDTH_STEP` — prose, confirming the body's correction about the earlier miscount. I count 25 files importing the module (24 external consumers plus the module's own unit test); none imports any of the seven. No namespace imports, no dynamic imports, no barrel re-exports, no string-keyed lookups.

**Unexport vs delete.** All seven are live internally: `controlRawToDisplay` 3 call sites (393, 580, 586), `controlDisplayToRaw` 2 (394, 591), one each for the five constants (595 ×2, 600, 601, 603). Unexporting rather than deleting is correct.

**Diff shape.** Every changed line is one of the seven declarations differing only by the removed `export` keyword (verified by filtering the diff).

**Prose.** Module header does not enumerate an exported surface; the `FILTER_WIDTH_*` docblock mention (~line 644) and the `controlRawToDisplay` docblock references describe behavior and remain true. Nothing false left behind.

**Body claims spot-checked.** Internal-use counts 3/2/1 ✓. `npm run check` output (4859 files, 0 errors, 0 warnings) ✓ reproduced locally. Filter-controls tests: 2 files, 41 passed ✓ reproduced. `pbtRawToHz`/`pbtHzToRaw` have live consumers and correctly stay exported (out of scope). Importer count: I measure 25 including the module's own test file, 24 external — the body's "24" holds under the external-consumer reading.

**Checks run locally at this head:** `npm --prefix frontend ci`; `npm --prefix frontend run check` → 4859 files, 0 errors, 0 warnings; `npm --prefix frontend run test -- filter-controls` → 41/41 passed.

**CI state as found (not waited on):** `visual` pass; `quick` pending on the self-hosted runner; `Agent Review Gate` failing only because no exact-head PASS directive existed yet — this comment is that directive. Required-check greenness at the exact head remains the merge operator's separate gate.

PR #3511, opened 2026-09-16
test(MOR-2492): add the shared CI-V radio fake and convert the Xiegu identity tests

morozsm:
## What

PR 1 of MOR-2492: one shared, stateful CI-V radio fake for wire-level tests, and the first five tests converted onto it.

(An earlier revision of this body said six. It is five: `test_open_failure_returns_false` was already literal-free and is correctly untouched. Corrected after independent review flagged it.)

**This PR touches no `src/`.** That is an owner decision, not an oversight — see "Why test-only" below.

| file | |
|---|---|
| `tests/civ_fake.py` | new — `CivRadioFake` (identity, state, command log) + `CivSerialPort` (the `(reader, writer)` adapter) |
| `tests/test_civ_fake.py` | new — 10 tests of the fake itself |
| `tests/test_discovery.py` | converts **four** of `TestProbeXieguModelId`'s five tests, plus `test_timeout_tries_next_baud` — five conversions in total |

3 files, +390/-24 against the merge base.

## Why this exists

Our suite has 20+ ad-hoc radio doubles across 43 files, nine of them independent re-implementations of "send a frame, get a scripted reply". Two consequences:

1. **A fault scenario cannot be expressed at all.** There is no way to say "NAK the next N frames of command X", "drop exactly this reply while the radio's state stands", or "the cable was pulled". The only vocabulary that exists is LAN-session-shaped and probabilistic (`drop_rate`), so a test cannot target one specific exchange.
2. **"Command X was never sent" cannot be asserted.** Four doubles keep a private log under four names (`sent_packets`, `sent_frames`, `received`, `writes`) and every test reimplements the check inline.

PR #3503 demonstrated the cost: a change that would have made **every real serial open fail** kept all 232 relevant tests green, because each one injects a fake opener that accepts anything.

## What the conversion actually buys

Before, `TestProbeXieguModelId` carried byte literals and nothing connected them to the radio they represent:

```python
_XIEGU_NAK_REPLY = bytes([0xFE, 0xFE, 0xE0, 0xA4, 0xFA, 0xFD])
```

After, an IC-705 refuses `0x1D 0x19` because that is what an IC-705 is:

```python
radio = CivRadioFake("IC-705")
```

The knowledge moves into one place where it is visible whose it is. The IC-705/X6200 CI-V address collision at `0xA4` is the entire subject of those tests, and the fake now encodes both halves of it: **both** models answer the transceiver-ID probe with `01 05` — the X6200 emulates an IC-705 there — and only the Xiegu-only `0x1D 0x19` opcode returns `62 00`. That is exactly why the hwid disambiguation exists.

## The seam, and why this one

`CivRadioFake` is transport-free. `CivSerialPort` is the one adapter, presenting it as an `(asyncio.StreamReader, asyncio.StreamWriter)` pair behind an opener callable.

Rejected: the `SerialCivLink` seam (structurally cannot serve discovery, which never constructs a link, and would fake the write queue and reconnect logic that `test_icom7610_serial_radio.py` exists to exercise) and the `MockTransport` seam (LAN session territory, whose fault vocabulary is correctly located there; 30 files; would drag ~600 lines of login/token/ping machinery in).

The argument for this seam does not rest on re-exercising the codec. `(reader, writer)` is the only contract production offers for a serial radio, and it is already faked **four** times with four incompatible semantics: `test_discovery.py` blocks on an empty queue, `test_serial_civ_link.py` blocks but raises a queued `BaseException`, `test_yaesu_cat_transport.py` raises `asyncio.TimeoutError` off a list index, and `test_serial_open.py` — which we added ourselves in #3503 — has its own pair. Rule of three is satisfied at this boundary and nowhere else.

## Codec ringfence — three rules a reviewer can check

1. The fake uses `parse_civ_frame` / `build_civ_frame` for the frame **envelope** and never touches `SerialFrameCodec`. `CivSerialPort` de-frames with ~10 deliberately dumb lines (split on `FE FE`, take to the first `FD`) — no resync-on-nested-SOF, no `0xFC` abort, no partial-frame timeout, no `max_frame_len`. It therefore **cannot** mask a `SerialFrameCodec` bug, which is the blindness that would actually matter.
2. Envelope shared, payload never. Every payload byte is a literal. The module imports from `rigplane.commands` exactly `parse_civ_frame`, `build_civ_frame`, `command_carries_sub`, `CONTROLLER_ADDR` — nothing else.
3. **`tests/test_serial_civ_link.py` and `tests/test_serial_stub_framing.py` are untouched** and must stay that way; they build their streams from byte literals and are the only thing that catches a stream-codec bug. Both are named as forbidden consumers in `civ_fake.py`'s module docstring, so the rule lives next to the code rather than in a PR body. `git diff --name-only` proves neither is in this diff.

## Faults live on the port, not the radio

One fault mode in this PR: `silent` — the port opens, but the device on it is not a radio. That is deliberate shape, not minimalism: an IC-705 NAKing a Xiegu opcode is **identity**, not an injected failure, and conflating the two would be the first wrong thing every later migration inherits. The radio class has no fault surface at all.

## Why test-only

The original plan paired this with a production fix. That pairing is gone because of our own work: #3503 already fixed MOR-2494's CI-V and Xiegu probe sites and already landed the fault tests this PR was going to write (`tests/test_serial_open.py: test_civ_probe_unexpected_open_error_logs_warning` and its three siblings). MOR-2494's four remaining blind sites are all Yaesu CAT (ASCII `readuntil(b";")`) or USB-audio resolution — a CI-V radio fake cannot reach any of them.

Owner decision, 2026-09-16: accept a test-only PR 1. The alternatives were pairing it with MOR-2495 (couples a TX-path behaviour change to a fixture, so a rollback takes both) and re-scoping around a Yaesu fake (would not exercise the CI-V seam at all).

## Mutation results

Each applied, the killed test recorded, then restored with a verified-clean tree:

| mutation | killed |
|---|---|
| IC-705 accepts `0x1D 0x19` | `test_handle_naks_opcode_the_model_does_not_implement`, `TestProbeXieguModelId::test_nak_returns_false` |
| `silent` yields bytes anyway | `test_silent_port_accepts_writes_but_reads_never_yield`, `test_timeout_returns_false`, `test_timeout_tries_next_baud` |
| `.commands` records raw bytes | `test_commands_records_parsed_frames_in_order`, `test_valid_reply_returns_true` |

The third matters most: if the log held bytes, the whole exercise would reduce to a fifth name for the same thing.

## Gates

- `uv run pytest tests/test_discovery.py tests/test_civ_fake.py tests/test_serial_open.py tests/test_serial_civ_link.py tests/test_serial_stub_framing.py -q` -> **165 passed**
- `tests/test_discovery.py` alone: **91 passed**, the same count as the merge base — conversions replace tests, they do not add
- `uv run ruff check src/ tests/` -> All checks passed!
- `uv run ruff format --check src/ tests/` -> 751 files already formatted
- `uv run lint-imports` -> Contracts: 5 kept, 0 broken

## Not in this PR, deliberately

The shared "was / was not sent" assertion helper (PR 2, once three call sites exist) · lifting `_ReceiverState` (zero readers here — it would be an 18-field orphan; it arrives with the serial-backend migration) · migrating `_FakeSerialCivLink` or `MockTransport` · a LAN adapter · deleting any existing double · the `open_error` fault (its tests already exist) · NAK-next-N, drop-this-reply, cable-pulled · Yaesu or Kenwood · any probabilistic fault.

`_XIEGU_MODEL_ID_REPLY` / `_XIEGU_NAK_REPLY` / `_XIEGU_MODEL_ID_CMD` **stay**: `TestIsXieguModelIdReply` and two `TestDiscoverSerialRadios` tests still use them, and they are outside this conversion's scope.

## Review findings, and what was done

Independent review PASSed the first head with four findings; three were acted on in `362b230c`.

- **An orphan.** `CivRadioFake.opcodes` had zero callers anywhere. Its natural consumer is the assertion helper deferred to PR 2. Deleted.
- **A branch nothing reached.** The reviewer wrote its own mutation expecting it to survive — and it did: deleting the unknown-opcode-on-broadcast silence branch left every test green. The behaviour is domain-correct (a radio does not NAK a broadcast), so the branch stayed and the unreachability went: one test now contrasts the two cases directly, same unknown opcode, direct → `FA` NAK, broadcast → silence. Re-run at the new head, that mutation kills `test_broadcast_unknown_opcode_gets_silence_direct_gets_nak`.
- **A docstring that overstated.** `handle` was described as "synchronous and pure"; it appends to `self._commands`. Now "deterministic, performs no I/O, and needs no event loop", with the append stated explicitly.

The fourth finding was this body claiming six conversions where there are five — corrected above.


morozsm:
Agent Review: PASS e386b3d047bd7b3d499dd84095dc2d9bbd3e1ba7

REQUIRED BEFORE MERGE / MANDATORY SQUASH-BODY CORRECTION findings are at the bottom (F1-F4). None reaches BLOCK level: no codec-masking behaviour, no protected file touched, no assertion quietly dropped.

## 1. The shape (check 1) — the split is right

`CivRadioFake` (transport-free identity + opcode table + parsed command log) / `CivSerialPort` (adapter + faults) accommodates the stated next steps without breaking open:

- **Stateful radio (frequency/mode):** the change is `_IDENTITY_TABLE`/`_opcodes` value type `bytes` → a responder callable reading state. Contained in one class; `handle(frame) -> bytes | None`, the constructor, the log, and every existing call site survive. Cheap.
- **Per-receiver MAIN/SUB:** production `parse_civ_frame` already unwraps Command29 (`src/rigplane/commands/_frame.py:517-549`), so `parsed.receiver` is visible inside `handle` today. No restructuring needed; receiver-keyed state is additive.
- **LAN adapter:** a sibling adapter class. The radio needs nothing, because the command log lives on the radio, not the port — "was X sent" assertions survive the adapter choice. That placement is the correctly inherited part.
- **More faults (NAK-next-N, drop-reply, cable-pull):** all land on the port, consistent with the identity/fault split. Additive knobs, no restructure.

Two edges worth naming for PR 2+, neither expensive today: (a) `handle` is strictly call-and-answer — a radio that emits unsolicited frames (meter/scope streams) would need a push channel in an adapter, not a change to the radio; (b) the fake never echoes the command onto the bus, while `_try_baud` explicitly reads past an echo (`src/rigplane/backends/discovery.py:217-234`). Current consumers scan past echo, so nothing is masked — but a future echo-tolerance test will want an `echo=` knob on the port.

## 2. Codec ringfence — all three rules hold

- (a) `_PortWriter.write` (tests/civ_fake.py:201-217) is exactly: find `FE FE` → first `FD` after it → slice to radio → consume. No nested-SOF resync, no `0xFC` abort, no partial-frame timeout, no length cap — versus `SerialFrameCodec.feed` (src/rigplane/backends/icom7610/drivers/serial_civ_link.py:56-95) which implements all four. It is genuinely dumber than the codec (it even clears a trailing lone `FE` the codec would keep), so it cannot mask a codec bug.
- (b) The module imports from `rigplane.commands` exactly `parse_civ_frame`, `build_civ_frame`, `command_carries_sub`, `CONTROLLER_ADDR` (tests/civ_fake.py:37-42). Every payload byte is a literal (`0xFA` at :137, table payloads at :59-63); the profile-sourced `civ_addr` is an address, not payload, and the docstring scopes that claim correctly.
- (c) `git diff --name-only 6bf81ed7..HEAD`: neither `tests/test_serial_civ_link.py` nor `tests/test_serial_stub_framing.py` appears. (The diff's three non-test files — `filter-controls.ts`, `serial_civ_link.py`, `serial_open.py` — come from main-side #3509/#3510 merged in by e386b3d0, not from this PR's commit e827df72; verified per-file with `git log 6bf81ed7..HEAD -- <path>`.)

## 3. Identity versus fault — encoded as claimed

IC-705 is `(b"\x01\x05", {})` — no `0x1D 0x19` entry, so the NAK emerges from the table lookup at civ_fake.py:133-137; the radio class has no fault parameter at all, `silent=` lives on `CivSerialPort`. Domain claim verified against the in-repo evidence: base `tests/test_discovery.py:104-114` has `_IC705_RESPONSE` = `...19 00 01 05` at 0xA4, and the X6200 scenarios (current :1146, :1175) feed `_IC705_RESPONSE` as the X6200's transceiver-ID reply — both models answer `0x19 0x00` with `01 05`; only `0x1D 0x19` yields `62 00` (`_XIEGU_MODEL_ID_REPLY`), with `_XIEGU_NAK_REPLY` for the IC-705. The fake agrees byte-for-byte.

## 4. Conversions — compared one by one against 6bf81ed7

Five conversions, not six — see F1. No assertion dropped:

| test | verdict |
|---|---|
| `test_timeout_tries_next_baud` | **Stronger.** Same result asserts; `seen_bauds == [19200, 9600]` beats `call_count == 2` (order + values); the 9600 reply is now causally produced through the port. |
| `test_valid_reply_returns_true` | **Equivalent, one byte narrower.** `writer.written[0] == _XIEGU_MODEL_ID_CMD` (7-byte equality) → parsed-frame tuple `(to_addr, command, sub, data)`. Preamble/terminator are still implicitly verified (de-frame + `parse_civ_frame` would fail otherwise); `from_addr == 0xE0` is no longer asserted but is a production constant. Gains: asserts on the semantic frame, in order. |
| `test_nak_returns_false` | **Stronger.** Same `is False`; the NAK is now causally produced by IC-705 identity. |
| `test_timeout_returns_false` | **Equivalent-or-stronger.** Same `is False`; now exercises write-accepted-but-no-reply via `silent=`. |
| `test_closes_writer` | **Equivalent.** Same `closed is True`, now against `_PortWriter.close()`. |

`test_open_failure_returns_false` (5th test of the class) was already literal-free and is untouched — correctly so. Count check: 91 test functions at base and head, diff shows 0 test defs added/removed; conversions genuinely replaced.

## 5. Mutations

| mutation | result |
|---|---|
| IC-705 accepts `0x1D 0x19` | **Killed** — `test_handle_naks_opcode_the_model_does_not_implement`, `TestProbeXieguModelId::test_nak_returns_false` (2 failed / 98 passed). Matches PR. |
| `silent` yields bytes anyway | **Killed** — `test_silent_port_accepts_writes_but_reads_never_yield`, `test_timeout_tries_next_baud`, `test_timeout_returns_false` (3 failed / 97 passed). Matches PR. |
| `.commands` records raw bytes | **Killed** — `test_commands_records_parsed_frames_in_order`, `test_valid_reply_returns_true` (2 failed / 98 passed). Matches PR. The log is pinned to parsed `CivFrame`s, not bytes. |
| **Mine:** delete the unknown-broadcast→silence branch (civ_fake.py:135-136) | **SURVIVED — 100 passed.** No test in this PR sends an unknown-opcode broadcast, so that branch is unreachable in this diff. See F3. |

Each restored with `git restore`; `git status --short` verified empty after every run.

## 6. Orphans

- `opcodes` property (civ_fake.py:106-109): **zero users repo-wide** (`rg '\.opcodes'` over src/ and tests/ finds nothing but `self._opcodes`). See F2.
- Broadcast-silence branch (civ_fake.py:135-136): no reaching test — proven by the surviving mutation. See F3.
- Everything else has a user in this PR: `model`, `civ_addr` (+ override), `model_id`, default `"IC-7610"`, `silent=`, all three `_IDENTITY_TABLE` entries, `_BROADCAST_ADDR`, every `_PortWriter` method.

## 7. Prose honesty

- Ringfence docstring rules 1-3: all three verified accurate against the code; both named forbidden files are indeed byte-literal-built and neither imports the fake.
- Module docstring calls `handle` "synchronous and pure" (civ_fake.py:5) — it is deterministic and I/O-free but appends to `self._commands`; not side-effect-free. Minor: see F4.
- Class docstring "no fault surface on this class": accurate.

## Gates (exact head, clean tree)

```
uv run pytest tests/test_discovery.py tests/test_civ_fake.py tests/test_serial_open.py tests/test_serial_civ_link.py tests/test_serial_stub_framing.py -q --tb=short --timeout=300 --timeout-method=thread
  165 passed in 2.09s
uv run pytest tests/test_discovery.py -q --tb=line
  91 passed in 0.89s        (base also 91: grep -c 'def test_' = 91 at 6bf81ed7 and at HEAD)
uv run ruff check src/ tests/
  All checks passed!
uv run ruff format --check src/ tests/
  751 files already formatted
uv run lint-imports
  Contracts: 5 kept, 0 broken.
```

## Findings

- **F1 — MANDATORY SQUASH-BODY CORRECTION.** PR body claims "the first six tests converted" and "converts `TestProbeXieguModelId` (5 tests)"; actual is **five** conversions: four of the class's five tests (`test_open_failure_returns_false` untouched) plus `test_timeout_tries_next_baud`. Whoever merges must correct the body/squash message and say so in the PR.
- **F2 — REQUIRED BEFORE MERGE.** `CivRadioFake.opcodes` (tests/civ_fake.py:106-109) is an orphan: no caller anywhere. Delete it, or give it a user before merge (its natural user — the "was never sent" helper — is explicitly PR 2).
- **F3 — REQUIRED BEFORE MERGE.** The unknown-broadcast→silence branch (`handle`, tests/civ_fake.py:135-136) is reached by nothing in this PR (surviving mutation proves it). Domain-correct per the comment, but per repo rules it is a branch nothing reaches: either pin it with one test (`radio.handle(unknown broadcast) is None`) or let it arrive with the migration that needs it. If F2/F3 are fixed by a commit, that new head needs its natural CI and fresh review per policy.
- **F4 — minor, fix opportunistically.** "synchronous and pure" overstates `handle`; "deterministic, no I/O" is accurate.

morozsm:
Agent Review: PASS 362b230c9aac4bea2b92a76a9fc031f54b54545b

Delta-only re-review of e386b3d0 → 362b230c (one commit, `tests/civ_fake.py` + `tests/test_civ_fake.py` only). Prior shape/identity/conversion conclusions stand; the delta touched none of them.

**F3 — mutation re-run (the one that survived last time).** Deleted `if parsed.to_addr == _BROADCAST_ADDR: return None` from the unknown-opcode path: `1 failed, 100 passed`, the failure exactly `tests/test_civ_fake.py::test_broadcast_unknown_opcode_gets_silence_direct_gets_nak` (broadcast half, line 59). The branch is now pinned. The new test is a genuine contrast, not one half: same unknown opcode 0x1D 0x19, direct to 0xA4 → `0xFA` NAK, broadcast to 0x00 → silence, both asserted. Working tree restored, `git status --short` empty.

**F2 — closed.** `opcodes` property deleted; `git grep '\.opcodes'` at e386b3d0 shows no caller ever existed, none remains. `_opcodes` still built at `civ_fake.py:90` and read at `:130`; the NAK/Xiegu-reply tests exercise it and pass.

**F4 — closed.** "synchronous and pure" replaced with "deterministic, performs no I/O, and needs no event loop", and `handle`'s docstring now explicitly says it records each parsed frame on `self._commands` — accurate about the append side effect. No pure/synchronous overstatement survives anywhere in the file (old phrase was only at old line 6).

**F1 — confirmed.** Body says five conversions; code matches: `test_valid_reply_returns_true`, `test_nak_returns_false`, `test_timeout_returns_false`, `test_closes_writer` converted, `test_open_failure_returns_false` correctly untouched (literal-free), plus `test_timeout_tries_next_baud`.

**Delta broke nothing.** Serial test files absent from `git diff --name-only e386b3d0..HEAD`; ringfence intact — imports from `rigplane.commands` still exactly `parse_civ_frame`/`build_civ_frame`/`command_carries_sub`/`CONTROLLER_ADDR`, de-framing still the deliberately dumb split (no resync/abort/timeout/cap, untouched hunk).

Gates (all at exact head 362b230c):
- 5-file pytest: `166 passed in 2.13s` (was 165; +1 = broadcast test)
- `tests/test_discovery.py`: `91 passed in 0.89s`
- `ruff check src/ tests/`: All checks passed!
- `ruff format --check src/ tests/`: 751 files already formatted
- `lint-imports`: Contracts: 5 kept, 0 broken

REQUIRED BEFORE MERGE (body edits only, no code change): the body's file table still says "9 tests of the fake itself" — the head has 10 — and the header stat "+382/-24" is now "+390/-24" against the merge base. Whoever merges should update both numbers when editing the body, and say so.

PR #3512, opened 2026-09-16
refactor: take the NB-depth range from the radio instead of a hardcoded default

morozsm:
**Guardrail exception, file count only, granted by the coordinating session under the owner's delegation of 2026-09-03: 4 code + 8 test = 12 files.** The line ceiling is untouched — 291 changed lines against 1000. **Three** of the eight test files reference removed symbols in code and would not compile otherwise — `filter-controls.test.ts` 16 such lines, `dsp-adapter.isolated.test.ts` 2, `mor1560-dsp-family-conformance` 1 — counting removed lines that mention a deleted symbol and are not comments. A fourth, `mor1565-vox-cw-family-conformance`, changes for **prose only**: its single mention is in a comment that this change makes false. The remaining four carry the new pins that make the change distinguishable from a silent regression. The change cannot be made in fewer files.

The counting rule is stated because three counts of this disagreed: the body first said 16/7/3/1, which counted comment lines as references; the independent review said 16/6/1/0; counting code lines only, with comment lines excluded by prefix, gives 16/2/1/0. The claim that carries the exception — that these files cannot stay unchanged — holds for three of them on compilation grounds and for the fourth on a false sentence.

The NB-depth control stops falling back to a hardcoded range and takes the published one, the way `manual_notch_freq` already does.

## What was wrong

`CONTROL_DEFAULTS.nb_depth` held `{rawMin: 0, rawMax: 9, displayMin: 1, displayMax: 10}` — a hand-written copy of what the only radio publishing the control already declares. `rigs/ic7610.toml` is the sole publisher, verified with `grep -l 'controls.nb_depth' rigs/*.toml`. For every other radio the default fabricated a 1..10 scale for a control the radio does not have. That is the "no hardcoded radio values" rule, broken in the data rather than in the code.

## What changes

`CONTROL_SPECS.nb_depth` becomes `{ capability: null, defaults: null }`, matching `manual_notch_freq`, and `CONTROL_DEFAULTS.nb_depth` goes. The three production call sites move onto `resolveControlContract(caps, 'nb_depth')`, after which `nbDepthRawToDisplay` and `nbDepthDisplayToRaw` have no callers and are deleted.

Each site handles a missing contract the way the notch path settled in #3502: `panel-adapters` projects "unavailable" rather than a number; `panel-props` yields 0 behind the existing `hasNbDepth` gate, which is untouched; `panel-commands` **refuses to dispatch** rather than sending a display value as a wire value.

## The trap this had to avoid

`controlRangeFromCapsOrDefault` throws for any key without a `CONTROL_DEFAULTS` entry, and its `nb_depth` call site in `panel-adapters.ts: projectDspRawValueToDisplay` sits inside a `try` whose `catch` returns `null`. Deleting the default without moving that call would have swallowed the throw and turned every NB-depth projection into "unknown" — for every radio including the IC-7610 that does publish the control. A silent loss of readings, not a crash, and no existing test was obliged to notice.

## Tests, each proved to discriminate

Mutations run, observed red, restored, with the failure text recorded:

- fabricate `+1` in `panel-props` → `expected 1 to be +0`
- drop the conversion in `panel-props` → `expected +0 to be 1` (IC-7610 pins 0→1, 5→6, 9→10)
- fabricate in `panel-adapters` → `expected { confirmed: 6, … } to match object { confirmed: null, … }`
- dispatch the display value verbatim in `panel-commands` → `expected [ [ 'set_nb_depth', { level: 6 } ] ] to deeply equal []`, plus reds on 1→0, 6→5, 10→9
- **restore the hardcoded default in `CONTROL_SPECS`** → red at all three levels. That is the pin that fails against a version keeping the fallback, which is the whole point.

Two of the pre-existing tests were pinning the fabricated behaviour itself and now pin its absence.

## Why the displayed value cannot move

`nb_depth` maps raw 0..9 to display 1..10 — ten values against ten, an exact mapping. On an IC-7610 the contract computes what the default computed. That is what separates this from PBT, whose scale is inexact and measurably wrong about the radio (MOR-2497) and which is deliberately untouched here.

## Checks

`npm --prefix frontend run test -- <changed files>` → 10 files, 414 passed. Wider `src/lib/runtime` + `src/lib/radio` → 91 files, 2428 passed. `check` → 0 errors, 0 warnings over 4859 files. `lint` clean.

Follow-up to MOR-2475


morozsm:
Agent Review: PASS a0744240e2b385bd7421ed070f5a85e98c74b374

Independent review at exact head a0744240e2b385bd7421ed070f5a85e98c74b374 (verified `git rev-parse HEAD`; tree clean after review). Attempted refutation on all six axes; could not break the change.

**IC-7610 displayed value does not move.** `rigs/ic7610.toml:1531` publishes `[controls.nb_depth]` as the legacy band 0..9 → 1..10. `resolveControlContract` resolves it through `legacyControlRange`/`legacyControlContract` (filter-controls.ts:445-448) into the same `controlRawToDisplay` formula the deleted default used: `round(((raw-0)/9)·9)+1` = raw+1, exact for all ten values both directions (encode is symmetric: display−1). Pins cover 0→1, 5→6, 9→10 and 1→0, 6→5, 10→9 plus endpoint round-trips. Note: `legacyControlContract` (filter-controls.ts:394-395) now passes `CONTROL_DEFAULTS['nb_depth']` = `undefined` as fallback — harmless because `range` is always supplied (the same shape `manual_notch_freq` already had), but a future edit that makes `range` optional there would dereference it.

**Silent-regression trap avoided.** The swallowing `try/catch` in `projectDspRawValueToDisplay` is deleted wholesale (panel-adapters.ts:629-636); the replacement has no catch and fail-closes through the contract. `rg` over all of `frontend/` finds **zero** remaining production callers of `controlRangeFromCapsOrDefault` — only `nr_level` test harnesses. No other catch reaches a missing `nb_depth` default. Observation (not blocking): that export is now production-orphaned; retire it with `nr_level` when that lane settles.

**Pins discriminate — mutations run by this reviewer, both red, then restored:**
- Restored `CONTROL_DEFAULTS.nb_depth` + `CONTROL_SPECS` default → 6 failures in 3 files: the empty-contract pins, `resolveControlContract` MOR-2475 pin, panel-props "fabricates no 1-10 reading", dsp-command-feedback "fails closed for nb no published control".
- Dropped the `wire === null` refusal in `onNbDepthChange` (dispatch display verbatim on unusable domain) → "refuses when the declared nb_depth entry carries no usable scale" red.

The rewritten pre-existing tests did not lose discrimination: the MOR-498 offset pin moved from null-caps fabrication to the published domain (same 0→1/5→6/9→10 assertions), and a new anti-pin asserts 0 with a comment naming the fabricated values it must not produce. mor1560's refusal walk keeps its discriminating gate.

**No-contract handling at the three sites is correct.** All three sites' `caps?.controls?.nb_depth` presence gates predate this PR (panel-adapters structural gate line 596, panel-props `hasNbDepth` MOR-502, panel-commands gate line 681), so for radios publishing nothing the visible behavior is unchanged. The only behavior change is the declared-but-unusable corner: panel-commands previously dispatched a wire value encoded against the fabricated scale onto the radio; refusing is strictly safer. panel-adapters "unavailable" and panel-props 0-behind-hidden-gate match the notch precedent.

**Deletion safety.** Comment-stripped scan (block, line, and `#` comments removed) across all 12 changed files: zero references to `nbDepthRawToDisplay`/`nbDepthDisplayToRaw`; raw frontend-wide grep also zero.

**Prose.** The false docblock sentence is corrected to "(`nr_level` today)"; mor1565's stale `filter-controls.ts:122-125` reference is fixed. Body claims verified: 12 files (4 code + 8 test), 291 lines (184+107), ic7610 sole publisher of `controls.nb_depth` (ic9700 explicitly absent, ic7300/ic705 carry commands only).

REQUIRED BEFORE MERGE (body correction — apply when editing the PR body or writing the squash body): the sentence "Four of the eight test files change because they reference symbols this PR removes and **would not compile otherwise** … 16 / 7 / 3 / 1" is not accurate. Only `filter-controls.test.ts` and `dsp-adapter.isolated.test.ts` fail compilation (deleted imports); mor1560's old `controlRangeFromCapsOrDefault('nb_depth', …)` compiles and fails at runtime; mor1565 references no removed symbol (prose-only change). Strict removed-symbol line counts are 16/6/1/0, and I could not reproduce 7/3/1 under any counting rule. Reword to "would fail otherwise" and correct or drop the per-file counts.

Checks: `npm --prefix frontend ci` clean; changed-file vitest at the exact head: 8 files, 401 tests passed (run twice, before and after mutations). CI as found: `quick` pending, `visual` pass, gate rows awaiting this directive.

PR #3513, opened 2026-09-16
docs: remove a merged PR's worktree and branch immediately

morozsm:
## What

Records a standing owner authority: after a PR merges, its worktree and branch are removed immediately.

Two files, because the rule is stated in both places today:

- `CLAUDE.md`, "Workspace lifecycle and context" — a new paragraph after the existing preservation rule
- `docs/internals/coordinator-policy.md` — one sentence extended, so the two do not drift

## Why

Owner instruction, 2026-09-16: "Нужно чтобы удалялась ветка и рабочая директория после каждого мерджа."

The prompt for it: a cleanup pass that day started from **80** accumulated worktrees in this repository. Sorting them took three passes and required per-branch evidence — classifying each by whether the symbols it added already exist in `main` — because by then nobody could tell what was finished from what was not. At the moment of merge that question has a known answer, so removing then costs nothing.

## Why this is not a reversal of the rule above it

The existing text withholds cleanup authority from three events — PR creation, a failed attempt, a context rollover — and requires four conditions before a worktree is removed: released ownership, retained work and evidence, verified state, existing authority. A merge *normally* satisfies all four, and the owner instruction supplies the fourth. The three withheld events do not, which is what separates them.

The new paragraph says so explicitly rather than leaving a reader to reconcile two rules that look opposed.

**An earlier revision of this PR said "all satisfied at that moment", full stop. Independent review blocked it, correctly**, with a case where ownership is *not* released at merge: when an open PR is based on the merged branch. Deleting that base auto-closes the child, and GitHub then refuses both reopen (base gone) and retarget (already closed). `AGENTS.md` already forbids it and mandates retargeting children first. Neither of the two limits then stated caught it — the parent's worktree is clean and nobody is working in it.

Stacked PRs are a lived pattern in this repository, which is why that retarget procedure exists at all. The text now lists it as the first of three conditions to check.

## Three conditions to check, not two

The first came from review (above). The other two are load-bearing and come from the same day's cleanup:

- **Uncommitted changes.** A worktree whose branch diff against `main` is empty can still hold unsaved work. One that day did: `pbt-pr4` looked safe by every automatic signal, and held 32 uncommitted lines belonging to an in-flight PR on a ticket that was open at that moment.
- **Another live session.** A second worktree was another Claude session's working directory, found by checking which processes had it as their cwd. Merged, clean, and still not ours to remove.

Also recorded, both raised by review:

- **Unpushed local commits are unretained work**, and are not "uncommitted changes" in the worktree — a separate thing the first draft conflated. `git branch -d` refuses them; `-D` discards them silently. The text now says not to reach for `-D` to make a refusal go away.
- `gh pr merge --delete-branch` run from inside a worktree can exit non-zero **after** the merge has already succeeded, because the CLI then fails to update the local branch. Confirm with `gh pr view --json state`, not the exit code. This bit twice in one day. Review noted this fact now lives in both `AGENTS.md` and `CLAUDE.md`; that is a duplicate to keep in sync, not a contradiction.

## Verification

Documentation-only, per the `AGENTS.md` exception: readback and diff proof, no citation, link, product, visual or full automation. No code paths touched, no `file:line` citations added, no Markdown links added.

Both edits read back above their surrounding text; the `coordinator-policy.md` sentence is a single-sentence extension of an existing one, verified in place.


morozsm:
Agent Review: BLOCKED 4be7170efdfb3a86f0d7ae1364690d56d4e08c8b

**Point 1 — the exception logic fails in one realistic case.** The new CLAUDE.md paragraph claims: "The preceding conditions are all satisfied at that moment, which is what distinguishes a merge from the cases above." Not always. When an open child PR is based on the merged PR's branch, ownership of that branch is NOT released at merge — AGENTS.md ("Protected main and review gate") forbids exactly this deletion: "Never merge with `--delete-branch` while another PR is based on that branch: GitHub auto-closes the child and then refuses to reopen (base gone) or retarget (already closed) it." Merging without `--delete-branch` and then obeying the new standing authority — "remove its worktree and delete its branch immediately" — produces the same auto-closed child and the same recovery procedure AGENTS.md describes. Neither stated limit catches it: the parent's worktree is clean and no live session works in it. Stacked PRs are a lived pattern here; the retarget-every-child-before-merging-the-parent procedure exists precisely because it happens. As written, the new text authorizes cleanup that the surrounding governance forbids.

Required fix (small): add the dependent-PR condition to the limits in BOTH files — the merge exception applies only once no open PR is based on the merged branch (i.e. after the retarget-and-push step AGENTS.md already mandates) — and qualify the "all satisfied at that moment" sentence accordingly.

Points 2–4 pass:
- The two files state the same rule with the same two exceptions; no drift.
- The limits are worded as binding, not advice: "Two limits still bind: ... is inspected before it is touched and left alone if either is true"; "unless ..., which is checked first."
- The `gh pr merge --delete-branch` exit-code claim is accurate, narrow ("can exit non-zero *after* the merge already succeeded"), and matches AGENTS.md's own record of the same behavior.

Minor, non-blocking:
- A merged branch can also carry unpushed local commits — work not retained, yet not "uncommitted changes" under the stated limit. `git branch -d` refuses, `-D` doesn't; the rule doesn't distinguish. Worth half a sentence while editing.
- The gh exit-code fact now lives in both AGENTS.md and CLAUDE.md — a third copy to keep in sync, though not contradictory.

Verification (documentation exception, no suites run): `git diff --name-only a1bd572c..HEAD` shows only `CLAUDE.md` and `docs/internals/coordinator-policy.md`; no `file:line` citation and no new Markdown link in the diff; both changed regions read in place at exact head `4be7170e`, base `a1bd572c`; PR head SHA matches.

morozsm:
Agent Review: PASS 9b98752530a14211653fba0b42e2cbb6742d11be

Delta review of 4be7170e..9b987525 (docs-only: CLAUDE.md, docs/internals/coordinator-policy.md).

**Blocker: closed.** The new first condition states the dependent-PR case correctly and bindingly ("Clean up only once no open PR is based on the branch"), names the mechanism (delete base -> auto-close child -> GitHub refuses reopen and retarget), and points at the right precondition — AGENTS.md's retarget-every-child-first procedure. It agrees with AGENTS.md's own wording; no drift.

**"Three" count: correct.** The three gated checks map onto the four preconditions (ownership released / work retained / state safe / authorized). The one additional realistic case — unpushed local commits — is named in both files as unretained work distinct from uncommitted changes, with the "-D silences a -d refusal" warning; it is also self-defending via the `-d` refusal. No unnamed fourth case found.

**Consistency: no drift.** coordinator-policy.md's prose carries the same three checks and the same unpushed-commits caveat; the -d/-D mechanics living only in CLAUDE.md is appropriate layering.

**Prose honesty: clean.** All added sentences check out. No new file:line citations, no new relative Markdown links.

PR #3514, opened 2026-09-16
feat(MOR-2497): add the measured PBT lattice conversion, pinned to bench data

morozsm:
Step 1 of MOR-2497. Adds the PBT raw↔Hz conversion that the radio actually implements, pinned against data taken off the radio. **No consumer changes and no profile changes** — the existing `pbtRawToHz`/`pbtHzToRaw` and all fifteen of their call sites are untouched, and this PR adds nothing to their path.

## Why

`rigs/ic7610.toml [controls.pbt_inner]` declares `display_min = -1200`, `display_max = 1200` over `raw_min = 0`, `raw_max = 255`. Measured on the bench 2026-09-16, every part of that is wrong: the span is ±filter_width/2 and therefore not a constant, the step is 50 Hz rather than one raw unit, and raw 0 and 255 are never reachable. The IC-7610 CI-V Reference Guide gives no Hz for PBT at all — only `0000=max. CCW, 0128=center, 0255=max. CW` — while it does give Hz for CW pitch and APF. The ±1200 was authored by us; `docs/capabilities-matrix.md` calls it "roughly".

## The measurement

Sweeping `set_pbt_inner(raw)` over every raw 0..255 and reading back — the radio snaps each write to its own lattice — gave the same result on the IC-7610 over LAN and the IC-7300 over serial: 73 positions `1..254` at a 3600 Hz filter, 37 at `3..252` at 1800, 11 at `11..244` at 500. Fifty Hz per step in all three. Confirmed against the radio's own display: at filter 3600 with inner at 254 the panel read `SFT +900` and `BW 1.8`, and those two readings discriminate — a 25 Hz step would have shown `SFT +450`.

## The model

```
P = filterWidthHz / 50 + 1        m = floor(255 / (2P))
raw(i) = m + round(i * (255 - 2m) / (P - 1))      round = half-up
Hz(i)  = (i - (P-1)/2) * 50
```

The centre lands on raw 128 on its own: at 3600, `i=36` gives `36*253/72 = 126.5`, half-up gives 127, so `raw = 1 + 127 = 128`. **There is no centre special case, and it would be dead code.** An earlier revision of this branch carried one, because the coordinator's analysis was done in Python, whose `round` is half-to-even and turns 126.5 into 126. Half-to-even puts the centre at 127 and does not match the radio; the measurement selects half-up. The archived evidence has been corrected accordingly.

## The pin is against the data, not the formula

The complete measured lattice — all 73 reachable positions at filter 3600 — is checked into `filter-controls.test.ts` as a fixture. A test that restates the formula proves nothing; this one fails when the formula stops describing the radio.

Three mutations, each a model that looks right, run and reverted:

- spread evenly over the full 0..255 instead of `[m, 255-m]` → 5 failures, `expected +0 to be 1`, `to be 3`, `to be 11`
- round half-to-even instead of half-up → 3 failures, `expected 127 to be 128`, at the centre cell on all three widths
- 25 Hz per step → 7 failures, `expected 1775 to be 1800`, `expected 73 to be 37`, `expected 21 to be 11`

The coordinator re-ran the second independently at this head and saw the same three failures.

## Decisions recorded

The step is a required parameter, not a literal: 50 Hz is measured on the IC-7610 and IC-7300, while `ic705` and `ic9700` were not on the bench, so applying it to them is inference and must be visible as such. A raw value between lattice points snaps to the nearest, ties toward the centre — the radio's own half-step rule was not measured, and the centre side never attributes a larger edge shift than the value can mean. Zero, negative, NaN and non-multiple-of-step filter widths, and raw outside 0..255, all yield `null` rather than a fabricated number.

## Checks

`npm --prefix frontend run test -- filter-controls` → 2 files, 56 passed. `check` → 4859 files, 0 errors, 0 warnings. `lint` clean. 2 files, +211, measured at this head.

Part of MOR-2497


morozsm:
Agent Review: PASS 07264bd7bed377e4adad14b1d6a8c724dc38374e

Independent exact-head review of PR #3514 (read-only checkout at 07264bd; head verified with `git rev-parse HEAD`). CI as found: `visual` pass, `quick` pending on the self-hosted runner, gate failing only for lack of this directive. Verdict is on the code; coordinator still confirms `quick` green before merge.

**Fixture independence — verified.** `PBT_LATTICE_3600` (filter-controls.test.ts:400-406) is 73 numeric literals; the loop at :409-416 feeds them to the implementation and derives only the Hz ladder `(i - 36) * 50` from literals 36/50, never from the module. The 1800/500 tests assert literal counts and literal endpoints (:434-439, :447-452). Nothing generates an expectation from the code under test. Nuance, not a defect: the interior *Hz* association is the model's centred-ladder assumption, anchored independently at the endpoints (254→+1800 via the SFT/BW display reading, 128→0 via the CI-V "0128=center") — and the PR's comment accurately describes the fixture as the measured raw positions, nothing more. The bench provenance of the literals is not verifiable from the repo; that is inherent.

**Mutations re-run at this head, each restored with `git checkout --`:**
- margin 0 → 5 failures: `expected +0 to be 1`, `expected +0 to be 3`, `expected +0 to be 11` (matches body).
- half-to-even in `latticeRaw` → 3 failures, each `expected 127 to be 128`, at the centre cell on all three widths (matches body; arithmetic checked: 36·253/72 = 126.5, half-up → 128 with margin 1; the two rounding rules differ at exactly one cell per width and the fixture catches all three).
- `PBT_MEASURED_STEP_HZ` 50 → 25 → 7 failures: `expected 1775 to be 1800`, `expected 73 to be 37`, `expected 21 to be 11` among them (matches body).

**Centre special case — genuinely absent.** `latticeRaw` (filter-controls.ts:159-162) is margin + `Math.round`, no `i === centre` branch; a pin would be dead code as claimed.

**Degenerates — verified.** Widths 0/-100/NaN/+Inf/125, steps 0/-50/NaN, width 13000 (260 steps > 255), raw -1/256/NaN all asserted null (:456-482). Return types are `number | null` (:185, :199); null flows out uncoerced. I additionally checked the untested infinite variants (−Inf width, ±Inf raw/hz) behave as null.

**Prose — clean.** No comment claims the radio "pins" the centre, none calls `pbtRawToHz` wrong, none promises a later PR. The half-up claim is tied to the fixture ("fails the swept-lattice fixture"), which mutation B confirms is true.

Minor (not merge-blocking): the Hz-side half-step tie (e.g. −1775 → raw 5, toward centre) is documented (filter-controls.ts:192-198) but has no direct test pin — only the raw-side tie is pinned (:484-490). The rule is deterministic, conservative, and honestly labelled as chosen rather than measured; a one-line test would close it.

PR #3515, opened 2026-09-17
docs: state the handover mechanics the succession protocol left unwritten

morozsm:
Documentation-only: three paragraphs added to the succession section of `docs/internals/coordinator-policy.md`. 1 file, 6 changed lines — inside both guardrails, no exception needed.

The section fixed the *order* of a handover — checkpoint, read-only bootstrap, ACK, RELEASE, ACCEPT, one writer — and left its *mechanics* unwritten. Three of them failed during the 2026-09-16 handover of the MOR-2478 line, in the same hour, and each is now stated.

## 1. A step is delivered, not only recorded

Adding a line to the handoff document notifies no session. The successor verified the checkpoint, wrote its ACK, and stopped to wait for a RELEASE nobody had been asked for; the source had no way to learn the ACK existed. Both sides waited on each other until the owner intervened.

The rule: whoever records ACK, RELEASE or ACCEPT also sends it to the counterpart in the same turn, saying where the line was written and what was verified; the counterpart is resolved by the worktree and branch the checkpoint names, against the host's session list, not by a session title; the message self-identifies so a misdelivery is harmless; and a successor that has ACKed reports a silent counterpart to the owner rather than waiting. **Established:** the peer listing's title did not carry the checkpoint's identifier — the source was found by `cwd` `…/worktrees/happy-agnesi-0ef76f` and branch `claude/strange-moser-f2af4e`.

## 2. Record the addressable id, not an internal one

The ACK recorded `376d4b1b-cc3d-4743-acde-20547170b373`, taken from a scratchpad path. It is not an address. The source could not reply to it and had to spend a message repairing the handover it was supposed to be ending. **Established:** sending to that id returns `No agent named '376d4b1b-…' is reachable`; `get_session self` reports the addressable id as `local_644a285e-14af-4b1b-929e-f7ae0a864a0a`.

## 3. The readback has to be able to fail

Confirming a handoff-document write from the mutation's own response does not confirm it. **Established:** `documentUpdate` returned `success: true` while echoing the `updatedAt` the document carried *before* the write — observed twice; a separate `document(id:)` query showed the new timestamp both times. The paragraph names what to compare: bytes against what was sent, heading count, the whole last sentence written, and the presence of the **oldest** section as well as the newest — a truncating write leaves the newest heading exactly where its author looks for it. The oldest-section check came from the outgoing session, which reports having lost roughly 90 KB of that document's history earlier the same day to a write checked only by its newest heading; that figure is its report and is not restated in the policy text.

## Scope

No behaviour, no code, no other document. Existing sentences untouched — nothing weakened, nothing restated. All three paragraphs are rules; the observations that produced them stay here in the PR body rather than in the policy.

## Checks

Documentation-only under the `AGENTS.md` exception: no citation, link, Markdown, product, visual or full automation applies, and the required `quick` context is expected to report server-side skipped/neutral. The substantive gate is the exact-head independent review. The rendered paragraphs were read back from the file after the edit.

Two commits, so the squash subject comes from the PR title rather than from a commit message — pass `--subject` at merge regardless.


morozsm:
Agent Review: PASS f3aefed1e12d8a774774ae8d4ec51af73fbf3c40

Verdict given by the repository owner directly, in the session that authored this PR, and relayed here verbatim because the gate reads a PR comment rather than a chat message. No review agent was run: the owner ruled that prose of this size does not warrant one.

Recorded for the file: the author of the change is not the author of this verdict.

PR #3516, opened 2026-09-17
fix(MOR-2497): stop clamping the derived IF shift to a bound the radio exceeds

morozsm:
`deriveIfShift` bounded every derived IF shift to ±1200 Hz. That bound is the same fabricated span MOR-2497 is about, and it truncates a state the radio actually reaches.

## Measured, not reasoned

IC-7610 over LAN, 2026-09-17 08:07 EDT. Filter set to 3600 and read back as 3600. For each target, `set_pbt_inner(raw)` then `set_pbt_outer(raw)`, then both read back; each read-back converted through the measured lattice with half-up rounding (`decimal.ROUND_HALF_UP` — Python's own `round` is half-to-even and produced a false conclusion on this same model once already).

| asked | read back | Hz per edge | shift |
|---|---|---|---|
| 128 | 128 / 128 | 0 | 0 |
| 192 | 191 / 191 | +900 | +900 |
| 224 | 222 / 222 | +1350 | **+1350** |
| 240 | 240 / 240 | +1600 | **+1600** |
| 254 | 254 / 254 | +1800 | **+1800** |
| 0 | 1 / 1 | −1800 | **−1800** |

**The radio accepts both edges at one extreme together** — it does not collapse the passband and refuse first, which was the open possibility. Maximum reachable |shift| at a 3600 Hz filter is 1800 Hz; everything from 1350 upward was reported as 1200.

Baseline restored and read back: `filter=3600 inner=128 outer=128`. No transmit at any point; the run only reads and writes levels.

## Why removing the clamp leaves nothing unbounded

**Corrected after review.** An earlier version of this section said both edges come from `measuredPbtRawToHz`. No production caller uses that function at this head — it was true of where the work is going, not of where it is.

What is true here: the four display-side callers (`panel-adapters.ts`, `radio-view-model-adapter.ts`, `scope-passband-display.ts`, `panel-props.ts`) pass two `pbtRawToHz` outputs taken at one declared range, and the mean lies inside it. Every profile the loader reads — `rigs/*.toml`, non-recursive, skipping `_`-prefixed and `.draft.toml` names, which is the rule in `rig_loader.py` and yields **eight radio profiles** — either declares `[controls.pbt_inner]` with `raw_center = 128` and ∓1200 (four of them: ic705, ic7300, ic7610, ic9700) or declares no PBT range at all and falls through to `PBT_DEFAULTS`, which is the same ±1200 (the other four). So those outputs span −1200..+1191, which the clamp never cut. An earlier version of this line said "four in their own block, seven through the same default, across all eleven `rigs/*.toml`" — wrong under every counting rule, and stated without one; the glob returns ten paths, two of which the loader skips. A caps payload declaring a wider range now yields the larger shift instead of a truncated one; that is the behaviour change, and the test exercises it.

The fifth caller is `mapIfShiftToPbt`, which receives raw 0..255 fields from `panel-commands.ts`. The mean of two raws cannot reach 1200 either, so the clamp never fired there and removal is a no-op — but checking that exception turned up a real defect on that write path, filed as **MOR-2500**: `onIfShiftChange` passes raws where Hz are expected, so moving the IF shift with PBT engaged collapses the passband. Measured, not reasoned — edges at raw 160/96 are a 600 Hz passband and come back 56 Hz. Pre-existing, not introduced or fixed here; it belongs to the writer half of step 2, which must convert that file to Hz anyway.

## Tests

Three new pins, from the measured read-backs rather than from a recomputation of the model: both edges at raw 254 → 1800, at raw 1 → −1800, and the case the bench already had — one edge at 254 with the other centred → 900, matching the front panel's `SFT +900`.

Mutation-checked: restoring `clampToBipolarRange` fails exactly the two extreme-shift cases, `expected 1200 to be 1800`.

**Correction, applied after review of the first head.** That first head said "nothing pinned the clamp at all — 105 tests in `src/lib/radio/` passed with it removed". That was false, and false in a way worth naming: it was a claim about the tree drawn from a run of one directory. `filter-passband-adapter.isolated.test.ts` pinned the clamp directly, in `src/lib/runtime/adapters/`, and the first head broke it — `expected 1984 to be 1200`, and CI `quick` went red on exactly that. The cause was the split: the change was separated from its tests by source file rather than by change-plus-its-tests. To be exact about it, since an earlier version of this body was not — that test is at the merge base, on `main`; it did not "stay on a sibling branch". The change simply broke a test that was already there, and the one-directory run did not reach it. That test is now on this branch, keeping its shape — the same deliberately wide declared range — and asserting the opposite outcome: nothing truncates at 1200 any more.

Also carries the half-step tie rule on the Hz side of `measuredPbtHzToRaw`, named as debt when step 1 merged: described in a comment, pinned by nothing. Mutation-checked by flipping the tie to resolve away from centre — `expected 254 to be 250`.

## The class, enumerated rather than expanded into

A BLOCKED finding naming one instance is a review of its class, so here is the class. Hardcoded ±1200 surviving in production code, by `git grep -n "1200" -- src` with CSS, layout and demo hits removed:

| Site | What it bounds | Why not here |
|---|---|---|
| `filter-controls.ts: FILTER_BIPOLAR_MIN/MAX` via `mapIfShiftToPbt` | the WRITE path | its call sites (`panel-commands.ts`) move in a later PR of this step; changing the symbol twice is worse than once |
| `filter-controls.ts: PBT_DEFAULTS` | the capabilities-store fallback range | fabricated, and removed with the store fallback itself when the converters retire |
| `FilterPanel.svelte: tableIfShiftDomain` / `defaultIfShiftDomain` | the fallback IF-shift domain | needs the filter width threaded into the panel |
| `FilterPanel.svelte` sliders (two `min`/`max` pairs) | the user-facing PBT travel | same; at a 3600 Hz filter these clip the control at 1200 where the radio reaches 1800 |
| `semantic/FilterSurface.svelte:45-47` | the same bound on the semantic surface | **missed by an earlier version of this table** and found by review; same dependency on the filter width |

All **five** need the same thing this PR could not give them — the filter width in scope at the site. (An earlier version of this line said "four" while the table already listed five; the fifth was added after review found it.) Recorded on MOR-2497 rather than pulled into this change.

## Checks

`npx vitest run src/lib/radio/filter-controls.test.ts` — **56 passed**. The whole directory, `npx vitest run src/lib/radio/`, is **109 passed across 6 files**. An earlier version of this body attached the 56 to the directory command; the two are different runs and the smaller number belongs to the single file. Focused, per the changed scope; the full suite is CI's.


morozsm:
Agent Review: BLOCKED a40e08e5bd17e4266262204a825e3671c234efaa

Independent review at the exact head. Three findings; the first is a red test at head, the second and third are false prose inside the diff.

## 1. BLOCKING — the change breaks an existing test; `quick` is red at this head

`frontend/src/lib/runtime/adapters/__tests__/filter-passband-adapter.isolated.test.ts:307-328` — "ifShift clamps to ±1200 via the real deriveIfShift even when the custom PBT scale would exceed it, while pbtInner/pbtOuter stay UNCLAMPED".

Reproduced locally at this head:

```
npx vitest run src/lib/runtime/adapters/__tests__/filter-passband-adapter.isolated.test.ts
 × ifShift clamps to ±1200 ...
 AssertionError: expected 1984 to be 1200
 at filter-passband-adapter.isolated.test.ts:319:39
 Tests  1 failed | 54 passed (55)
```

CI agrees: run 35220784059 (`quick`) fails on exactly this file, `1 failed` in `filter-passband-adapter.isolated.test.ts`. The PR body reports only the two green focused runs and does not mention this.

Both directions checked. Restoring the clamp in `deriveIfShift` (`filter-controls.ts`) turns that file green again (55 passed) while failing the two new extreme-shift rows — so the adapter test is a real pin on the removed behaviour, not a coincidence, and it was green before this commit.

**Required fix (commit):** decide what the adapter contract now is and update `filter-passband-adapter.isolated.test.ts` accordingly — its name, its comment at :308-311, `expect(Math.abs(expectedIfShift)).toBe(1200)` at :319 and `value: 1200` at :327 all assert the removed clamp. Deleting the assertion is not enough on its own: the test also exists to prove `pbtRawToHz` is unclamped while `deriveIfShift` was clamped, and with the clamp gone that discriminator is dead.

## 2. BLOCKING — the bounding argument in the new comment is false of every shipped call site

`filter-controls.ts`, docstring on `deriveIfShift`:

> "Both edges come from `measuredPbtRawToHz` at one filter width, which cannot exceed +/-filter_width/2, so their mean cannot either."

`grep -rn 'measuredPbtRawToHz' frontend/src` outside tests returns only its own definition and this comment. **No production caller uses it.** Every shipped caller of `deriveIfShift` feeds it `pbtRawToHz` output:

- `lib/runtime/props/panel-props.ts` — `pbtRawToHz(rx?.pbtInner ?? 128)` (store range), then `deriveIfShift(pbtInner, pbtOuter)`
- `lib/runtime/adapters/panel-adapters.ts: buildIfShiftFeedback` — `toHz = (raw) => pbtRawToHz(raw, pbtRangeFromCaps(caps))`, three `deriveIfShift` calls
- `lib/runtime/adapters/radio-view-model-adapter.ts` — `deriveIfShift(pbtInnerHz, pbtOuterHz)` from the caps-derived scale
- `lib/runtime/adapters/scope-passband-display.ts` — `deriveIfShift(pbtRawToHz(inner, scale), pbtRawToHz(outer, scale))`
- `filter-controls.ts: mapIfShiftToPbt` — internal; its result is re-clamped, so unaffected

The sentence is not merely premature: as written it states the bound that justifies the removal, and it holds for zero shipped callers. The `±filter_width/2` property of `measuredPbtRawToHz` itself is correct (`(positions-1)/2 * stepHz` = `filterWidthHz/2` via `pbtLattice`) — the false part is "both edges come from" it.

The next sentence is also false:

> "At a call site still on the retired `pbtRawToHz`, the outputs span -1200..+1191 ... removing it changes nothing those call sites can observe."

The span is right **for the shipped profiles only**. I enumerated every `[controls.pbt_inner]`/`[controls.pbt_outer]` block in `rigs/*.toml` and `rigs/examples/*.toml`: `ic705`, `ic7300`, `ic7610`, `ic9700` all declare `raw_center = 128`, `display_min = -1200`, `display_max = 1200`; `ftx1`, `tx500`, `x6100`, `x6200` and the three examples declare no PBT control block. So `pbtRawToHz` spans −1200..+1191 there and the clamp never fired — that half is CONFIRMED. But `pbtRangeFromCaps` reads the range from the profile, and a profile declaring a wider display span produces 1984 Hz at raw 255 — which is precisely finding 1, an observable change in `view.filterPassband.ifShift.reading` at a `pbtRawToHz` call site. "Changes nothing those call sites can observe" is refuted by a test in this repository.

**Required fix (commit):** per CLAUDE.md, delete these two claims rather than soften them, or replace them with a bound established by a command actually run.

## 3. Class sweep — the ±1200 shape occurs downstream and is not addressed

The read path can now emit ±1800, but the IF-shift row that displays it still defaults to −1200..1200: `components-v2/panels/FilterPanel.svelte` (`tableIfShiftDomain`, `defaultIfShiftDomain`) and `semantic/FilterSurface.svelte` (`['ifShift', 'IF shift', -1200, 1200, 25]`), pinned by `FilterPanel.isolated.test.ts` (`aria-valuemin`/`aria-valuemax`) and `FilterSurface.test.ts`. The PR's "Not fixed here" section names only `mapIfShiftToPbt`. Report where the rest of the class lands; no scope expansion required here.

## What I could not refute

- **New tests discriminate.** (a) restoring `clampToBipolarRange`: `expected 1200 to be 1800` and `expected -1200 to be -1800`, 2 failed — matches the PR's reported mutation exactly. (b) negating `measuredPbtRawToHz`: 8 failed including all three new cases. (c) `deriveIfShift` returning its first argument: `expected 1800 to be 900`, caught by "still reports the mean when only one edge is displaced". The two `it.each` rows pass identical arguments so they cannot distinguish mean-vs-first-argument — expected, since they pin the clamp, not the mean.
- **`Math.round` is not a behaviour change**: `clampToBipolarRange` already rounded (`filter-controls.ts`), so rounding is preserved.
- **Check counts in the body are accurate.** `npx vitest run src/lib/radio/filter-controls.test.ts` → 56 passed (1 file); `npx vitest run src/lib/radio/` → 109 passed across 6 files. Both reproduced; the corrected body matches.
- **The bench measurement** cannot be re-run from here. The test expectations (254 → +1800, 1 → −1800, one-edge → +900) and the numeric lattice are properties of `measuredPbtRawToHz` as it already shipped in #3514, so they stand on the code. What rests purely on the read-back claim is the premise that the radio *accepts both edges at one extreme together* — the whole motivation for removing the clamp. One inference in the docstring goes past what was measured: "so the true shift is +1800 Hz" is the mean by definition, not something read back from the radio; the body reports a front-panel `SFT +900` only for the one-edge case.

```
cd frontend
npx vitest run src/lib/radio/ src/lib/runtime/adapters/ \
  src/components-v2/panels/__tests__/FilterPanel.isolated.test.ts \
  src/semantic/__tests__/FilterSurface.test.ts
```
At this head that is `1 failed | 2045 passed` across 64 files. Then the required `quick` run on the new head.

Tree left byte-identical; `git status --short` clean at a40e08e5.

morozsm:
Agent Review: PASS 4f1775b5d4583618f715e4528ecac11ea33d58fd

Delta re-review at the new head, worktree detached at 4f1775b5, `git status --short` clean before and after every mutation.

## The three blocking findings

**1. CLOSED — the pin is inverted and still discriminates.**
`filter-passband-adapter.isolated.test.ts: 'carries the derived ifShift unclamped, including past the retired ±1200 bound'` keeps its wide declared range (`display_min/max` ±2000) and now asserts `expectedInnerHz` (1984) instead of 1200. Mutated both directions in the worktree:
- reinstating `clampToBipolarRange` in `filter-controls.ts: deriveIfShift` → `3 failed`: `expected 1200 to be 1984` at the adapter test, plus `expected 1200 to be 1800` / `expected -1200 to be -1800` at the two new `filter-controls.test.ts` rows. Reverted, green again.
- `deriveIfShift` returning its first argument (`Math.round(pbtInner)`) → `3 failed`: `still reports the mean when only one edge is displaced`, plus two existing `radio-view-model-adapter` rows. So mean-vs-first-argument is caught; the two `it.each` rows pass equal edges and cannot distinguish it, which is expected — they pin the clamp, not the mean.
- Control: rewriting the body as `Math.round((pbtOuter + pbtInner) / 2)` — a legitimate no-op change — stays green (1855 passed). The check reddens on the bug and not on correct work.

CI at this head: `quick` pass (5m3s), `visual` pass. `quick-v2-observe` is red with "producer and publisher metadata bindings disagree" — an observation-layer binding, not a suite failure; coordinator's call.

**2. CLOSED on the substance; one clause is narrower than written.**
The docstring on `deriveIfShift` no longer claims both edges come from `measuredPbtRawToHz`. Checking its three assertions:

- *(b) CONFIRMED.* "for every shipped profile the range is +/-1200 and the outputs span -1200..+1191". Enumerated by `grep -rn -A6 '\[controls.pbt_inner\]|\[controls.pbt_outer\]' rigs/` over all of `rigs/*.toml` and `rigs/examples/*.toml` (11 files): `ic705`, `ic7300`, `ic7610`, `ic9700` declare `raw_center = 128`, `display_min = -1200`, `display_max = 1200`; `ftx1`, `tx500`, `x6100`, `x6200` and the three examples declare no PBT block and so fall through to `filter-controls.ts: PBT_DEFAULTS`, which is the same triple. `pbtRawToHz` is `Math.round((raw - rawCenter) * (displayMax / rawCenter))`, so raw 0 → −1200 and raw 255 → round(127 × 9.375) = 1191. Span confirmed arithmetically, not by sampling.
- *(c) CONFIRMED.* The wider declared range producing the larger shift is exactly what the adapter test now pins, and the mutation above shows it fails when the clamp returns.
- *(a) NARROWED.* "Every production caller today passes two `pbtRawToHz` outputs taken at ONE declared range" is true of four of the five production call sites (`panel-props.ts`, `panel-adapters.ts: buildIfShiftFeedback` ×3, `radio-view-model-adapter.ts`, `scope-passband-display.ts`) and false of the fifth: `filter-controls.ts: mapIfShiftToPbt` calls `deriveIfShift(currentPbtInner, currentPbtOuter)` on values that arrive from `panel-commands.ts` as `activeRx.pbtInner` / `activeRx.pbtOuter` — raw 0..255 state fields, not `pbtRawToHz` outputs (`panel-props.ts` applies `pbtRawToHz` to the same field, which is what fixes their unit). Not blocking: the mean of two raw 0..255 values is ≤ 255, so the removed clamp never fired there either, the conclusion the sentence supports is unaffected, and the very next paragraph names `mapIfShiftToPbt` explicitly, so no reader is sent past it. Recorded rather than made into a round-trip.

**3. NARROWED — the enumeration misses a site the blocking comment itself named.**
I re-ran the body's own method, `git grep -n "1200" -- src` in `frontend/`, dropping test, CSS, z-index, layout and window-width hits. It returns the four sites the table lists **and** `src/semantic/FilterSurface.svelte:45-47`:

```
['ifShift', 'IF shift', -1200, 1200, 25],
['pbtInner', 'PBT inner', -1200, 1200, 25],
['pbtOuter', 'PBT outer', -1200, 1200, 25],
```

These are neither CSS, layout nor demo, so the stated filter does not exclude them, and the previous BLOCKED comment named `FilterSurface.svelte` by name in its finding 3. "Not here" is honest for it — like the `FilterPanel` rows, the row bounds need the filter width threaded into the surface — but the table claims to be the class and is missing a member of it.

## PR-body corrections — REQUIRED BEFORE MERGE (editable in place; no commit needed, head SHA unaffected)

1. **The body still carries the two sentences the previous review refuted.** Section "Why removing the clamp leaves nothing unbounded" is unchanged from the first head: "Both edges come from `measuredPbtRawToHz` at one filter width…" and "the clamp never fired there and this is a no-op for them". The code comment was corrected; the body was not. `grep -rn 'measuredPbtRawToHz' frontend/src` outside tests still returns only its definition and comments — no production caller — and the "no-op" half is refuted by the adapter test in this very PR. Replace this section with the corrected account now in the docstring.
2. **Add `src/semantic/FilterSurface.svelte` (the `ifShift` / `pbtInner` / `pbtOuter` rows) to the class table**, with the same "needs the filter width at the site" reason.
3. **"the pin stayed on the sibling branch"** — the pin was not on a sibling branch. `filter-passband-adapter.isolated.test.ts` is present at the merge-base (94361667) and the first head simply did not touch it. Say that instead.

## What I verified rather than took on trust

- **The body's own correction is accurate.** `userContentEdits` shows the body at 12:22:55Z and 12:32:52Z read "Before this PR nothing pinned the clamp at all — 105 tests in `src/lib/radio/` passed with it removed"; the correction added at 12:40:05Z quotes it faithfully and its diagnosis matches the tree.
- **Both quoted counts reproduce.** `vitest run src/lib/radio/filter-controls.test.ts` → **56 passed (1 file)**; `vitest run src/lib/radio/` → **109 passed (6 files)**.
- **`Math.round` is not a behaviour change**: `clampToBipolarRange` already rounded half-up before clamping.
- The bench read-backs cannot be re-run from here; the lattice values the new tests assert are properties of `measuredPbtRawToHz` as it shipped in #3514 and hold on the code.

## Checks run

```
cd frontend && ./node_modules/.bin/vitest run src/lib/radio/ src/lib/runtime/adapters/
```
→ **62 files, 1855 passed** at the unmodified head; 3 failed under each of the two mutations above; 1855 passed again under the control change and after revert. Focused to the affected scope as dispatched; the full suite is CI's.

morozsm:
Agent Review: PASS 133fe3f1b66c51172fc6ab1ac2c5703efe650d4f

Delta re-review at the new head. Worktree detached at 133fe3f1, `git status --short` clean before and after every probe. The delta is one commit and it touches only the docstring on `deriveIfShift` — no executable line changed (`git diff 4f1775b5 133fe3f1 --stat`: 1 file, 19+/9-, all inside a block comment), so nothing the previous PASS established about behaviour can have moved. I re-ran the affected scope anyway: `npx vitest run src/lib/radio/ src/lib/runtime/adapters/` — **1855 passed, 62 files**, exactly the count the commit message claims.

## 1. The corrected comment, sentence by sentence

**The list of four display-side callers is complete and correct — CONFIRMED.** `git grep -n deriveIfShift -- 'frontend/src/**'` with tests removed returns exactly five call sites plus the definition and two prose mentions: `lib/runtime/props/panel-props.ts:424`, `lib/runtime/adapters/scope-passband-display.ts:202`, `lib/runtime/adapters/radio-view-model-adapter.ts:671`, `lib/runtime/adapters/panel-adapters.ts:1019/1023/1027` — and `filter-controls.ts: mapIfShiftToPbt`. Four display-side plus the one named exception; there is no sixth.

**`mapIfShiftToPbt` is fed raws — CONFIRMED.** `panel-commands.ts: onIfShiftChange` passes `activeRx.pbtInner` / `activeRx.pbtOuter` straight in, and those are wire fields: the same state fields go through `pbtRawToHz` at `scope-passband-display.ts` and `radio-view-model-adapter.ts`. The comment names the right caller.

**"the outputs span -1200..+1191" — CONFIRMED.** `pbtRawToHz` is affine with a final `Math.round`, so endpoints bound it: raw 0 → −1200, raw 255 → round(127 × 9.375) = 1191. Measured, not read: my probe printed `spanMin:-1200, spanMax:1191`.

**"The mean of two raws cannot reach 1200" — CONFIRMED.** `deriveIfShift(255, 255)` = 255. Raws are capped at 255 by `pbtHzToRaw`'s own `Math.min(255, …)`.

**"every shipped profile declares +/-1200 (four in their own `[controls]` block, seven through the same default)" — NARROWED, non-blocking.** The guarantee is true; the population is stated without its counting rule and resolves two ways.

- Four declare it: `git grep -ln '\[controls.pbt_inner\]' -- '*.toml'` → `ic705`, `ic7300`, `ic7610`, `ic9700`, each `raw_center = 128 / display_min = -1200 / display_max = 1200`. Exactly four; CONFIRMED.
- Seven fall through only if `rigs/examples/*.toml` counts as a shipped profile. The profiles the loader can actually load are eight: `profiles/rig_loader.py: discover_rigs` globs `*.toml` **non-recursively** and skips `_`-prefixed names, so `rigs/examples/ftx1.toml`, `sdrplay_rspdx.toml`, `xiegu_x6200.toml` are never loaded and can never produce a caps payload. On that set the split is 4 + 4, not 4 + 7.
- The reading that makes seven correct is "shipped in the wheel": `pyproject.toml [tool.hatch.build.targets.wheel.force-include] "rigs" = "rigplane/rigs"` copies the directory whole, `examples/` included — 11 non-underscore files, 4 + 7.

Not blocking, because the three extra files are a superset: they declare no PBT block either, so the bound the sentence exists to establish (no shipped caps payload declares a range wider than ±1200, hence the removed clamp never cut) holds under both readings, and I verified it under both. Worth one commit only if another correction is being made anyway; if so, the fix is to name the rule — e.g. "four of the eight profiles `discover_rigs` loads declare it in their own `[controls]` block; the other four, and the three `rigs/examples/` templates, fall through to `PBT_DEFAULTS`".

No sentence in the new block states more than can be checked beyond that one count.

## 2. The MOR-2500 numbers — CONFIRMED, reproduced independently

Own throwaway probe against the real exports (`mapIfShiftToPbt`, `pbtRawToHz`, `pbtHzToRaw`), written into the worktree, run, then deleted; tree confirmed clean afterwards. Output:

```
innerHz 300, outerHz -300, widthBefore 600
mapIfShiftToPbt(300, 160, 96) -> { pbtInner: 332, pbtOuter: 268 }
pbtHzToRaw -> 163 / 157   widthAfter 56
mapIfShiftToPbt(300, 300, -300) -> { pbtInner: 600, pbtOuter: 0 }   widthAfter 600
```

Every number in the commit message and in the PR body matches: 600 → 56 Hz through the raw path, raws 163/157, 600 Hz preserved through the Hz path. The measured claim reproduces from a clean checkout.

## 3. Out of scope — agreed, and for a stronger reason than the commit gives

The commit defers this to the writer half of step 2 because that file must convert to Hz anyway. That is right, and it understates the case. Passing Hz at `panel-commands.ts:1010-1014` is not a one-line fix in isolation: `mapIfShiftToPbt` applies `clampToBipolarRange` (±1200) to all three of its values, harmless while they are raws bounded by 255, but immediately live once they are Hz. Feeding it Hz today would newly truncate the IF shift at ±1200 — the exact fabricated bound this PR removed after measuring 1800 Hz on the radio. A correct fix must change the units **and** that clamp together, and the clamp is explicitly deferred so its symbol is not changed twice. Deferral confirmed correct.

## 4. The three body corrections

- **Both refuted sentences replaced — CONFIRMED.** The body no longer claims `measuredPbtRawToHz` feeds the edges; it states that no production caller uses it at this head, which I re-verified (`git grep measuredPbtRawToHz` outside tests returns its definition at `filter-controls.ts:185` and the docstring reference — nothing else).
- **`semantic/FilterSurface.svelte` in the class table — CONFIRMED, and the table is complete.** Own sweep: `git grep -n 1200 -- 'frontend/src/**'` minus tests and `21200`. Every production bound it returns is in the table — `FILTER_BIPOLAR_MIN/MAX`, `PBT_DEFAULTS`, `FilterPanel.svelte:62-63`, `FilterPanel.svelte:466/467/478/479`, `FilterSurface.svelte:45-47`. The rest are CSS `z-index`/breakpoints, `ControlButtonDemo.svelte` demo state, a 1200 ms debounce at `panel-commands.ts:1161`, and prose. No sixth site.
- **The merge-base account — CONFIRMED.** `git merge-base origin/main 133fe3f1` is `94361667`, and `git show 94361667:…/filter-passband-adapter.isolated.test.ts` contains `it('ifShift clamps to ±1200 …')` with `expect(view.filterPassband!.ifShift.reading).toEqual({ status: 'known', value: 1200 })`. The test was at the merge base and on `main`; "stayed on a sibling branch" was indeed wrong and is now stated correctly.

Body counts also reproduce exactly: `filter-controls.test.ts` **56 passed**; `src/lib/radio/` **109 passed across 6 files**.

## REQUIRED BEFORE MERGE (PR body edit; head SHA unaffected)

1. The class table now has **five** rows but the sentence under it still reads "All four need the same thing this PR could not give them." Stale neighbour from adding the `FilterSurface.svelte` row. Change "All four" to "All five".
2. "checked across all eleven `rigs/*.toml`" — the shell glob `rigs/*.toml` yields ten files, two of them `_`-prefixed non-profiles; the git pathspec `'rigs/*.toml'` recurses and yields thirteen. Eleven is the non-underscore count *including* `rigs/examples/`. State the rule, e.g. "all eleven non-underscore profile files under `rigs/`, `rigs/examples/` included — the eight `discover_rigs` loads plus three example templates".

## CI at this head

`quick` pass (4m56s), `visual` pass (1m22s). `quick-v2-observe` red with "pull request base is not the live main head" and the two v2-observation publishers red — observation-layer bindings, not suite failures, same shape the previous review reported; coordinator's call. Agent Review Gate red pending this directive.

morozsm:
Agent Review: PASS 5338ebea5c784c23be693f64dcf338563b112fce

Delta re-review of the single prose commit `133fe3f1..5338ebea` (`frontend/src/lib/radio/filter-controls.ts`, `deriveIfShift` doc comment). Nothing further is required; the observation from the previous cycle is answered.

**1. The profile guarantee — CONFIRMED, established from the repository, not the commit message.**

Skip rule read at `src/rigplane/profiles/rig_loader.py: discover_rigs`: `sorted(directory.glob("*.toml"))`, then `continue` on `path.name.startswith("_") or path.name.endswith(".draft.toml")`. Non-recursive glob, exactly as the comment states. `discover_available_rigs` reads package resources `rigplane/rigs`, which `pyproject.toml` `[tool.hatch.build.targets.wheel.force-include]` maps from the repo-root `rigs` directory — one corpus, not two.

`rigs/*.toml` returns ten paths; two are `_`-prefixed (`_ctcss_tables_v1.toml`, `_keyboard-default.toml`); no `.draft.toml` exists. Eight admitted, each checked for `[controls.pbt_inner]`:

| profile | `[controls.pbt_inner]` |
|---|---|
| ic705 | `raw_center = 128`, `display_min = -1200`, `display_max = 1200` |
| ic7300 | same |
| ic7610 | same |
| ic9700 | same |
| ftx1 | absent (no `pbt` token in the file at all) |
| tx500 | absent |
| x6100 | absent (only prose comments explaining the dropped cap) |
| x6200 | absent (same) |

No admitted profile declares a PBT range other than centre 128 / ∓1200. The fallback is `filter-controls.ts: PBT_DEFAULTS` = `{rawCenter: 128, displayMin: -1200, displayMax: 1200}`, reached by `pbtRange()` whenever `raw_center`/`display_min`/`display_max` are not all present. The `-1200..+1191` span is arithmetic from `pbtRawToHz`: `Math.round((raw - 128) * (1200/128))` gives `-1200` at raw 0 and `1190.625 → 1191` at raw 255. Recomputed here, not taken from the body.

**2. Falsifiability.** The sentence names the rule, the corpus and the property, so a reader can refute it with one `ls` and eight greps — which is what I did. A future profile with a wider declared range would falsify the enumeration but not the conclusion: the very next sentence already states that a wider caps range now yields the larger shift. The claim is not load-bearing beyond the historical "the clamp never cut them", which is what it says.

**3. Body edits — both CONFIRMED, and the new body text is accurate.** "All **five**" now matches the five-row class table. `git grep -n 1200 -- frontend/src` at this head finds no production `±1200` bound outside those five sites (remaining hits are CSS `z-index`, a `1200` ms timeout at `panel-commands.ts`, test fixtures, and comments); `semantic/FilterSurface.svelte:45-47` holds the three `-1200, 1200` rows the fifth entry claims. The restated profile claim — eight profiles, four declaring (ic705, ic7300, ic7610, ic9700), four falling through, glob returns ten of which the loader skips two — matches the enumeration above exactly. Also re-checked the unchanged clause it sits beside: the four display-side callers do pass Hz (`panel-props.ts` line 378 `pbtRawToHz(rx?.pbtInner ?? 128)` feeding line 424), and `deriveIfShift` has exactly those four external callers plus `mapIfShiftToPbt`.

**4. Focused run at this head**, scratch checkout detached at `5338ebea`, `node_modules` symlinked: `npx vitest run src/lib/radio/ src/lib/runtime/adapters/` — **62 files, 1855 tests passed**. No mutation was run for this delta and none is owed: the commit changes only comment text, which no test can observe. The mutation evidence for the behavioural part of this PR was checked in the earlier cycle at `133fe3f1`, and those files are byte-identical here.

Both trees `git status --short` clean.

Non-blocking, stylistic only, no action expected: the rewritten block wraps at ~76 columns except line 810, which runs to ~99. No `max-len`/`printWidth` rule exists in `frontend/eslint.config.js`, so no check reddens on it.

PR #3518, opened 2026-09-17
refactor(MOR-2497): read PBT off the measured lattice in the passband consumers

morozsm:
**Guardrail exception — file count: 15 code + 0 regenerated baselines = 15 files** (hard ceiling 10). Granted by the coordinating session under the owner's delegation of 2026-09-03, file count only, in three steps as the work showed what it needed: 12 for the prop threading to the renderer, 13 for a fixture that read PBT without a step, 15 for the blocked review's fix (`SpectrumPanel.component.test.ts`, `panel-props.test.ts`). The line ceiling is not waived and is not reached: 718 changed lines against the merge-base (`git diff --shortstat origin/main...HEAD` → 15 files, 597 insertions, 121 deletions).

**Why this is one unit (soft threshold crossed at 718 lines).** Three consumers turn raw PBT into Hz, and two of them are compared for equality: `scope-passband-display.ts` checks `strict.ifShiftHz === shiftHz` against the value derived through `radio-view-model-adapter.ts`. If they take the lattice step from different sources the passband display can never reach `current`. The third consumer, the audio-spectrum renderer, is a pure drawing function, so the step has to be threaded to it through `panel-props.ts` → `AudioSpectrumPanel.svelte` → `AudioSpectrumCanvas.svelte`; those three files are prop threading of one field. Splitting would ship a knowingly wrong 50 Hz step in AM in whichever consumer went second.

## What

Step 2 of MOR-2497, re-sequenced behind #3519 and #3520.

1. **The PBT lattice step comes from the rig profile, per mode**, at all three reading sites — `FilterModeConfig.pbtStepHz` resolved with the existing `resolveFilterModeConfig` — instead of the constant `PBT_MEASURED_STEP_HZ` this branch was first written against. 50 Hz in SSB/CW/RTTY, 200 Hz in AM (IC-7300 Advanced Manual p.40; owner ruling: AM is taken from the manual, not re-measured).
2. **A mode with no step has no reading.** No fallback to 50. FM is now refused because the mode has no twin PBT, not because a 15 kHz width happens to form no lattice — the wrong reason the first review of this PR caught.
3. **The semantic view model stops offering PBT in such a mode.** In `radio-view-model-adapter.ts` the PBT fields, and the derived IF-shift path that depends on them, become non-structural when the current mode has no twin PBT. This is what closes the owner's FM symptom on the default skin `desktop-v2`, which #3520 could not reach: that skin does not show `FilterPanel` and reads the view model instead.
4. **A mode without twin PBT has a known shift of 0, not an unknown one.** In `scope-passband-display.ts: inspect`, FM no longer invalidates the passband display through a missing shift: the tuple carries 0, and the strict comparison there pins `strict.ifShiftHz === null`, which is what the view model reports once its PBT fields are non-structural. Modes with twin PBT keep the exact `strict.ifShiftHz === shiftHz` equality. A legacy payload with no step anywhere still fails closed — PBT may really be engaged there.
5. **Two absences stay apart**, by reusing the one definition from #3520 — `modeHasTwinPbt` is now exported from `panel-props.ts` and imported, not copied. Some mode declares a step and this one does not → no twin PBT here. No mode declares one → a payload from before the field existed: structure stays radio-wide as before, and there is still no Hz reading because there is no step to read it with.

`PBT_MEASURED_STEP_HZ` stays exported: no production code uses it any more, but `filter-controls.test.ts` and others use it as the bench-measured SSB value.

## Review history

First review, at `0f3aeaca`: BLOCKED with three findings. Two were fixed in `d3e9cedd`; the third grew into #3519 (the per-mode step contract) and this PR went to draft behind it.

Second review, at `a9d28882`: BLOCKED — `quick` red, three `SpectrumPanel.component.test.ts` cases, because a missing step invalidated the whole passband display rather than a PBT number. Fixed in `1e2ab8a9`…`f0cc7771`: the known-zero shift above; the three tests now drive the production selector with the captured IC-7300 capabilities plus the per-mode steps the server publishes today (composed in a local helper — the captured JSON is byte-faithful and was not edited, and #3520's step-less legacy test is untouched); a pin on `toAudioSpectrumProps` emitting 50 / 200 / undefined, which two surviving mutants had shown was unpinned; a false fixture comment deleted. `main` (with #3521, which edits the same file on other lines) merged in cleanly afterwards.

## Two things found on the way

- `semantic-pbt-continuity.component.test.ts` was red on this branch since `d3e9cedd`, before this work. Fixed in `d532f80a` by keeping the outer PBT edge constant across the transient-loss accept, not by weakening the assertion.
- `freshness-invisibility.component.test.ts` went red, correctly: its fixture is legacy-shaped (no `filterConfig`) and relied on PBT sliders having Hz readings. Its fixture now publishes the USB step. No assertion was changed. That is the 13th file. Two workers committed that same line a minute apart into one tree — a coordination slip of mine — so the history carries two same-subject commits and a third, `a9d28882`, removing the duplicated key.

## Checks

At this head, from `frontend/`, single files only — directory runs are left to `quick`:
`scope-passband-display.test.ts` → 210 passed · `SpectrumPanel.component.test.ts` → 157 passed · `npx tsc --noEmit` → exit 0 (coordinator, after merging `main`).
Before that merge, by the builder: `filter-passband-adapter.isolated.test.ts` 58 · `scope-adapter.authority.isolated.test.ts` 69 · `panel-props.test.ts` 99 · `semantic-pbt-continuity.component.test.ts` 9 — all passed.

Mutations measured by the second reviewer at `a9d28882` over `src/lib/radio/ src/lib/runtime/ src/components-v2/panels/audio-scope/ src/components-v2/wiring/` (baseline 3320 passed): constant 50 at site 1 → 4 fail, at site 2 → 3, at site 3 → 2; collapsing the two absences → 6; Task B made a no-op → 2; resolving the other receiver's mode → 25; an equivalent rewrite → 0. Added for the fix, single-file: FM shift back to undefined → the FM pin fails; the zero extended to the legacy payload → the legacy pin fails; the `pbtStepHz` emission deleted → the two emission pins fail.

## Not covered

- The PBT **Reset** button in `FilterPanel.svelte` still sits outside `{#if hasPbt}` and stays live in FM. It goes with the writers.
- The write path (`panel-commands.ts`, `mapIfShiftToPbt`, MOR-2500) is untouched.
- The threading `panel-props.ts` → panel → canvas: the emission end is now pinned — `panel-props.test.ts` asserts `toAudioSpectrumProps` emits `pbtStepHz` 50 in USB, 200 in AM and none in FM, against a fixture whose `stepHz` differs, so reading the wrong field fails too (the reviewer's two formerly surviving mutants now fail 2 tests each). The two Svelte hops in between are covered by `tsc` and the renderer pins only.
- `panel-props.ts: toFilterProps` still converts with the old `pbtRawToHz`, so the v2 / LCD skins show a different Hz from the semantic skin — the reviewer measured raw 200 → 675 Hz there in every mode, against 700 Hz (USB/2400) and 2500 Hz (AM/9000) semantically. It goes with the writers.
- In FM the scope shows no passband rectangle on `main` either, for an unrelated reason measured live today: the radio reports no filter width in FM (`main.filterWidth: null`). That is MOR-2503; the known-zero shift here is what lets the rectangle appear once a width exists.
- PSK / PSK-R on the IC-7610 have no filter-table entry, so they publish no step. The consequence after this PR is sharper than "no PBT shown": `modeHasTwinPbt` is false there, so the PBT controls are withheld **and the scope passband is drawn at a confidently-zero shift** — while the radio does have twin PBT in PSK on a 50 Hz step (measured today). A mode with twin PBT but no profile table is indistinguishable from a mode without it; that is a profile gap, and MOR-2504 (in progress) is what makes it right. The comment in `scope-passband-display.ts` calling this "the KNOWN zero of a passband that cannot be displaced" is wider than the code for exactly this case; narrowing it is noted on MOR-2502, which edits the same file.
- No live run on a radio at this head yet. The expected live result — PBT controls gone in FM on `desktop-v2`, readings unchanged in USB, and AM readings on a 200 Hz lattice — is not yet observed.

Merging this auto-closes MOR-2497 through the branch name; it is not finished and must be reopened.


morozsm:
Agent Review: BLOCKED 0f3aeaca99c8d448af14209c10e879e0cd3996d5

Independent review at the frozen head. Baseline reproduced exactly: `npx vitest run src/lib/radio/ src/lib/runtime/ src/components-v2/panels/audio-scope/ src/components-v2/wiring/` -> **3305 passed, 132 files**, matching the body.

Mutations run against `measuredPbtRawToHz` (tree restored and `git status --short` clean after each):
- (a) off by one lattice position -> **19 failed / 6 files**, including all four touched suites.
- (b) ignore the width argument, fix the lattice at 2400 -> **17 failed / 3 files**, including `scope-passband-display.test.ts > refuses a PBT reading at a width off the 50 Hz lattice` and `filter-passband-adapter > the declared range no longer sets the scale, and the filter width does`.
- (c) fail-closed returns 0 instead of null -> **9 failed / 2 files**.
- control (algebraically equivalent rewrite of the same return) -> **3305 passed**, so the suite does not redden on correct work.

The coupling claim is CONFIRMED mechanically, not taken on reading: `scope-passband-display.ts: inspect` builds `strict` from `toSpectrumAuthority(state, caps)`, which reads `filterPassband.ifShift` off the converted adapter, and compares `strict.ifShiftHz === shiftHz` (plus `strict.filterWidthHz === width`). Converting one alone does pin the display at non-`current`.

The counting claims are CONFIRMED. Rule used: occurrences of `pbtRawToHz`/`pbtHzToRaw` under `frontend/src` at the merge-base `3a9d2837`, excluding comments and import statements. 15 invocations on 14 lines in 6 files -- `audio-spectrum-renderer.ts` 3/3, `panel-adapters.ts` 1/1, `radio-view-model-adapter.ts` 2/2, `scope-passband-display.ts` 2/1, `panel-commands.ts` 5/5, `panel-props.ts` 2/2. Seven converted here. The stated transposition (panel-adapters one, scope-passband-display two) is correct.

---

## Blocking (in the diff; only a commit fixes these)

### 1. The refusal enumeration is false for the width that actually occurs on hardware

`frontend/src/lib/runtime/adapters/scope-passband-display.ts:207`

> `// (absent, zero, not a multiple of the step) yields `null` here and the`

That set is incomplete, and the missing member is the one that fires on every shipped rig. `filter-controls.ts: pbtLattice` also returns `null` at `frontend/src/lib/radio/filter-controls.ts:154` (`if (positions - 1 > 255) return null;`). A width of 15000 Hz **is** a whole number of 50 Hz steps -- 300 of them -- and is refused by that guard, not by any of the three causes the comment names.

15000 Hz is not hypothetical. All four Icom profiles in the repo declare it as the FM FIL1 default:

- `rigs/ic7300.toml` `[filters.width.FM] fixed = true / defaults = [15000, 10000, 7000]`
- `rigs/ic7610.toml:980-982`, `rigs/ic705.toml:375-377`, `rigs/ic9700.toml:385-387` -- identical.

Reproduced: setting `bareState`'s `filterWidth` from 2400 to 15000 in `filter-passband-adapter.isolated.test.ts` turns three previously-`known` readings into `unknown`:

```
AssertionError: expected { status: 'unknown' } to deeply equal { status: 'known', value: 700 }
  > pbtInner fresh, pbtOuter stale-but-observed -- ifShift derives from its last value
  > both pbtInner and pbtOuter observed -- ifShift derives and reports known
  > pbtOuter ABSENT from the receiver object ...
```

So in FM the semantic surface loses its PBT and derived IF-shift readings on IC-7300 / IC-7610 / IC-705 / IC-9700. **Risk:** this is a behaviour change on live hardware that the diff's own prose describes as impossible, the PR body describes as impossible ("forms no lattice -- not a whole number of 50 Hz steps"), and no test covers.

I could not establish whether the scope RF passband overlay also disappears in FM: my probe at width 15000 through `scope-passband-display.test.ts` returned non-`current` under the old conversion too, because that fixture is a USB row whose declared segments top out at 3600 Hz. The FM path needs its own fixture, which a read-only review cannot add.

**Required fix:** (i) correct or delete the parenthetical at `scope-passband-display.ts:207` -- the honest form names the positions-per-raw guard, or names nothing; (ii) decide and state what FM does. If refusing is intended, pin it with a test at width 15000 and say so in the body; if not, the guard or the model needs to cover it. (iii) Check `rigs/ic705.toml`'s `[filters.width.WFM]` block for the same shape -- I did not read it.

**Checks:** `npx vitest run src/lib/runtime/adapters/ src/components-v2/panels/audio-scope/` after adding the 15000 Hz pin.

### 2. The continuity fixture collapses the two values the test is named for

`frontend/src/components-v2/wiring/__tests__/semantic-pbt-continuity.component.test.ts:59`, `:84`, `:96`

The remap replaces `state(100, -200)` with `state(100, 100)` -- inner and outer become the *same raw* -- in `it('retains independent fresh values through transient loss, ...')`. The comment added at `:38-47` says the remap kept "each value on the side of centre it was on, which is what the assertions below are about". It did not: what the assertion is about is that the two values are independent, and they are now indistinguishable.

Reproduced both directions, at this head, without touching anything else:

- cross-wire `radio-view-model-adapter.ts: deriveFilterPassband` so `pbtInnerHz` reads `pbtOuterRaw` -> `describe('mounted PBT continuity is fenced by the live control session (MOR-1706)')` stays **fully green (4 passed)**. (Four tests in the file's *other* describe do catch it, which is why the whole-file run reddens -- the weakened describe does not.)
- same cross-wire, plus the fixture changed to `state(100, 156)` -> **1 failed**: `retains independent fresh values through transient loss, then synchronously clears coalesced same-epoch reconnects`.

**Risk:** an inner/outer swap in the adapter is now invisible to the test written to catch it. **Required fix:** use two distinct reachable raws on opposite sides of centre (156 works and is on the lattice at 2400 Hz), and drop the sentence that says the remap preserved the assertion's subject.

### 3. A tautological assertion with a comment claiming it discriminates

`frontend/src/lib/runtime/adapters/__tests__/filter-passband-adapter.isolated.test.ts:331-339`

```ts
const atDefaultWidth = measuredPbtRawToHz(50, 2400, PBT_MEASURED_STEP_HZ);
...
expect(measuredPbtRawToHz(50, 2400, PBT_MEASURED_STEP_HZ)).toBe(atDefaultWidth);
```

Identical call compared against itself: `x === x`. No mutation of `measuredPbtRawToHz` can redden it, because both sides move together. The comment at `:329` states the opposite -- "a regression to a range-driven scale fails the first half" -- which cannot be true: `measuredPbtRawToHz` takes no range argument, so there is nothing for a range-driven regression to change between the two calls. Only the second half (`1800` vs `2400`) discriminates, and that is the half mutation (b) killed.

Secondary: the test now calls no adapter function at all, while living inside `describe('pbtInner/pbtOuter/ifShift parity with the real filter-controls helpers')`. It is a `filter-controls` unit test in an adapter-parity suite.

**Required fix:** delete the tautological assertion and the sentence claiming it fails on a range-driven regression; if the range-independence of the *adapter* is worth pinning, assert it through `toRadioViewModel` with two different declared `pbt_inner` ranges at one width, which does discriminate.

---

## REQUIRED BEFORE MERGE (PR body -- edit in place, no commit needed, head stays valid)

**B1. "Literals rather than a recomputation through `measuredPbtRawToHz`, so a wrong model cannot satisfy the test by being wrong in both places" is NARROWED to false for the largest rewritten file.** `filter-passband-adapter.isolated.test.ts:253-257` adds `const hz = (raw, widthHz = 2400) => measuredPbtRawToHz(...)` and uses it in six assertions; the parity row computes `expectedInnerHz`/`expectedOuterHz` through the same function. Measured under mutation (a): that file loses **1** test (the one written with the literal `1800`), while `scope-passband-display.test.ts`, which does use literals, loses **4**. The sentence is true of `scope-passband-display.test.ts` and of the renderer suites; it is not true of the file where most of the rewriting happened.

**B2. "Where the width is absent, or forms no lattice -- not a whole number of 50 Hz steps -- there is no honest Hz."** Same defect as finding 1: 15000 Hz is a whole number of 50 Hz steps and still yields no Hz.

**B3. The "Not here" section names only the write side; two READ-side producers remain on the retired conversion and are not named.**

- `panel-props.ts: toFilterProps` (2 invocations, still `pbtRawToHz(rx?.pbtInner ?? 128)` with the store fallback). It is live: `MobileRadioLayout.svelte:78`, `panels/FilterPanel.svelte` (via `deriveFilterProps`), `panels/lcd/AmberScope.svelte:62`, `panels/lcd/AmberCockpit.svelte:74`. At raw 200 / width 2400 the v2 and LCD skins now show **675 Hz** where the semantic skin shows **700 Hz** (at raw 140: 113 Hz vs 100 Hz). Different skins, so not on screen simultaneously -- but the body should say the divergence exists and when it closes.
- `panel-adapters.ts: getIfShiftControlFeedback` (1 invocation). Its docstring at `panel-adapters.ts:989` claims it is "not a second, independent re-derivation of the confirmed reading"; after this PR it is exactly that, computed with a different conversion. It has **no production caller** at this head (`grep` finds only its own definition and a prose mention in `stores/commands.svelte.ts:449`), so nothing user-visible diverges -- which is why this is a body/follow-up item and not a blocker.

---

## Tried to refute and could not

- **The renderer's fail-closed path does not fabricate or leak NaN.** `audio-spectrum-renderer.ts` falls back to `0` for geometry only, which is byte-for-byte the behaviour an absent `pbtRange` already produced; a never-reported `filterWidth` arrives as `NaN`, `pbtLattice`'s `Number.isFinite` check rejects it, and nothing reaches a label. The adapter's `undefined` becomes `{status:'unknown'}`, which `FilterSurface.svelte:284-288` renders as the unknown span with the slider withheld -- honest degradation, no fabricated 0, no NaN in a readout.
- **Averaging in Hz rather than in raw** (`audio-spectrum-renderer.ts:172-181`) is correct and its comment survives: `deriveIfShift` is `Math.round((a + b) / 2)`, and on lattice values the rounding is a no-op, so the renderer's unrounded mean does match.
- **The width sources agree.** `scope-passband-display.ts` reads `read('...filterWidth', rx.filterWidth)` and the adapter reads `numOrUndef(rx?.filterWidth)`; the same `strict` comparison at `:264` already required those to be equal before this PR, so the two conversions see one width.
- **The read/write round trip was already broken and is not made worse.** `panel-commands.ts` still writes through `pbtHzToRaw`, but the radio snapped those writes to the same lattice before this PR too; the reading is now correct where it was previously wrong. No regression, and the body discloses the write side as the next PR.
- **No shipped rig loses PBT for want of a `filter_width` capability**: all four profiles declaring `pbt` also declare `filter_width`.

morozsm:
Converted to draft, and re-sequenced behind a change that did not exist when it was opened. Two of the three BLOCKED findings are fixed in `d3e9cedd`; the third turned out to be the tip of something larger.

## The third finding, and what it led to

The review found that a 15000 Hz filter width forms no lattice, so PBT and IF-shift readings vanish — and that all four Icom profiles declare 15000 Hz as FM FIL1. That is correct and I could not have argued it away.

Going to the manual rather than the bench (the owner is operating the radio) settled it, and settled more than was asked. IC-7300 Advanced Manual, p. 40, *Using the Twin PBT*:

> **SSB, CW, RTTY and AM modes**
>
> The PBT is adjustable in 50 Hz steps in the SSB, CW, and RTTY modes, and **200 Hz in the AM mode**. In this case, the center shift value changes in 25 Hz steps in the SSB, CW, and RTTY modes, and 100 Hz in the AM mode.

Three consequences:

1. **FM has no twin PBT at all.** The missing reading is therefore right, and the retired conversion — which fabricated Hz from a declared ±1200 in FM — was wrong. But this branch reaches that right answer for the wrong reason: it refuses because the lattice cannot be built, not because the mode has no PBT.
2. **The 50 Hz step is confirmed independently of our bench sweep**, and "the center shift changes in 25 Hz steps" is exactly what a mean of two edges on a 50 Hz lattice produces — the model already gives that.
3. **The step is 200 Hz in AM, not 50.** The bench measured SSB only. This branch hardcodes `PBT_MEASURED_STEP_HZ = 50`, so in AM it would ship a knowingly wrong conversion: at a 9000 Hz AM width a 50 Hz step computes 181 positions over 256 raw values, which collide under rounding.

## Why that re-sequences this PR

The owner's ruling: the step comes from the profile, and AM is taken from the manual rather than re-measured. So the per-mode step becomes profile data before any consumer reads it, which inverts the order this PR assumed.

The plumbing already exists and needs one field, not a new mechanism: `[filters.width.<MODE>]` in the rig TOML → `FilterWidthRule` → the `filterConfig` payload → `FilterModeConfig` in the browser, which the frontend already consumes through `resolveFilterModeConfig`.

## What happens to this branch

It stays as the draft record of the consumer work and its review. When the contract and the profile data have landed, it returns with `PBT_MEASURED_STEP_HZ` replaced by the step resolved from the radio's own profile for the current mode, and with the FM case refused for the stated reason rather than by accident. The two fixed findings and their mutation evidence carry over unchanged.

morozsm:
Agent Review: BLOCKED a9d2888270c1dd02fac18e15eb2eea86169b77d6

Independent review at the frozen head by a verifier that did not author the change. Body counts confirmed: 13 files, 531 insertions, 117 deletions, 648 changed lines against `b83e7c81`.

## Blocking

**1. The required `quick` check is red at this head — three real regressions outside the scopes the PR's own checks ran.** Run `35254986236`: 3 failed / 12205 passed, all in `frontend/src/components/spectrum/__tests__/SpectrumPanel.component.test.ts`. Reproduced both ways in separate worktrees: `npx vitest run src/components/spectrum/__tests__/SpectrumPanel.component.test.ts` → 3 failed / 154 passed at `a9d28882`, 157 passed at `b83e7c81`.

Cause: `scope-passband-display.ts: inspect`. With no `pbtStepHz` for the current mode, `toHz` returns `null`, `shiftHz` is undefined, and the `typeof shiftHz !== 'number'` guard turns the **whole passband display** into `invalid-observation`. The RF passband overlay and its resize handle disappear — not only a PBT number. The three tests drive the production selector with the captured pre-#3519 fixture (no step in any mode) in AM at 10 000 Hz. So the body's "structure stays radio-wide as before" is false for the scope overlay.

**2. Added by the coordinator after reading that code — the same path fires on a current payload in FM.** FM has no step by definition, so after this PR a real radio in FM would show no passband rectangle on the scope at all, and the comment above that block states this as the intent. It is the wrong intent: a mode without twin PBT has a PBT-derived shift of exactly 0, which is a known value, not a missing one. The reviewer verified that FM closes correctly on `FilterSurface`; nobody had looked at the scope overlay in FM. Not yet confirmed by a test or a live run — it follows from the code path in finding 1.

**Required fix (in progress):** mode without twin PBT → `shiftHz = 0`, display valid, PBT controls still absent; mode with a step → convert as now; legacy payload with no step anywhere → keep failing closed, and give the three tests a payload matching what the server serializes today, since the frontend ships in the same package as the server and that payload cannot occur in production. Both sides of the `strict.ifShiftHz === shiftHz` comparison must agree on the 0.

## Confirmed by the review

- FM closes on the default skin: `FilterSurface` over `toRadioViewModel` in FM/15000 has `pbtInner.availability.structural=false`, both PBT rows, the Reset button and the IF-shift row absent; USB/2400 and AM/9000 render live values.
- Regression sweep, 4 `pbt`-declaring profiles × every mode × dataMode 0/1 = 88 rows through the real resolvers: only FM (all four), WFM and DV (IC-705), PSK / PSK-R (IC-7610) lose structure — the accepted set. No non-Icom profile declares `pbt`.
- AM is really 200: `measuredPbtRawToHz(100, 9000, 200) = -900` against `-950` at 50; lattices form at 9000 / 6000 / 3000.
- The two coupled sites resolve from the same active receiver; mutation (b) below kills the lockstep pin.
- `PBT_MEASURED_STEP_HZ` has no production use; `modeHasTwinPbt` has one definition; the body's check counts (132 files / 3320, 41 / 886, `tsc` 0) reproduce verbatim. `semantic-pbt-continuity` was genuinely red at `d3e9cedd` and was fixed at its input, no assertion touched; the freshness fixture change is 5 truthful lines.

## Mutations (body's own scope, baseline 3320 passed, each restored)

| mutation | body said | measured |
|---|---|---|
| constant 50 at site 1 | 3 | 4 |
| constant 50 at site 2 | 3 | 3 |
| constant 50 at site 3 | 2 | 2 |
| collapse the two absences | 2 | 6 |
| Task B made a no-op | — | 2 |
| resolve the other receiver's mode | — | 25 |
| cross-wire inner / outer | — | 16 |
| equivalent rewrite (control) | — | 0 |
| `toAudioSpectrumProps` reads `stepHz` instead of `pbtStepHz` | — | **0 — survives** |
| delete that emission entirely | — | **0 — survives** |

The last two are a coverage gap, disclosed in the body: the one line that carries the step to reading site 3 has no pin.

## Required before merge, body only

Correct the mutation counts to 4 / 3 / 2 and 6, naming the scope. Add to "Not covered" that `panel-props.ts: toFilterProps` still uses `pbtRawToHz`, so the v2 / LCD skins show a different Hz from the semantic skin — at raw 200: 675 Hz there in every mode and width, against 700 Hz (USB/2400) and 2500 Hz (AM/9000) semantically.

## Prose (does not block)

The comment in `semantic-pbt-continuity.component.test.ts` claiming the pair sits "on opposite sides of centre 128" is false — `state(100, 40)` has both below 128 — and cross-wiring inner/outer fails nothing in that describe on `main` either, so it should be deleted rather than narrowed. `scope-passband-display.ts` still lists the lattice-refusal causes incompletely.

## Not established

No live-radio run. Whether FTX-1's `CW-U` / `DATA-L` resolve any width table — outside this PR.

morozsm:
**Coordinator correction to finding 2 of the BLOCKED review above — it does not hold as a regression of this PR.** I flagged it there as not yet confirmed; I have now run it live and it is wrong in its consequence.

Live, 2026-09-17 14:44 EDT: real IC-7610 over LAN, server and production frontend build from `main` (`b83e7c81` build, server at `1d5517a0`), default skin, 1440×900, MAIN 14.137.200, RX only. Mode switched USB → FM → USB through the page's own mode buttons; the radio was returned to USB FIL1 3600 Hz, PBT 128/128, and nothing was transmitted.

| | USB on `main` | FM on `main` |
|---|---|---|
| `main.filterWidth` from `/api/v1/state` | 3600 | **`null`**, still `null` 11 s after the change |
| `.passband-overlay` on the scope | present | **absent** |
| `.tune-line` | present | present |
| `filter-pbtInner`, `filter-pbtOuter`, `filter-pbt-reset` on the filter surface | visible | **visible and enabled** |

So on `main` the scope already shows no passband rectangle in FM, for a reason that has nothing to do with PBT: the radio reports no filter width in FM, and `inspect`'s `!positive(width)` guard refuses the display before the shift is ever looked at. `rigs/ic7610.toml` declares `[filters.width.FM] fixed = true, defaults = [15000, 10000, 7000]`, but nothing turns that into a width for the display. This PR therefore cannot make FM worse on a real radio, and "after this PR a real radio in FM would show no passband rectangle" described `main` as much as this head.

What stands:
- Finding 1 stands unchanged — `quick` is red at `a9d28882`, three `SpectrumPanel` tests, and that needs the commit now in progress.
- The ruling given to the builder stands as the correct semantics: a mode without twin PBT has a PBT-derived shift of exactly 0, not an unknown one, so the shift must not be what invalidates the display there. Its visible effect waits on a width in FM, which is a separate gap and outside this PR.
- The last row of the table is the owner's FM symptom, reproduced live on `main`: the PBT controls are live in FM on the default skin. That is what this PR closes. Owner, same day, on the acceptance criterion: the controls may stay visible as long as they are unavailable, if that is easier to build. This PR removes them, which also satisfies it.

Also seen, not this PR's: after switching back to USB the overlay took between 6 and 10 s to return — the slow recapture after a legitimate retirement that MOR-2440 tracks.

morozsm:
**Live run at this head, `442d28b2` — the owner's FM symptom is closed on the default skin.**

2026-09-17 15:15 EDT, real IC-7610 over LAN, production frontend build of `442d28b2` served by a server from `main`, default skin `desktop-v2`, 1440×900, MAIN 14.137.200, RX only. Modes switched through the page's own mode buttons; the semantic filter surface read by `data-testid`. `/api/v1/capabilities` published `pbtStepHz` 50 for USB, LSB, USB-D, LSB-D, CW, CW-R, RTTY, RTTY-R, 200 for AM, and none for FM.

| mode (radio state) | `filter-pbtInner` / `filter-pbtOuter` | `filter-pbt-reset` | scope passband |
|---|---|---|---|
| USB, FIL1 3600 | present, enabled, read 0 | present | present |
| **FM**, FIL1, width `null` | **absent** | **absent** | absent |
| AM, FIL2 6000 — entered from USB, at 8 s, 18 s and 28 s | present, enabled, read 0 | present | present |
| USB again | present, enabled, read 0 | present | present |

On `main`, measured the same way an hour earlier, all three PBT controls were visible and enabled in FM. They are now gone there, including the Reset button on this surface — so the "Reset stays live in FM" item in the body applies to `FilterPanel.svelte` only, not to the semantic surface.

Two things seen that are not this PR's:
- The scope shows no passband in FM because the radio reports no filter width there — the same on `main` (MOR-2503).
- Entering **AM directly from FM**, the PBT controls were present but **disabled** and the passband absent at 8 s; entering AM from USB they were enabled and the passband present at 8 s. So the recovery after a mode that has no PBT reading is slower than after one that has. Whether `main` does the same was not measured, so this is an observation, not a regression claim; it is the slow recapture MOR-2440 tracks.

Not done live: moving PBT in AM to watch the 200 Hz lattice reading — a second LAN connection would have contended with the server's. That conversion rests on the unit pins and on the reviewer's arithmetic check.

The radio was returned to USB, FIL1 3600 Hz, PBT 128/128, 14 137 200 Hz and read back; `lastActuation` is `null` — nothing was transmitted.

morozsm:
Agent Review: PASS 442d28b26e592e1cd23fcb6db1412883e5e27624

Delta re-review of `a9d28882..442d28b2` by the verifier that BLOCKED `a9d28882`. Own throwaway worktree, single test files plus `tsc`; the wide run read from CI.

**The blocker is cleared, measured two ways:** `SpectrumPanel.component.test.ts` → 157 passed at this head (3 failed / 154 at `a9d28882`), and `quick` passes here (run `35263012062`), `visual` passes.

**The mode-aware comparison is exactly as narrow as claimed.** `pbtShiftZero = !native && pbtStep === undefined && !modeHasTwinPbt(caps, modeConfig)` is false whenever the mode declares a step, false for the legacy payload, false on a native IF-shift radio. No input found where a mode with twin PBT reaches the `null` branch or where the zero fires for the legacy payload. The `null` on the authority side is real: with the PBT fields non-structural, `knownSemanticReading` yields `ifShiftHz: null`. The FM pin asserts `display.state === 'current'` and the view model through the real `toRadioViewModel`.

| mutation — single file, restored each time | result |
|---|---|
| remove the zero branch (FM back to `undefined`) | 1 failed — the FM pin |
| extend the zero to the legacy payload | 1 failed — the legacy pin |
| strict always `=== null` | 155 failed |
| strict always `=== shiftHz` (pre-fix) | 1 failed — the FM pin |
| drop the `!native` guard | 1 failed — the native AM row |
| control: reorder the same conjuncts | 210 passed |
| `toAudioSpectrumProps` reads `stepHz` | 2 failed in `panel-props.test.ts` |
| delete that emission | 2 failed in `panel-props.test.ts` |

The two mutants that survived the first review now die; the new pin uses USB `stepHz: 100` against `pbtStepHz: 50`, so it catches the wrong field, not only absence.

**The #3521 interaction composes.** Three throwaway cases added to the suite and removed again (213 passed): FM reaches `current` with `shiftHz 0` and translates across the band edge as stale + translated; USB→FM retires; FM→USB retires. Mode is still a fence.

**`IC7300_STEPPED_CAPABILITIES` is correct mode by mode** — it reproduces `rigs/ic7300.toml` (USB/LSB/CW/RTTY 50, AM 200, FM none) and what `_serialize_filter_config` emits. The captured JSON is unedited. No assertion weakened; #3520's step-less legacy test is untouched.

**Body:** all seven check numbers reproduced exactly — 210 · 157 · 99 · 58 · 69 · 9 · `tsc` 0; 15 files / 597+ / 121− / 718 lines; the guardrail line is in the required form; the mutation paragraph quotes the first review's numbers correctly. Both required body edits — the stale "`panel-props.test.ts` is not among the touched files" bullet, and the PSK bullet that understated the consequence — are applied.

**Prose, does not block.** P1: the comment "the KNOWN zero of a passband that cannot be displaced" is wider than the code — `modeHasTwinPbt` is also false for a mode that has twin PBT but no profile table (IC-7610 PSK), where the display is `current` with shift 0 while PBT may be engaged. The code follows the coordinator's ruling; the instance is MOR-2504. P2: the incomplete refusal list is gone. P3: the false "opposite sides of centre" clause is deleted; the MOR-1706 describe still does not discriminate an inner/outer swap, identical to `main`.

**Not established by this review:** the two live measurements the body cites (MOR-2503, MOR-2504), and the coordinator's live run at this head, which is posted separately above.

PR #3519, opened 2026-09-17
feat(MOR-2497): publish the per-mode twin-PBT step from the rig profile

morozsm:
10 files, 237 insertions, no deletions — above the 6-file soft threshold, so: it is one unit of work because a contract field that no profile declares is unreadable, and profile data that no contract carries is unreachable. Splitting them would ship half a path.

## Why the step cannot be a constant

MOR-2497 step 1 landed the measured PBT lattice with the step as a module constant, `PBT_MEASURED_STEP_HZ = 50`, taken from a bench sweep in SSB. The IC-7300 Advanced Manual, p.40, *Using the Twin PBT*, says otherwise:

> **SSB, CW, RTTY and AM modes**
>
> The PBT is adjustable in 50 Hz steps in the SSB, CW, and RTTY modes, and **200 Hz in the AM mode**. In this case, the center shift value changes in 25 Hz steps in the SSB, CW, and RTTY modes, and 100 Hz in the AM mode.

Two consequences a constant cannot express:

- **AM is 200 Hz.** A 50 Hz constant puts every AM reading on a lattice four times finer than the radio's, so the conversion returns numbers and they are wrong. **Corrected after review:** an earlier version of this body said the 181 positions a 50 Hz step computes at a 9000 Hz AM width "collide under rounding". They do not — the collision count is 0 and the minimum gap is 1 raw unit, and the `positions - 1 > 255` guard makes collisions impossible by construction. The harm is the wrong step, not a malformed lattice.
- **FM is not in that list at all.** It has no twin PBT, so there is nothing to declare — not a default to apply. This also answers, for the right reason, a finding on #3518: readings vanish at FM widths and that turns out to be correct behaviour rather than the regression it looked like, since the retired conversion fabricated Hz from a declared ±1200 in FM. **Corrected after review:** an earlier version said the lattice "cannot be formed" at the FM widths. That is true only of 15000 Hz; 10000 and 7000 form lattices of 201 and 141 positions. The reason FM has no reading is the mode, not the arithmetic — which is exactly why the step belongs to the mode.

The manual also confirms the 50 Hz step independently of our own sweep, and "the center shift changes in 25 Hz steps" is exactly what a mean of two edges on a 50 Hz lattice produces — the model already gives that without being told.

## What this adds

One field through plumbing that already exists, not a new mechanism:

`[filters.width.<MODE>].pbt_step_hz` → `FilterWidthRule.pbt_step_hz` → `filterConfig[MODE].pbtStepHz` → `FilterModeConfig` in the browser, which consumers already reach via `resolveFilterModeConfig`.

Declared for **all four PBT-capable profiles**. An earlier head declared only ic7610 and ic7300 and deferred the rest — review showed that made the absence of a step ambiguous between "this mode has no twin PBT" and "this profile is not migrated yet", and that a consumer obeying the documented reading would have taken PBT away from the ic705 and ic9700 with nothing to redden. The split created the ambiguity, so the split is gone.

ic7610 and ic7300 carry the measured provenance. ic705 and ic9700 say in their own note that their values are inference from two agreeing bench radios, that neither rig was measured, and that nothing has confirmed its AM step. The `-D` and `-R` variants are inference in every profile — the manual names the base modes, not each variant — and all four notes say so.

## Two pins, and why the second one exists

**Absence is a signal.** FM declares no `pbt_step_hz`, and both the loader test and the payload test assert its absence, so a later "sensible default" has to delete a test to land.

**`stepHz` and `pbtStepHz` are different quantities that collide.** In AM both are 200 — the width step and the PBT step — which is exactly the setup in which someone reuses one for the other. So USB is pinned beside it, where the width step is *absent* (its widths come from segments) and the PBT step is 50.

A third pin, added after review: a walk over **every shipped rig** asserting that a profile declaring the `pbt` capability declares a step somewhere. Review found the previous pins loaded ic7610 alone — corrupting ic7300 left the suite green — so naming four profiles would not have been enough either; adding a fifth rig has to fail rather than slip through.

Mutation-checked, each reverted: corrupting ic7300's AM step fails 1 test, removing every step from ic705 fails 2, giving ic9700's FM a step fails 1. All three were invisible at the previous head.

## Checks

- `uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py` — **460 passed, 1 skipped**.
- `uv run ruff check` and `ruff format --check` on the changed paths; `uv run mypy --strict src/rigplane/web` — clean.
- `npx tsc --noEmit` and `npx vitest run src/lib/types/` — **62 passed**.

Focused on the changed scope; the full suite is CI's.

## What follows

#3518 returns from draft with `PBT_MEASURED_STEP_HZ` replaced by the step resolved from the radio's own profile for the current mode, and the FM case refused because the mode has no PBT rather than because the arithmetic ran out of raw values.


morozsm:
Agent Review: BLOCKED ab2ab8854d00c2d6d6b3951aad5690012e1d0641

Independent review at the exact head, in a detached worktree at that SHA. The manual quotation is accurate word for word and every declared value matches it; one added doc sentence is false against the tree it lands in, and one of the two radios it declares data for is unpinned.

## What the manual actually says

Read `~/Documents/radio-manuals/IC-7300_Advanced_Manual_11a.pdf` with pypdf. PDF page 40 (the manual's own page label is **4-5**, not "p.40" — the index entry is "Twin PBT / Using … 4-5"; cite both). Section heading: "Using the Twin PBT" / "SSB, CW, RTTY and AM modes". Bullet, verbatim: "The PBT is adjustable in 50 Hz steps in the SSB, CW, and RTTY modes, and 200 Hz in the AM mode. In this case, the center shift value changes in 25 Hz steps in the SSB, CW, and RTTY modes, and 100 Hz in the AM mode." CONFIRMED, quotation and heading both. Nothing on pages 39-41 contradicts it. The IC-7610 CI-V Reference Guide p.4 gives only `0000 ~ 0255` for inner/outer TWIN PBT and states no step: it neither confirms nor contradicts.

`rigs/ic7300.toml` declares exactly the manual's mode set (USB/LSB/CW/RTTY at 50, AM at 200, FM none) — CONFIRMED. `rigs/ic7610.toml` additionally declares `USB-D`, `LSB-D`, `CW-R`, `RTTY-R` at 50; the quoted heading does not name them, and the same manual treats `SSB-D` as a distinct mode elsewhere (contents p.36 "SSB-D, CW and RTTY modes"; IF-filter table p.41 lists SSB-D and AM-D separately). Those four rows are a reasonable inference, but an **unstated** one.

## BLOCKING (in the diff; needs a commit)

**1. `frontend/src/lib/types/capabilities.ts:47-51` and `src/rigplane/profiles/__init__.py:208-213` state a guarantee wider than the tree.** "Absent means the mode has no twin PBT at all" / "`None` means the mode has no twin PBT at all". False at this head for two shipped profiles. Reproduced:

```
$ uv run python -c "... load_rig('rigs/ic705.toml').to_profile() ..."
ic705  pbt in capabilities: True
  {'USB': None, 'LSB': None, 'CW': None, 'CW-R': None, 'RTTY': None, 'RTTY-R': None,
   'AM': None, 'FM': None, 'WFM': None, 'DV': None}
ic9700 pbt in capabilities: True
  {'USB': None, 'LSB': None, 'CW': None, 'CW-R': None, 'RTTY': None, 'RTTY-R': None,
   'AM': None, 'FM': None}
```

`rigs/ic705.toml` and `rigs/ic9700.toml` both declare the `pbt` capability and `get/set_pbt_inner`/`_outer` (`0x14 0x07`/`0x08`), so 18 mode entries across two profiles have an absent step while the radio demonstrably has twin PBT. The PR body concedes this ("`ic705` and `ic9700` follow in their own change"). A consumer obeying the sentence as written — which is precisely what #3518 is being told to do — removes PBT entirely on IC-705 and IC-9700, and no test fails. `rigs/ftx1.toml` and `rigs/x6200.toml` have `[filters.width.*]` blocks with no step either; the class is four profiles, not one.

Required fix: delete the absolute clause, or narrow it to something checkable — absence means "this profile declares no PBT step", and only `ic7610`/`ic7300` declare one at this head, so a consumer must treat absence as *unknown* except where the profile also says the mode is fixed. The FM-scoped version of the same sentence inside the two TOMLs is fine as written: within those two files it is true.

**2. `rigs/ic7300.toml`'s five new values are entirely unpinned** — both new tests load `TEMPLATE_PATH = rigs/ic7610.toml` (`tests/test_rig_loader.py:32`). Mutation on the tree under review, reverted afterwards: setting `[filters.width.AM].pbt_step_hz = 50` **and** adding `pbt_step_hz = 50` to `[filters.width.FM]` in `rigs/ic7300.toml` → `458 passed, 1 skipped` — fully green. Parametrise the loader pin over both rig files while fixing (1).

## Mutation results (tree restored, `git status --short` clean at ab2ab88)

| Mutation | Result |
|---|---|
| (a) `pbt_step_hz = 50` on ic7610 `[filters.width.FM]` | 2 failed — `test_pbt_step_is_declared_per_mode_and_absent_where_there_is_no_pbt`, `test_capabilities_include_filter_config` |
| (b) ic7610 AM `200` → `50` | 3 failed — both loader pins + the payload pin |
| (c) delete the `pbtStepHz` line from `_serialize_filter_config` | 1 failed — `test_capabilities_include_filter_config` |
| (d) same as (a)+(b) applied to **ic7300** | **458 passed, 1 skipped — green** (finding 2) |
| control: reorder `pbt_step_hz` below `min_hz` in ic7610 USB | 458 passed, 1 skipped — the pins do not redden on a legitimate edit |

## MANDATORY SQUASH-BODY CORRECTION (commit message; replace at merge time, no commit needed)

Paragraph 2 of `ab2ab88`: "a 50 Hz constant computes 181 lattice positions over 256 raw values at a 9000 Hz AM width, **which collide under rounding**." REFUTED against the real code (`measuredPbtHzToRaw`, run through `vite-node`):

```
9000 50   positions 181  distinctRaw 181  collisions=0
9000 200  positions  46  distinctRaw  46  collisions=0
15000 50  LATTICE_NULL        (probe can report failure)
```

Minimum raw gap at (9000, 50) is 1; collisions are *impossible* whenever `pbtLattice` returns non-null, because its own guard is `positions - 1 > 255`. The real objection stands but is a different one: 181 reachable positions where the radio has 46, i.e. values the radio cannot produce. State that instead. Also "Mutation-checked: declaring `pbt_step_hz = 50` for FM fails both new pins" is true for ic7610 only (see (d)).

## REQUIRED BEFORE MERGE (PR body edits; head SHA unaffected)

- Same "collide under rounding" correction as above.
- "at the FM widths (15000/10000/7000 Hz, `fixed = true`) the lattice cannot be formed, readings vanish" — REFUTED at two of three widths: `10000/50` → 201 positions, `7000/50` → 141 positions, both form and return Hz; only `15000/50` returns `null`. The FM conclusion is right, but not for this reason — the reason is the manual's mode list.
- Say that `USB-D`/`LSB-D`/`CW-R`/`RTTY-R` at 50 are inference from the same DSP family, not from the quoted heading, and that the 2026-09-16 sweep was in SSB. Worth a clause in the two TOML comments as well while fixing the blockers.
- `ic7610.toml`'s provenance block cites another radio's manual; naming that explicitly ("no IC-7610 Advanced Manual in the local archive; its CI-V guide states no step") costs one clause and saves the next reader the check.

## Observations, not blockers

- `copyFilterConfiguration` (`frontend/src/lib/runtime/adapters/radio-view-model-adapter.ts`) whitelists `minHz`/`maxHz`/`stepHz`/`segments`/`table` into `ActiveFilterConfiguration` and silently drops `pbtStepHz`. The commit's stated route via `resolveFilterModeConfig` does carry it, so this is not wrong here — but #3518 must not consume the field through that copier.
- Orphan rule: acceptable sequencing. The field is data with pins, `web/server.py` is the only serializer, and `FilterWidthRule` is only ever constructed by keyword, so the class is swept.
- The 2026-09-16 bench sweep asserted in the TOML comments could not be verified from the repository — the evidence lives in private handoff artifacts. It restates a claim already in `PBT_MEASURED_STEP_HZ`'s comment from #3514.

## Checks reproduced at this head

`uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py` → 458 passed, 1 skipped (matches the commit). `ruff check` + `ruff format --check` on the five changed Python paths → clean. `uv run mypy --strict src/rigplane/web` → clean, 27 files. `npx tsc --noEmit` → clean; `npx vitest run src/lib/types/ src/lib/radio/filter-controls.test.ts` → 118 passed, 4 files. CI at this head as found: `visual` pass, `consumer-contracts-gate` pass, `quick` pending.

morozsm:
Agent Review: BLOCKED e592b6bebbe4b3bfcd0c4e95feac9dcc70db2314

Delta re-review of `ab2ab885..e592b6be`. Two of my three prior findings are
answered and reproduced as answered. One is answered at the named instance but
not across its class, and the new prose states the wider guarantee anyway while
naming the narrower test as its enforcement. That sentence is in the diff, so it
needs a commit.

## Finding 1 (prior) — CONFIRMED FIXED, by my own enumeration

I enumerated every profile the loader actually reads rather than the four
expected. `discover_rigs` is non-recursive over `rigs/*.toml`, skipping
`_`-prefixed and `.draft.toml`, so the corpus is exactly: FTX-1, IC-705,
IC-7300, IC-7610, IC-9700, TX-500, X6100, X6200. `rigs/examples/*.toml` are not
reachable (all three fail `load_rig` with `missing required section [filters]`)
and declare no `pbt`.

Declaring `pbt`: IC-705, IC-7300, IC-7610, IC-9700 — four, and all four now
declare steps. Non-declaring: FTX-1, TX-500, X6100, X6200 (X6200 drops the cap
deliberately, per its own note). **No live counterexample.** Measured values:

- IC-705 — USB/LSB/CW/CW-R/RTTY/RTTY-R 50, AM 200; FM/WFM/DV absent.
- IC-9700 — USB/LSB/CW/CW-R/RTTY/RTTY-R 50, AM 200; FM absent.

These match what the new per-profile pin asserts and what the provenance notes
claim for them.

## Finding 3 (prior) — CONFIRMED FIXED; provenance verified at source

I read the cited page rather than trusting the citation.
`~/Documents/radio-manuals/IC-7300_Advanced_Manual_11a.pdf` PDF page 40, footer
`4-5`, section head **"Using the Twin PBT"**, sub-head **"SSB, CW, RTTY and AM
modes"**, body: *"The PBT is adjustable in 50 Hz steps in the SSB, CW, and RTTY
modes, and 200 Hz in the AM mode. In this case, the center shift value changes
in 25 Hz steps in the SSB, CW, and RTTY modes, and 100 Hz in the AM mode."*
Every element of the citation — page, own-numbering, heading, wording — is
exact, in the TOML notes and in the PR body blockquote.

The ic705/ic9700 notes are honest: they label themselves INFERENCE, say the
manual is the IC-7300's, say the rig was never on the bench, and say nothing has
confirmed its AM step. That is a fair description of what I can and cannot
verify.

## Finding 2 (prior) — mutations re-run independently; all counts below are mine

Baseline `uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py`:
**460 passed, 1 skipped**. Each mutation reverted; `git status --short` clean
after every one.

| Mutation | Result |
|---|---|
| ic7300 AM `200` → `250` | **2 failed** (`..._declared_by_every_shipped_pbt_capable_profile`, `..._values_of_every_shipped_profile`) |
| ic7300 AM `200` → `50` (stays inside the allowed set) | **1 failed** (values pin only) |
| strip all 8 `pbt_step_hz` from ic705 | **2 failed** |
| ic7300 FM given `pbt_step_hz = 50` | **1 failed** (values pin only) |
| ic9700 FM given `pbt_step_hz = 50` | **1 failed** (values pin only) |
| **new 5th profile, `pbt` cap, no step anywhere** (`rigs/ic8888.toml`, model `IC-8888`) | **`..._declared_by_every_shipped_pbt_capable_profile` FAILED** — the walk catches exactly the case it was built for |

Control (the other direction): a legitimately-added 5th profile that *keeps* its
correct steps leaves all 6 `-k pbt` tests green. The walk does not redden on
correct work.

`test_pbt_step_declared_by_every_shipped_pbt_capable_profile` exists, is
collected, and reddens — that is not taken on its reading, it is the row above.

## BLOCKING — the class was not swept, and the comment claims it was

`frontend/src/lib/types/capabilities.ts:51-54`:

> Every shipped profile that declares the `pbt` capability **declares this for
> the modes that have it**, enforced on the server side by
> `test_pbt_step_declared_by_every_shipped_pbt_capable_profile`

The named test asserts only that `steps` is non-empty — a step **somewhere** —
plus `set(steps.values()) <= {50, 200}`. It does not assert per-mode coverage.

Reproduced (mutation E): a 5th profile cloned from ic9700 under model `IC-8890`,
`pbt` capability declared, `pbt_step_hz` kept **in USB only** and stripped from
AM, CW, CW-R, RTTY, RTTY-R:

```
pbt cap: True   steps: {'USB': 50}
uv run pytest tests/test_rig_loader.py -k pbt  ->  6 passed, 399 deselected
```

Green. A browser consumer obeying the documented reading would see `pbtStepHz`
absent in AM and conclude "this radio has no twin PBT in AM" — which is the
exact wrong conclusion the previous BLOCK was about, one level down:
partially-migrated instead of wholly-unmigrated. The sentence can become false
with no test failing, and it names a test as the thing that stops that.

Note the asymmetry inside this same delta: `src/rigplane/profiles/__init__.py`
(`FilterWidthRule.pbt_step_hz`) states it correctly — "no PBT-capable profile is
*merely unmigrated*" — which is precisely what the walk enforces. The TypeScript
neighbour restates it wider. So the correct wording already exists in the change;
one of the two copies drifted.

**Required fix (either, one commit):**

1. Narrow `capabilities.ts:51-54` to match the Python comment — drop "for the
   modes that have it" and say what the walk actually carries: no PBT-capable
   profile is merely unmigrated. This is the cheap fix and needs no new test; or
2. Strengthen the walk to assert per-mode coverage against the declared PBT
   mode set, and keep the sentence.

**Checks to run after the fix:** `uv run pytest tests/test_rig_loader.py
tests/test_web_capability_guards.py`; `npx tsc --noEmit`; and if option 2, re-run
mutation E above and confirm it reddens.

The PR carries exactly two commits, so both messages land in the squash body.

**In `ab2ab885`:** *"a 50 Hz constant computes 181 lattice positions over 256 raw
values at a 9000 Hz AM width, **which collide under rounding**."* The collision
clause is false and must not land. I recomputed it through the shipped code
(`measuredPbtHzToRaw` / `measuredPbtRawToHz` in
`frontend/src/lib/radio/filter-controls.ts`), driving every lattice position:

```
width=9000  step=50   positions=181  collisions=0  minGap=1  maxGap=2  first=0  last=255
width=15000 step=50   -> LATTICE CANNOT BE FORMED (null)
width=10000 step=50   positions=201  collisions=0  minGap=1  maxGap=2
width=7000  step=50   positions=141  collisions=0  minGap=1  maxGap=2
EXHAUSTIVE positions 2..400: formable=255, collision-bearing=0
```

181 positions is right; 0 collisions, minimum gap 1 raw unit. The comparison is
capable of reporting inequality — at step 200 the same harness reports
`positions=46 minGap=5 maxGap=6 first=2 last=253`, so it is measuring, not
returning a constant. The exhaustive row also establishes the PR body's stronger
claim that `positions - 1 > 255` makes collisions impossible **by construction**:
across all 255 formable lattices, none has a gap below 1.

Replace the clause with the PR body's own corrected wording: the harm is the
wrong step, not a malformed lattice.

**Also in `ab2ab885`, now stale:** *"Declared here for the two radios that were
on the bench; ic705 and ic9700 follow separately."* They do not follow
separately — `e592b6be` brings them into this same PR and says so. Self-corrected
later in the same squash body, but worth dropping at merge time.

The other claim I previously refuted — *"at FM widths the lattice cannot be
formed"* — is **not** in either commit message. I grepped both. It lived only in
the PR body, where it is already corrected, and the correction is accurate:
15000 forms nothing, 10000 and 7000 form 201 and 141 positions. Verified above.

## REQUIRED BEFORE MERGE (PR-body / squash-body edits, no commit needed)

`e592b6be` says *"corrupting ic7300's AM step fails 1 test"*; the PR body repeats
it. True only when the corrupted value stays inside `{50, 200}`. Corrupting to
`250` fails **2**, because the walk's value-set assertion also catches it. Not
false as written, but it understates the pins; worth stating the condition.

## Non-blocking, prose imprecision in the diff

`rigs/ic7300.toml` gained *"The -D and -R variants take the SSB/CW/RTTY figure by
inference"*. `ic7300.toml` declares **no** `-D` and **no** `-R` filter tables at
all — its modes are USB, LSB, CW, RTTY, AM, FM. The sentence is vacuous there,
describing rows the file does not have. Same half-issue in `rigs/ic705.toml` and
`rigs/ic9700.toml`, which have `CW-R`/`RTTY-R` but no `-D` mode. Only
`rigs/ic7610.toml` (USB-D, LSB-D, CW-R, RTTY-R) has both. Vacuously true rather
than false, so not blocking — but if this is touched anyway, drop the `-D` clause
where no `-D` table exists.

## Checks at this head

- `uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py` — **460 passed, 1 skipped**
- `uv run ruff check` / `ruff format --check` on the changed Python paths — clean, 2 files already formatted
- `uv run mypy --strict src/rigplane/web` — no issues, 27 files
- `npx tsc --noEmit` — clean
- CI at this head as found: `quick` pass (6m18s), `visual` pass (1m20s),
  `consumer-contracts-gate` / `grep-gate` / `classify` / `changes` pass. The two
  Agent Review Gate contexts are red pending this directive. `quick-v2-observe`
  reports *"Unknown: producer and publisher metadata bindings disagree"* — an
  observer-metadata condition, not a code result; I am reporting it as found and
  not resolving it.

Size accounting checked against the PR's own base, not a stale local `main`:
10 files, +237/-0. At the 10-file hard ceiling, not over it; the soft-threshold
justification is present as the first line of the body.

Review worktree was detached at the head under review and left byte-identical;
`git status --short` clean after every mutation and at the end.

morozsm:
Agent Review: BLOCKED 67461a497bfb0267cf7ebe99a5c6b600f3f1eb95

Delta re-review of `e592b6be..67461a49`. The finding I blocked on last cycle is
answered: the walk now enforces the guarantee, and the two hostile shapes I
named are killed. One new defect, in the correction's own prose.

## What I reproduced

Mutations applied to `rigs/ic7300.toml` in a throwaway worktree at this head,
each reverted with `git checkout --`; `tests/test_rig_loader.py -k pbt_step`:

| mutation | `..._declared_by_every_shipped_pbt_capable_profile` |
|---|---|
| step kept in USB only, dropped from LSB/CW/RTTY/AM (my last counterexample) | **FAILED** |
| `pbt_step_hz = 50` added to `[filters.width.FM]` | **FAILED** |
| AM step changed 200 -> 100 | **FAILED** |
| `[filters.width.PSK]` added with no step (mode in neither list) | **FAILED** |
| `[filters.width.FM]` section deleted entirely | passed (see below) |
| control: harmless comment added next to the AM step | passed |

So the strengthened walk discriminates in both directions. The one shape it
does not catch — a PBT-capable profile that omits FM from `filter_config`
altogether — is caught by its sibling `test_pbt_step_values_of_every_shipped_profile`,
whose exact per-profile dict comparison reddened on that mutation. The pair is
sound for the four shipped rigs.

## BLOCKING — in the diff, needs a commit

`src/rigplane/profiles/__init__.py:217` and
`frontend/src/lib/types/capabilities.ts:55` both state that the test
"requires the modes without a step to be **exactly** FM, WFM and DV".

That is false twice over, and I measured both halves.

The test requires `missing <= {"FM","WFM","DV"}` and
`not (declared & {"FM","WFM","DV"})` — a subset plus a disjointness, never an
equality. And the equality does not hold of the shipped data. Enumerated over
every `rigs/*.toml` that declares the `pbt` capability, at this head:

```
ic705.toml:  missing=['DV','FM','WFM']  equals_FM_WFM_DV=True
ic7300.toml: missing=['FM']             equals_FM_WFM_DV=False
ic7610.toml: missing=['FM']             equals_FM_WFM_DV=False
ic9700.toml: missing=['FM']             equals_FM_WFM_DV=False
```

A test written the way these two comments describe it would fail on three of
the four profiles this PR ships. The comment that must be trusted for a
consumer to read an absent `pbtStepHz` as "no PBT here" is the one stating a
rule the data does not satisfy.

The test's own docstring gets it right — `tests/test_rig_loader.py:1750` says
"exactly the ones with no twin PBT ... which leaves FM, and the FM variants
WFM and DV that the IC-705 adds". Both comments compressed that into a bare set
equality and dropped the qualifier that made it true. Class swept: those are
the only three sites in the tree carrying this sentence (`grep -rn "WFM and DV"`),
and only the two comments are wrong.

**Required fix (one commit, two lines).** State what the test enforces, e.g.
"every mode a profile declares without a step must be one of FM, WFM or DV, and
none of those may carry one". Keep both copies word-identical in force; they
are identical now, and that part of the fix held — there is no third variant.

## Also fix in that same commit — narrowed claim

`frontend/src/lib/types/capabilities.ts:52-53`: "Every shipped profile that
declares the `pbt` capability declares this for every mode that has twin PBT".
CONFIRMED over modes present in `filter_config`; **NARROWED** over the
profile's mode list, which is what "every mode" says. `rigs/ic7610.toml`
`[modes].list` contains `PSK` and `PSK-R`, and neither has any
`[filters.width.*]` entry, so neither carries a step; `resolveFilterModeConfig`
(`frontend/src/lib/runtime/props/panel-props.ts`) has fallbacks for `-D`,
`SSB`, `CW-R`->`CW` and `RTTY-R`->`RTTY` but none for `PSK`, so a consumer gets
`null` for that mode. The walk cannot see it: `missing` is computed from
`filter_config` keys only. Whether the IC-7610 has twin PBT in PSK is
**unestablished** — only its CI-V reference guide is in the local manual
archive, not its full manual — so I am not calling the sentence false. Scope it
to the modes a profile gives a filter rule to, which is what is enforced.
(`rigs/ic7300.toml` lists `CW-R`/`RTTY-R` with no filter entry too, but those
resolve to `CW`/`RTTY` on both sides and do carry a step.)

Still uncorrected in `ab2ab885`'s message, and this repo squash-merges with
`COMMIT_MESSAGES`, so it lands on `main` unless the body is edited at merge
time. The paragraph reads:

> AM is 200 Hz, so a 50 Hz constant computes 181 lattice positions over 256 raw
> values at a 9000 Hz AM width, which collide under rounding.

"which collide under rounding" is false, as established last cycle. Delete the
clause at merge time; the sentence stands without it, since the defect a
constant causes is the wrong step, not colliding positions. Everything else in
that paragraph I re-verified: `IC-7300_Advanced_Manual_11a.pdf` PDF page 40
(its own "4-5") is headed "Using the Twin PBT / SSB, CW, RTTY and AM modes" and
reads "The PBT is adjustable in 50 Hz steps in the SSB, CW, and RTTY modes, and
200 Hz in the AM mode." The citation in all four TOML headers is correct.

## Checks at this head

`uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py`
-> 460 passed, 1 skipped. `uv run ruff check` / `ruff format --check` on the
five changed Python paths -> clean. `npx tsc --noEmit` -> exit 0. Worktree left
byte-identical (`git status --short` empty). CI at this head as I found it:
visual, classify, changes, grep-gate, consumer-contracts-gate green; `quick`
pending; the review gate red for want of this directive.

morozsm:
Agent Review: BLOCKED 1064ba39aca2bb6d0295c5390a70b34bf541a62e

Delta re-review of `67461a49..1064ba39` (prose only, two comments). The
previous finding is fixed, and a new false claim was introduced on the
TypeScript side only.

## What the fix got right (CONFIRMED)

Both comments now state the guarantee as a subset plus a disjointness, and I
checked that clause against the assertions in
`test_pbt_step_declared_by_every_shipped_pbt_capable_profile`:

- "every mode ... other than FM, WFM and DV declares a step" == `missing <= no_twin_pbt`.
- "and none of those three does" == `not (declared & no_twin_pbt)`.
- "over every shipped rig" == `discover_rigs(RIGS_DIR)` + `checked >= 4`.

The comments claim *less* than the test in two places (the `{50, 200}` value
constraint and `checked >= 4` are enforced but unstated). That is not a defect.

Enumerated over all four PBT-capable profiles (`IC-705`, `IC-7300`, `IC-7610`,
`IC-9700`) via the real loader: the new wording is true of every one, where the
old "exactly FM, WFM and DV" was false of three.

Mutation-tested the walk in an extracted copy (`git archive 1064ba39 | tar -x`),
tree under review untouched:

- baseline: pass.
- drop `pbt_step_hz` from `rigs/ic7300.toml` `[filters.width.CW]` -> FAIL,
  `IC-7300: modes with no pbt_step_hz must be modes with no twin PBT, but ['CW'] are not`.
- add `pbt_step_hz = 50` to `rigs/ic705.toml` `[filters.width.WFM]` -> FAIL,
  `IC-705: ['WFM'] have no twin PBT and must declare no step`.
- restore -> pass.
- control A: add a correct `[filters.width.PSK]` with `pbt_step_hz = 50` to
  `rigs/ic7610.toml` -> still pass.
- control B: `max_hz = 10000` -> `9000` in `rigs/ic7610.toml` -> still pass.

The walk reddens in both directions and stays green on legitimate nearby edits.

## BLOCKING: `frontend/src/lib/types/capabilities.ts:58`

> A mode with no entry in the table at
> all is outside the guarantee and reaches this type as no config.

The second conjunct is false. `resolveFilterModeConfig`
(`frontend/src/lib/runtime/props/panel-props.ts`) — the function
`ab2ab885`'s own commit message names as the path by which consumers reach
`FilterModeConfig` — does not look the mode up directly. It builds a candidate
list with fallbacks: `<MODE>-D` when `dataMode > 0`, `USB`/`LSB` -> `SSB-D`/`SSB`,
`CW-R` -> `CW`, `RTTY-R` -> `RTTY`.

The IC-7300's filter table is `AM, CW, FM, LSB, RTTY, USB`; its `[modes]` also
lists `CW-R` and `RTTY-R`. Those two have no entry in the table, yet they reach
this type as a full config. Probe against the real serialized payload
(`_serialize_filter_config` on the loaded IC-7300 profile, fed to the real
`resolveFilterModeConfig` under vitest, in scratch space):

```
CW-R   -> {"defaults":[1200,500,250],"fixed":false,"pbtStepHz":50,"minHz":50,"maxHz":3600,...}
RTTY-R -> {"defaults":[2400,500,250],"fixed":false,"pbtStepHz":50,"minHz":50,"maxHz":2700,...}
CW     -> {"defaults":[1200,500,250],"fixed":false,"pbtStepHz":50,...}
PSK    -> null
```

So the sentence is NARROWED, not CONFIRMED: true of the one instance the Python
comment names (IC-7610 `PSK`, which has no fallback candidate and does return
`null`), false of the whole `-R`/`-D` family it also claims. A reader of this
type who obeys it will treat a present `pbtStepHz: 50` on IC-7300 `CW-R` as
impossible.

This is also the same drift that blocked `67461a49`, in the other direction:
the two comments are again **not** identical in force.
`src/rigplane/profiles/__init__.py:219` is correct — "nothing here speaks for
it" is a claim about the dataclass and the loader, where a mode absent from
`[filters.width.*]` genuinely produces no `FilterWidthRule`. The TypeScript
copy alone adds a consumer-path claim its own consumer path contradicts.

Case: **in the diff**. It lands in the repository and only a commit changes it.

### Required fix

Delete the clause — the sentence's first half is true and self-sufficient:

```
 *  nor "this mode was not filled in". A mode with no entry in the table at
-*  all is outside the guarantee and reaches this type as no config.
+*  all is outside the guarantee.
```

Per CLAUDE.md's delete-before-narrow ruling, deletion is the default here. If a
replacement is wanted instead, it must be established against
`resolveFilterModeConfig`, not asserted — e.g. "whether such a mode reaches this
type as a config at all depends on `resolveFilterModeConfig`, which falls back
for the `-D` and `-R` families" — and I have not established any wording beyond
the probe above.

### Checks to run on the new head

Comment-only, so `cd frontend && npx tsc --noEmit` is sufficient; no Python path
changes, so the suites below need no rerun.

## Everything else at this head

- `git diff --stat 67461a49 1064ba39`: 2 files, +12/-8, both comment bodies.
  Nothing else moved.
- `uv run pytest tests/test_rig_loader.py tests/test_web_capability_guards.py`
  -> 460 passed, 1 skipped.
- `uv run ruff check` / `ruff format --check` on
  `src/rigplane/profiles/__init__.py` -> clean.
- `npx tsc --noEmit` -> exit 0. `npx eslint` on the changed file -> clean.
  (Prettier is not a tool of this repo — not in `frontend/package.json` nor any
  workflow — and an ad-hoc run reports `origin/main` equally unformatted, so I
  discarded it as noise.)
- CI at this head: `quick` pass (6m16s), `consumer-contracts-gate`,
  `grep-gate`, `classify`, `changes` pass; the two Agent Review Gate contexts
  are red only for the absent directive.

## MANDATORY SQUASH-BODY CORRECTION (not blocking; fix at merge time)

`ab2ab885`'s message contains, in the paragraph beginning "Two things follow
that a single constant cannot express":

> AM is 200 Hz, so a 50 Hz constant computes 181 lattice positions over 256 raw
> values at a 9000 Hz AM width, **which collide under rounding.**

The arithmetic is right — `9000 // 50 + 1 == 181` — but the trailing clause is
not the consequence it names, and this repo squash-merges with
`COMMIT_MESSAGES`, so it would land on `main`. 181 distinct positions do not
collide with each other; what is many-to-one is the 256 raw values mapping onto
them, and that is equally true of a correct 200 Hz step (`9000 // 200 + 1 == 46`
positions over the same 256 raw values). The clause therefore describes
something that happens either way and cannot be the reason the constant is
wrong.

Delete the four words `, which collide under rounding` so the sentence ends at
"...at a 9000 Hz AM width." If a replacement is wanted, the defect a 50 Hz
constant actually causes in AM is the opposite one — it computes 181 positions
where the radio has 46, so 135 of them are positions the radio cannot reach —
and that arithmetic is the whole of my evidence for the replacement.

No REQUIRED BEFORE MERGE items in the PR body or comments.

Tree under review left byte-identical; `git status --short` clean at
`1064ba39aca2bb6d0295c5390a70b34bf541a62e`.

morozsm:
Agent Review: PASS 6d2488fea76bb8a22230d62c38a53056fc9745eb

Delta re-review of the one prose-only commit `1064ba39..6d2488fe` (2 files, +14/-4).
The clause I blocked on — "A mode with no entry in the table at all is outside the
guarantee and reaches this type as no config" — is gone, and its replacement is
true on every clause I could check.

## The corrected clause, verified against the code

Read `resolveFilterModeConfig` in `frontend/src/lib/runtime/props/panel-props.ts`
at this head, not the commit message. Candidate order: `<MODE>-D` (when
`dataMode > 0`), `<MODE>`, then `SSB-D`/`SSB` for `USB`/`LSB`, `CW` for `CW-R`,
`RTTY` for `RTTY-R`; the loop returns the first present entry, else `null`.

- CONFIRMED `CW-R` -> `CW` and `RTTY-R` -> `RTTY`: both branches are in the
  resolver, and the shipped pin `radio-view-model-adapter.test.ts` exercises
  exactly those rows (`['CW-R', 0, 'CW'], ['RTTY-R', 0, 'RTTY']`).
- CONFIRMED "the SSB variants to `SSB`": `USB`/`LSB` push `SSB-D`/`SSB`; the same
  pin covers `['LSB', 0, 'SSB'], ['LSB', 1, 'SSB-D']`.
- CONFIRMED ic7300's `CW-R` carries 50 despite no entry of its own:
  `rigs/ic7300.toml` `[modes].list` includes `CW-R`, its `[filters.width.*]`
  sections are USB/LSB/CW/RTTY/AM/FM only, and `[filters.width.CW]` declares
  `pbt_step_hz = 50`. (Stated in the commit message, not in the comments — the
  comments do not name it, which is fine.)
- CONFIRMED the IC-7610's `PSK` has neither entry nor fallback:
  `rigs/ic7610.toml:871` lists `"PSK", "PSK-R"`; there is no `[filters.width.PSK]`
  and no PSK branch in the resolver, so it resolves to `null`. `PSK-R` is in the
  same position; the comments say "as the IC-7610's `PSK` is", which is
  illustrative rather than an exhaustiveness claim, so this is not a defect.

## Identical in force

Read side by side (`src/rigplane/profiles/__init__.py` `FilterWidthRule.pbt_step_hz`
and `frontend/src/lib/types/capabilities.ts` `FilterModeConfig.pbtStepHz`). Same
three propositions in the same order: outside the guarantee / not necessarily
without a step / only no-entry-and-no-fallback resolves to nothing. The only
wording difference is "resolves to nothing" vs "resolves to no config at all",
which is the same claim in each language's vocabulary. No drift.

## The class, swept

I checked every other statement that exists in both copies:

- The guarantee sentence ("every mode in the filter table other than FM, WFM and
  DV declares a step, and none of those three does", attributed to
  `test_pbt_step_declared_by_every_shipped_pbt_capable_profile`). The two copies
  agree, and both match the test's actual assertions: `missing <= no_twin_pbt`
  plus `not (declared & no_twin_pbt)`. Measured over the shipped profiles —
  ic705/ic7300/ic7610/ic9700 each declare 50 for every non-FM/WFM/DV mode, 200
  for AM, and `None` for FM/WFM/DV; no other profile declares the `pbt`
  capability.
- "the two are both 200 in AM and differ in USB" / "the two collide at 200 in AM
  and differ in USB". Agree, and measured true: in AM `step_hz = pbt_step_hz =
  200` in all four profiles; in USB `step_hz` is absent while `pbt_step_hz = 50`.
- The four TOML provenance blocks (ic705/ic7300/ic7610/ic9700) are two variants,
  and the difference is deliberate and correct: the bench-measured pair says
  "independently confirmed by a bench sweep", the inferred pair carries the
  explicit "INFERENCE, not measurement, for this radio" paragraph.

## Checks

Fresh detached worktree at this SHA. `uv run pytest tests/test_rig_loader.py
tests/test_web_capability_guards.py` -> 460 passed, 1 skipped, reproducing the
count in the commit message exactly. `ruff check` and `ruff format --check` on
`src/rigplane/profiles/__init__.py` clean. `npx tsc --noEmit` exits 0. No
mutation run and none required: the delta changes only comment text, so there is
no behaviour it could have broken and no pin whose discriminating power it could
have weakened — the behavioural claims it makes were checked against the resolver
source and the shipped TOML instead. Tree left byte-identical; `git status
--short` clean. CI at this head: all required contexts SUCCESS except `quick`,
which was IN_PROGRESS when I looked.

## MANDATORY SQUASH-BODY CORRECTION (commit message; replace at merge time — no commit needed)

Unchanged from the previous two cycles and still uncorrected in `ab2ab885`'s
message. This repo squash-merges with `COMMIT_MESSAGES`, so it lands on `main`
unless the body is edited in the merge dialog. In `ab2ab885`, paragraph 2 reads:

> AM is 200 Hz, so a 50 Hz constant computes 181 lattice positions over 256 raw
> values at a 9000 Hz AM width, which collide under rounding.

**Delete the four words `, which collide under rounding`** (keep the sentence,
ending it at "AM width"). The clause is false: driving `measuredPbtHzToRaw` over
every lattice position gives width=9000 step=50 -> 181 positions, 181 distinct
raw values, 0 collisions, minGap=1. Collisions are impossible whenever
`pbtLattice` returns non-null, since its own guard is `positions - 1 > 255`. If a
replacement is wanted rather than a deletion, the true objection is "…181
reachable positions where the radio has 46, i.e. values the radio cannot
produce."

Earlier non-blocking items from cycles 2–4 (the vacuous "-D and -R variants"
sentence in `rigs/ic7300.toml`, and `e592b6be`'s "fails 1 test" which is 2 when
the corrupted value leaves `{50, 200}`) stand as recorded; neither blocks.

PR #3520, opened 2026-09-17
fix(MOR-2497): block the PBT controls in modes that have no twin PBT

morozsm:
## What

`toFilterProps` reported `hasPbt` from the radio-wide `pbt` capability alone, so on a radio that has PBT the FilterPanel kept its two PBT controls live in FM. Owner ruling, 2026-09-17: in FM the controls must be blocked, because the radio has no twin PBT there (IC-7300 Advanced Manual p.40 heads the section "SSB, CW, RTTY and AM modes").

`hasPbt` is now `hasCap(caps, 'pbt') && modeHasTwinPbt(caps, filterConfig)`. The per-mode fact is the `pbtStepHz` the profile publishes since #3519: present for a mode with twin PBT, absent for one without.

## The one decision in it

Two absences are kept apart, in `panel-props.ts: modeHasTwinPbt`:

- some mode in the payload declares `pbtStepHz` and the current one does not → no twin PBT here, controls off;
- no mode declares it → a payload from before the field existed, so the radio-wide capability decides as it did before.

The second case is a real payload, not a hypothetical: `frontend/src/lib/runtime/adapters/__tests__/fixtures/ic7300-capabilities.json` contains no `pbtStepHz` (`git show <head>:<path> | grep -c pbtStepHz` → 0, with `grep -c filterConfig` → 1 as the control that the read worked).

## Scope — what this does NOT gate

`hasPbt` has one consumer, `components-v2/panels/FilterPanel.svelte` (`git grep hasPbt` over `frontend/src`, tests excluded). Three other places still decide on the radio-wide capability and are untouched here:

- `adapters/radio-view-model-adapter.ts` — `hasPbtCap`, which makes the semantic view model's `pbtInner`/`pbtOuter` structural;
- `adapters/panel-adapters.ts: pbtStructural`;
- `commands/panel-commands.ts` — the write path.
- the PBT **Reset** button in the same `FilterPanel.svelte` (`pbt-reset-button`): it sits in `filter-actions`, after the `{#if hasPbt}` block closes, so it stays live in FM. Three PBT controls in that panel; this PR gates two.

**This PR does not clear the FM symptom on the default skin.** `skins/registry.ts` sets `DEFAULT_SKIN_ID = 'desktop-v2'`; `presentation/layouts/desktop-declarations.ts` declares the `filter` zone; and `LeftSidebar.svelte` renders `FilterPanel` only when `!declared.has('filter')`. So on `desktop-v2` the panel this PR gates is not shown at all, and the filter surface there reads PBT off the semantic view model, which still follows the radio-wide capability. What this PR fixes is `FilterPanel` wherever it is shown; the default skin follows with the reader step (#3518). Those belong to the reader and writer steps still open on MOR-2497 (#3518 and what follows it), and are listed here so the remainder is visible rather than implied closed.

## Checks

At `669e9ab2`, from `frontend/`:
`npx vitest run src/lib/runtime/props/__tests__/panel-props.test.ts src/lib/runtime/adapters/__tests__/` → 57 files, 1842 tests passed. That directory is included because it holds the fixture conformance test the legacy-payload case protects.

The branch is one commit on top of `main` at `c02874b2` (0 behind, 1 ahead). 2 files, 92 additions, 1 deletion.

Note for whoever merges: the ticket reference in the branch and title auto-closes MOR-2497 on merge, and MOR-2497 is not finished. Reopen it.


morozsm:
Agent Review: PASS 669e9ab2269b2ad5a1ca5e8536d24de29f98d060

Independent review by a verifier that did not author the change, at the frozen head.

**`modeHasTwinPbt` checked over the whole shipped input space, not sampled.** A script reimplementing `resolveFilterModeConfig`'s candidate walk against `tomllib`-parsed `rigs/*.toml`, over every `pbt`-capable profile x every mode in its `[modes]` list x dataMode in {0,1}: 58 combinations across ic7300, ic7610, ic705, ic9700. The modes that lose the controls are FM (all four), WFM and DV (ic705), and PSK / PSK-R (ic7610). No mode with twin PBT loses them. The fallback worry is refuted concretely: ic7300 has no `USB-D` / `CW-R` / `RTTY-R` filter entries, each falls back to `USB` / `CW` / `RTTY`, and all three declare `pbt_step_hz`. PSK / PSK-R going dark is the known open question on MOR-2497, not a finding here.

**`!== undefined` is the right predicate.** `src/rigplane/web/server.py: _serialize_filter_config` is the only emitter and omits the key when `pbt_step_hz is None`, so `null` does not reach the wire.

**Mutations, each in a throwaway worktree, tree clean after each:**

| mutation | failed |
|---|---|
| baseline | 0 — 57 files / 1842 tests, matching the body |
| restore `hasPbt: hasCap(caps,'pbt')` | 2 |
| collapse to `modeConfig?.pbtStepHz !== undefined` | 2, including the `mor1562` fixture conformance test |
| `return anyModeDeclares` (inverted) | 4 |
| control: operands swapped, equivalent | 0 |

`npx tsc --noEmit` exit 0. The commit message's own mutation counts reproduce exactly.

**One behaviour change nothing pins:** with capabilities published but `state === null`, `hasPbt` was `true` and is now `false`. Transient, before the first state arrives. Not blocking.

**Required before merge, both in the PR body and both applied:** that the default skin `desktop-v2` does not show `FilterPanel`, so this PR does not clear the FM symptom there; and that the PBT Reset button sits outside `{#if hasPbt}` and stays live in FM. The coordinator re-verified both by command before editing the body.

**Not established:** which skin the owner had open when reporting the symptom; and whether the radio physically ignores PBT writes in FM — no bench run, the manual establishes only that twin PBT is documented for SSB, CW, RTTY and AM.

PR #3521, opened 2026-09-17
fix(MOR-2367): keep the scope passband through band-edge and out-of-band tuning

morozsm:
## What the operator saw

Tuning with the arrow keys made the passband overlay on the scope vanish for 1–6 s and come back by itself. The tuning marker stayed. It looked timing-dependent; it was not. It happened whenever the confirmed frequency left a profile band, moved while outside every band, or came back in.

## Cause

`scope-passband-display.ts: inspect` built `continuityIdentity` with the profile band the frequency falls in. Outside every band that identity is `null`, and `projectScopePassbandDisplay` allowed `translating` and `holdingFrequency` only for an equal, non-null identity. So a confirmed step across a band edge — or any step taken while outside the bands — counted as a context change and retired the tuple to `unknown/retired`. `SpectrumPanel.svelte` then had no tuple, `passbandHz` went to 0 and `.passband-overlay` left the DOM. Re-capture needs strictly newer observation markers on all three geometry paths (`main.filterWidth`, `main.pbtInner`, `main.pbtOuter`), hence the wait for the slowest poll.

## Contract (owner ruling, 2026-09-17)

The passband's width is determined only by the selected filter and the modulation mode. Band membership is therefore **not** a context boundary for the passband geometry. Continuity is now decided by a band-less context (`bandlessIdentity`): in→in, in→out, out→in and out→out all translate the last coherent geometry to the new frequency as `stale` + `translated`, keeping `widthHz` and `shiftHz`, until fresh observations arrive. Nothing is invented. Every other fence is unchanged: mode, filter, DATA, session, epoch, receiver, slot, frame mode, span, A→B→A. A one-step jump to another band with mode and filter unchanged also translates, and is pinned.

Two existing cases that expected `unknown` for `'band'` and `'outside profile bands'` are restated as positive pins of the new contract; their old meaning is carried by new pins that a mode change and a filter change at the band edge still retire. `'band definitions'` — the profile's band table itself changing — still retires: that is a capabilities change, not tuning. Profiles with no band table at all keep their previous behaviour (`bandlessIdentity` is `null` there), pinned by the existing "retains complete … edges" cases.

## Live evidence — real IC-7610 over LAN, USB FIL1 3600 Hz, scope CTR, default skin, RX only

Instrument: a `requestAnimationFrame` sampler recording each frame whether `.tune-line` and `.passband-overlay` exist with non-zero width; trusted `ArrowRight`/`ArrowLeft` key events, 1 kHz step. The 20 m band ends at 14 350 000 Hz in `rigs/ic7610.toml`. Production build of each commit, server from `main`.

| build | step | passband |
|---|---|---|
| `main` `b83e7c81` | slow ×9, inside the band | never missing |
| `main` | bursts crossing 14.350, 8 runs | missing 1.25–6.2 s, 8 of 8 |
| `main` | 30+30 burst, stays 3 kHz short of the edge | never missing |
| first commit `02683998` | +10 kHz out across the edge | never missing |
| `02683998` | −10 kHz back in (first steps still out of band) | **missing 6.3 s** |
| head `11ff6df5` | 14.239 → 14.345 → 14.355 → 14.375 → 14.365 → 14.345 → 14.239 | **0 of 1251 samples missing, 38.4 s** |

The `02683998` row is why there are two commits: the first fix required at least one side to be inside a band, and a temporary probe logging the reducer's flags on the live page showed both sides `null` while tuning outside the bands. The probe was never committed. The tune line was missing in 0 samples throughout. The radio was returned to its starting frequency after every run; nothing was transmitted.

Limits: one radio, one transport, one browser engine; the pane delivers 10–35 frames per second, so a gap shorter than ~100 ms would be missed; synthetic key repeat is more regular than a held key.

## Checks

From `frontend/`, single file only: `npx vitest run src/lib/runtime/adapters/__tests__/scope-passband-display.test.ts` → 206 passed at head `11ff6df5`; `npx tsc --noEmit` → exit 0. Both reproduced by the independent reviewer. `quick` and `visual` pass at this head.

Mutations, measured by the reviewer at head `11ff6df5` against that one file, each restored: band triple put back into continuity → 8 tests fail; the null/null guard alone restored → 7; the empty-band-table guard dropped → 3. Every other component of the context is pinned — removing it fails at least one test: mode 4, filter 3, DATA 2, session epoch 2, receiver 2, slot 4, frame mode 3, span 1, capability identity (including band definitions) 4, provider generation 3, transport epoch 1. The builder's own counts, taken on the intermediate commit with fewer tests, were lower and are not quoted here. Run against `main`'s code, the head's test file fails 8.

## Left for a follow-up

`continuityIdentity` is now written into the state and read by no decision (`git grep -n continuityIdentity -- frontend/src` finds only this adapter; `bandlessIdentity` as the control finds the reads). Its removal is MOR-2502, kept out of this PR so that a reviewed, live-accepted head does not change.

Two corners the review found unpinned and benign: with a profile that has no band table at all, an interrupted selection recovery is now kept through a tune where `main` cancelled it — unreachable with any shipped profile, all of which carry `freq_ranges`; and a far jump during an interrupted selection recovery is no longer exercised by a test. A band recall that arrives frequency-first shows the old width as `stale`+`translated` until the filter observation lands, then retires — the same as on `main`.

PR #3518 edits the same file on other lines; whichever merges second merges `main`.


morozsm:
Agent Review: PASS 11ff6df5ca0de360d59f651fd2cff9995c7f1aa6

Independent review by a GLM-5.3 worker that did not author the change (author: Kimi K3), at the frozen head. Single-file test runs only; the wide run was read from CI (`quick` and `visual` pass at this head).

**Scope of commands.** `git rev-parse HEAD` → `11ff6df5ca0de360d59f651fd2cff9995c7f1aa6`. All test evidence: `cd frontend && npx vitest run src/lib/runtime/adapters/__tests__/scope-passband-display.test.ts` (single file only); `npx tsc --noEmit` → exit 0. CI read via `gh pr checks 3521`: `quick` pass, `visual` pass at head; Agent Review Gate fail only awaits this directive.

**1. No other fence got looser.** Old `continuityIdentity` = `[providerGeneration, capabilityIdentity(caps), session.epoch, receiver, slot, mode, filter, data, 'hardware', transportEpoch, frame.mode, partition.name, partition.start, partition.end, span]`, null when no partition. New `bandlessIdentity` = same list minus the partition triple, null only when `flatBands.length === 0`. The `context` array is untouched by the diff (`git diff origin/main...HEAD` shows no edit between lines 242–244). Every fence discriminates (table below); no unpinned fence.

**2. `recoveryContext`.** Old non-null / new null exactly when a candidate exists and `flatBands.length === 0` (old fell back to `identity`). Scratch test (slotted, `freqRanges = []`, A→B then tune during recovery): main → recovery `{context: identity}`, tune **cancels** it; head → recovery `{context: null}`, tune **keeps** it; both sequences end `unknown → current` after marker renewal, no blank regression. Unreachable with shipped profiles (all `rigs/*.toml` carry `freq_ranges` with bands; the no-band `fixture()` is single-slot). Not pinned by any test — gap noted, benign.

**3. No-band profiles.** No shipped rig lacks a band table (`rg -c "freq_ranges" rigs/`); `rig_loader.py:2185` defaults `freq_ranges` to empty, so only hand-written custom profiles hit it, and there tuning still blanks — by design. Pinned: mutation N (drop the empty-table null guard) fails the three `retains complete … edges` tests.

**4. Translation honesty.** Band-recall arriving frequency-first: display is `stale+translated` with the old width until the filter observation lands (poll lag, ~1.5–3 s per profile), then `changedGeometry` + context change retire → blank until all three markers advance. The `current`-with-old-width escape (`translating && crossedFloors && strict && axisAligned`) is byte-identical on `origin/main` — pre-existing MOR-2437 behaviour, not introduced here. No mixed width/shift tuple is possible (retire fires in the same frame).

**5. Restated tests.** `'band'`/`'outside profile bands'` retire cases replaced one-for-one by positive pins plus new mode/filter-at-edge retire pins; `'band definitions'` still retires (mutation K fails its pin). Lost corner: a far jump (7.1 MHz) during an *interrupted selection recovery* is no longer exercised — unpinned, behaviour follows from the same rule.

**6. Dead field.** `git grep -n continuityIdentity -- frontend/src` → only the adapter itself (interface, write at 248, copy at 370); control grep `bandlessIdentity` → reads at 295/308. `ScopePassbandDisplayState` consumers: module + test only. Removal is a trivial mechanical deletion — either remove it in this PR or name the follow-up ticket (currently none).

**Fence-by-fence mutations** (each: edit → single-file run → `git checkout HEAD --` → clean):

| component dropped / mutated | failed tests |
|---|---|
| band triple + null guard (old identity) | 8 |
| band triple only | 8 |
| null-guard only (`!partition`) | 7 |
| mode | 4 (selection-recovery, hard retire, band-edge, markers) |
| filter | 3 |
| DATA | 2 |
| session.epoch | 2 |
| receiver | 2 |
| slot | 4 (incl. A→B→A replay pin) |
| frame.mode | 3 |
| span (bandless only) | 1 |
| capabilityIdentity (incl. band definitions) | 4 |
| providerGeneration | 3 |
| transportEpoch | 1 |
| empty-table guard | 3 |

**Prose findings / REQUIRED BEFORE MERGE (body edit):** the Checks paragraph's mutation counts are not reproducible: "band back into continuity → 4 pins fail" measures **8**; "null/null guard → 3 out-of-band pins fail" measures **7**; "red first: 5 failed on main" measures **8** against head's test file (the 5 plausibly reflects an intermediate test file — as stated it is false). Fix the three numbers before merge. Also state for item 6 whether the dead field gets a ticket or an in-PR removal (body edit or commit).

**Not established:** the live IC-7610 evidence table (no hardware run in this review — read, not endorsed); behaviour of the far-jump-during-recovery corner beyond code tracing; `02683998` intermediate red counts.

`git status --short` is empty; HEAD unchanged at `11ff6df5ca0de360d59f651fd2cff9995c7f1aa6`.
**Coordinator note:** the three counts flagged above are corrected in the PR body to the reviewer's measurements at this head, and the dead field is now MOR-2502. The live-evidence table was produced by the coordinator on a real IC-7610 and, as the reviewer says, was read and not re-run by this review.

PR #3522, opened 2026-09-17
fix(MOR-2500): write twin-PBT edges on the measured lattice

morozsm:
## What is wrong today

On a radio with twin PBT the write path still speaks the profile's fabricated ±1200 Hz scale, and one handler mixes units:

- **The ±1200 Hz bound — operator-visible.** `mapIfShiftToPbt` clamped the target and both edges to ±1200, and `pbtHzToRaw` scaled by the declared 1200/128. The radio's real span is ±filterWidth/2, so a written value lands elsewhere. Live on an IC-7610 (USB, 3600 Hz, default skin, PBT Inner slider to +1200 Hz): `main` writes raw 254, which is +1800 Hz; this PR writes raw 212.
- **MOR-2500 — latent.** `panel-commands.ts: onIfShiftChange`, in its PBT branch, passed the state's **raw 0..255** edges into `filter-controls.ts: mapIfShiftToPbt`, which expects **Hz**. With unequal edges that narrows the passband instead of moving it. The independent review established that no shipped surface reaches that branch: it is the `else` of the `if_shift` capability, every surface gates its IF-shift control on that same capability (`radio-view-model-adapter.ts: ifShiftControlStructural`, `panel-props.ts: toFilterProps` `hasIfShift`), and the one ungated caller, `FilterPanel.svelte: handleTableReset`, sits inside `{#if isTableMode}`, which only `rigs/ftx1.toml` enables — and the FTX-1 declares `if_shift`. The handler is corrected and pinned anyway.

## What changes

The writers move onto the measured lattice the readers went onto in #3518, reusing what exists: `measuredPbtRawToHz` and its inverse `measuredPbtHzToRaw` (already pinned both ways against the 73-position bench sweep in `filter-controls.test.ts`), and the same `resolveFilterModeConfig(caps, mode, dataMode)` resolution for the ACTIVE receiver, so a reader and a writer cannot disagree about which mode's step applies.

- `onIfShiftChange` (PBT branch) reads the current edges in Hz on the lattice, moves the passband in Hz, and writes both edges back on the same lattice. The native `if_shift` branch is untouched.
- `mapIfShiftToPbt` works in Hz end to end. The requested shift snaps to a whole multiple of the step, and **the shift is what gets clamped, never one edge alone**: a shift that would push an edge past ±filterWidth/2 stops at the boundary, so the passband width survives the move.
- `onPbtInnerChange`, `onPbtOuterChange`, `onPbtReset` write the nearest reachable lattice position.
- **No lattice, no write.** `activePbtLattice()` returns `null` when the current mode declares no `pbtStepHz` (FM), when no mode declares one (a legacy payload), or when the width is not a finite number; all four handlers then return before dispatching anything. This is deliberately stricter than the readers' `modeHasTwinPbt`, which keeps the radio-wide capability for the legacy payload: a write with no lattice to land on has no honest fallback.

`FILTER_BIPOLAR_MIN/MAX` and `clampToBipolarRange` are deleted with their last caller.

## Known consequence: Reset where the modelled lattice has no centre

`onPbtReset` used to write raw 128 unconditionally; it now writes the modelled lattice position nearest 0 Hz. The model (`width // step + 1` positions) has an even count where width/step is odd, and then no centre: CW 250 Hz resets to raw 106, AM 9000 Hz to raw 125. The reset test runs at 3600 Hz, where both give 128.

Measured on the IC-7610 after the review raised this (2026-09-17, RX only, `set_pbt_inner` swept raw 0..255 and read back): the **radio** always has a centre at 128 — 250 Hz has 5 positions (25, 76, 128, 179, 230), 350 Hz has 7, not the modelled 6 and 8 — and it snaps a written 106 (at 250 Hz) or 125 (at AM 6000 Hz) to 128. So Reset still lands on centre on the radio. The model's position count is wrong for odd width/step, which also affects the readers merged in #3518; that is a separate fix to `measuredPbtRawToHz` / `measuredPbtHzToRaw`, recorded on MOR-2497.

## Not in this PR (PR-W2)

`panel-props.ts: toFilterProps` still converts with the old formula for the v2/LCD skins; `FilterPanel.svelte`'s Reset button sits outside `{#if hasPbt}`; the sliders on both surfaces are still bounded at ±1200 Hz, so the wider reach this PR makes writable is not yet reachable by dragging. `pbtRawToHz` stays exported for those remaining users. `pbtHzToRaw` is left with no caller in `src` (`git grep -n "pbtHzToRaw(" -- src` finds only its definition); it is deleted in PR-W2.

## Checks

At `6ce82fed`, from `frontend/`, single files only — directory runs are left to `quick`:
`panel-commands.intent.isolated.test.ts` → 142 passed · `filter-controls.test.ts` → 59 passed · `npx tsc --noEmit` → exit 0 (run by the coordinator). `git grep -ln mapIfShiftToPbt -- src` finds only those two test files and the two source files.

The builder's run was cut short by its sandbox at its first mutation check, so **this PR carries no mutation table from the author**; the stray mutation was reverted by `git checkout HEAD --` and the tree is clean. The independent review ran it: six mutations killed, one survivor shown unreachable by construction, one equivalent-rewrite control green; see the review comment.

4 files, 323 insertions, 41 deletions.


morozsm:
**Live, before and after, on a real IC-7610 — the ±1200 Hz write bound is a visible defect on `main`, and this PR removes it.**

2026-09-17 15:30 EDT, LAN, server from `main`, default skin `desktop-v2`, 1440×900, MAIN 14.137.200 USB, FIL1 3600 Hz, RX only. The semantic surface's PBT Inner range input (`min=-1200 max=1200 step=25`) was set through the page and the radio's raw value read back from `/api/v1/state` 5 s later. Expected raws come from the lattice formula in `pbt-evidence/README.md` at 3600 Hz / 50 Hz (73 positions, margin 1): `raw = 1 + round_half_up((hz/50 + 36) · 253/72)`.

| slider set to | `main` frontend (`9f1af0e2`) | this PR (`6ce82fed`) | lattice raw for that Hz |
|---|---|---|---|
| +1200 Hz | **254** — the radio's top position, **+1800 Hz** | **212** | 212 |
| −400 Hz | not run | 99 | 99 |
| 0 Hz | 128 | 128 | 128 |

On `main` the old conversion sends raw 255 for "+1200 Hz" and the radio snaps it to 254, so the operator who asks for +1200 Hz gets +1800 Hz — 600 Hz further than shown. With this PR the radio lands on the asked-for lattice position. 99 is in the measured position list from the 2026-09-16 sweep; 100 is not.

Not exercised live: `onIfShiftChange`'s PBT branch. On this radio and skin no control reaches it — the semantic surface renders no IF-shift row for a radio without the `if_shift` capability, and `FilterPanel.svelte` disables its IF-shift slider when `hasPbt`. So the MOR-2500 collapse is fixed and pinned at the handler level; whether any shipped surface can trigger it on a PBT radio was **not established**, and the reviewer is asked to say. Also not run: AM's 200 Hz step, the refusal in FM, SUB active.

The radio was returned to PBT 128/128, USB, FIL1 3600 Hz, 14 137 200 Hz and read back; `lastActuation` is `null` — nothing was transmitted.

morozsm:
Agent Review: PASS 6ce82fedcbcb75c5c7b788c053dcbcd9affe8fe7

Independent review by the internal `verifier` (Opus); author: GLM-5.3. Throwaway detached worktree at the exact head, removed afterwards.

**Baselines confirmed at the head:** `filter-controls.test.ts` 59 passed, `panel-commands.intent.isolated.test.ts` 142 passed, `npx tsc --noEmit` exit 0.

**Mutations, one test file each, restored with `git checkout HEAD --`:**

| # | Mutation | Result |
|---|---|---|
| a | feed raw edges to `mapIfShiftToPbt` again (the MOR-2500 defect) | killed, 3 failed |
| b | reinstate the ±1200 clamp on both edges | killed in both files |
| c | drop the no-step refusal (`?? 50`) | killed, 2 failed |
| d | hardcode step 50 | killed, 2 failed |
| e | clamp each edge alone instead of the shift | killed in both files |
| f | resolve MAIN while SUB is active | killed, 1 failed |
| g | dispatch inner before the outer-null check | survived — unreachable: both raws come from the same lattice and are null only together |
| h | control: equivalent rewrite of the clamp | green, 59 + 142 |

**Enumerated, not read:** `mapIfShiftToPbt` over all 73×73 edge pairs at 3600/50 × 161 targets (857k cases) preserves width exactly, keeps both edges in ±1800 and on the grid, and reaches the optimal mean — no half-step drift. Hz↔raw round trips are the identity over 16 lattices. The native `if_shift` branch is byte-identical to `main`.

**Findings, all non-blocking; body edits applied before merge:**
1. `onPbtReset` no longer writes raw 128 where the modelled lattice has an even position count (CW 250 → 106, AM 9000 → 125). Now disclosed in the body together with a bench measurement made in response: the radio snaps those writes to 128, and the model's position count is wrong for odd width/step — a separate fix.
2. MOR-2500 is latent: no shipped surface reaches the PBT branch of `onIfShiftChange`. Body reworded.
3. `pbtHzToRaw` is orphaned by this change, not "still used". Body corrected; deletion rides with PR-W2.
4. The "56 Hz" figure in two in-diff comments and in commit `24e038de` mixes units: on the lattice the collapse is 900 → 100 Hz (raws 160/96 → 163/157). Corrected in the squash body; the two comments are a follow-up commit in PR-W2.
5. Outside the diff: `commands.svelte.ts: rawReceiverEchoCommandDescriptor`'s comment still says the PBT handlers call `pbtHzToRaw`. Follow-up in PR-W2.

**Not established by the review:** any hardware behaviour. (Coordinator's live IC-7610 evidence is in the earlier comment.)

PR #3523, opened 2026-09-17
fix(MOR-2504): teach the Mode enum and the IC-7610 profile PSK and PSK-R

morozsm:
## What

The IC-7610 can be put into PSK and PSK-R from its front panel, and its profile lists both modes, but the `Mode` enum had no member for CI-V mode bytes `0x12` / `0x13`.

- **Read path:** with the radio in PSK every mode frame failed to decode. The RX loop swallows that `ValueError` at debug level, so the state store kept the previous mode indefinitely and the UI showed a stale mode. The request/response path (`IcomRadio.get_mode_info`) let `ValueError: 18 is not a valid Mode` propagate to the caller — observed on the bench on 2026-09-17.
- **Profile:** `rigs/ic7610.toml` had no `[filters.width.PSK]` entry, so the mode had no filter table and no twin-PBT step, and the PBT controls stayed off in PSK although the radio has twin PBT there.

## Change

- `core/types.py: Mode` gains `PSK = 0x12` and `PSK_R = 0x13`.
- `rigs/ic7610.toml` gains `[filters.width.PSK]` and `[filters.width.PSK-R]`, mirroring the CW entry's shape, with `pbt_step_hz = 50`.

## How the values were established

- Mode bytes: IC-7610 CI-V Reference Guide, command `06`, data `12` = PSK, `13` = PSK-R.
- Filter table: IC-7610 Basic Manual, "Selecting the IF filter" — CW and PSK share one row (FIL1 1.2 kHz, FIL2 500 Hz, FIL3 250 Hz).
- Twin PBT in PSK: same manual, "Using the Twin PBT", headed "SSB, CW, RTTY, PSK and AM modes".
- 50 Hz step: measured on a real IC-7610 over LAN on 2026-09-17, PSK at FIL2 / 500 Hz, PBT Inner swept raw 0..255 and read back: 11 positions, raw 11..244, 50.0 Hz per step. PSK-R was **not** measured; its entry is an inference and the profile comment says so.

## Tests

Single files, run by the builder at the head: `tests/test_radio.py` 369 passed; `tests/test_rig_loader.py` 397 passed, 1 skipped; `tests/test_web_capability_guards.py` 64 passed; `tests/test_mode_label_coercion.py` 43 passed, 1 xfailed (X6200 `DIGI`). `ruff check`, `ruff format --check`, `lint-imports` clean.

`tests/test_mode_label_coercion.py` already walked every CoreRadio-routed profile's mode list through `IcomRadio._coerce_mode` and carried IC-7610 `PSK` / `PSK-R` as `xfail(strict=True)`; the first CI run on this PR failed on exactly those two XPASSes, which is that guard doing its job. The two entries are retired here. A walking test this PR had first added to `tests/test_rig_loader.py` duplicated that guard and was deleted.

Mutations (independent reviewer, one pinning file each, at `14f02baf`): removing `PSK` from the enum, or changing `0x12` to `0x11`, fails `test_set_mode_psk_emits_civ_psk_byte`; changing the PSK `pbt_step_hz` to 200, or deleting the `PSK-R` table, fails `test_psk_filter_config_mirrors_cw`.

## Not in this change

- rigctld still refuses Hamlib PSK/PSKR (`rigctld/contract.py` maps are untouched): passband semantics for PSK are not established.
- X6100 / X6200 profiles list a `DIGI` mode the enum cannot express. `tests/test_mode_label_coercion.py` keeps it as a strict xfail; it wants its own ticket.

Refs MOR-2504


morozsm:
Agent Review: PASS 14f02baf26072c9daec8b0ac300432a681a7cc17

Independent review by GLM-5.3 (author: Kimi K3), in a detached worktree at the exact head; tree left clean.

**Mutations, each run against one pinning file and restored:**

| Mutation | File | Result |
|---|---|---|
| remove `PSK = 0x12` | `tests/test_radio.py` | fails `test_set_mode_psk_emits_civ_psk_byte` |
| `0x12` → `0x11` | `tests/test_radio.py` | fails the same test |
| PSK `pbt_step_hz` 50 → 200 | `tests/test_rig_loader.py` | fails `test_psk_filter_config_mirrors_cw` |
| delete `[filters.width.PSK-R]` | `tests/test_rig_loader.py` | fails `test_psk_filter_config_mirrors_cw` |

**Verified:** PSK/PSK-R segments are identical to `[filters.width.CW]` and match the manual's shared CW/PSK row; `Mode.PSK_R` ↔ `"PSK-R"` takes the same paths as `CW_R` in both directions; rigctld still refuses PSK (`HAMLIB_MODE_MAP` has no key) before and after.

**Non-blocking, pre-existing, not introduced here:** `IcomRadio.set_mode` is gated by the enum only, never by the profile's mode list, so a hand-crafted web request of `"PSK"` against a rig whose profile does not list it now reaches the wire instead of failing with `Unknown mode`. `DV` has the identical exposure today against the IC-7610; the capabilities payload advertises profile modes only, so no shipped UI offers it.

The reviewer could not confirm the "PDF p.33 / p.34" citations from the excerpt it was given. Coordinator check: `pdftotext -f 33 -l 33` of `IC-7610_Basic_Manual_ENG_3.pdf` contains "Using the Twin PBT"; `-f 34 -l 34` contains "Selecting the IF filter".

morozsm:
Agent Review: PASS 220061114329a44c00c1a922bbc749e5065ffc08

Delta review of `14f02baf..22006111` (tests only, 2 files) by GLM-5.3; author Kimi K3. The full review at `14f02baf` is in the earlier comment.

- The deleted `test_every_civ_profile_mode_is_expressible_by_the_mode_enum` checked enum membership only; the surviving guard in `tests/test_mode_label_coercion.py` is stronger per label (it calls `IcomRadio._coerce_mode` and asserts the round trip). Mutation: with `PSK = 0x12` removed, `uv run pytest tests/test_mode_label_coercion.py` → `1 failed, 42 passed, 1 xfailed` (`[IC-7610:PSK]`); restored, tree clean.
- Dropped with the deleted test, both outside the survivor's documented scope: the X6100 labels (`backends/factory.py: create_radio` rejects X6100, so they cannot reach `_coerce_mode`) and a `checked >= 6` profile-count pin.
- Every sentence of the edited module docstring verified true at this head; `git grep` finds no remaining reference to the deleted test.

PR #3524, opened 2026-09-17
fix(MOR-2505): resolve the filter rule for reversed modes named with an underscore

morozsm:
## What is wrong

Live on an IC-7610 (`main` 66e1b7b9, 2026-09-17, RX only): in PSK-R `get_filter_width()` returned **9** — the raw CI-V width index — where PSK returned 500 Hz.

`profiles/__init__.py: RadioProfile.resolve_filter_rule` looks the mode up by the profile's hyphenated key (`"CW-R"`, `"RTTY-R"`, `"PSK-R"`), and its own CW-R → CW / RTTY-R → RTTY fallbacks compare against the hyphen form. Every Icom runtime caller passes `Mode.name`, which has an underscore (`"CW_R"`). No candidate matches, the rule is `None`, and the width is neither decoded nor encoded through the mode's table. So all three reversed modes are affected, on read and on write.

Callers, as established by the builder with `git grep` (file + symbol): `runtime/radio.py: IcomRadio.set_filter_width` and `get_filter_width`, `runtime/_civ_rx.py: _decode_filter_width`, `web/server.py` (FFT passband) — all pass the underscore form. `backends/yaesu_cat/radio.py` passes hyphenated Yaesu names and is unaffected. `web/radio_poller.py` already emits the hyphen form to the frontend; nothing changes there.

## Change

One normalisation at the single lookup point: `str(mode).upper().replace("_", "-")`. Callers untouched, no helper added.

## Tests

`tests/test_rig_ic7610.py`: 7 new tests on the real IC-7610 profile — `CW_R`, `RTTY_R`, `PSK_R` resolve to the same rules as the hyphen forms, and a width index 9 decodes to 500 Hz. Six of them are red at 66e1b7b9 (6 failed / 64 passed) and green with the fix (70 passed); the seventh, `test_hyphen_forms_still_resolve`, is a non-regression guard that passes both ways. Mutation, repeated by the reviewer: removing the normalisation → 6 failed. Single files run by the builder: `test_rig_ic7300.py` + `test_dsp_filter_family.py` + `test_mor1545_filter_width_dedupe_key.py` 185 passed; `test_rig_loader.py` 397 passed, 1 skipped. `ruff`, `lint-imports` clean.

Live acceptance is in the PR comments: PSK-R, CW-R and RTTY-R read back 500 Hz on the IC-7610.

Write side, established by the reviewer by calling the real `encode_filter_width` before and after: `CW_R` 500 Hz and `RTTY_R` 2400 Hz raised `CommandError: no filter-width mapping` before (nothing was sent) and now encode to the same byte the hyphen form always produced. No path sends a different byte than before.

Refs MOR-2505


morozsm:
Live acceptance at `abe8af803981b0c7f8c4c9da7cf8d800d9863ce9`, IC-7610 over LAN, RX only, 2026-09-17: `set_mode` then read back after 1.5 s — PSK-R, CW-R, RTTY-R and CW (control) each report FIL2, width **500**; at `main` 66e1b7b9 the same read in PSK-R returned 9. Radio restored and read back: USB, FIL1, 3600, PBT 128/128. No transmit.

morozsm:
Agent Review: PASS abe8af803981b0c7f8c4c9da7cf8d800d9863ce9

Independent review by Kimi K3 (author: GLM-5.3), detached worktree at the exact head, left clean.

- **No key can newly miss.** `grep -rn '^\[filters' rigs/` enumerates all 47 `[filters.width.*]` headers in the 9 shipped profiles: none contains an underscore, and no mode label does either (Yaesu labels are already hyphenated), so `replace("_", "-")` cannot break a lookup that matched before.
- **Write side: exception → correct byte, never byte → different byte.** Real `encode_filter_width` on `rigs/ic7610.toml`, pre-PR resolver vs head: `CW_R` 500 Hz `CommandError` → `b'\x09'`; `RTTY_R` 2400 Hz `CommandError` → `b'\x28'`; both equal to what the hyphen forms produced before. `USB` with data mode and all hyphen forms are byte-identical before and after. Cross-profile (ic705, ic9700, ic7300, x6200, ftx1): underscore and hyphen forms resolve identically.
- **Mutation:** normalisation removed, only `tests/test_rig_ic7610.py` run → 6 of the 7 new tests fail; `test_hyphen_forms_still_resolve` passes both ways, as a non-regression guard should. Restored, 70 passed.
- The added comment verified sentence by sentence against the callers (`runtime/radio.py`, `runtime/_civ_rx.py`).

Correction applied to the PR body on the reviewer's note: the PR adds 7 tests, not 6.

PR #3525, opened 2026-09-17
fix(MOR-2497): model the twin-PBT lattice the radio actually snaps to

morozsm:
## What is wrong

The twin-PBT lattice model merged in #3514 (`filter-controls.ts: measuredPbtRawToHz` / `measuredPbtHzToRaw`) assumed `width/step + 1` positions. It was derived from sweeps at 3600, 1800 and 500 Hz — all widths where that count is odd. The #3522 review noticed that where the count comes out even the model has no centre, and Reset writes raw 106 (CW 250) or 125 (AM 9000) instead of 128.

Measured on an IC-7610 over LAN on 2026-09-17, RX only, `set_pbt_inner(raw)` for every raw 0..255, each read back:

| mode / width | positions | reachable raws |
|---|---|---|
| USB 500 | 11 | 11 … 128 … 244 |
| USB 250 | **5** (model said 6) | 25, 76, 128, 179, 230 |
| USB 350 | **7** (model said 8) | 18, 54, 91, 128, 164, 201, 237 |
| USB 3600 | 73 | 1 … 128 … 254 |
| AM 6000, step 200 | 31 | 4, 12, 20 … 128 … 251 |

The radio always has a centre at raw 128. `P = 2·floor(width / (2·step)) + 1` and `raw(i) = floor((i + 0.5)·256 / P)` reproduce all five sweeps exactly; the merged model reproduced only USB 500 and USB 3600 (wrong count at 250/350, right count but wrong raws at AM 6000).

Operator-visible at `main`: with an odd width/step — the shipped CW FIL3 250 Hz and AM FIL1 9000 Hz among them — a centred PBT reads half a step off centre.

## Change

- The model inside `filter-controls.ts` is replaced by the measured one. Signatures unchanged; "nearest position, ties toward centre" kept. `width == step` gives the single position raw 128 / 0 Hz.
- `mapIfShiftToPbt` bounds the shift by the reachable span `floor(width / (2·step))·step` instead of `width/2`, so at 250 Hz an edge cannot be sent to an unreachable ±125 Hz. Pinned at 250.
- The prose block stating the old formula ("half-up is load-bearing") is deleted.

## Tests

New pins for the four new sweeps in `filter-controls.test.ts` (every measured raw ↔ Hz, raw 128 → 0 Hz at every width, between-position snap, P = 1, the 250 Hz shift bound): 9 red on the old model, 71/71 green. The existing 73-position sweep still passes.

Two old expectations changed, both in `panel-commands.intent.isolated.test.ts`: AM 9000 / step 200, write +150 Hz → raw 130 becomes **133** (MAIN and SUB). 130 came from the old 46-position grid; on the measured grid (P = 45, centre 128) +150 Hz snaps to +200 Hz, position 23. **AM 9000 itself was not swept on the radio — 133 is the formula's value, not a measurement** (AM 6000 was measured).

Single files run by the builder, green without edits: `filter-passband-adapter.isolated` 58, `scope-adapter.authority.isolated` 69, `scope-passband-display` 210, `audio-spectrum-pbt-range.isolated` 2, `audio-spectrum-renderer` 31, `radio-view-model-adapter` 164, `panel-commands.intent.isolated` 142. `npm run check`: 0 errors. eslint clean on the 3 files.

Mutations (builder, pinning file `filter-controls.test.ts`): P back to `width//step + 1` → 7 failed; `floor` → `round` → 11 failed; centre offset 0.5 → 0 → 19 failed.

3 files, 126 insertions, 47 deletions.

Refs MOR-2497


morozsm:
Two more sweeps after the PR was opened (IC-7610 over LAN, RX only, 2026-09-17; no width written, modes as found; script `measure-pbt-am9000-cw250-ic7610.py` in the coordinator's evidence directory, which computes the model in the PR and compares):

| case | width | positions | model P | every raw equals the model |
|---|---|---|---|---|
| AM FIL1 | 10000 (this radio's FIL1 is set to 10 kHz, not the 9 kHz default) | 51 | 51 | yes |
| CW FIL3 | 250 | 5 | 5 | yes — 25, 76, 128, 179, 230 |

That makes seven sweeps the model reproduces exactly (USB 250 / 350 / 500 / 3600, CW 250, AM 6000 / 10000). **AM 9000 is still not measured**, so the raw 133 expectation stays the formula's value. Radio restored and read back equal to the baseline (USB, FIL1, 3600, PBT 128/128); the script exits non-zero otherwise. No transmit.

morozsm:
Agent Review: PASS e9c83b716c13e161ed1cda5507b38c622d5e56fb

Independent review by GLM-5.3 (author: Kimi K3), detached worktree at the exact head, left clean. Everything below was measured by calling the real exported helpers from a scratch test file, deleted afterwards.

- **Ground truth.** All five bench sweeps map raw ↔ Hz exactly in both directions; every raw 0..255 reads as the nearest measured position. The pinned arrays in the test file are identical to the bench data. Exact ties exist (1 at 250 Hz, 37 at 3600 Hz); the radio data cannot tell the tie direction, the code resolves toward the centre and its comment says so.
- **Every width the shipped profiles can produce** (91: step 50 at 50–500 and 600–3600, step 200 at 200–10000): non-null, identity on positions, raw 128 ↔ 0 Hz, strictly increasing, within 0..255 — zero failures.
- **`mapIfShiftToPbt`**, exhaustive over lattice edge pairs at 250 / 350 / 500 / 3600 (1.18 M cases at 3600): no edge leaves ±floor(w/2s)·s, width always preserved.
- **AM 9000, +150 Hz → raw 133** recomputed by hand from the model: P = 45, position 23, floor(23.5·256/45) = 133.
- **Mutations** (only `filter-controls.test.ts`): shift bound back to width/2 → killed; tie-break flipped on the raw side → killed; on the Hz side → killed; `positions > 256` guard removed → killed.
- Callers: no behavioural `span = width/2` remains; Reset now writes raw 128 at every width.

**Non-blocking, for the PR-W2 sweep:** two comments outside this PR's files (`scope-passband-display.ts`, `audio-spectrum-renderer.ts`) still say the lattice's span is the current filter width; the `PBT_MEASURED_STEP_HZ` comment dates the front-panel `SFT +900` / `BW 1.8` reading to 2026-09-17 where MOR-2497 records it on 2026-09-16.

PR #3526, opened 2026-09-17
fix(MOR-1680): take the manual-notch feedback echo from the field the value uses

morozsm:
## What is wrong

On the FTX-1 the manual-notch position has two state fields. The periodic poll feeds `manualNotchFreq` (the only one in the acquisition policy, `rigs/ftx1.toml`), while `notchFilter` is filled only by the post-write readback (`core/command_service.py: _command_target`). The **value** the notch control shows already prefers `manualNotchFreq` when the profile publishes its display domain (`panel-props.ts: manualNotchReading`). The **confirmed echo** in the command-feedback strip did not: `panel-adapters.ts: getDspControlFeedback('notchFilter')` read the literal `main.notchFilter`.

Consequences, both reproduced as red tests before the fix: when the two fields disagree (the operator turned the knob on the radio), the confirmed value stays at the old `notchFilter`; and before the first web write the control is reported unavailable although the radio is reporting a position.

## Change

`manualNotchReading` is exported and returns which source field it chose; `getDspControlFeedback` builds its confirmed value from that same choice. One field choice, no second copy. With no domain published (every Icom profile) the path is unchanged — pinned by an Icom control case.

## Tests

`dsp-command-feedback.isolated.test.ts`: 3 new cases (fields disagree → the polled 1600 Hz; never written → available at 3200 Hz; Icom control → raw value as before); 2 red before the fix, 37/37 after. Neighbouring single files green: `notch-freq-domain` 12, `notch-nb-depth-control-domain` 9, `panel-props.test` 99, `semantic-dsp-wiring.component` 41. `npm run check` 0 errors; eslint clean on the 3 files. Mutation: dropping the source choice fails the 2 new FTX-1 cases, the Icom control stays green.

3 files, 98 insertions, 15 deletions. Not live-verified: the FTX-1 is not on the bench today; the bench check is on the FTX-1 list of MOR-1680.

Refs MOR-1680


morozsm:
Agent Review: PASS c1a39d701610f968cf0e3cb2f6d7a7652e28246f

Independent review by Kimi K3 (author: GLM-5.3), detached worktree at the exact head, left clean.

- **Units — the hypothesis I was asked to treat as blocking, refuted as a defect mechanism.** On a domain-publishing radio the feedback object is now mixed: `confirmed` is display Hz, `target` / `requestedTarget` are still raw codes (`panel-adapters.ts` spreads the descriptor). Every consumer was traced: no numeric comparison ever involves `target` (`continuous-scalar.svelte.ts` compares canonical / draft / lastDispatch, all display-unit; renderers position `view.displayed` only). The one visible effect is text: while an FTX-1 notch command is pending, `DspScalarHost.svelte: status()` can read "requested 160; confirmed 1600". Cosmetic; before this PR a raw `confirmed` sat inside a Hz domain and misplaced the thumb, which was worse. Follow-up recorded on MOR-1680.
- `DspScalarHost.svelte: domain()` consumes the published `notchFreqDomain` at this head; `[0,255]` is only the no-domain fallback, so the Hz value is neither clamped nor misrendered. (The builder's note that it is hardcoded was stale.)
- **Icom path identical:** with no domain, `manualNotchReading` returns `source: 'notchFilter'` and the very same frozen descriptor object is used.
- **Staleness:** a never-observed `manualNotchFreq` cannot produce an available control; stale-but-observed resolves to available, which is the uniform doctrine for every DSP control.
- **Mutations** (only `dsp-command-feedback.isolated.test.ts`, baseline 37/37): confirmed from `notchFilter` again → 2 failed; decoding skipped → 2 failed; Icom also taking `manualNotchFreq` → 1 failed. Each killed by exactly the intended case.
- Added prose verified, including "the poll feeds `manualNotchFreq`" against `backends/yaesu_cat/poller.py`. Outside the diff, now stale: the `SemanticRadioSurfaces.svelte` comment saying notch "stays wire-raw".

PR #3527, opened 2026-09-17
refactor(MOR-2497): move the v2 filter surface onto the measured PBT lattice and delete the ±1200 converters

morozsm:
Guardrail exception, file count only, granted by the coordinating session under the owner's delegation of 2026-09-03: 11 code + 0 regenerated baselines = 11 files. 380 insertions + 166 deletions = 546 changed lines, under the line ceiling. The eleventh file is `pbt-if-shift-terminal-outcome-retention.test.ts`, a pre-existing test whose fixture had no PBT lattice: CI `quick` failed on it at 5bb5e88b, and the honest fix is 12 lines in that file — there is no shared fixture among the other ten.

It is one unit because deleting the fabricated ±1200 Hz converters (`pbtRawToHz`, `pbtHzToRaw`, `pbtRange()`, `PBT_DEFAULTS`) cannot land before their last three readers move, and each reader brings its pinning test file; splitting would leave `main` with either dead converters or a reader with no scale.

## What is wrong

After #3518 / #3522 / #3525 the default skin reads and every surface writes twin PBT on the measured lattice, but the v2 / LCD surface still showed it on the fabricated proportional ±1200 Hz scale:

- `panel-props.ts: toFilterProps` converted `pbtInner` / `pbtOuter` with `pbtRawToHz` — at a 3600 Hz filter raw 254 read +1191 Hz where the radio is at +1800 Hz.
- `panel-adapters.ts: getIfShiftControlFeedback` derived the IF-shift feedback on the same scale.
- `FilterPanel.svelte` hardcoded the PBT slider bounds at ±1200 Hz step 25, and rendered Reset outside `{#if hasPbt}`, so Reset stayed on screen in FM, where the radio has no twin PBT.

## Change

- One derivation, `filter-controls.ts: measuredPbtDisplayDomain(width, step)`, next to the lattice helpers: min / max / step of one edge, `±floor(width / (2·step))·step`. Searched first for an existing one (`pbtLattice` — positions only; the `halfSpan` inside `mapIfShiftToPbt` — private): nothing returned a slider domain.
- `toFilterProps` and `getIfShiftControlFeedback` read off the lattice; `toFilterProps` emits the domain and `FilterPanel.svelte` takes its bounds and step from it. Reset moves inside `{#if hasPbt}`.
- The four fabricated-scale symbols are deleted with their tests. `PbtRange` / `pbtRangeFromCaps` stay: they still serve as the structural-availability gate in `panel-commands.ts`, `panel-adapters.ts`, `radio-view-model-adapter.ts`, `scope-passband-display.ts` and `toAudioSpectrumProps`.
- The `PBT_MEASURED_STEP_HZ` comment's date is corrected to the session MOR-2497 records (2026-09-16).

## Tests

Each behaviour red first, then green (10 commits). Single files run by the builder: `filter-controls.test.ts` 79 (8 new: bounds at USB 3600 ±1800/50, USB 500 ±250, CW 250 ±100, AM 6000 ±3000/200, no lattice in FM), `panel-props.test.ts` 105 (6 new), `FilterPanel.isolated.test.ts` 97 (3 new), `pbt-if-shift-command-feedback.isolated.test.ts` 18, `audio-spectrum-renderer.test.ts` 25 (6 removed with the deleted converter), `scope-passband-display.test.ts` 210; witnesses green: `panel-commands.intent` 142, `filter-passband-adapter` 58, `no-fabricated-defaults` 51, mor1561 14, mor1536 15, `audio-spectrum-pbt-range` 2. `npm run check` 0 errors, eslint clean.

Mutations (builder): bounds back to ±1200 → 6 failed in `filter-controls.test.ts`; step hardcoded 50 → 2 failed (AM 200); Reset moved outside the gate → 1 failed in `FilterPanel.isolated.test.ts`.

`main` was merged in (5bb5e88b) to resolve one import-line conflict with #3526 in `panel-adapters.ts` — both imports kept. After the merge, run by the coordinator: `npm run check` 0 errors; `dsp-command-feedback.isolated.test.ts` 37 passed; `pbt-if-shift-command-feedback.isolated.test.ts` 18 passed.

Disclosed by the builder: one edit (a rename inside the pbt-if-shift test) was made with a script from the shell instead of the edit tool; the result was read back.

## After the first review and the first CI run (5bb5e88b → head)

`quick` failed on one test the builder had never run: `pbt-if-shift-terminal-outcome-retention.test.ts` expected `failed` and got `unavailable`. Its fixture declared no mode, no filter width and no `pbtStepHz`, so the feedback failed closed before the terminal outcome could be reported. The fixture now carries a real lattice (USB, 3600, step 50); the `failed` assertion is untouched. The behaviour itself — one side failed, lattice present → phase `failed` — was already pinned green in `pbt-if-shift-command-feedback.isolated.test.ts`.

Every other test file that calls what changed was then run singly by the builder: `MobileRadioLayout.component.svelte.test.ts` 47, `AmberPanels.honesty-migration.isolated` 27, `AmberLabels.profile-data.isolated` 13, `lcd-availability.isolated` 4, `auxiliary-tune-authority.isolated` 4 — green.

The two mutations that survived the independent review are now pinned: PBT slider `step` hardcoded 25 → 2 failed in `FilterPanel.isolated.test.ts` (ArrowRight from centre: SSB +50, AM +200); `getIfShiftControlFeedback` ignoring `dataMode` → 1 failed in `pbt-if-shift-command-feedback.isolated.test.ts` (USB-D step 100 vs USB 50).

## Left for W2b

From the review: the private `pbtHzToRaw` in `frontend/tests/e2e/v2-ui-interactive.impl.ts` still encodes the retired formula; comments in 7 files name the deleted functions as live. Default-skin slider bounds (`semantic/radio-view-model.ts`, `radio-view-model-adapter.ts`, `FilterSurface.svelte`); the "56 Hz" comments in `panel-commands.ts` and its test; the `pbtHzToRaw` mention in `commands.svelte.ts`; "span is the CURRENT filter width" in `scope-passband-display.ts`, `audio-spectrum-renderer.ts`, `radio-view-model-adapter.ts`. Then the profiles' `display_min` / `display_max`.

Refs MOR-2497


morozsm:
Agent Review: PASS 5bb5e88bc09c7d5c2ca46921c108b2004f7a77ad

Independent review by Kimi K3 (author: GLM-5.3), detached worktree at the exact head, left clean. Measured through the real functions in scratch test files, deleted afterwards. The reviewer disclosed applying one mutation with `perl -pi` first, reverting it before any test ran, and redoing it with the edit tool.

- **Deleted symbols:** no live import or call of `pbtRawToHz` / `pbtHzToRaw` / `PBT_DEFAULTS` remains anywhere in the repo.
- **Read/write agreement on the v2 surface:** at 250/50, 500/50, 3600/50 and AM 6000/200 every lattice raw reads the bench Hz through `toFilterProps`, and `onPbtInnerChange` writes the same raw back; the emitted slider stops equal the lattice exactly and both ends are reachable (120 points).
- **No-lattice cases** (FM, legacy payload, width null/NaN/0/negative, caps/state null): no NaN, no crash. On a legacy payload `hasPbt` stays true and `FilterPanel` falls back to its explicit ±1200/25 constants while the write path dispatches nothing.
- **IF-shift feedback:** 254/128 at 3600 now gives 900 (the front-panel `SFT +900`), where the old scale gave 591; the v2 feedback and the default-skin view model agree on all six states tested.
- **Removed tests:** each of the six accounted for — either superseded by a lattice pin or pinned a fabricated number.
- **Mutations:** other receiver's width → killed (3); domain = width/2 → killed (2); **slider `step` hardcoded 25 → survived** (the panel tests pin min/max, not step); **`getIfShiftControlFeedback` ignoring `dataMode` → survived** (inert on every shipped profile, fixture has no `-D` split).

**Non-blocking, carried into W2b:** the two surviving mutations need pins; `frontend/tests/e2e/v2-ui-interactive.impl.ts` keeps a private `pbtHzToRaw` with the retired formula whose 300 Hz expectation (raw 160 ± 1) contradicts the write path at 3600 Hz (149) — not run by PR CI, it needs `RIGPLANE_V2_URL`; comments in 7 files still name the deleted functions as live (`commands.svelte.ts`, `radio-view-model-adapter.ts`, `semantic/radio-view-model.ts`, `filter-controls.ts: controlRawToDisplay` doc, `band-plan.ts`, `FilterSurface.svelte`, `audio-spectrum-renderer.ts`); `AudioSpectrumProps.pbtRange` doc; `min: -0` in the degenerate domain.

morozsm:
Agent Review: PASS 69ef5ed60fee960121fb11986da36d08f2df8c2a

Delta review of `5bb5e88b..69ef5ed6` (tests only, 3 files) by Kimi K3; author GLM-5.3. The full review at `5bb5e88b` is in the earlier comment. Edit tool only; tree clean.

- **The retention test is the same test.** Only its fixture changed (mode USB, width 3600, `pbtStepHz: 50`); the lifecycle and every assertion are untouched. It can still fail: making `panel-adapters.ts: mergeTerminalOutcome` prefer the better outcome → 1 failed (`'confirmed'` instead of `'failed'`).
- **Failing closed before the outcome hides nothing:** in a mode with no lattice every PBT handler returns before `dispatchRadioIntent` and the controls are not rendered, so a failed PBT command cannot originate there.
- **Both previously surviving mutations now die:** slider `step` hardcoded 25 → 2 failed in `FilterPanel.isolated.test.ts`; `dataMode` ignored → 1 failed in `pbt-if-shift-command-feedback.isolated.test.ts`. Both pins observe behaviour rather than restating a value.
- Added prose verified. Unmutated at head: FilterPanel 99/99, feedback 19/19, retention 1/1.

PR #3528, opened 2026-09-17
fix(MOR-2497): bound the default skin's PBT sliders by the measured lattice

morozsm:
Guardrail exception, file count only, granted by the coordinating session under the owner's delegation of 2026-09-03: 11 code + 0 regenerated baselines = 11 files. 257 insertions + 78 deletions = 335 changed lines. The eleventh file is `freshness-invisibility.component.test.ts`, a pre-existing test that pinned the old `step="25"` on the PBT slider; it is the blast radius of this change, not new scope.

It is one unit: the default-skin slider domain, plus the prose and the one e2e helper that three independent reviews of #3522 / #3525 / #3527 asked to be corrected in the files this change already touches.

## What is wrong

After #3527 the v2 / LCD panel bounds its PBT sliders by the measured lattice, but the default skin's sliders (`semantic/FilterSurface.svelte`) were still ±1200 Hz step 25: at a 3600 Hz filter the operator could not drag to the ±1800 Hz the writers reach, and at 500 Hz the slider offered ±1200 where the radio has ±250.

## Change

- `radio-view-model-adapter.ts: deriveFilterPassband` emits `pbtDomain` from the same `measuredPbtDisplayDomain(width, step)` call `toFilterProps` uses — no second copy of the formula. `FilterPassbandViewModel.pbtDomain?` with a validator modelled on `ifShiftDomain`. `FilterSurface.svelte: passbandLimits` takes PBT min / max / step from it; the component computes nothing.
- No lattice (FM — fields non-structural; a legacy payload with no step; a width that is not a multiple of the step): the key is absent and the component keeps its row constants `{min: -1200, max: 1200, step: 25}`, exactly `FilterPanel`'s fallback after #3527.
- `measuredPbtDisplayDomain` returns `min: 0`, not `-0`, when width equals step.
- `frontend/tests/e2e/v2-ui-interactive.impl.ts`: the private `pbtHzToRaw` that re-implemented the retired ±1200 formula is deleted. The spec now verifies a round trip — after the commands settle, the PBT Inner slider's `aria-valuenow` must read the requested Hz. (Importing the real helper is not possible in the Playwright runtime: `filter-controls` transitively pulls a runes module.) Not run by PR CI — it needs `RIGPLANE_V2_URL`; typechecked.
- Prose: "56 Hz" → the lattice figure (900 → 100 Hz, raws 163/157) in `panel-commands.ts` and its intent test; comments naming the deleted `pbtRawToHz` / `pbtHzToRaw` as live, and "span is the CURRENT filter width", corrected or deleted in the files this PR touches.

## Tests

Red first: `filter-controls.test.ts` 79/80 → 80; `filter-passband-adapter.isolated.test.ts` 61/66 → 66; `FilterSurface.test.ts` 98/101 → 101. The builder then ran, one file at a time, every test file that calls a changed symbol (from `git grep`): 30 files, 1715 tests, green. `npm run check` 0 errors (includes the e2e typecheck); eslint clean on the changed `src` files.

Mutations (builder): adapter emits ±1200 again → 6 failed; `FilterSurface` step hardcoded 25 → 4 failed; domain taken from the other receiver's width → 6 failed.

## Left over (would have been files 12+)

Comment-only: `audio-spectrum-renderer.ts` (dead name, span), `scope-passband-display.ts` (span), `stores/commands.svelte.ts` (`pbtHzToRaw` in a doc comment), `props/panel-props.ts` (`AudioSpectrumProps.pbtRange` doc), `lib/radio/band-plan.ts` (dead name). Then the profiles' `display_min` / `display_max`.

Refs MOR-2497


morozsm:
Agent Review: PASS 0d44bb1516a0f143c4b506283e9f04745d536d56

Independent review by Kimi K3 (author: GLM-5.3), detached worktree at the exact head, left clean; measured through the real functions and a mounted `FilterSurface` in scratch test files, deleted afterwards.

- **Read/write agreement on the default skin** at CW 250, CW 500, USB 3600 and AM 6000: the emitted `pbtDomain` stops equal the lattice Hz list exactly (5 / 11 / 73 / 31), both ends are reachable, and every stop survives slider → `onPbtInnerChange` → raw → adapter read unchanged.
- **Receiver:** domain, value and writer all resolve the same active receiver (MAIN 3600 / SUB 500 probe: SUB gives ±250 and a write with `receiver: 1`).
- **No write without a gesture — the case asked to be treated as blocking.** Mounted with value 1800 under a ±250 domain, and with the domain shrunk ±1800 → ±250 under a live reading: no handler call in either case; `changePassband` fires only from `oninput`. On a 3600 → 500 width change the adapter re-derives the raw on the new lattice in the same update, so the reading cannot leave the domain.
- **Validator:** shape-only, like `ifShiftDomain` (it would accept min > max or step 0), and a rejected domain throws in dev rather than falling back; unreachable from the sole producer. Non-blocking.
- **No lattice** (FM, legacy payload, width not a multiple of the step): no key; the default skin shows its `pbt-unknown` marker rather than a dead slider.
- **e2e rewrite is meaningful:** `aria-valuenow` on that control is the confirmed reading, never the local draft, so the old ±1200 write path would fail it. Caveat, non-blocking: the target is captured right after the click without a settle wait.

PR #3530, opened 2026-09-17
fix(MOR-2508): emit the profile's hyphen mode labels in the public web state

morozsm:
## What is wrong

Owner, bench IC-7610, 2026-09-17: with the radio in PSK-R no button on the mode panel lights, and the status badge and the CW panel's `RX MODE` read `PSK_R`.

Path, established before the fix: `runtime/_civ_rx.py` stores `Mode(...).name` in the state store — `"CW_R"`, `"RTTY_R"`, `"PSK_R"` — and `web/runtime_helpers.py: _build_public_state_payload_from_dict` passes it through to both `/api/v1/state` and the WebSocket state envelope. The frontend compares with strict equality against the capability labels, which are the profile's hyphen forms (`ModePanel.svelte`: `active={currentMode === mode}`), and shows `main.mode` verbatim in the badge and in `CwKeyerSurface.svelte`. So it is one defect for **all** reversed modes — CW-R shows `CW_R` and lights no button today too.

`docs/api/web.md` documents the wire vocabulary with the hyphen form (capabilities `"modes": [… "CW-R", "RTTY-R"]`); the underscore table in `docs/api/types.md` is the Python enum. The only path that already emitted the hyphen form was the band-stack recall in `web/radio_poller.py`.

## Change

One normalisation where the public state payload is finalised, covering `main` / `sub` `mode` and the VFO-slot modes. The store keeps the enum name internally. The way back is already correct: `IcomRadio._coerce_mode("PSK-R")` → `Mode.PSK_R`. No frontend change is needed; the mode panel already offers PSK / PSK-R (`ModePanel.isolated.test.ts`).

## Tests

New `tests/test_web_state_mode_wire_labels.py`, 6 cases, all red before the fix. Run singly, green: `test_state_pipeline_contracts` 203, `test_web_runtime_helpers` 65, `test_civ_rx_coverage` 429, `test_field_status_frontend_fixture` 4, `test_mor1408_registry_projection` 6, `test_radio_poller_coverage` 262, `test_yaesu_web_projection` 42, `web/test_state_schema_conformance` 26, `test_mode_label_coercion` 43 + 1 xfailed, `test_rig_ic7610` 70, `test_radio` 369. `ruff`, `mypy --strict src/rigplane/web`, `lint-imports` clean. Mutation: normalisation removed → 6 failed.

2 files, 112 insertions.

Refs MOR-2508


morozsm:
Agent Review: PASS 4943054ecca3a7f8bf47a4aabbf14df743367d86

Independent review by the internal `verifier` (Opus); author GLM-5.3. Chosen over an OpenCode reviewer because the change alters a string in the public web state. Throwaway detached worktree at the exact head, removed afterwards.

- **Whole input space, not a sample:** the base and head `web/runtime_helpers.py` were loaded side by side and every leaf of both payload producers diffed over all 12 `Mode` members with every slot populated on both receivers: exactly 42 differences, all `CW_R|RTTY_R|PSK_R` → `CW-R|RTTY-R|PSK-R`; the other 9 modes are byte-identical. A leak scan of the head payload with `CW_R` injected everywhere finds 0 underscore leaves; the same scan on the base finds 8 and 6, so the scan can report hits.
- **Vocabulary:** every mode label of every `rigs/*.toml` enumerated — none contains `_`, so the blanket replace cannot corrupt a label. Only `main` / `sub` and their `vfo_a` / `vfo_b` / `unselected_vfo` `mode` keys are touched; `scope_controls.mode` is not.
- **Producers:** GET `/api/v1/state`, the WS full state, delta and baseline all funnel through the fixed builder. `mode_changed` events echo the client's own string or already use the hyphen form, and the frontend has no consumer for them. rigctld never imports `rigplane.web` and reads the store, which this diff cannot change.
- **Mutations** (only the new test file): normalisation deleted → 6 failed; SUB dropped → 3 failed; VFO slots dropped → 3 failed; behaviour-identical refactor → 6 passed. The tests call the real builders.
- **Round trip:** `IcomRadio._coerce_mode` maps every new wire label back to its enum member; pinned in `tests/test_radio.py` and `tests/test_mode_label_coercion.py`.

**Non-blocking, same class outside this ticket's surface:** `rigplane mode --json` on the CLI still prints `CW_R` (`IcomRadio.get_mode` returns `Mode.name`).

**The reviewer could not check rigplane-pro.** Coordinator, on the local checkout of `rigplane-pro` (HEAD `900e180`, a feature branch — not necessarily its `main`): `git grep -E "\"(CW|RTTY|PSK)_R\"|'(CW|RTTY|PSK)_R'"` outside tests and Markdown finds nothing.

PR #3531, opened 2026-09-17
fix(MOR-2507): a focused slider or radiogroup owns its arrow keys

morozsm:
## What is wrong

Owner, bench acceptance 2026-09-17 (IC-7610 and IC-7300, default skin): while tuning with the keyboard arrow keys the AF volume sometimes rises.

One arrow press was handled twice. `components-v2/layout/KeyboardHandler.svelte` listens on `window` and its only guard, `keyboard-map.ts: shouldIgnoreEvent`, recognised the native tags `INPUT` / `TEXTAREA` / `SELECT`. The value controls are not native inputs: every renderer (HBar, Knob, DualParam, Bipolar, Discrete, ProfessionalKnob) is a `div role="slider" tabindex=0` with its own `onkeydown`. With one of them focused the global shortcut tuned **and** the slider moved — and the same applies to sliders that write to the radio (RF gain, PBT, notch).

The builder then enumerated the class (`git grep` for arrow handlers in `frontend/src`) and found the same shape in `SegmentedButton` and `ActiveReceiverToggle` (`role="radiogroup"` / `role="radio"`: an arrow moved the segment and also changed the tuning step) and in the spectrum split divider (`role="separator"`, masked until now by its own `stopPropagation`).

## Change

One guard, in the place the global handler already decides. It yields an **unmodified** arrow key when the focused element — or an ancestor of it — declares that it consumes that arrow through a `data-owns-arrows` hook (`horizontal` / `vertical` / `both`, plus `shift` where Shift+Arrow is the fine step). The hook is set by the widgets that actually handle arrows: the shared value-control mount (`wheel-control.ts`, all six slider renderers — `both shift`), `SegmentedButton` and `ActiveReceiverToggle` (`both`), the spectrum split divider (`vertical`). A first version of this PR gated on ARIA roles instead; the independent review showed that nine plain `role="radio"` buttons handle no arrows and would have lost tuning, and that Ctrl+Arrow (AF level) and Ctrl+Shift+Arrow (RF gain) from `rigs/_keyboard-default.toml` were swallowed. Both are now pinned on the real `AttenuatorControl` and the real bindings. The second review pass found that a value control that is not editable still declared the hook (reachable: `RadioLayout.svelte: toggleTxSettings` focuses the first slider of a TX popover programmatically); the declaration is now set reactively by `wheel-control.ts` only while the control is editable, pinned on a real `ValueControl` mounted disabled and across the editable → non-editable transition. The rule — a focused control that consumes arrows owns them — is what keyboard-only operation (MOR-1413 scenario 9) needs, and `docs/guide/web-ui.md` already states the analogous rule for PTT and Space / Enter; the rule is now stated there for arrows too.

Deliberately unchanged: a focused **button** still lets arrows tune (pinned — otherwise the guard would silently grow into "ignore everything once anything is focused" and arrow tuning would die after any click); the frequency readout (`role="group"`) keeps global tuning when no digit is selected, as the MOR-2364 tests require. Home / End are consumed by these widgets but bound by no global shortcut; PageUp / PageDown are bound globally but consumed by no widget — nothing to do for either.

## Tests

Red first, three times. `KeyboardHandler.test.ts` 51 passed, `keyboard-map.test.ts` 46 passed; run singly and green: `ValueControl.controlled` 94, `ValueControl.test` 48, `DualParamRenderer.controlled` 7, `SegmentedButton` 45, `ActiveReceiverToggle` 28, `AttenuatorControl` 4, `SpectrumPanel.component` 157, `semantic-desktop-migration.component` 159, `MobileRadioLayout.component.svelte` 47, `tab-traversal.isolated` 5. `npm run check` 0 errors; eslint clean. Mutations: role proxy instead of the hook → 5 failed (including the plain-radio pin); ctrl/alt/meta check removed → 1 failed per file; unconditional declaration → 2 failed (disabled and transition pins). `KeyboardHandler.test.ts` 53, `keyboard-map.test.ts` 46.

The frequency readout (`role="group"`, arrows only with a digit selected) is outside the declared class and still double-fires there; pre-existing, its own ticket.

Pre-existing, identical on `main`, filed as MOR-2512: the widgets themselves step on Ctrl/Alt/Meta+Arrow that the global map owns; the frequency readout double-fires with a digit selected.

9 files, 551 insertions, 2 deletions.

Refs MOR-2507


morozsm:
Agent Review: BLOCKED 997242e4cbcd59aa5ee45bcabeae44256fa13879

Independent review by the internal `verifier` (Opus); author GLM-5.3. The fix is right for the reported defect (a focused `role="slider"` no longer double-handles arrows; the pin dies under mutation), but the selector is wider than the behaviour it assumes. The widening to `radiogroup` / `radio` was the coordinator's instruction, not the builder's idea.

**Blocking.**
1. **`role="radio"` / `role="radiogroup"` is not evidence that arrows are consumed** (`keyboard-map.ts: ARROW_KEY_OWNER_SELECTOR`). Only `SegmentedButton` and `ActiveReceiverToggle` handle arrows; nine other radiogroups (`AttenuatorControl` ×2, `AudioRoutingControl`, `LcdContrastControl`, `LcdDisplayModeControl`, `RfFrontEndInstrumentHost`, `RxAudioInstrumentHost` ×3, `AntennaInstrumentHost`, `CwKeyerSurface` ×2, `ScopeControlsSurface` ×2) are plain buttons with `onclick`. With one focused (Tab reaches them; `RadioLayout.svelte` focuses them programmatically) all four arrows now do nothing: reproduced on the real `AttenuatorControl` — control changes 0, global actions 0. Before this PR they tuned. Same shape, narrower: `SpectrumPanel: handleSplitKeydown` handles only Up/Down, the guard yields all four on `role="separator"`.
2. **Modifier-blind** (`keyboard-map.ts: focusedElementOwnsArrowKey` tests `ARROW_KEYS.has(key)` only). `rigs/_keyboard-default.toml` binds Ctrl+Arrow → `adjust_af_level` and Ctrl+Shift+Arrow → `adjust_rf_gain`; with a `role="radio"` focused both vanish (baseline `[adjust_af_level, adjust_rf_gain]` → `[]`), and on a focused slider Ctrl+ArrowUp moves the slider instead (`value-control-core.ts: handleKeyboardStep` ignores ctrl/alt/meta).
3. **The added paragraph in `docs/guide/web-ui.md` claims wider than the code:** "a segmented (radiogroup) control … the arrow keys act on that control" is false for the nine sites; "an adjustable separator" is false for Left/Right; the lead-in is falsified by the frequency readout below.

**Non-blocking.** The frequency readout (`frequency-interaction.svelte.ts: handleKeyDown`) double-fires today when a digit is selected — pre-existing, not a regression; a blanket `role="group"` would redden three MOR-2364 pins, but those cover only the no-digit path. The suite cannot tell `document.activeElement` from a safe `event.target` variant, nor a selector that adds `[role="button"]` (used in production in four files). `docs/guide/web-ui.md` still lists `Tab → Swap VFO`, which no profile binds and `KeyboardHandler` refuses before binding resolution — one row above the new paragraph. There is no focus-release affordance (Escape is `rit_xit_clear`).

**Required for a new head:** gate on what actually consumes arrows (an explicit hook set by the two arrow-handling components, or handlers added to the nine), yield unmodified arrows only, pin `role="button"` and the focus source, delete or narrow the doc paragraph and the stale Tab row.

morozsm:
Agent Review: BLOCKED 058c6193672d25adddecb1b706d632e426cae9d6

Second pass by the same `verifier` (Opus). Round 1's three blocks are refuted by reproduction: the role proxy is gone and a plain `role="radio"` (real `AttenuatorControl`) tunes again; the Ctrl+Arrow and Ctrl+Shift+Arrow bindings from `rigs/_keyboard-default.toml` fire again; the false `Tab` row is deleted. Six mutations die, an equivalent rewrite stays green. Two new findings block, both in the diff.

**Blocking.**
1. **A non-editable value control still declares `data-owns-arrows`** — `wheel-control.ts: wheelControl` sets it unconditionally, without `view.editable`. Measured on a real `ValueControl` mounted `disabled: true`: `tabindex=-1`, `aria-disabled=true`, hook `both shift` → four arrows do nothing and tune nothing. Reachable: `RadioLayout.svelte: toggleTxSettings` ends with `querySelector('[role="slider"], input, button')?.focus()`, which parks focus on such a slider when the first value control of a TX popover is not yet editable. Same class as round 1 (a declaration asserting consumption where the widget does not consume). Fix on either side — omit the declaration while not editable, or have `focusedElementOwnsArrowKey` refuse a declaration on an `[aria-disabled="true"]` element — and pin it.
2. **`docs/guide/web-ui.md` names "the segmented receiver/mode controls"** as declaring the hook. Receiver: true (`ActiveReceiverToggle`). Mode: false — `SegmentedButton` is imported only by the component gallery (`ControlButtonDemo.svelte`); every mode / AGC / NR / preamp / attenuator control in the cockpit is hand-rolled `role="radiogroup"` markup with no hook, and after fix 1 they correctly keep tuning. Narrow (user-facing guide): name the active-receiver toggle, drop "mode"; and state that the splitter owns Up/Down only, as the PR's own test pins.

**Non-blocking, pre-existing (identical on `main`), for a follow-up ticket:** the widgets themselves do not honour "modified arrows stay global": with the AF value control focused, Ctrl+ArrowUp both steps the slider and fires `adjust_af_level` (`value-control-core.ts: handleKeyboardStep` and the five `lease.key` call sites ignore ctrl/alt/meta; same in `SegmentedButton`, `ActiveReceiverToggle`, `SpectrumPanel: handleSplitKeydown`). The frequency readout (`frequency-interaction.svelte.ts: handleKeyDown`) still double-fires with a digit selected. Hook parsing fails closed on every malformed value (typo keeps global tuning). No container-with-plain-button-child instance exists in the tree.

morozsm:
Agent Review: PASS 56ac4effb6f2e663f798586574b3d032fb3f57f3

Third pass by the same `verifier` (Opus); author GLM-5.3. The two blocks of the second pass are fixed, established against the real components rather than the builder's own pin — which matters: that pin calls `action.update()` by hand, so it proves the helper works when invoked, not that Svelte invokes it.

- **Non-editable value control:** real `ValueControl` mounted `disabled: true`, slider focused → hook absent, `aria-disabled=true`, arrows tune (`['tune','tune']`), control unchanged. The transition driven by a reactive prop on a real mount, focus never leaving the slider: editable → non-editable → editable, the declaration disappears and returns synchronously after the flush; repeated on the hbar, bipolar, knob and discrete renderers. `update(next)` assigns before it syncs; the reversed order is pinned (mutation dies). A null view can never carry a declaration: two renderers gate the element behind `{#if view != null}`, the other four hold a non-nullable view.
- **Docs paragraph** re-read claim by claim against the code and `rigs/_keyboard-default.toml` — every sentence true; "mode" gone; splitter stated as Up/Down only.
- **Mutations** (only `KeyboardHandler.test.ts`, baseline 53): declaration unconditional → 2 failed; sync on mount only → 1 failed; sync before assignment → 1 failed; equivalent rewrite → green.

**Carried forward, MOR-2512 (pre-existing, unchanged by any head of this PR):** widgets step on Ctrl/Alt/Meta+Arrow that the global map owns; the frequency readout double-fires with a digit selected. **Coverage note:** the transition pin exercises the action directly; a regression in Svelte's propagation of the editability flip would not be caught by the suite.

PR #3532, opened 2026-09-17
feat(MOR-1687): wire command feedback into PBT inner/outer and IF shift on both surfaces

morozsm:
Guardrail exception, file count only, granted by the coordinating session under the owner's delegation of 2026-09-03: 12 code + 4 regenerated baselines = 16 files. 776 insertions + 46 deletions = 822 changed lines (PNGs count no lines), under the line ceiling. The eleventh and twelfth files are pre-existing substitutes of `panel-adapters` that lacked the three new producers and made `quick` red: the literal mock in `semantic-vfo-indicator-wiring.component.test.ts` and the fixture-capture stub module `frontend/fixtures/stubs/panel-adapters.ts`; three stub lines each.

It is one unit of work because the three passband controls share one feedback producer family and one wiring seam on each surface; wiring one control without the others would leave the seam half-typed. Of the 12 code files, 5 are production (`panel-adapters.ts`, `FilterSurface.svelte`, `FilterPanel.svelte`, `SemanticRadioSurfaces.svelte`, `fixtures/stubs/panel-adapters.ts`) and 7 are tests or test fixtures, each pinning one of the two surfaces. The 4 PNGs are the LCD cockpit baselines (`unified-instrument` and `panadapter-first`, desktop and 1100×800): the v2 panel no longer substitutes 0 Hz for a PBT reading the server has not published, and those fixtures publish none, so their PBT rows now show the shared bipolar renderer's unknown state. The PNGs are the `-actual.png` images of `visual` run 35292148408 at 63403357, byte for byte. Against the previous baselines, counting pixels with any channel differing by more than 30/255: 4266–4608 per image, of which 3317 (identical in all four) are the two PBT rows and 949–1265 are the LCD header (`MAIN A ●` → `MAIN ●`, band chip `?` → `20m`, a `⏻ UNKNOWN` power chip). The header part predates this branch: captured locally with the spec's own procedure, the base `3d81dff3` and this head render the header pixel-identically (0 differing pixels above the S-meter strip, three captures each), and under the comparator's own metric (`threshold: 0.2`, YIQ) the header alone is 213–349 px against an allowance of 1024 (`maxDiffPixelRatio: 0.001`), which is why #3533's run passed these four tests against the old baselines and this branch's did not.

## What is wrong

The beta known-limitations register lists it (MOR-1691): "PBT Inner/Outer lack gesture-local draft and pending feedback. Numbers follow delayed readback rather than the thumb." The producers `panel-adapters.ts: getPbtInnerControlFeedback / getPbtOuterControlFeedback / getIfShiftControlFeedback` existed and were tested, but nothing called them (MOR-1687 scoping, 2026-09-17).

## Change

Both surfaces now wire command feedback (pending / confirmed / failed / unavailable) into PBT Inner, PBT Outer and IF shift, the way the DSP scalars already do:

- default skin: `SemanticRadioSurfaces.svelte` passes the three feedbacks into `semantic/FilterSurface.svelte`; while the command is in flight the thumb shows the feedback's `target` (Hz, marked by `data-command-phase`), the canonical readout stays on the confirmed value, and on failure the thumb returns to it. Display priority: draft → pending target → view value, so the MOR-1706 continuity fence still owns the value when no gesture or command is live;
- v2 / LCD: `FilterPanel.svelte` shows the requested value as a `PENDING <target> Hz` chip next to the canonical readout — the same idiom that panel already uses for table-mode Width — while `aria-valuenow` stays on the confirmed value. On a held (stale) reading the v2 controls stay live and dispatch through the reading, as the default skin does (MOR-2425 / R40); the two surfaces are pinned to agree on that state.

The retention pins mint a **new** lifecycle id at the gesture, as production does (`radio-intents.ts: dispatchRadioIntentWithResult` → `makeCommandId()`); the first version of this PR held the id constant and its pins passed with the surface reverted to `main` — found by the independent review.

**One unit:** PBT feedback is converted from raw on the measured lattice in one place next to the producers, so `confirmed`, `target` and `requestedTarget` are all in Hz — the mixed-unit object flagged in today's notch review is pinned against (`confirmed !== 131` for raw 131). In a mode without a lattice (FM) the feedback is unavailable, never NaN. No new abstraction: the existing continuous-scalar model (canonical / draft / displayed) carries it.

## Tests

Red first: `pbt-if-shift-command-feedback.isolated` 25, `FilterSurface.test` 116, `FilterPanel.isolated` 106. Every substitute the wiring is compiled or mounted against was swept — the 32 `vi.mock` files, the fixture-capture stub module, and the runtime stubs of the component tests — and run singly; the second review pass found the last two (`fixture-capture-root-cwd.isolated` and `cw-keyer-zone-lifecycle.component`, the latter because the three producers were called without the wiring's `controlSession`, unlike their neighbours); both green at 63403357, where `quick` is green. `visual` was red at 63403357 on exactly the four LCD baselines above, for the reason given there; the head that re-pins them needs its own run. `npm run check` 0 errors; eslint clean.

Mutations (builder): `FilterPanel.svelte` reverted to `main` → 6 failed (all new v2 pins incl. the pending chip); pending display removed from `FilterSurface` → 6 failed; stale-reading fallback removed → 1 failed. Verified by the review at the first head and unchanged since: units all Hz through the real producer; no write without a gesture across receiver switch, width change, FM and failure; on a lattice-neighbour snap the surface ends on the confirmed value.

## Not pinned, disclosed

If `SemanticRadioSurfaces` stopped passing the three new props, no test would fail today (confirmed by the review under mutation): component and adapter pins exist, a wiring pin does not — part 2's composition test. The builder disclosed one expectation edit made with a shell script instead of the edit tool; the result was read back.

## Known follow-up from the mechanism audit

`panel-adapters.ts: pbtFeedbackOnMeasuredLattice` (new) and `projectDspControlFeedbackToDisplay` (existing) both project a feedback's three value fields into display units; the one reachable divergence is a target raw that does not convert (only from a foreign command — `panel-commands.ts` emits on-lattice raws). One core projector with two facades is the consolidation; recorded here so the pair is not rediscovered.

## Part 2

i18n of the "PENDING" label; `control-feedback-composition.component.test.ts` (with the wiring pin above); polite announcements for the three passband rows; `breakInDelayValueText` in `CwKeyerSurface.svelte`. Filter-shape and Reset feedback are out of beta scope.

Refs MOR-1687


morozsm:
Agent Review: BLOCKED 05e17f269ff0af9aa04eb7518d59f67ee93ebdc3

Independent review by the internal `verifier` (Opus); author GLM-5.3. Throwaway worktree at the exact head, removed. Four blocking findings, all in the diff.

1. **The head is red.** `semantic-vfo-indicator-wiring.component.test.ts` mocks `panel-adapters` with a literal factory and lacks the three new producer stubs, so mounting `SemanticRadioSurfaces` throws — `quick` failed at this head on exactly that, reproduced locally. The diff patched this shape in `mor1536-armed-adoption.test.ts` and stopped; the reviewer swept all 42 files mocking `panel-adapters` and this is the whole remaining class.
2. **The headline behaviour does not happen in production, and the pins cannot tell.** Both retention pins hold `feedback.lifecycleId` constant across the gesture; production mints a new command id per dispatch (`radio-intents.ts: dispatchRadioIntentWithResult` → `makeCommandId()`), and `panel-adapters.ts` builds the lifecycle id from it, so it always changes — and then `continuous-scalar.svelte.ts: reconcile` retires the draft at once for the `native-range` / `bipolar` policies the new rows use. Measured on the real components with a fresh id: semantic `FilterSurface` shows the stale readback (25) instead of the requested 600; v2 `FilterPanel` reads "0 Hz" right after a +50 gesture. `FilterPanel.isolated.test.ts`'s retention pin **passes with `FilterPanel.svelte` reverted to `main`**. Either implement the pending display from the feedback's `target` (no surface reads it today), or drop the claim and the two pins; either way the pins must mint a new lifecycle id at the gesture.
3. **v2 / LCD regression:** `FilterPanel.svelte` now gates PBT Inner / Outer / IF shift on `feedback.availability === 'available'`; `getReceiverEchoControlFeedback` requires a `current` observation, so on a held (stale) reading the controls go dead — on `main` they were live and dispatching. `FilterSurface.svelte` keeps the MOR-2425 / R40 reading fallback on the same state, so the two surfaces now disagree. Give `FilterPanel` the same fallback or record the exception.
4. **False prose:** `FilterSurface.svelte` says the fallback is "the same fallback `RfFrontEnd.svelte`'s `rfGainInput` established"; that branch is a disabled unknown triggered on `feedback === null`, it does not keep commanding. Delete the clause.

**Verified and holding:** units — `confirmed`, `target`, `requestedTarget` all in Hz through the real producer (raw 254 → 1800 at 3600/50; SUB raw 132 → 50 with receiver 1); no write without a gesture across receiver switch, 3600 → 500 width change, FM and command failure (dispatch count stays at the gesture's 1); on a lattice-neighbour snap the surface ends on the confirmed value; the freshness-invisibility loosening did not blind that test (a painted leak still reddens it); the mor1536 stubs are mechanically required.

**Body corrections owed:** "17 files, all green" is false at this head; the retention sentences overstate what the pins establish; the v2 held-reading change is unmentioned.

morozsm:
Agent Review: BLOCKED 93be4e8ebd7149437a196fff680a14816aacc2eb

Second pass by the same `verifier` (Opus). The three behavioural blocks of the first pass are fixed and were reproduced on the real components: with a **fresh** lifecycle id the semantic thumb holds the requested 600 while a stale 25 arrives and the `<output>` stays canonical, then ends on the confirmed 550; the v2 panel shows the `PENDING 50 Hz` chip with `aria-valuenow` on the confirmed value and clears it on confirm; a held reading keeps both surfaces enabled and dispatching once per gesture; the false `rfGainInput` clause is gone. `FilterPanel.svelte` reverted to `main` now fails 6 pins including the retention one that survived before. No write without a gesture across receiver switch, width change, FM and failure; a null or off-domain pending target produces no NaN and no dispatch.

**Blocking — the substitute class was swept too narrowly ("files that `vi.mock` `panel-adapters`"); `quick` is red at this head on exactly these two files, and reverting only `SemanticRadioSurfaces.svelte` turns both green:**
1. `frontend/fixtures/stubs/panel-adapters.ts` — the non-test stub the fixture-capture build compiles the wiring against — lacks the three new producers: `fixture-capture-root-cwd.isolated.test.ts` fails with `"getIfShiftControlFeedback" is not exported`. Add the three exports in the module's existing `unavailable` shape.
2. `SemanticRadioSurfaces.svelte` calls the three producers with **no argument**, unlike its neighbours `getCwPitchControlFeedback(controlSession)` two lines below; `getIfShiftControlFeedback` then falls through to `runtime.controlSession` and reads `session.epoch` unguarded — `cw-keyer-zone-lifecycle.component.test.ts` fails 3 tests with `Cannot read properties of undefined (reading 'epoch')`. Pass `controlSession` at the three call sites.

The file count will rise to 12; the exception is restated with the new count before merge. **Body corrections owed:** state the substitute class as "every substitute the wiring is compiled or mounted against"; restate the green-files claim against the new head.

morozsm:
Agent Review: PASS 43207ea85060a50b16d80ac3e3ed6c1a9fca4f21

Independent review (verifier, Opus), five passes on this PR; the last two at 63403357 and 43207ea8. Read-only; measurements by the reviewer, restated here by the coordinator.

**Code (at 63403357, delta-reviewed since):** both pass-2 blocks fixed — the three feedback producers are called with the wiring's `controlSession` like their neighbours, and the fixture-capture stub module exports them (removing the stubs fails all 30 fixture-wiring pins; flipping one to `available` changes nothing, so they are structural, not value-bearing). Through the real producer with an IC-7610-shaped fixture, `getPbtInnerHzControlFeedback(connected)` and `()` return byte-identical Hz feedback; with the disconnected snapshot both PBT and IF-shift return `unavailable` instead of throwing. Mounted re-measurements on both surfaces: fresh lifecycle id at the gesture, thumb on the pending target, readout on the confirmed value, stale readback while pending changes nothing, confirmation at a value other than the requested one lands the thumb there, one gesture → one dispatch. `fixture-capture-root-cwd.isolated` 1, `cw-keyer-zone-lifecycle.component` 3, `FilterSurface.test` 116, `FilterPanel.isolated` 106, `semantic-vfo-indicator-wiring.component` 30, each run singly.

**Mechanism audit (skill method, scoped to the PR's additions):** no dead code — every added symbol has a consumer outside its definition, counted. Three consolidation findings, none blocking: F1 `pbtFeedbackOnMeasuredLattice` (new) beside `projectDspControlFeedbackToDisplay` (existing) in `panel-adapters.ts`, one reachable divergence and it is benign (a target raw that does not convert, only from a foreign command) — recorded in the PR body as a follow-up; F2 four more copies of the PENDING chip markup in `FilterPanel.svelte`; F3 the width row still on its own lease plumbing beside `createPassbandRow`. Cleared: the lattice derivation is a recorded ruling (`deriveIfShift`, MOR-2497) and the PR reduced in-module copies from one inline to one helper; `measuredPbtRawToHz` has a single definition; the receiver-echo gate is reused, not rewritten.

**The four PNGs (43207ea8):** exactly the delta (`4 files changed, 0 insertions, 0 deletions`), each `cmp`-identical to the `-actual.png` of `visual` run 35292148408. Mechanism confirmed by reading: `main`'s `FilterPanel.svelte` substituted `p.pbtInner ?? 0`; the head feeds `{ status: 'unknown' }`; the LCD fixtures publish no PBT; `BipolarRenderer.svelte` (untouched) renders an unknown value as `—` at the origin. The header part of the delta predates the branch — base `3d81dff3` and head render it pixel-identically in local captures with the spec's procedure (0 differing pixels above the S-meter strip), and under the comparator's metric the header alone (213–349 px) sits under the 1024-px allowance; the reviewer reproduced these counts with its own script. The reviewer's earlier attributions (stale baselines; header originating in the branch) were withdrawn on that evidence.

Required at this head: `quick` pass (run 35293253470); `visual` (not required) pass.

PR #3479, opened 2026-09-15
fix: hide unsupported power toggle and neutralize unknown power state

morozsm:
## Behaviour

- Radios without the `power_control` capability (e.g. FTX-1, whose profile does not declare it) keep the power button **rendered but disabled** in `StatusBar` — per the owner decision of 2026-09-15, hiding it shifted the status-bar layout and changed the MOR-1400 production design-language screenshots. The unsupported-radio button stays in place (no layout shift) and renders with the standard disabled styling (the generic `.control-btn:disabled` dimming applies); it keeps the known-state label and classes (`OFF` / `is-on`), and the tooltip truthfully reports the limitation via new key `core.statusbar.power.unsupported` ("This radio does not support power control") in en-US / ru-RU / ja-JP. Clicking never confirms or dispatches.
- With `power_control` and an unknown power state (`radioPowerOn === null`): the button is rendered **neutral** (`power-unknown` class overrides the green base and red `is-on` styling, including the `desktop-control-face` skin variants), is `disabled`, and its label/tooltip truthfully report the unknown state — `core.statusbar.power.labelUnknown` ("UNKNOWN") and reworded `core.statusbar.power.toggleUnknown` ("Radio power state unknown", no "click to toggle"). When no capability AND unknown state coincide, the unsupported tooltip wins and the label/styling follow the unknown state.
- Defence in depth: `handlePowerToggle` returns before any `confirm()` or `powerOn()/powerOff()` dispatch when the capability is absent or the state is unknown.
- Known ON / known OFF behaviour (labels, styling, confirm dialogs, `powerOff()`/`powerOn()` calls) is unchanged.

## Files

- `frontend/src/components-v2/layout/StatusBar.svelte` — button always rendered, disabled when unsupported or unknown, unsupported tooltip branch, handler guard, neutral unknown CSS.
- `frontend/src/components-v2/layout/__tests__/StatusBar.power-toggle.component.test.ts` — NEW component test (5 cases: unsupported rig keeps a disabled button with the truthful tooltip and known-state label/classes, unsupported+unknown precedence, capability+unknown incl. disabled-bypass pin, known ON, known OFF).
- `frontend/src/lib/i18n/locales/{en-US,ru-RU,ja-JP}.json` — only the `core.statusbar.power.*` block touched (`toggleUnknown` reword, added `labelUnknown`, added `unsupported`).
[...]

PR #3533, opened 2026-09-18
fix(MOR-1673): render no power control on a radio that cannot switch power over CAT

morozsm:
6 code + 4 regenerated baselines = 10 files, at the file ceiling, not over it.

## What is wrong

Bench FTX-1, 2026-09-17: the status bar still rendered a disabled power button labelled "UNKNOWN" with the title "This radio does not support power control". MOR-1673's acceptance is that a radio whose power cannot be switched over CAT shows **no** power control.

`StatusBar.svelte` keyed the `disabled` attribute on `hasCapability('power_control')` (from #3479) but always rendered the button. `rigs/ftx1.toml` declares no `power_control` and lists `global.tx_state.power_on` as unsupported; `ic7300.toml` / `ic7610.toml` declare it.

## Change

The same capability gate now decides the render (`{#if powerControlSupported}`), and the dead "unsupported" tooltip / handler branches and their i18n key (`core.statusbar.power.unsupported`, three locales) go with it. With `power_control` declared, unknown and known states render as before. The tab-traversal conformance pin mocks the capability so the (now conditional) button still exists for it to pin.

## Second finding, not fixed here

`/api/v1/state` publishes `powerOn: true` for the FTX-1 although the radio never reports it: the payload is seeded from `RadioState().to_dict()` whose `power_on` defaults to `True` (`core/radio_state.py`, mirrored by `web/state_schema.py`), and no observation ever overwrites the seed. The frontend masks it through field availability, which is why the UI showed UNKNOWN rather than ON. A dataclass default, not a profile line — filed separately.

## Tests

`StatusBar.power-toggle.component.test.ts` 4 passed (1 red before the fix); singly green: `tab-traversal.isolated` 5, `StatusBar.bad-link-chip` 3, `StatusBar.radio-link-chip` 1, `StatusBar.managed-tot` 2, i18n `faceplate-invariant` 3, i18n `index` 22. `npm run check` 0 errors; eslint clean; `npm run i18n:check` OK. Mutation: render regardless of capability → 1 failed.

**Baselines:** the four production design-language PNGs (fieldline / studioline × dark / light) pinned the button; `quick` run 35289571143 failed on exactly those four, and they were re-pinned from that run's diagnostics artifact (the `-actual` attachments). The diff image shows only the header cluster shifting left where the button was.

6 code files, 26 insertions, 69 deletions, plus 4 PNGs.

Refs MOR-1673


morozsm:
Agent Review: PASS 9a8ce7824dc3df4d2ed4e0dc3b1c71b1e8c10804

Independent review by GLM-5.3 (author: DeepSeek Flash) at `e22372c9`, plus a delta pass at this head.

- `StatusBar.svelte` is the only render site of a power control (the mobile / LCD / amber layouts and the settings panel carry RF power only). With `power_control` present and the state unknown the rendering is byte-identical to `main` (the "unsupported" branch was unreachable there). The removed i18n key has zero remaining references; the i18n invariant files pass singly.
- Non-blocking: the tab-order without a power control is not pinned (removing the element cannot create a gap; `panel-commands.ts: onPowerOff` still gates dispatch on the capability). Outside the delta: `AppGlobalHost.svelte` renders its power-on overlay on `radioPowerOn === false` without a capability gate — inert on the FTX-1 because the field is masked to null.
- Mutations (only `StatusBar.power-toggle.component.test.ts`): unconditional render → 1 failed; gate on `powerOn` availability instead of the capability → 1 failed.
- **Delta at `9a8ce782`:** exactly the four PNGs the reviewer had named, no code; IHDR sizes unchanged at 1280×800, each file ~1 KB smaller — consistent with a small button leaving the status bar.

PR #3539, opened 2026-09-18
fix(MOR-2511): show ATT and preamp as unavailable with a hint on a receiver whose profile leaves them undeclared

morozsm:
10 files, 236 insertions + 39 deletions = 275 changed lines (against merge-base 3ee73d7a): at the file hard ceiling (5 source + 3 locales + 2 tests), well under the line threshold. One unit of work: one reason code from the adapter that derives it, through the label map, to the host that shows it, with its pins; the three locale files move together.

## What is wrong

On the FTX-1 the SUB receiver has no CAT form for the attenuator or the preamp, and the server says so: `fieldStatus['sub.att'].availability === 'undeclared'` (likewise `sub.preamp`), while `main.att` / `main.preamp` are `available`. The default skin did not read that — `radio-view-model-adapter.ts` took `structural` from the radio-wide capability and `operational` from "seen and fresh", so SUB's ATT and preamp rendered as *not yet observed* ("?"), which is false: they will never be observed. Owner ruling 2026-09-17 (MOR-2511): show them unavailable, with a hint.

## Change

- **Adapter**: `fieldUndeclared(state, path)` reads the published availability. On the `rfFrontEnd` fields `structural` stays the radio-wide capability (the host renders a row only when it is true) and `operational` is false for an undeclared field; a new reason code `receiver-lacks-control` is pushed on `rfFrontEnd.attenuator` / `rfFrontEnd.preamp` when the ACTIVE receiver's field is undeclared. On the receiver-indicator strips the undeclared badge is structurally absent (no room for a hint there). MAIN and every profile that declares both are unchanged. The first head of this PR set `structural=false` on the `rfFrontEnd` fields too, which made the hint unreachable (review finding); corrected.
- **Label map** `rf-front-end-instruments.ts: DISABLED_REASON_LABEL` now holds i18n keys (`core.disabledReason.receiverLacksControl` = "Not available on this receiver"; the existing mutex label moved to `core.disabledReason.mutuallyExclusiveControl` so every label goes through `t()`), in en-US / ru-RU / ja-JP.
- **Host** `RfFrontEndInstrumentHost.svelte`: both rows stay rendered, disabled, with `data-disabled-reason="receiver-lacks-control"`, the `aria-describedby` hint paragraph and `blocked` seats on the finite path; the "not yet observed" placeholder is replaced by the hint when the reason is present. A code with no label renders no hint (`reasonLabel()` guards the partial map).
- **v2 panel** `RfFrontEnd.svelte` already hides a structurally-absent ATT/preamp (`activeFieldShown` returns false for undeclared) — untouched; on v2 the ruling is met by omission rather than by a hint.

`radio-view-model-adapter.test.ts` 167 (FTX-1-shaped state, active SUB: `rfFrontEnd.attenuator|preamp.availability` = `{structural: true, operational: false}`, reading unknown, both reasons present; active MAIN: no reason, MAIN unchanged), `RfFrontEndInstrumentHost.isolated.test.ts` 55 (the three MOR-2511 cases build the view with the REAL `toRadioViewModel` on a two-receiver state with `sub.att`/`sub.preamp` undeclared: active SUB → both rows present, `data-disabled-reason="receiver-lacks-control"`, hint text in the described-by target, no unknown placeholder, controls disabled; active MAIN → rows present, no reason, no hint), `RfFrontEndSurface.test.ts` 69 (the DIGI-SEL pin asserts on the rendered `t()` text: setting the en-US value to "Disabled: DIGI-SEL is active" fails exactly that test), `RfFrontEnd` panel test 28. Mutations (outputs in the commit body of 658c93f9): deleting the attenuator reason push in the adapter fails exactly the attenuator host test (1 failed / 54); the earlier head's report claimed host coverage it did not have — corrected in this head; `npm run i18n:check` OK; `npm run check` 0 errors; eslint clean.

Builders: `opencode/nemotron-3-ultra-free` for the first three heads (its final report claimed three things it had not done — caught by the review), GLM-5.3 for the last commit with mutation outputs required in the report.

Refs MOR-2511


morozsm:
Agent Review: BLOCKED e1095c03b1dd2b9cd0697657d8d7e30f0ce7874b

Independent review (verifier, Opus), read-only in a throwaway worktree; coordinator relays. Four findings in the diff.

1. **The hint is unreachable.** `RfFrontEndInstrumentHost.svelte` gates both rows on `availability.structural`; the adapter sets exactly that to false whenever the field is undeclared — the same predicate that pushes the reason. So the reason exists only when the row is gone. Measured with the real `toRadioViewModel` → `renderFinite` on an undeclared state: rows absent, hint absent, `data-disabled-reason` never rendered; the two new host tests pin a state the adapter cannot produce, and the attenuator hint path is dead code (the mutex reason is only ever emitted on `rfFrontEnd.preamp`). Ruling to implement (owner 2026-09-17, "unavailable with a hint"): the row STAYS, disabled, with the hint — `structural` stays true on `rfFrontEnd.*`, `operational` false, and the host renders a disabled row with the reason text instead of the "not yet observed" placeholder.
2. **The production structural term is unpinned.** Mutating the `rfFrontEnd` structural expression back to `hasAttCap` leaves all three test files green (167 / 54 / 69); only the receiver-indicator twin is pinned. A test must assert `rfFrontEnd.attenuator|preamp` — the object the host reads — on an FTX-1-shaped state.
3. **Hollow neighbour pin.** `RfFrontEndSurface.test.ts` "the label text never mentions DIGI-SEL" now asserts on the i18n key; setting the en-US value to "Disabled: DIGI-SEL is active" keeps 69/69 green. Assert on `t(key)`.
4. **False comment** in the host: "matches any disabled reason on 'rfFrontEnd.attenuator' (mutex or receiver-lacks-control)" — no mutex reason is emitted on that field.

Confirmed: 10 files +184/−35; IC-7610-shaped, no-`fieldStatus` and null states deep-equal to `3ee73d7a` (comparison proven able to report inequality on the FTX-1-shaped state); active MAIN untouched; v2 `activeFieldShown` already hides undeclared fields at `3ee73d7a`; every `DISABLED_REASON_LABEL` consumer renders through `t()`; no raw key reaches the DOM; `check` 0, eslint 0, `i18n:check` 297 keys; `quick` pass at the head.

morozsm:
Agent Review: BLOCKED 2f0dc9903128210285a9dfd95e89716e73d6f5c8

Independent review (verifier, Opus), delta over e1095c03; coordinator relays.

Findings 1, 2 and 4 of the first pass are fixed and measured: `deriveRfFrontEnd` structural terms byte-identical to `main`; with the real `toRadioViewModel` → host on an undeclared SUB state, both rows are present with `data-disabled-reason="receiver-lacks-control"`, the hint text (en-US and ru-RU), no unknown placeholder, disabled controls, `aria-describedby` set; reason-absent states render identically to `main` bar an empty Svelte anchor.

Still blocking: finding 3 was not touched — `RfFrontEndSurface.test.ts` "the label text never mentions DIGI-SEL" asserts on the i18n key and cannot fail (setting the en-US value to "Disabled: DIGI-SEL is active" keeps 69/69 green). Also `preampMutexReason` in the host is now defined and never called.

The PR body's Tests section carried three claims from the builder's report that the review refuted: the host's MOR-2511 cases do not go through the real adapter (they inject reasons onto `finiteBase()`), the DIGI-SEL pin does not assert on `t()` text, and dropping the reason push fails only the adapter test (host 54/54). Corrected in the body; the fix commit will make the first two true.

morozsm:
Agent Review: PASS 658c93f9e6d81ce4531e5d68fda938ce61df7a1f

Independent review (verifier, Opus), three passes: full at e1095c03 (BLOCKED: the hint was unreachable because the same predicate that pushed the reason hid the row; production structural term unpinned; a hollow DIGI-SEL pin; a false comment), delta at 2f0dc990 (rows kept, hint measured on both surfaces with the real adapter; DIGI-SEL pin still hollow), delta at 658c93f9 (fixed). Read-only in throwaway worktrees; coordinator relays.

Measured at the head: `deriveRfFrontEnd` structural terms byte-identical to `main`; with the real `toRadioViewModel` → host on a two-receiver state with `sub.att`/`sub.preamp` undeclared and SUB active, both rows present, `data-disabled-reason="receiver-lacks-control"`, the hint text (en-US "Not available on this receiver", ru-RU "Недоступно на этом приёмнике") in the described-by target, no unknown placeholder, controls disabled; active MAIN and IC-7610-shaped states render identically to `main` bar an empty Svelte anchor; no raw i18n key in the DOM. Mutations reproduced: en-US mutex label set to "Disabled: DIGI-SEL is active" fails exactly the DIGI-SEL pin (1/69); deleting the attenuator reason push fails exactly the attenuator host case (1/55) — at the previous head the same mutation left the host file green. `preampMutexReason` gone. Singles: adapter 167, host 55, surface 69, v2 panel 28; `npm run check` 0 errors; eslint 0; `i18n:check` 297 keys OK.

Builder note for the record: the first three heads came from `opencode/nemotron-3-ultra-free`, whose final report claimed the host tests used the real adapter, the DIGI-SEL pin used `t()` and the host mutation failed — none was true at 2f0dc990; the last commit (GLM-5.3) made the first two true and quoted the mutation outputs.

PR #3540, opened 2026-09-18
fix(MOR-2511): forward the receiver from the Yaesu command loop to the radio setters

morozsm:
2 files, 66 insertions + 20 deletions = 86 changed lines.

## What is wrong

On the owner's FTX-1 at `3ee73d7a` (web server, 2026-09-18 00:15 EDT): `POST /api/v1/commands {"name":"set_agc","params":{"mode":3,"receiver":1}}` changed MAIN's AGC (1 → 3) and left SUB at 6; `set_nb_level` with `receiver: 1` likewise moved MAIN. So every DSP / AGC / width write the operator made on SUB today landed on MAIN — a wrong-receiver write, worse than a refusal. The backend is right in isolation (`YaesuCatRadio.set_agc(3, receiver=1)` → `GT13;`, `set_nb_level(3, receiver=1)` → `NL1003;`, run on the same machine); the receiver is dropped in the Yaesu command loop, `backends/yaesu_cat/poller.py: YaesuCatPoller._execute_command`, whose arms read the value and not the receiver the command dataclass carries (`case SetAgc(mode=mode): await radio.set_agc(mode)`). AF / RF gain / squelch never go through this loop (they have receiver-aware descriptors in `core/command_dispatch.py`), which is why those worked on SUB while AGC and DSP did not.

## Change

Ten arms bind and forward `receiver=rx` exactly as `SetFreq` / `SetMode` in the same loop already do: `SetAgc`, `SetNB`, `SetNR`, `SetNBLevel`, `SetNRLevel`, `SetAutoNotch`, `SetManualNotch`, `SetNotchFilter` (→ `set_manual_notch_freq`), `SetIfShift`, `SetFilterWidth`. Every target setter accepts `receiver: int = 0`. No other arm touched.

## Tests

`tests/test_yaesu_cat_poller.py::test_execute_command_forwards_the_receiver_to_the_radio`, ten commands × receiver 1 and 0, asserting the radio setter is awaited with the explicit receiver: RED 10 failed at the head → GREEN. Single files: `test_yaesu_cat_poller` 191, `test_canonical_receiver_levels` 38, `test_mor2161_command_dispatch` 51, `test_ftx1_radio` 402. Ruff check and format clean. `mypy --strict` on `poller.py`: 1 error at head, the same 1 at base (line 449, outside the diff). Mutation: reverting the `SetAgc` arm alone fails exactly the `set_agc` row.

Bench acceptance after merge: with SUB active, an AGC or NB change on the web must move `sub.*` and leave `main.*` unchanged (the same API probe as above).

Refs MOR-2511


morozsm:
Agent Review: PASS 64cb6ac2705dba44c42c09898e96eab68735da30

Exactly ten arms changed, every other line of `_execute_command` byte-identical; all ten target setters accept `receiver` (signatures quoted in the review). Defect reproduced at the base and measured fixed on the real `YaesuCatRadio` + bundled profile through the real `YaesuCatPoller._execute_command`: at `3ee73d7a` `SetNBLevel(5, rx=1)` → `NL0005;`, `SetAgc(3, rx=1)` → `GT03;`, `SetIfShift(200, rx=1)` → `IS00+0200;`, `SetFilterWidth(…, rx=1)` → `SH0012;`; at `64cb6ac2` → `NL1005;` / `GT13;` / `IS10+0200;` / `SH1012;`, rx=0 unchanged. Class swept by cross-joining every `Set*` dataclass carrying `receiver` with every arm: the remaining non-forwarding arms either raise `NotImplementedError` (SetFilter, SetFilterShape, SetIpPlus, SetPbtInner) or target setters that ignore the receiver internally with no SUB form (SetAttenuator, SetDataMode — the owner's "unavailable with a hint" ruling); `web/radio_poller.py` already forwards on its sibling match; `core/command_dispatch.py` holds only twelve descriptors, none for `set_agc` or the DSP setters, which is exactly why its receiver-aware gate never fired. Pin awaits the real loop on a real poller with a mocked radio and asserts the exact `receiver` kwarg; mutation reverting the `SetIfShift` arm fails exactly that row. Singles 191 / 38 / 51 / 402; ruff clean; mypy 1 = 1 at base (line 449, outside the diff). `quick` pass at the head.

Separate follow-up recorded on MOR-2511: `set_narrow` has a receiver-aware backend method and the `NA1` form, but no `SetNarrow` intent type exists and no arm dispatches it — narrow cannot be written on either receiver from any consumer.

PR #3544, opened 2026-09-18
fix(MOR-2512): widgets ignore Ctrl/Alt/Meta+Arrow the global keyboard map owns

morozsm:
8 files, 110 insertions + 2 deletions = 112 changed lines (4 code + 4 test): over the soft file threshold, one unit of work — one predicate and the three widgets that must honour it, each with its pin.

## What is wrong

After MOR-2507 (#3531) the global keyboard layer leaves Ctrl/Alt/Meta+Arrow to the global bindings (`rigs/_keyboard-default.toml`: Ctrl+ArrowUp/Down → `adjust_af_level`, Ctrl+Shift+ArrowUp/Down → `adjust_rf_gain`), but the widgets' own key handlers ignore the modifiers: measured with a focused widget, Ctrl+ArrowUp both steps the widget and fires the global action (MOR-2512, review finding on #3531). `keyboard-map.ts: focusedElementOwnsArrowKey` already embodied the rule inline.

## Change (MOR-2512 part 1 of 3)

- `keyboard-map.ts`: the inline test extracted into `hasCommandModifier(event)` (ctrl / alt / meta), used by `focusedElementOwnsArrowKey` (behaviour unchanged) and exported for widgets.
- `SegmentedButton.svelte`, `vfo/ActiveReceiverToggle.svelte`, `spectrum/SpectrumPanel.svelte` (`handleSplitKeydown`): return before any arrow handling when a command modifier is held — the event propagates untouched, so the global layer acts once. `SegmentedButton` also declares `data-owns-arrows="both shift"`: it acts on Shift+Arrow, and the global layer yields Shift+Arrow only to widgets that declare it (the ticket's smaller finding).
- Parts 2 and 3 (the frequency readout with a digit selected; the six value-control renderers) follow separately.

`keyboard-map.test` 48 (the predicate; the ctrl pin on `focusedElementOwnsArrowKey` unchanged), `SegmentedButton.test` 47, `ActiveReceiverToggle.test` 29, `SpectrumPanel.component.test` 158 — each with a new pin: focus, dispatch the widget's arrow with `ctrlKey: true` → no value/selection change and `defaultPrevented === false`; the unmodified pins stay. RED at the base for the three widget pins. `KeyboardHandler.test` 53 unchanged; `npm run check` 0 errors; eslint clean on the eight files. Mutation: removing the guard in `SegmentedButton` alone fails exactly its ctrl pin.

Refs MOR-2512


morozsm:
Agent Review: PASS f3f9c0c09ac1fd5ffbaf1bf48e1cd2ca2324f900

Shape exactly as described (8 files +110/−2; the guard is the first statement of each of the three handlers, before any `preventDefault`/`stopPropagation`; `focusedElementOwnsArrowKey` unchanged in meaning). Double fire measured end to end with a real `KeyboardHandler` (Ctrl+ArrowUp → `adjust_af_level`) and a real focused widget: at the base SegmentedButton and ActiveReceiverToggle each step AND the global action fires; at the head the global action fires once and the widget is unchanged; plain ArrowUp still steps the widget with no global action (MOR-2507 preserved). `both shift` is a correction, not a steal: `SegmentedButton.handleKeydown` never inspects `shiftKey`, so Shift+Arrow steps it — with a Shift+ArrowUp global binding the base double-fired and the head does not. Pins per widget assert no change and `defaultPrevented === false`; removing the guard from ActiveReceiverToggle alone fails exactly its ctrl test while SegmentedButton stays 47/47. The `hasCommandModifier` doc comment's consumers and pinning sites exist and fail under mutation. Singles 47 / 29 / 48 / 158 / 53; `npm run check` 0 errors; eslint 0.

PR #3545, opened 2026-09-18
fix(MOR-2512): value-control renderers ignore Ctrl/Alt/Meta+Arrow the global keyboard map owns

morozsm:
8 files, 139 insertions + 0 deletions (6 renderers + 2 test files): over the soft file threshold, one unit of work — the same one-line guard in every value-control renderer, each pinned.

## What is wrong

MOR-2512 part 3 (parts 1 and 2: #3544, and the readout PR). The six value-control renderers' `handleKeyDown` call `lease.key({ key, fine: shiftKey })` and `preventDefault` for any arrow, ignoring ctrl / alt / meta — so a focused slider steps AND the global Ctrl+Arrow binding (`adjust_af_level`, `adjust_rf_gain`) fires (measured on the AF control in the #3531 review).

## Change

`if (hasCommandModifier(e)) return;` as the first statement of the key handler in `BipolarRenderer`, `DiscreteRenderer`, `DualParamRenderer`, `HBarRenderer`, `KnobRenderer` and `skins/ProfessionalKnob` (predicate from #3544, `keyboard-map.ts`). Nothing else in those files; `value-control-core.ts` untouched.

`ValueControl.test.ts` 53 — `it.each` over hbar / bipolar / discrete / knob / professional: focus, ArrowUp with `ctrlKey: true` → `aria-valuenow` unchanged and `defaultPrevented === false`; `DualParamRenderer.controlled.svelte.test.ts` 8 — the same for the dual-param renderer; `ValueControl.controlled.svelte.test.ts` 94 unchanged; `npm run check` 0 errors; eslint clean. RED at the base for the six new cases. Mutations: removing the guard from `HBarRenderer` alone fails exactly the hbar case; from `DualParamRenderer` alone exactly its case.

Builder: `opencode/nemotron-3-ultra-free`; every claim in its report was checked against the tree by the coordinator (guard and import counted per file, no other change in the six files, both test files run locally: 53 and 8).

Refs MOR-2512


morozsm:
Agent Review: PASS 99e8c49f50160163aafe99a837622af0490d1e74

Shape exactly as described: two lines per renderer (import and the guard as the first statement of the key handler, before `lease.key` and `preventDefault`), nothing in `value-control-core.ts` or `ValueControl.svelte`. Double fire measured with a real `KeyboardHandler` (Ctrl+ArrowUp → `adjust_af_level`) and a real focused `ValueControl`: at the base hbar / bipolar / knob / professional each call `onChange` AND the global action fires; at the head the global fires once and `onChange` is not called; plain ArrowUp unchanged (MOR-2507). Pins: five rows in `ValueControl.test.ts` and the dual-param case each assert Ctrl, Alt and Meta leave the value and `defaultPrevented` alone, plus the plain-arrow leg; removing the guard from `KnobRenderer` alone fails exactly the knob row, while the professional row stays green because the skin's own `onKey` carries its own guard — correct discrimination. Singles 53 / 94 / 8; `npm run check` 0 errors; eslint 0.

PR #3546, opened 2026-09-18
fix(MOR-2512): the frequency readout owns the arrow keys only while a digit is selected

morozsm:
6 files, 162 insertions + 12 deletions = 174 changed lines against merge-base 825ee060 (readout, interaction, keyboard-map, one new lib module, two test files): on the soft file threshold; one unit of work — the readout's arrow ownership and the predicate move it needs.

## What is wrong

MOR-2512 §2 (part 2 of 3; parts 1 and 3: #3544, #3545). `StandardFrequencyReadout.svelte` mounts `frequency-interaction.svelte.ts: handleKeyDown` on a `role="group"` root; with a digit selected ArrowUp steps the digit AND the global `adjust_tuning_step` fires, because the readout declares no arrow ownership. A blanket `role="group"` exclusion was rejected: it reddens the three MOR-2364 pins that cover the no-digit path. The readout also ignored ctrl / alt / meta like the widgets did.

## Change

- `StandardFrequencyReadout.svelte`: `data-owns-arrows="vertical"` present on the root ONLY while a digit is selected; absent otherwise, so the MOR-2364 no-digit path is unchanged.
- `frequency-interaction.svelte.ts: handleKeyDown`: returns first when a command modifier is held.
- `hasCommandModifier` moved to `frontend/src/lib/utils/keyboard-modifiers.ts` — `primitives/` may not import `components-v2/` (ADR 2026-04-12, enforced by eslint; the builder stopped on that boundary and the coordinator ruled the move) — and re-exported from `keyboard-map.ts`, so the #3544/#3545 consumers are untouched; its comment describes consumers by rule, not by list.

`StandardFrequencyReadout.isolated.test.ts` 10 (no digit → no `data-owns-arrows`; digit selected → `vertical`; digit selected + Ctrl+ArrowUp → no step and `defaultPrevented === false`; plain ArrowUp steps), `KeyboardHandler.test.ts` 54 (the three MOR-2364 pins untouched; one composition pin: real `KeyboardHandler` + real `VfoSurface`, ArrowUp → `adjust_tuning_step` with a digit selected fires 0 times, without one fires once — under mutation it fired twice, i.e. the pin catches the §2 defect, not the attribute text), `keyboard-map.test.ts` 48, `SegmentedButton.test.ts` 47 and `ValueControl.test.ts` 48 (re-export consumers), `FrequencyRendererSeat.isolated` 2, `frequency-tuning` 13, `FrequencyDisplayInteractive.component` 17; eslint on the primitive passes the boundary with `$lib/utils`; `npm run check` 0 errors. Mutations: removing the attribute fails the ownership and composition pins; removing the guard fails the ctrl pin.

Refs MOR-2512


morozsm:
Agent Review: PASS d44b57d7738bc627bd1e6f25f87c7ca93579a726

Shape exactly as described; the predicate's body is byte-identical after the move and `keyboard-map.ts` re-exports it. Boundary genuine: `$lib/utils` is absent from `FORBIDDEN_PRIMITIVES_IMPORTS`, and rewriting the import back to `components-v2/layout/keyboard-map` makes eslint fail with the primitives rule — the move is the fix it claims to be. Composition measured with a real `KeyboardHandler` and a real `VfoSurface` readout: no digit → global once, no tune (base and head identical); digit selected + ArrowUp → base double-fires (global 1 + tune 1), head tunes once with no global; digit selected + Ctrl+ArrowUp → base `adjust_af_level` + tune, head `adjust_af_level` only. The three MOR-2364 pins are byte-identical (the test diff has no deletions) and green. Mutations: removing the attribute fails the ownership pin and the composition pin; removing the guard fails exactly the ctrl pin. Merge with current main (37b7eda1, #3545): `merge-tree` clean; on the merged tree `ValueControl.test` 53, `KeyboardHandler` 54, readout 10, `keyboard-map` 48, `npm run check` 0 errors — the six renderers keep importing from `keyboard-map`. Head singles all green; eslint 0.

PR #3547, opened 2026-09-18
refactor(MOR-2502): drop the write-only continuityIdentity from the scope passband display state

morozsm:
1 file, 3 insertions + 10 deletions. Three commits: the field; then the `partition` lookup that only fed it; then the `band` local and the `findActiveBand` import that only fed `partition` — the orphan chain swept until every remaining local has a reader (two review findings of the same class).

Since #3521 `scope-passband-display.ts` decides passband continuity by `bandlessIdentity` alone; `continuityIdentity` was still computed in `inspect()` and copied into `ScopePassbandDisplayState` with no reader (established by the #3521 review: `git grep continuityIdentity -- frontend/src` found only the adapter — interface, one write, one copy — with `bandlessIdentity` as the control that finds the reads). Removed from the `Candidate` type, the state type, the initial value, the computation and the returned state. No decision logic touched. `git grep continuityIdentity -- frontend/src` is empty at the head; `scope-passband-display.test.ts` 210 passed before and after (no test named the field); `npm run check` 0 errors; eslint clean.

Builder: `opencode/nemotron-3-ultra-free`; the coordinator verified the diff, the grep and the test count on the tree.

Refs MOR-2502


morozsm:
Agent Review: PASS 2408cce078d4014fcdb595f3922158d6194576e3

Independent review (verifier, Sonnet), three passes: at f131e52e the field removal was confirmed write-only tree-wide (five write sites, no reader; `bandlessIdentity` as the control with real comparisons) but the `partition` lookup it fed was left as a dead local; at b0df3dfd that lookup's removal orphaned `band` / `findActiveBand` in turn; at 2408cce0 the class is swept programmatically — every local in `inspect()` and at module scope has a reader (the two module symbols with one in-file occurrence are exports consumed elsewhere), `findActiveBand` keeps eight other call sites, the single test file is 210 at every head, eslint clean, one file +3/−10.

PR #3548, opened 2026-09-18
feat(MOR-1687): localize the break-in delay value text

morozsm:
5 files, 57 insertions + 8 deletions against merge-base 59b94be6 (surface, three locales, one test file).

MOR-1687 part 3 (the last item on the ticket). `CwKeyerSurface.svelte: breakInDelayValueText` — the break-in delay slider's `aria-valuetext` — was built from five hard-coded English strings. They now come from `core.cwKeyer.breakInDelay.valueText.{unavailable,draft,requested,outcome,confirmed}` in en-US / ja-JP / ru-RU through `t()`; `{phase}` keeps the raw phase word as before.

No behaviour delta: the first head added a `displayed === null` branch and an alias; both are gone (the review showed `displayed` is null only when `confirmed` is, which the function already returns on; the busy branch passes `displayed ?? confirmed` to satisfy the `t()` parameter type without a non-null assertion). Drive-by, disclosed: `core.cwKeyer.breakIn.blocked` was missing from ja-JP and is added (`npm run i18n:check` tolerates missing keys, so nothing required it).

Tests: `CwKeyerSurface.test.ts` 103 — the two `aria-valuetext` pins keep LITERAL expected strings (the first head compared against `t(key)`, which the review showed is tautological: a corrupted en-US string left the file green; now the same mutation fails both pins — reproduced by the coordinator), plus one ru-RU case that switches the locale and asserts the Russian literal; `npm run i18n:check` OK (302 keys); `npm run check` 0 errors; eslint clean.

Builder: `opencode/nemotron-3-ultra-free`; the coordinator read the diff and ran the test file and `i18n:check` on the tree.

Refs MOR-1687


morozsm:
Agent Review: PASS adffa01682712225f118aabc2b429a37db98e84f

Independent review (verifier, Sonnet), two passes; coordinator relays. First head BLOCKED: the two `aria-valuetext` pins compared against `t(key)` and were tautological (a corrupted en-US string left the file green), plus a dead `displayed === null` branch, an alias and an undisclosed ja-JP drive-by key. At this head: the pins are literal strings — corrupting the en-US `unavailable` string fails exactly the two assertion sites (one direct, one through a six-row `test.each`), corrupting the ru-RU string fails exactly the ru-RU case; the branch and alias are gone; the busy branch uses `displayed ?? confirmed` with no non-null assertion; the drive-by key is disclosed. The five strings are character-identical to the original literals after substitution in all three locales; `CwKeyerSurface.test.ts` 103; `npm run i18n:check` 302 keys OK; eslint clean; 5 files +57/−8 against the merge-base.

PR #3549, opened 2026-09-18
refactor(MOR-1687): one projector for feedback display units, two facades

morozsm:
3 files, 86 insertions + 20 deletions.

Mechanism-audit follow-up from the #3532 review (MOR-1687): `panel-adapters.ts` carried two projectors of a `ControlFeedback`'s three value fields into display units — `projectDspControlFeedbackToDisplay` (any non-null raw that fails to project → unavailable) and `pbtFeedbackOnMeasuredLattice` (only a failed `confirmed` → unavailable; a failed `target`/`requestedTarget` silently became `null`).

## Change

One core `projectControlFeedbackToDisplay(feedback, project)` with the DSP semantics — a null raw stays null; any non-null raw whose projection is null fails the whole feedback closed (`unavailableControlFeedback`); otherwise a frozen copy with the three projected fields — and both existing functions become thin facades over it (`projectDspRawValueToDisplay` and `measuredPbtRawToHz` respectively; the PBT lattice-null / availability guard stays in front). No caller touched.

Intended behaviour change, pinned: a PBT feedback whose target raw does not convert on the lattice is now unavailable instead of available-with-null-target. Unreachable from `panel-commands.ts` (it emits on-lattice raws only); reachable from a foreign command.

`pbt-if-shift-command-feedback.isolated.test.ts` — new case: pending `set_pbt_inner` with raw 9999 through the real producer → `availability: 'unavailable'` (RED at the base: `available` with `confirmed: 50`); the DSP feedback test file — direct unit test of the core over its four rules with a per-slot fail-closed `it.each`; `FilterSurface.test.ts` and `FilterPanel.isolated.test.ts` (consumers) unchanged and green; four files 301 passed total; `npm run check` 0 errors; eslint clean. Mutations: dropping only the `target` check is caught by the DSP per-slot case (through the real PBT producer `requestedTarget` carries the same raw and masks it); restoring the old PBT behaviour (null both targets) is caught by the PBT case.

Refs MOR-1687


morozsm:
Agent Review: PASS 8d9306fdec7417b906495c3cd539e7a63f469f90

Independent review (verifier, Opus — the same reviewer whose #3532 mechanism audit raised this finding), read-only in a throwaway worktree; coordinator relays.

Shape: three hunks in `panel-adapters.ts` (new core, two facades, docs), no caller changed. DSP facade equivalence measured on the full 3×3×3×2 grid through the real exported function with the file swapped to `dc4a3b03`: 54 lines each, 0 differing (the instrument flagged the PBT row, so it can report inequality). PBT delta through the real producer on a 3600/50 lattice: exactly one reachable shape differs — on-lattice confirmed with an off-lattice target: base available `[50, null, null]`, head unavailable — the intended fix; the other eight shapes identical. A function-level difference (base failed closed on `confirmed === null` while available) is unreachable because the producer marks a missing confirmed reading unavailable first. Pins: the PBT case runs through the real producer; the DSP `it.each` covers each slot; dropping the `requestedTarget` clause fails two DSP cases while the PBT file stays green for the stated reason. Singles 26 / 42 / 122 / 111 / 3 / 40 / 12; `npm run check` 0 errors; eslint 0.

PR #3550, opened 2026-09-21
fix(MOR-1679): read and write FTX-1 SH width codes as CAT 2508-C Table 5 defines them

morozsm:
Part 1 of 2 for MOR-1679 (backend truth). The browser side follows in a second PR; the ticket stays open across both and the hardware rerun.

## What was wrong

The FTX-1 profile decoded the `SH` width answer against zero-based tables, while the CAT Operation Reference 2508-C, Table 5, numbers the widths from code 01 (code 00 is "(Default)"). Every width was therefore published one step high and written one step low, the tables ended in a 4500 Hz entry the manual does not have, and FM was 12500 Hz where Table 5 gives 16000.

Separately, a width table resolved for only 4 of the 17 modes in the profile's `[modes] list` (LSB, USB, FM, AM). For the other 13, `YaesuCatRadio.read_filter_width` returned the raw code as if it were Hz.

Seen on a live FTX-1 on 2026-09-21 (RX only): in USB the radio's own screen showed 3000 Hz while the published state said 3200; a receiver in DATA-U published `filterWidth: 19`.

## What this changes

- `rigs/ftx1.toml`: tables rewritten to Table 5; `first_code = 1` under `[filters]`; one `[filters.width.<MODE>]` entry per covered mode, so the existing exact-match lookup in `RadioProfile.resolve_filter_rule` finds them with no Yaesu mode names added to shared code; fixed entries AM-N 6000, AM / FM-N / DATA-FM-N 9000, FM / DATA-FM 16000; `width_max_hz` 4000.
- `profiles/rig_loader.py`, `profiles/__init__.py`: `filter_width_first_code` (default 0) read from `[filters] first_code`.
- `backends/yaesu_cat/radio.py`: read and write offset the code by `first_code`; a mode without a table (C4FM-DN / C4FM-VW), a code-00 answer, or a code outside the table reads as `None` instead of the raw code; `set_filter_width` refuses in a fixed-width mode. The two polling callers already carry `None` as "not available" (`observations.py: poll_medium` through `_safe_read`, and the legacy poller's `filter_width: int | None`), so neither was edited. A third caller does not: `validation/hardware.py: _check_filter_width_set` passes the read value to `_nudge_filter(width: int)`, which raises `TypeError` on `None`. It fails before any write and `execute_hardware_checks` catches it, so the hardware validator reports that check as failed rather than nudging a fake width; making it skip cleanly is left for the follow-up (found by the independent review).
- The SH command templates are unchanged: `SH0{code:03d};` for MAIN and `SH1{code:03d};` for SUB in `rigs/ftx1.toml` — one three-digit field covering the manual's fixed P2 = 0 and the two-digit P3.

## Pins

Literal values, RED before / GREEN after: USB code 10 -> 1950, code 20 -> 3000, code 23 -> 4000; CW-U code 21 -> 4000; DATA-U code 19 -> 3200; set 2400 Hz -> `SH0013;` (MAIN) and `SH1013;` (SUB); full-table round trips for the SSB and CW families on MAIN and SUB; fixed-mode reads and refused writes; C4FM-DN, code 00 and out-of-range read as `None`; a loop over the TOML's own `[modes] list` that fails when a mode other than the two C4FM ones has no rule; parity of the seven CW-family tables.

## Size

8 files, +394 / -56, measured with `git diff --stat $(git merge-base origin/main HEAD)`. Over the 6-file soft threshold because the table fix, the code numbering and the four test files that pinned the old shifted values are one contract: any subset leaves `main` with tests asserting the wrong Hz.

## Checks run by the builder (single files, locally)

`tests/test_ftx1_radio.py` 419 passed; `tests/test_dsp_filter_family.py` 77 passed; `tests/test_ftx1_sub_acquisition.py` 3 passed; `tests/test_yaesu_cat_observation_adapter.py` 225 passed; `tests/test_rig_loader.py` 397 passed, 1 skipped; `ruff check` and `ruff format --check` on the changed paths clean; `lint-imports` 5 kept, 0 broken.

## Not in this PR

`rigs/_schema.md` has no row for the new `[filters] first_code` key yet (review follow-up, goes with PR 2). The `-> int` annotations on `radio_protocol.py: DspControlCapable.get_filter_width` and on the `read_filter_width` Protocol stub in `yaesu_cat/observations.py` were left as they are to stay inside the file budget; the browser's rendering of an unavailable width and its 4000 Hz bound are PR 2. No hardware was touched by this change; the one-step shift itself was confirmed against the radio's screen (see the ticket).

Linear: MOR-1679 (remains open).


morozsm:
Agent Review: PASS acb3dcd4282b4d8f28eae5c6b6158041d40114bb

Independent review at the exact head. No blocking defect. The Table 5 claim holds
over the whole input space, and every new pin was shown capable of going red.

## What I measured (not read)

**Primary source, checked myself.** I read the SH WIDTH / Table 5 page of the
2508-C CAT reference text copy directly, not the brief's transcription. Both
lists match `rigs/ftx1.toml` value for value: LSB/USB 01-23 (300 … 4000, 23
entries), CW-L/CW-U/DATA-L/DATA-U/RTTY-L/RTTY-U/PSK 01-21 (50 … 4000, 21
entries), AM-N 01=6000, AM/FM-N/DATA-FM-N 02=9000, FM/DATA-FM 03=16000, C4FM
absent, P1 0=MAIN/1=SUB, P2 fixed 0, P3 00-23.

**Exhaustive enumeration, not samples.** A read-only probe drove the real
`backends/yaesu_cat/radio.py: YaesuCatRadio.read_filter_width` over all 17 modes
in `[modes] list` x codes 00-25 x both receivers = 884 outcomes, compared against
the manual: **0 mismatches**. The write path was driven over every Hz entry of
each mode's own table on both receivers: every frame exact
(`SH{P1}{code:03d};`), the six fixed modes refuse with `CommandError`, and the
only frames that are not structurally valid are the two C4FM modes (below).

**Mutations** (each applied to the tree, run, then `git checkout --`; tree
verified clean at the end, `git status --short` empty, HEAD still acb3dcd4):

| mutation | test file | result |
|---|---|---|
| drop `+ filter_width_first_code` on the write path only | `tests/test_ftx1_radio.py` | RED, 6 failed |
| `radio.py` no-table branch `return None` -> `return code` | `tests/test_ftx1_radio.py` | RED, 1 failed |
| `radio.py` `except ValueError: return None` -> `return code` | `tests/test_ftx1_radio.py` | RED, 1 failed |
| delete the fixed-mode `CommandError` refusal | `tests/test_ftx1_radio.py` | RED, 6 failed |
| `rig_loader.py: load_rig` stops reading `first_code` | `tests/test_dsp_filter_family.py` | RED, 3 failed |
| drift one CW-family value (RTTY-U 4000 -> 4200) | `tests/test_dsp_filter_family.py` | RED, 2 failed (manual pin + parity) |
| **control:** equivalent rewrite of the offset expression + an added TOML comment | both files | GREEN, 419 / 77 |

The checks redden on the bug and stay green on legitimate nearby work.

**Class sweeps done:** `first_code` and `encoding = "table_index"` occur in no
other rig (`grep -rn "first_code\|table_index" rigs/`), so the default 0 is inert
elsewhere. No 4500 or 12500 pin exists anywhere in tests, fixtures, goldens,
frontend fixtures or docs. `rigs/_schema.md` is the only documentation of the
`[filters]` section; no doc cites the FTX-1 tables. The frontend needs no offset:
`components-v2/panels/FilterPanel.svelte` converts index<->Hz against its own
copy of `table` and emits Hz, so `web/server.py: _serialize_filter_config` not
carrying `first_code` is correct, not an omission.

## CI as found

`quick` **success** at this head, 16740 passed / 219 skipped / 13 xfailed in 71s
(run 35630016984), including the FTX-1 validation dry-run golden gate,
import-linter and Ruff. `Agent Review Gate` red pending this directive;
`quick-v2-observe` pending. The `mypy --strict src/rigplane/web` step is gated on
`core && frontend` and did not run, because no frontend path changed.

## REQUIRED BEFORE MERGE — PR body corrections (edit in place; no commit needed, head stays valid)

1. **"Both existing callers already carry `None` … so neither was edited" is
   false.** There is a third production caller:
   `src/rigplane/validation/hardware.py: _check_filter_width_set` passes
   `dsp.get_filter_width(0)` into `_nudge_filter(width: int)`. I reproduced it:
   `_nudge_filter(None)` raises
   `TypeError: '<=' not supported between instances of 'NoneType' and 'int'`.
   It is not fatal — `execute_hardware_checks` has an `except Exception`
   backstop, so the check reports FAIL and the matrix still completes, and the
   TypeError happens before any write, so nothing bogus is sent to the radio
   (at the base commit the same path nudged a raw code and *did* write it). But
   the sentence as written says the None path was checked everywhere and it was
   not. Reachable only when the radio answers code 00 or the mode is C4FM.
2. **"`SH{P1}0{P3:02d};` serialization is unchanged"** — unchanged is true, but
   the rendering misdescribes the template. `rigs/ftx1.toml` writes
   `SH0{code:03d};`: one 3-digit field spanning P2+P3P3, not a literal `0`
   followed by a 2-digit P3. That matters, because it is why an out-of-range
   code overruns P2 rather than being clipped.

## Follow-up that does need a commit (not blocking this head)

3. `rigs/_schema.md`, `[filters]` "Additional optional fields" table: the PR adds
   a new `[filters]` key and adds no row for it. A future Yaesu profile author
   reading the documented schema will not know `first_code` exists, and the
   silent default of 0 reproduces exactly the off-by-one this PR fixes. Not
   blocking: nothing in the diff becomes false, no gate enforces the table, the
   key is inert for every other profile, and the ticket stays open for PR 2 —
   but the row should land in PR 2 at the latest. Neighbour check on that same
   table: the existing `style` row documents a key `load_rig` never reads from
   `[filters]` (it reads only `list`, `width_min_hz`, `width_max_hz`,
   `encoding`, `first_code`, `width`). Pre-existing, not this PR's doing.

## Notes (no action required)

4. `tests/test_ftx1_radio.py: test_fixed_mode_width_read_and_set_refusal` feeds
   `"SH0002"` for all six rows, but Table 5 gives AM-N only code 01 and
   FM/DATA-FM only code 03. I proved the stimulus is inert: the six rows still
   pass with `"SH0099"`. The assertions are correct and the rows do distinguish
   different `defaults[0]` values, so this is not a hollow test — the `rule.fixed`
   branch simply cannot reach code decoding. Only the literal is misleading;
   feeding each mode's own legal code would say what the docstring claims.
5. The kept sentence "When no table is defined, `width_hz` is sent as the raw
   index (compat fallback)" is true, and I checked what it costs: in C4FM-DN /
   C4FM-VW a set still emits `SH03000;` or `SH0500;`, which are not valid SH
   frames. Pre-existing — at the base commit the same fallback covered 13 of the
   17 modes; this PR narrows it to 2. A refusal for a mode with no Table 5 row
   is the natural follow-up.
6. `[filters.width.DATA-L]` / `[filters.width.DATA-U]` get
   `defaults = [500, 300, 50]`, a mechanical copy of the CW row, and neither the
   body nor the commit message mentions it. I checked whether this is a
   regression and it is not: the `[filters.width.DATA]` entry it replaces
   (`[3000, 500, 50]`) was unreachable dead config — `[modes] list` has no
   `DATA` mode, and neither `profiles/__init__.py: RadioProfile.resolve_filter_rule`
   nor `frontend/src/lib/runtime/props/panel-props.ts: resolveFilterModeConfig`
   maps DATA-L/DATA-U to DATA (their only fallbacks are USB/LSB->SSB,
   CW-R->CW, RTTY-R->RTTY). So these are new reachable values, not changed ones.
   They are still the panel's factory presets for the digital modes and have no
   cited source; worth confirming against the radio in the PR 2 / hardware pass.
7. `int | None` now contradicts `core/radio_protocol.py: DspControlCapable.get_filter_width -> int`
   and the `read_filter_width -> int` stub in `backends/yaesu_cat/observations.py`.
   The body discloses this. Confirmed no gate catches it (mypy's scope is
   `src/rigplane/web` and its step did not run here), and the two poll callers are
   safe: `core/radio_state.py` declares `filter_width: int | None`, and the
   browser adapter already maps a missing width to `NaN`.

## Guardrails

8 files, +394/-56 = 450 changed lines: over the 6-file soft threshold, under the
10-file / 1000-line hard ceiling, and the body carries the one-unit-of-work
justification the soft threshold requires. No new abstraction (one profile field
with a real consumer), no orphan, no frontend or `src/rigplane/web/` change,
import-linter green.

morozsm:
Coordinator: applied both REQUIRED BEFORE MERGE corrections from the review to the PR body — the caller sentence now names the third caller (`validation/hardware.py: _check_filter_width_set` → `_nudge_filter`) and what happens there on `None`; the serialization line now quotes the real templates (`SH0{code:03d};` / `SH1{code:03d};`). The commit message carries neither claim, so the squash body needs no correction. The `rigs/_schema.md` row for `first_code` and the validator's clean skip go with PR 2.

PR #3551, opened 2026-09-21
fix(MOR-2514): a selected frequency digit is released by Escape and by focus leaving the readout

morozsm:
Closes the defect the owner hit on a live radio page on 2026-09-21 (MOR-2514): a digit of the frequency readout, once selected, could not be released — clicking the waterfall or the VFO block left it selected, the readout kept declaring `data-owns-arrows="vertical"`, and bare ArrowUp/ArrowDown never returned to the global tuning binding.

## What this changes

- `primitives/frequency/frequency-interaction.svelte.ts`: Escape releases the selected digit (consumed only while a digit is selected). A click on a digit always selects it, including the already-selected one — a first head made that second click toggle the selection off, which broke two existing pins in `src/semantic/__tests__/VfoSurface.test.ts` (they click the digit before each wheel step) and would have cost an operator the selection by clicking again before scrolling; the second commit removes it. The existing internal `reset()` is reused as the release; before this change its only callers were the lease's `clearOwner()` and the context/receiver-change effect.
- `primitives/frequency/StandardFrequencyReadout.svelte`: focus leaving the readout group releases the selection (`focusout` with a `relatedTarget`-outside-the-group check, so moving between digits does not release); after release `data-owns-arrows` is absent again.
- A hint on the selected digit says how to leave, through the i18n catalog (en-US, ru-RU, ja-JP), passed down through `FrequencyDisplayInteractive.svelte`, `frequency-instrument.svelte.ts` and `semantic/VfoSurface.svelte`.
- Modified arrows stay ignored by the readout (unchanged, MOR-2512).

## Tests

`src/primitives/frequency/__tests__/StandardFrequencyReadout.isolated.test.ts` on the real readout: RED → GREEN `15 passed` (the independent review re-measured RED at the merge-base as 4 failed | 11 passed) (one of them now pins that re-clicking the selected digit keeps it selected). `src/semantic/__tests__/VfoSurface.test.ts`: `2 failed | 192 passed` at the first head → `194 passed`. `src/components-v2/layout/__tests__/KeyboardHandler.test.ts`, composition with the real `KeyboardHandler`: RED `1 failed | 54 passed` → GREEN `55 passed` (before release a bare ArrowUp does not reach the global tuning action, after release it does). Existing lines in both test files are untouched (0 deletions), including the MOR-2364 no-digit pins. `npm run check`: 0 errors, 0 warnings.

## Size

10 files, +235 / -3, measured with `git diff --stat $(git merge-base origin/main HEAD)` — at the file ceiling, not over it; three of the ten are one-line locale entries. Over the 6-file soft threshold because the hint has to travel from the catalog through the instrument binding to the readout; the behaviour change itself is two files plus two test files.

## Known gap

Only `semantic/VfoSurface.svelte` passes the hint. Two other mounts of the readout do not: `semantic/ReceiverInstrumentHost.svelte`, and `components-v2/vfo/VfoPanel.svelte` (reached through `VfoHeader.svelte` and `DualVfoDisplay.svelte`, i.e. the v2 skin). Release by Escape and by focus leaving works on all of them; the hint text is absent on those two. Left out to stay inside the file ceiling; follows as a small PR on the same ticket (the second mount was found by the independent review). Not verified in a real browser: that a pointer click gives the readout group DOM focus (the tests use synthetic events plus a programmatic `.focus()`); the owner's check on the live page covers it.

Linear: MOR-2514.


morozsm:
Agent Review: PASS 63568b2607dfc7cd0d1e675fbae1ff06850d1035

Independent review, read-only, no edits left behind (`git status --short` empty; `git diff 63568b26` empty at the end).

## Mutations — every new pin discriminates

Each mutation applied in place, one test file run, restored with `git checkout --`.

| Mutation | File that went red |
|---|---|
| (a) Escape branch deleted from `frequency-interaction.svelte.ts:81-86` | `StandardFrequencyReadout.isolated.test.ts` 1 failed \| 14 passed **and** `KeyboardHandler.test.ts` 1 failed \| 54 passed |
| (b) `onfocusout=` binding deleted from `StandardFrequencyReadout.svelte:77` | `StandardFrequencyReadout.isolated.test.ts` 1 failed \| 14 passed |
| (c) inside-the-group guard deleted from `StandardFrequencyReadout.svelte:56` | `StandardFrequencyReadout.isolated.test.ts` 1 failed \| 14 passed |
| (d) click-toggle reinstated in `handleDigitClick` (`frequency-interaction.svelte.ts:75`) | `StandardFrequencyReadout.isolated.test.ts` 1 failed \| 14 passed **and** `VfoSurface.test.ts` 2 failed \| 192 passed |
| (e) en-US hint value changed by one character | `KeyboardHandler.test.ts` 1 failed \| 54 passed — the pin is a literal, not `expect(t(key))` |
| negative control: `handleFocusOut` rewritten to an equivalent `insideGroup` form | `StandardFrequencyReadout.isolated.test.ts` 15 passed — the check does not redden on correct work |

(d) independently reproduces the reason commit 63568b26 exists: the toggle is exactly what broke the two `VfoSurface.test.ts` pins.

## Claims checked

- **MOR-2364 pins untouched** — CONFIRMED. `git diff --numstat` shows `78 0` and `101 0`: zero deletions in both test files.
- **`data-owns-arrows` absent after release, as in the never-selected state** — CONFIRMED. `StandardFrequencyReadout.svelte:69` yields `undefined` (attribute omitted) once `selectedDigitIndex` is `null`; pinned both ways in the two suites.
- **Escape with no selection stays global** — CONFIRMED by reading: `frequency-interaction.svelte.ts:80` returns before the Escape branch, so neither `preventDefault` nor `stopPropagation` runs; pinned by `expect(bare.defaultPrevented).toBe(false)`. Swept the class: no shipped keyboard binding uses Escape (`grep` over `src/lib` found none); the Escape consumers are dialogs/popovers (`RadioLayout.svelte:126`, `BottomSheet.svelte:154`, `SendReportDialog`, `FrequencyEntryDialog`, `ScopeSettingsPopover`) and each moves focus into itself on open, which itself releases the selection through the new `focusout`.
- **Modified arrows still ignored** — CONFIRMED, `hasCommandModifier` remains the first line of `handleKeyDown`; unchanged by the diff.
- **Stale lease cannot release the current owner's selection** — CONFIRMED. `createFrequencyInteractionLease.releaseSelection` uses the same `active(true)` gate as every sibling method; `active()` returns `false` at its first line when `activeRevocation.get(owner) !== revoke`, so `latch`/`clearOwner` is never reached from a superseded lease. No new hazard over the pre-existing methods.
- **Layering** — CONFIRMED. The only new import in `primitives/` is `import type { DigitInfo } from './frequency-tuning'`. The hint arrives as an already-localized string; `t()` is resolved in `semantic/VfoSurface.svelte`, the same place `pendingAnnouncement` is. `npm run lint` and `npm run check` run inside the green `quick` job at this head.
- **i18n** — CONFIRMED. All three bundled locales carry `core.vfo.freq.selectedDigitHint` (read directly out of the three JSON files). Parity is gated by `npm run i18n:check` in `quick.yml:212`, not by a vitest file — `locale-contract.test.ts` is locale *resolution* and stays green even with the ru-RU key deleted, so it is not the parity gate.
- **Class sweep for the selection mechanism** — `handleDigitClick` has exactly one consumer in the tree (`StandardFrequencyReadout.svelte:47`), so no second digit-selecting renderer needs the same fix.
- **Size** — CONFIRMED: 10 files from `gh pr view 3551 --json files`, +235/-3. At the hard ceiling, not over; the soft threshold is crossed and the body justifies it.

## REQUIRED BEFORE MERGE (PR body — editable in place, head SHA unaffected)

1. **The "RED 5 failed | 10 passed" figure is not reproducible at this head.** Checking out the five production files at the merge base and running `npx vitest run src/primitives/frequency/__tests__/StandardFrequencyReadout.isolated.test.ts` gives **4 failed | 11 passed**, not 5/10. The fifth row — "keeps the digit selected when it is clicked again" — passes at the merge base, because the toggle it guards against never existed on `main`; it was red only at the abandoned first head. Replace the figure with `4 failed | 11 passed → 15 passed`. The test is still worth its place: mutation (d) shows it reddens against the regression.
2. **The "Known gap" list is narrower than the gap.** `components-v2/vfo/VfoPanel.svelte:184` mounts `FrequencyDisplayInteractive` with no `selectedDigitHint` either, and it is production-reachable through `LegacyVfoPanelAdapter` from `components-v2/layout/VfoHeader.svelte:163` and `components-v2/panels/vfo/DualVfoDisplay.svelte:50`/`:64`, neither of which passes `onFrequencyClick` — so `frequencyEntryButton` is false there and the digits really are selectable. Release works (it lives in the readout component); only the title text is missing, exactly as for `ReceiverInstrumentHost.svelte`, and consistent with `VfoPanel` already not passing `pendingAnnouncement`. Name both mounts.

## MANDATORY SQUASH-BODY CORRECTION (commit 30af7001)

Squash merge concatenates branch commit messages, so these two sentences from 30af7001 would land on `main` describing behaviour 63568b26 removed. Drop or rewrite them when composing the squash body:

- "`frequency-interaction.svelte.ts`: handleDigitClick toggles the selection off when the clicked digit is already selected" — false at this head; `handleDigitClick` assigns unconditionally.
- "Tests: StandardFrequencyReadout.isolated pins Escape release and consumption, focusout-outside release vs digit-to-digit retention, **the toggle click**, the title, and group focus on selection" — the pin is now the opposite: a repeated click keeps the selection.

Everything else in both messages checks out, including "tabindex unchanged" (`StandardFrequencyReadout.svelte:73` is not in the diff).

## Minor, in the diff — would need a commit, not blocking on its own

`src/lib/i18n/locales/ru-RU.json:200` — "Esc или щелчок вне разряда снимает выделение" is narrower than the two sibling strings and has a reachable counterexample: a click on a *different* digit is a click "вне разряда" and moves the selection rather than removing it (the behaviour the `keeps the digit selected … and moves on a different digit` pin asserts). en-US "click away" and ja-JP "範囲外のクリック" both point at the readout area instead. Practical harm is near zero — nobody deselects by clicking a neighbouring digit — so this is not worth a round trip by itself; fold "щелчок вне поля частоты" (or similar) into any commit this branch still takes.

## UNVERIFIED — needs the live page, not a block

These follow from real browser focus semantics and I could not exercise them here (jsdom does not implement click-focus; the tests use synthetic events plus a programmatic `.focus()`).

1. **Click on the waterfall releases.** `.freq` carries `tabindex=0` when not inert, so a pointer press on a digit focuses the group; `components/spectrum/SpectrumPanel.svelte: handleDragStart` does **not** call `preventDefault`, so the press moves focus off the group and `focusout` fires with `relatedTarget === null`, which `handleFocusOut` treats as outside. Check: select a digit, click the waterfall, confirm the highlight clears and bare ArrowUp tunes globally again.
2. **Tab away releases.** Same handler, `relatedTarget` non-null and outside. Check: select a digit, press Tab, confirm the highlight clears.
3. **Mid-tuning re-render does not release.** Focus sits on the `bind:this` root, which a frequency update never re-creates (the digits are text updates inside unkeyed `{#each}` blocks), so no `focusout` should fire. Check: select the 1 Hz digit, hold ArrowUp for several steps, confirm the selection survives.
4. **PTT does not release** — `components-v2/panels/TxPanel.svelte:254` calls `preventDefault()` on `pointerdown`, which suppresses the focus change, so a PTT press leaves the digit selected. Almost certainly desirable; worth an eyeball.
5. **Window blur / tab switch does not fire `focusout`** in any major engine, so the selection survives an app switch. Stated as intended behaviour, not a defect.
6. **Ticket mechanism, for the owner's own record.** `components-v2/layout/keyboard-map.ts: focusedElementOwnsArrowKey` resolves ownership from `document.activeElement.closest('[data-owns-arrows]')`, so once focus genuinely leaves the readout the global arrows already work even on `main`. The part of MOR-2514 that reproduces from code reading is the stuck highlight plus arrows being re-stolen the next time the readout takes focus — including by Tab, with nothing selected by the operator. This PR fixes that; if the live page showed arrows staying captured with focus demonstrably elsewhere, there is a second cause worth a separate ticket.

## CI as found

`quick` pass (6m26s), `visual` pass (1m15s), `classify`/`grep-gate`/`changes` pass. `Agent Review Gate` fail — no exact-head directive existed yet, which this comment supplies. `quick-v2-observe` fail: "pull request base is not the live main head" — the branch's merge base is `21fa3063` while `origin/main` is now `2d128601`. Main movement alone does not invalidate this review or the green required runs; the coordinator should still re-verify required statuses at the exact head immediately before merging.

morozsm:
Coordinator: applied the review's REQUIRED BEFORE MERGE corrections to the PR body — the RED count that did not reproduce is replaced by the review's measurement, and the known-gap section now names both mounts that do not pass the hint (`ReceiverInstrumentHost.svelte` and `components-v2/vfo/VfoPanel.svelte`). The squash body is written by hand at merge without the two sentences about the removed click-toggle. The ru-RU wording note and the two missing hint mounts go into a follow-up PR on the same ticket.

PR #3552, opened 2026-09-21
fix(MOR-1679): encode the FTX-1 width write against the radio's live mode, not the legacy mirror

morozsm:
Follow-up to #3550 for MOR-1679, found on hardware the same day. The ticket stays open (browser side still to come).

## The defect, as measured on a live FTX-1 (RX only, 2026-09-21)

SUB receiver in DATA-U at 3200 Hz. `set_filter_width(500, receiver=1)` sent the radio to **100 Hz**; `set_filter_width(2900, receiver=1)` sent it to **3200 Hz**. Both are what the SSB table produces: 500 Hz → nearest SSB entry 400 = code 02, which in the CW/DATA family is 100 Hz; 2900 Hz = SSB code 19, which in the CW/DATA family is 3200 Hz. The second request was chosen to test that prediction. The MAIN / USB write checked earlier was correct only because there the mirror and the radio agreed.

Cause: `YaesuCatRadio.set_filter_width` chose the width table — and decided the fixed-mode refusal — from the mode held in the legacy `self._state` mirror, which the observation adapter never writes. The read path already avoided this by reading the mode from the radio.

## What this changes

- `backends/yaesu_cat/radio.py: set_filter_width` reads the receiver's live mode with `read_mode(receiver)`, as `read_filter_width` does, and encodes against that mode's table. The fixed-mode refusal uses the same live mode.
- A live mode with no width table now refuses with `CommandError` instead of sending the Hz value as the SH code. On the FTX-1 the only such modes are C4FM-DN / C4FM-VW, and the independent review measured that their mode answer does not fit `read_mode`'s one-character parse template (`MD0{mode};`), so on the radio the write is refused earlier, by `CatParseError` from `read_mode`; either way nothing is written. That `read_mode` cannot parse the C4FM answers is a separate defect, tracked on its own.
- `_filter_width_table` loses its mirror fallback and its `receiver` parameter: after this change its only caller is `read_filter_width`, which always passes the resolved mode.
- Encodings other than `table_index` still send the value as given. No Yaesu profile uses one today (`rigs/ftx1.toml` is the only Yaesu rig and declares `table_index`); the branch is kept because the loader's default encoding is not `table_index`, and one test pins it.

Cost: one extra mode query per width write.

## Pins (RED before, GREEN after; the review re-measured RED as 7 failed / 1 passed with `-k set_filter_width`)

Mirror says USB, radio answers DATA-U, set 500 Hz on SUB → `SH1010;` (was `SH1002;`). Mirror says USB, radio answers CW-U, set 4000 Hz → `SH0021;`. Mirror says USB, radio answers FM → refused, nothing written. Mirror says FM, radio answers USB, set 2400 Hz → `SH0013;` (no false refusal). `read_mode` returning C4FM-DN → refused, nothing written. The existing `SH0013;` / `SH1013;` pins stay green.

## Checks run by the builder (single files)

`tests/test_ftx1_radio.py` 425 passed; `tests/test_dsp_filter_family.py` 77 passed; `ruff check` and `ruff format --check` clean on the changed paths; `lint-imports` 5 kept, 0 broken.

3 files, +155 / -44, measured with `git diff --stat $(git merge-base origin/main HEAD)`. Not verified on hardware yet — the repeat of the measured step (SUB DATA-U, set 500 Hz, radio's screen) follows the merge.

Linear: MOR-1679 (remains open).


morozsm:
Agent Review: PASS c4c77b78c2f9737ff1c5d354422e99ee737e75ea

Independent review at the exact head (`git rev-parse HEAD` and `gh pr view 3552 --json headRefOid` agree; base `cb5af903`, 3 files +155/-44 per `git diff --stat $(git merge-base origin/main HEAD)`). The production change is correct, is an improvement over the base, and its new pins discriminate. No blocking code defect. Three corrections below: one squash-body, one PR-body, one prose item the coordinator decides on.

## Code verification

- **Exhaustive enumeration of the reachable input space.** Every single-char MD answer `0`..`F` driven through `backends/yaesu_cat/radio.py: YaesuCatRadio.set_filter_width` with the real `rigs/ftx1.toml` profile and a mocked transport (throwaway script, nothing committed), `set_filter_width(500)`: LSB/USB -> `SH0002;`; CW-U/CW-L/DATA-L/DATA-U/RTTY-L/RTTY-U/PSK -> `SH0010;`; FM/AM/FM-N/AM-N/DATA-FM/DATA-FM-N -> `CommandError` (fixed); unmapped `0` -> `CommandError` (no table). 500 Hz is code 10 in DATA-U, and code 02 — what the mirror path sent — is 100 Hz in that table: the bench symptom reproduces from the tables alone.
- **Mode strings match.** `profiles/__init__.py: RadioProfile.resolve_filter_rule` upper-cases and maps `_`->`-`, and `read_mode` returns the profile's own hyphenated names, so no case/underscore gap. `None` and `UNKNOWN(x)` now refuse instead of writing a raw code — strictly better than the base, which wrote `width_hz` as the SH code.
- **Concurrency.** `backends/yaesu_cat/transport.py` serialises every `write`/`query` through one exchange gate, so the new MD query and the SH write are two atomic exchanges. A poll may interleave between them (the mode could change under us); no exchange can be corrupted. Acceptable.
- **Failure path.** Neither `web/radio_poller.py` (SetFilterWidth case) nor `backends/yaesu_cat/poller.py: _execute_command` wraps the call, so a failing `read_mode` propagates and `filter_width_changed` is not emitted. No silent success on the page. Cost is the one extra MD exchange per write the body states.
- **Narrowed helper.** `grep -rn "_filter_width_table" src/ tests/` -> definition plus exactly one call, in `read_filter_width`. Nothing passes the removed `receiver` argument. The body's "its only caller" claim is true.
- **Kept raw path.** Pre-existing behaviour preserved, not a new orphan: the base already sent `width_hz` raw whenever no table resolved, and `profiles/rig_loader.py`'s default encoding is `segmented_bcd_index`. `rigs/ftx1.toml` is the only rig declaring `encoding = "table_index"`. `test_set_filter_width_raw_index_keeps_non_table_encoding` pins that compat path and reddens when it is broken (below).

## Mutation probes — production mutated, tests untouched, restored after each

| mutation | result |
|---|---|
| setter reads the mode from `self._state` again | `tests/test_ftx1_radio.py` RED, 15 failed |
| `+ filter_width_first_code` dropped in the setter | `tests/test_ftx1_radio.py` RED 9 failed; `tests/test_dsp_filter_family.py` RED 2 failed |
| no-table refusal replaced by the base raw-write fallback | `tests/test_ftx1_radio.py` RED, 1 failed (`test_set_filter_width_refuses_live_mode_without_table`) |
| fixed-mode refusal reads the mirror | `tests/test_ftx1_radio.py` RED, 8 failed |
| raw-index path off by one | `tests/test_ftx1_radio.py` RED, 1 failed (`test_set_filter_width_raw_index_keeps_non_table_encoding`) |
| whole base `radio.py` reinstated under the head tests | `-k set_filter_width` RED, 7 failed 1 passed |
| control: behaviour-preserving refactor of the index expression | GREEN both files (425 / 77) |

At head both files are green: `tests/test_ftx1_radio.py` 425 passed, `tests/test_dsp_filter_family.py` 77 passed. Working tree left byte-identical, `git status --short` clean.

## Corrections

1. **MANDATORY SQUASH-BODY CORRECTION.** The commit body says "a live mode with no width table (C4FM-DN/C4FM-VW) now refuses with CommandError instead of sending width_hz as the raw code". Measured against the real profile: the mode parse template is `MD0{mode};`, a single character, so the radio's actual C4FM answers `MD010;` / `MD011;` raise `CatParseError` inside `read_mode` before the new branch is reached. Those two modes therefore never produce the `CommandError` the sentence promises; on this profile the branch's only reachable trigger is an unmapped single-char code. Reword at merge time to "a live mode that resolves to no width table", without naming C4FM.

2. **REQUIRED BEFORE MERGE (PR body edit).** "Pins (RED before: `6 failed, 2 passed`)" does not reproduce. With the base `radio.py` restored under the head tests: `uv run pytest tests/test_ftx1_radio.py -q -k "set_filter_width"` -> **7 failed, 1 passed**; `-k "set_filter_width or fixed_mode_width"` -> **13 failed, 1 passed**. Replace the stated counts with a measured pair and name the selector.

3. **Prose, coordinator's call (one-line commit if taken).** The same unreachable-C4FM framing is in the new docstring at `src/rigplane/backends/yaesu_cat/radio.py:1724`. It mirrors untouched pre-existing wording in the neighbouring `read_filter_width` docstring at line 1676, so an honest fix drops the `(C4FM)` parenthetical from both. Not blocking: prose only, and the code it describes is correct.

## Notes, no action required

- Class sweep for "a Yaesu write decides from the legacy `self._state` mirror" (`grep -n "self\._state" src/rigplane/backends/yaesu_cat/radio.py`). Fixed here: `set_filter_width`. Still open, pre-existing and out of scope: `YaesuCatRadio.set_nb` and `YaesuCatRadio.set_nr` (restore the previous level from `self._state.main.nb_level` / `nr_level`, MAIN-only, ignoring `receiver`), and `YaesuCatRadio.get_data_mode` (derives the DATA flag from `self._state.main.mode`, consumed by `rigctld/handler.py` and `rigctld/server.py`).
- `tests/test_ftx1_radio.py: _FIXED_MODE_HEX` is a hand-maintained map beside the derived `[modes] list` in `rigs/ftx1.toml`. All six entries check out against that list (1-based index -> hex nibble). But a row only has to land on *some* fixed mode to stay green, so a future drift between two fixed modes would not redden the file.
- CI as found (`gh pr checks 3552`): `quick` pass 1m37s, `classify` / `grep-gate` / publishers pass, `quick-v2-observe` pending, `Agent Review Gate` fail for the missing directive this comment supplies. Confirming green at this exact head immediately before merge remains the coordinator's step.

morozsm:
Coordinator: applied the review's REQUIRED BEFORE MERGE corrections — the PR body no longer states RED counts that do not reproduce and now says that on the radio the C4FM refusal comes from `read_mode`'s parse, not from the `CommandError` branch; the squash body is written by hand at merge with the same correction. The docstring's C4FM example is corrected with the separate `read_mode` C4FM fix, not here.

PR #3553, opened 2026-09-21
fix(MOR-1679): pin the FTX-1 filter-width rendering and write path in the browser against CAT Table 5

morozsm:
Browser half of MOR-1679's acceptance box "full-table MAIN/SUB and browser render/write tests". The backend halves are #3550 and #3552.

## Defect found and fixed

`components-v2/panels/FilterPanel.svelte`: in a fixed-width mode (FM 16000, AM 9000, AM-N 6000 on the FTX-1) the width control stayed operable — `filterWidthInput().enabled` was true and `handleTableReset` emitted a width write (the RED test saw a call with `[3200]`). Since #3552 the backend refuses such a write with `CommandError`, so the page was offering a control that can only fail. The control is now not operable in a fixed mode and emits nothing.

No defect was found in the mode resolver, the props layer, the command layer or `AmberFilterGhost`; their new tests were green at once and are pins.

## Tests

- `lib/runtime/adapters/__tests__/ftx1-filter-width-conformance.test.ts` (40): every FTX-1 mode name resolves to the right `filterConfig` entry — LSB/USB → SSB (23 values, 300..4000), each CW-family mode → its own 21-value table (50..4000), fixed modes → fixed, C4FM-DN / C4FM-VW → none.
- `components-v2/panels/__tests__/FilterPanel.ftx1-table5.isolated.test.ts` (11; the independent review measured 4 of them RED with both production hunks reverted): stepping only emits table values, bounds are the table's ends (never 4500), fixed modes are not operable and emit nothing, a `null` width renders the existing unavailable treatment and emits nothing.
- `components-v2/panels/lcd/__tests__/AmberFilterGhost.ftx1.isolated.test.ts` (3): normalization uses the corrected bounds.
- `lib/runtime/commands/__tests__/panel-commands.intent.isolated.test.ts` (+8): the emitted command is `set_filter_width` with the table's Hz and the active receiver, through the real command layer.

Fixture `ftx1-capabilities.json`: produced offline by running the real `WebServer._serve_capabilities` over the real `rigs/ftx1.toml` with a fake radio (the scheme `tests/test_web_capability_guards.py` uses). It is a PROJECTION of that output — the sections the filter-width contract reads, values untouched — because the full capture is 1573 lines and would not fit the PR ceiling. It is therefore not a byte-for-byte capabilities fixture like the ic7300 one.

`npm run check`: 0 errors, 0 warnings.

## Size

7 files, +587 / -3 = 590 changed lines at head `65554ad3`, from `gh pr view 3553 --json additions,deletions,files` — over the 6-file soft threshold, under the 600-line one: one production file, five test/fixture files that exercise one contract (the Table 5 widths as the browser sees them), and one existing test file extended.

## Not covered

The presets path in the settings modal can still send `set_filter_width` for a mode with no `filterConfig` entry (C4FM); pre-existing, shared, outside this PR.

Linear: MOR-1679.


morozsm:
Agent Review: BLOCKED f8578e50400ee3f6d3f6769ec6b3641227b30aaf

Independent verification at head `f8578e50`. The production change, the fixture
and the pins are sound — I could not refute them. Two statements in the prose
this PR adds are false as written, and only a commit can change them.

## What I verified (no defect found)

- **Fixture is genuinely the serializer's output.** Rebuilt `/api/v1/capabilities`
  offline with the `tests/test_web_capability_guards.py` scheme (`_make_radio("FTX-1")`
  → `WebServer._serve_capabilities` → `_parse_json_body`) and compared to
  `fixtures/ftx1-capabilities.json`: 25 fixture keys, **all 25 value-identical**,
  0 keys absent from the real payload, 18 real keys deliberately projected out.
  No hand-edited or invented value. Negative control: perturbing `filterConfig.SSB.table[0]`
  and `filterWidthMin` in a copy made the same comparison report inequality, so the
  "identical" result is a measurement.
  Header claims also checked: backend HEAD `60d05a42` is the merge base and has
  `2d128601` (#3552) / `cb5af903` (#3550) as ancestors; the "retired 4500-era ceiling"
  really was in `rigs/ftx1.toml` until `cb5af903`; "full capture 1573 lines" reproduces.
- **No other-radio regression.** `grep -n '^table = ' rigs/*.toml`: the only other hits
  are `table = "standard_50"` (CTCSS) in ic705/ic7300/ic9700. The FTX-1 is the **only**
  profile with table-mode filter widths, so `isTableMode` is false for every Icom.
  IC-7300/IC-7610 `[filters.width.FM] fixed = true` has **no** `table`, so it renders the
  `{:else}` branch: read-only `BW` row, no `[role=slider]`, no Reset button — neither
  changed line is reachable there. I also checked the one path where the unrendered
  binding is still consumed in non-table mode (`FilterPanel.svelte: consumeReadOnlyWidthStatus`
  via the `$effect` on `filterWidthBinding.view`): `politeAnnouncement` comes from
  `projectControlFeedbackPresentation(input.feedback, …)`, which does not read `input.enabled`,
  so the FM live-region status is unaffected. `FilterPanel.isolated.test.ts` 111/111 green.
- **Mutations** (each applied to the tree, run, reverted):
  | mutation | result |
  |---|---|
  | revert the `enabled` guard | RED — `FilterPanel.ftx1-table5.isolated.test.ts`, 3 failed |
  | revert the `handleTableReset` guard | RED — same file, 1 failed (observed call value `3200`) |
  | both reverted | RED — same file, **4 failed** / 7 passed |
  | alias `DATA-U` → `SSB` in `panel-props.ts: resolveFilterModeConfig` | RED — `ftx1-filter-width-conformance.test.ts`, 3 failed |
  | `ftx1-capabilities.json` SSB `table[0]` 300→350 | RED — conformance 3 failed **and** FilterPanel 2 failed |
  | `adjacentWidthChoice` returns `4500` past the table end | **green** — not a hollow pin: the candidate is re-quantised by `policy.normalize` (`nearestWidthChoice`) back to 4000, so 4500 cannot reach the handler through that branch |
  | append `4500` to `currentWidthTable` (makes 4500 genuinely reachable) | RED — 3 failed, incl. both "never 4500" rows |
  | control: equivalent refactor (reorder the `enabled` conjuncts, rename `defaultWidth`) | green, as it should be |
- MAIN and SUB are both on the write path (`receiver: 0` and `receiver: 1`) in the
  MOR-1679 describe of `panel-commands.intent.isolated.test.ts`. The `null`-width
  rendering is the pre-existing `formatExactWidthDisplay(NaN)` → `--- Hz` treatment,
  not invented.
- `vi.mock('$lib/runtime/adapters/panel-adapters')` resolves: same specifier
  `FilterPanel.svelte` imports, and all seven imported symbols are stubbed — a
  non-resolving specifier would have thrown in `getFilterHandlers()`.
- CI at this head as found: `quick` **pass** (4m55s; the "Frontend install + check +
  tests + build" step ran and succeeded, so `npm run check` is covered — I did not
  rerun it), `visual` **pass**, `classify`/`changes`/`grep-gate` pass,
  `quick-v2-observe` pending, `Agent Review Gate` red for want of this directive.
- Guardrails: `gh pr view --json files,additions,deletions` → 7 files, +597/-3 = 600.
  Over the 6-file soft threshold and at the 600-line one; the PR body justifies the
  unit of work, and the 10-file/1000-line hard ceiling is not approached.

## BLOCKING — false prose in the diff (needs a commit)

**B1. `frontend/src/lib/runtime/adapters/__tests__/ftx1-filter-width-conformance.test.ts:8`**
> "Literal expectations only — never derived from the fixture under test."

Contradicted 53 lines later in the same file, at
`ftx1-filter-width-conformance.test.ts:61`:
`expect(resolved).toBe(caps.filterConfig?.[key]);` — that expectation *is* the object
read out of the fixture under test. Risk: the sentence tells the next reader the whole
suite is independent of the fixture, so they stop checking; a future row that really is
self-satisfying then inherits the reassurance. (The row itself is fine — the
DATA-U→SSB mutation killed it.)
**Required fix:** delete the sentence, or narrow it to the assertions it actually
describes (e.g. "every expected value is a literal; the key-mapping rows assert
identity against the resolved fixture entry").

**B2. `frontend/src/lib/runtime/adapters/__tests__/fixtures/ftx1-profile.ts:17-20`**
> "Pins against this fixture: `ftx1-filter-width-conformance.test.ts`,
> `FilterPanel.ftx1-table5.isolated.test.ts`, and the MOR-1679 describe in
> `panel-commands.intent.isolated.test.ts`."

The list is closed ("and"), and it is already incomplete at the head that introduces it:
`frontend/src/components-v2/panels/lcd/__tests__/AmberFilterGhost.ftx1.isolated.test.ts:15`
imports `FTX1_CAPABILITIES` from this very loader and pins against it — a fourth
consumer, added by this same PR. `grep -rn "FTX1_CAPABILITIES" frontend/src` returns
four consumer files, not three. Risk: this is a hand-maintained list at a boundary;
whoever re-generates the fixture will use it to decide what to re-run and will miss the
ghost test.
**Required fix:** add the fourth file, or delete the enumeration (nothing depends on it —
the grep above is authoritative).

## REQUIRED BEFORE MERGE — PR body (editable in place, no commit)

**P1.** "`FilterPanel.ftx1-table5.isolated.test.ts` (11; **2 RED** before the fix)".
Measured: reverting both production hunks at this head leaves **4** failing tests
(`FM`/`AM`/`AM-N` fixed-mode rows plus the Reset row), 7 passing. Replace "2 RED" with
"4 RED", or state the pre-parametrisation count it actually refers to.
Command: revert both hunks, then
`npx vitest run src/components-v2/panels/__tests__/FilterPanel.ftx1-table5.isolated.test.ts`.

## Notes (no action required)

**N1. The fixture sits outside the repository's fixture drift guard.**
`conformance/fixture-provenance.guard.test.ts` runs only over `conformance/profiles.ts`,
whose `PROFILES` has the single `ic7300` entry, and its sidecar cross-check is hardcoded
to `ic7300-provenance.json`. The FTX-1 fixture joins neither (joining the first would
need a state fixture it does not have). Consequence, confirmed by construction: a future
correction to `rigs/ftx1.toml` turns no frontend test red, and these files keep pinning
the old Table 5 while their headers say "CAT 2508-C Table 5". A `ftx1-provenance.json`
sidecar carrying the backend SHA would at least make the staleness visible. Out of scope
for this PR.

**N2. `FilterPanel.svelte: handleTableReset` still hardcodes `const defaultWidth = 3200`.**
Pre-existing and not worsened — 3200 is a member of every non-fixed FTX-1 Table 5 row
(SSB and all seven CW-family rows), and the FTX-1 is the only table-mode radio — so it
never sends a value the server must snap. Worth a ticket only if a second table-mode
profile ever lands.

**N3.** `FilterPanel.svelte:252` describes the fixed family as "AM/AM-N/FM family"; the
profile's fixed rows are AM, AM-N, FM, FM-N, DATA-FM, DATA-FM-N. Read as shorthand it is
fine; if the sentence is touched for B1/B2, "the six fixed rows" is exact.

**N4.** `expect(x).not.toBe('4500')` immediately after `expect(x).toBe('4000')` is
strictly implied by its neighbour and cannot fail on its own. Harmless as documentation
of intent; the real discrimination comes from the `toBe('4000')` assertion, which the
append-4500 mutation killed.

- `npx vitest run src/lib/runtime/adapters/__tests__/ftx1-filter-width-conformance.test.ts`
- `npx vitest run src/components-v2/panels/__tests__/FilterPanel.ftx1-table5.isolated.test.ts`
- nothing else: B1/B2 are comment-only, so the required CI for the new head is the
  natural `quick` run.

Tree left byte-identical: `git status --short` clean after every mutation and at the end.

morozsm:
Agent Review: PASS 65554ad3838205b520db20906f98c284152e1ebd

**Delta review** on top of my full review of `f8578e50`
(https://github.com/rigplane/rigplane-core/pull/3553#issuecomment-5765684483). Covered:
the one commit `f8578e50..65554ad3` only. Everything established in the full review —
fixture-vs-serializer equality, the seven mutations, the other-radio analysis, guardrails
— is unaffected by a comment-only delta and is not re-litigated here.

## 1. The delta is comment-only — established, not assumed

`git diff --stat f8578e50 HEAD` → 5 files, +16/-26, no file added or removed.

Instrument: `git archive` of both revisions into scratch, then for each of the 5 files
drop every line whose trimmed form starts with `//`, `*`, `/*` or `<!--` and compare the
remainder byte-for-byte:

| file | non-comment lines |
|---|---|
| `FilterPanel.svelte` | SAME, 959 = 959 |
| `FilterPanel.ftx1-table5.isolated.test.ts` | SAME, 231 = 231 |
| `fixtures/ftx1-profile.ts` | SAME, 4 = 4 |
| `ftx1-filter-width-conformance.test.ts` | SAME, 106 = 106 |
| `panel-commands.intent.isolated.test.ts` | SAME, 2776 = 2776 |

**Negative control** (the comparison must be able to say "different"): the same check
against a scratch copy of the new `FilterPanel.svelte` with `const defaultWidth = 3200`
changed to `3201` reports `DIFFERS-NONCOMMENT` and prints that exact hunk. So "SAME" is a
measurement.

Second, independent instrument: `ts.createScanner(…, skipTrivia=true)` token streams of
both revisions. Identical for `FilterPanel.ftx1-table5.isolated.test.ts` (1479 tokens),
`ftx1-profile.ts` (23) and `ftx1-filter-width-conformance.test.ts` (968). For
`FilterPanel.svelte` and `panel-commands.intent.isolated.test.ts` the scanner mis-lexes a
backtick and swallows whole regions into template-literal tokens, so it reports a
difference; I read those two token payloads and the difference inside them is exactly the
comment text — the instrument degrades there rather than disagreeing, which is why the
line-filter above is the load-bearing one.

## 2. B1 and B2 are gone, and nothing equally false replaced them

`grep -rn "Literal expectations\|Pins against this fixture" frontend/src` → no matches.

I read every `+` line of the delta and asked of each whether it could be false without a
test failing:

- `FilterPanel.svelte`, width-control comment — now cites
  `'MOR-1679 FTX-1 fixed-width modes are not operable'`.
  `grep -n "describe('MOR-1679 FTX-1 fixed-width modes are not operable'"` →
  `FilterPanel.ftx1-table5.isolated.test.ts:197`. Exact match. The remaining claim
  ("the control reads the fixed Hz disabled") is what that describe asserts.
- `FilterPanel.svelte`, `handleTableReset` comment — now cites
  `'the table-branch Reset button resets IF shift but emits no width write'`.
  `grep -n "it('the table-branch Reset button resets IF shift but emits no width write'"` →
  same file, line 213. Exact match. Both citations were also confirmed live: reverting
  each production hunk reddens exactly those tests (full review).
- `FilterPanel.ftx1-table5` header — the three surviving claims (table-end bounds, fixed
  modes not operable, `--- Hz` for null) each have an assertion in the file.
- `fixtures/ftx1-profile.ts` — "values are the serializer's own output, unedited" is the
  claim I reproduced value-by-value in the full review. Still true.
- `ftx1-filter-width-conformance.test.ts` header — "Pins `resolveFilterModeConfig` and
  `toFilterProps` over the mode table below" is weaker and establishable; it no longer
  claims completeness or literal-only expectations.
- `panel-commands.intent.isolated.test.ts` describe doc — the deleted sentence was about
  server-side SH encoding, which nothing in this file enforces; what remains is what the
  four `it`s assert.

No new unenforced claim. The two test-name citations can go stale under a rename without a
test failing, but naming an enforcing test is exactly the tie the repository asks for, and
matches the existing convention in this file (the ru-RU chip comment).

## 3. Tests re-run at this head, one file at a time, from `frontend/`

| `npx vitest run <file>` | result |
|---|---|
| `src/lib/runtime/adapters/__tests__/ftx1-filter-width-conformance.test.ts` | 1 file, **40 passed** |
| `src/components-v2/panels/__tests__/FilterPanel.ftx1-table5.isolated.test.ts` | 1 file, **11 passed** |
| `src/components-v2/panels/lcd/__tests__/AmberFilterGhost.ftx1.isolated.test.ts` | 1 file, **3 passed** |
| `src/lib/runtime/commands/__tests__/panel-commands.intent.isolated.test.ts` | 1 file, **150 passed** |

No parse/transform error in either the edited `.svelte` or the edited `.ts` files.

## REQUIRED BEFORE MERGE — PR body (edit in place, no commit)

**P2. The Size section is stale at this head.** It says "over the 6-file soft threshold and
**at the 600-line one**" and cites `git diff --stat $(git merge-base origin/main HEAD)`.
Run at `65554ad3`, that command gives **7 files, +587 / -3 = 590 changed lines**
(`gh pr view 3553 --json files,additions,deletions` agrees: 7 / 587 / 3). So the PR is now
*under* the 600-line soft threshold and only the 6-file one is crossed; and the cited
command produces that 587/3 total, not the two per-commit figures listed beside it.
Suggested: "7 files, +587 / -3 = 590 changed lines at head `65554ad3`
(`git diff --stat $(git merge-base origin/main HEAD) HEAD`) — over the 6-file soft
threshold, under the 600-line one."
(P1 from the previous round is applied and correct.)

The squash body concatenates both commit messages. Two things in it must not land on
`main` as written:

**S1 (first commit, `f8578e50`).** "resolveFilterModeConfig / toFilterProps pins for all 17
FTX-1 state modes". Only `resolveFilterModeConfig` is pinned over all 17 — the
`expectedKey` table in `ftx1-filter-width-conformance.test.ts` has 17 rows. `toFilterProps`
is pinned for **8** of them (LSB, USB, CW-L, DATA-U in the bounds table; AM-N, AM, FM in
the fixed table; C4FM-DN in the null-width case). CW-U, RTTY-L, RTTY-U, DATA-L, PSK, FM-N,
DATA-FM, DATA-FM-N and C4FM-VW never reach `toFilterProps` in that file. Write it as
"`resolveFilterModeConfig` pins for all 17 FTX-1 state modes; `toFilterProps` pins for 8 of
them".

**S2 (second commit, `65554ad3`).** Its closing line "No behaviour change, no test change."
is true of that commit and false of the squashed change, which *does* change behaviour
(the disabled fixed-mode width control). Scope it — e.g. "No behaviour change in this
correction; comments only." — or drop it.

Everything else in both commit bodies checks out: the 14 `filterConfig` keys, the
Table 5 tables and the 50/4000 bounds are the serializer's real output; the "full 23-entry
sweep", the MAIN 0 / SUB 1 routing, the off-table rejection and the AmberFilterGhost 4000
bound are all asserted; and every deletion the second commit claims to have made is absent
from the tree at this head.

## Note carried forward (no action)

The FTX-1 fixture still sits outside `conformance/fixture-provenance.guard.test.ts`
(`PROFILES` holds only `ic7300`), so a future `rigs/ftx1.toml` correction reddens nothing
in the browser. Unchanged by this delta; out of scope for this PR.

## CI at this head, as found (not waited on)

`visual` **pass** (1m22s) · `quick` **pending** · `quick-v2-observe` pending ·
`classify`/`changes`/`grep-gate`/observation publishers pass · `Agent Review Gate` red for
want of this directive. The previous head's `quick` covered `npm run check`; the new head's
run had not finished when this was written, so the body's "`npm run check`: 0 errors, 0
warnings" is exact-head-unverified until `quick` reports.

Tree left byte-identical; `git status --short` clean, nothing written inside the worktree.

morozsm:
Coordinator: applied the delta review's REQUIRED BEFORE MERGE correction — the Size section now states 7 files, +587 / -3 = 590 at head `65554ad3`, under the 600-line threshold. The squash body is written by hand at merge and contains neither "toFilterProps pins for all 17 modes" nor "No behaviour change, no test change".

PR #3554, opened 2026-09-21
feat(MOR-2509): VFO panel skeleton — frequency and meter on one line, content-sized rows, compact control block

morozsm:
Guardrail exception (file count only, granted by the coordinating session under the owner's delegation of 2026-09-03): 12 code + 18 regenerated baselines = 30 files. The 18 are 15 PNGs, `manifest.json` and the two baseline READMEs' provenance records. Changed text lines are 566 + 220 = 786, under the line ceiling and above the 600-line soft threshold: this is one unit because the panel's markup, the bridge that must line up with it, the browser geometry spec that selects on that markup and the baselines that picture it cannot land separately without a red `main`.

Slice 1 of MOR-2509 (VFO panel redesign; layout approved by the owner 2026-09-21). Structure and sizing only — no new indicators, no colours, no power scale, no TX row.

## What this changes

- `components-v2/vfo/VfoPanel.svelte`: three rows per panel, marked `data-vfo-row="identity|main|chips"` — identity (name, mode, filter; the filter chip moved up from the control strip), main (frequency readout left, S-meter right, on one line), and the chip row, which wraps instead of being clipped. Row heights come from content; the panel fills its tile.
- `components-v2/layout/RadioLayout.svelte`: `.receiver-deck { overflow: hidden }` now applies only to the non-semantic root.
- `components-v2/layout/vfo-layout-tokens.ts`: the five fixed row-height tokens the old panel consumed and the `vfoMeterScale` URL override that fed one of them are removed with their last consumer.
- `semantic/VfoSurface.svelte`, `semantic/VfoOperationGroup.svelte`, `components-v2/vfo/ActiveReceiverToggle.svelte`: the centre bridge lays its controls out in two columns with a 24 px minimum control height and takes its block margin from `--vfo-instrument-inset-block`, the custom property the panel wrappers use as block padding. `.receiver-instrument` is a column flex container whose tile grows to fill it. Controls, intents, disabled reasons and accessible names are unchanged.
- `frontend/tests/e2e/i18n/desktop-geometry.spec.ts`: selects panel rows through the `data-vfo-row` hooks; its clipped-descendant scan covers the new rows and skips elements that generate no box.

## Tests

New `semantic/__tests__/VfoSurface.panel-rows.test.ts` on the real `VfoSurface`. One existing pin in `VfoPanel.isolated.test.ts` changed because the filter chip moved rows.

## Visual baselines

Eleven cockpit captures and the four production-root scenes are re-pinned from Linux runs at head `40a01126`; which runs, which files and the per-scene pixel counts are in the two READMEs' new provenance sections. Six regenerated files that differed in bytes but passed the comparison at that head were not copied.

At a 1024 px wide window the meter wraps below the frequency instead of sharing its line; that is the intended fallback at that width.

Linear: MOR-2509 (remains open).


morozsm:
Agent Review: BLOCKED 2ee3e14022d8a8e966c135822ca221772cbf572f

Independent review of the net diff against `b49e385886628602087be791ca4c8de2dd8d60ac` (7 files, +460/-146, confirmed via `gh pr view 3554 --json files,additions,deletions`). Working tree left byte-identical (`git status --short` empty).

## Blocking

**B1 — `quick` is RED at this head, and not for the reason the PR body gives.** `gh pr checks 3554`: `quick` **fail** (3m31s). Step 16 "Frontend install + check + tests + build" passed; step 20 "Frontend i18n visual smoke (RP-ML-006)" failed with **10 failures, 84 tests did not run** (`maxFailures` stopped the run, so the true count is unknown). All 10 are hard DOM assertions in `frontend/tests/e2e/i18n/desktop-geometry.spec.ts` — a file this PR does not touch (`git diff b49e3858..HEAD --name-only -- frontend/tests/` → 0 files) — and `quick` on the base commit `b49e3858` is `success`. Two independent causes, both created by this diff:

- **B1a — `.panel-body` was deleted.** `frontend/tests/e2e/i18n/desktop-geometry.spec.ts:242`: `const primary = instrument.querySelector('.panel-body [data-vfo-freq]');`. `frontend/src/components-v2/vfo/VfoPanel.svelte` removed the `.panel-body` wrapper and its rule, so `primary` is `null` on every Standard page. 6 tests red — `standard {1024,1200,1440} topology-{1-single,2-main-sub} painted grid` — failing at spec:649/650 ("primary frequency glyphs are painted", received 0) and then `TypeError: Cannot read properties of null (reading 'rect')` at spec:653.
- **B1b — mode and filter left the control strip.** `desktop-geometry.spec.ts:592-595` reads `strips.nth(n).locator('.mode-badge-wrapper')` and `toContainText(/CW.*FIL3/)` / `/USB.*FIL1/` off `.control-strip`. Both chips now live in `.panel-identity` (`VfoPanel.svelte`, the new `.panel-identity` block). 4 tests red — `Standard {900,1024,1200,1700} compact absolute VFO pair keeps every bridge control`, `Error: element(s) not found`.

Risk: a required check is failing and the PR body attributes the expected redness to the pixel-diff run only. Required fix: update `desktop-geometry.spec.ts:242` to a selector that survives the restructure (e.g. `.display-row [data-vfo-freq]`) and `:592-595` to locate the mode chip and the mode/filter text in `.panel-identity`, then re-run. Checks: `npm run test:e2e:i18n` (or the single spec, `npx playwright test -c ./playwright.i18n.config.ts tests/e2e/i18n/desktop-geometry.spec.ts`) plus `npx vitest run src/semantic/__tests__/VfoSurface.panel-rows.test.ts`.

**B2 — same class, one site further: a pin that can no longer redden where the change put the risk.** `desktop-geometry.spec.ts:555` builds the per-card clipped-descendant scan from `'.panel-header, .panel-meter, .vfo-freq, .control-strip, .control-strip *'`, and spec:579 asserts `cardOverflow` is `[[], []]` ("both full card strips fit without clipped descendants"). The new identity row and everything this PR moved into it — `.panel-identity`, `.mode-badge-wrapper`, the mode and filter `StatusIndicator`s — are in none of those selectors, so the one browser assertion that guards against clipped chips is now blind to the row the change created. Sweep of the class (`grep -rn "panel-body\|mode-badge-wrapper\|panel-header" src tests`): `desktop-geometry.spec.ts` is the only e2e consumer; no other `.panel-body` reference belongs to `VfoPanel` (the other hits are the unrelated `components-v2/panels/*` panels). Required fix: add `.panel-identity, .panel-identity *` to the spec:555 list in the same commit as B1.

## Non-blocking (report, do not expand scope)

- **N1 — orphaned layout tokens.** `src/components-v2/layout/vfo-layout-tokens.ts:298-306` still emits `--vfo-panel-header-height`, `--vfo-panel-meter-height`, `--vfo-panel-body-height`, `--vfo-display-row-height`, `--vfo-control-strip-height`. After this diff no stylesheet reads any of them (`grep -rn -- "--vfo-panel-header-height" src` etc. returns only the emitter and two tests). Per-skin row-height tuning is inert, and `layout/__tests__/vfo-layout-tokens.test.ts:121-122` and `top-row-visual-regression.isolated.test.ts:374-393` still pin their emission — pins that no longer discriminate anything about layout. CLAUDE.md § Scope discipline, "No orphans".
- **N2 — "one source" is two literals.** `semantic/VfoSurface.svelte`: `.instrument-panel { --vfo-instrument-inset-block: 6px }` and `[data-vfo-appearance='standard'] .instrument-panel:not(:has(> .standard-pair-bridge)) { --vfo-instrument-inset-block: 8px }`. Two values that can drift, plus two `var()` fallbacks (`6px` base, `8px` standard) that are both unreachable because `.instrument-panel` always declares the property. Behaviour check against the removed `[data-vfo-appearance='standard'] .standard-receiver[data-standard-vfo-slot] { padding: 6px }`: pair slots keep 6px, non-pair standard keeps 8px — equal to before, **except** a pair slot whose `fixed.slot.kind !== 'slotted'` (no `data-standard-vfo-slot` attribute, `VfoSurface.svelte: standardInstrument`), which went 8px → 6px.
- **N3 — `:has()` is fine here.** Precedent already ships in `src/presentation/languages/studioline/studioline.css:181` and `fieldline/fieldline.css:201`. Where unsupported, `:not(:has(…))` invalidates the whole rule and the inset stays 6px — a sane 2px-tighter fallback. Not a defect.
- **N4 — the deck-overflow change is close to inert, and the prose oversells it.** `RadioLayout.svelte:1105` sets `.desktop-control-face > .receiver-deck { display: contents }`, so on `desktop-v2` and `sdr-test` that element generates no box and its `overflow` never clipped anything. The only reachable effect is on `.radio-layout.semantic-deck:not(.desktop-control-face)`, which already carried `overflow-y: auto` (`RadioLayout.svelte:1064`): computed overflow moves from (hidden, auto) to (auto, auto), i.e. a horizontal scrollbar instead of horizontal clipping. **Clipping did not become overlap** — that deck is still a scroll container, so it cannot paint over the row below. The provenance half of the new comment checks out: `git show 84d5e41d` does add `.receiver-deck { overflow: hidden }` alongside `.receiver-deck :global(.vfo-header) { height: 100% }`.
- **N5 — guardrails.** 7 files · 606 changed lines crosses the soft threshold on **both** numbers (6 · 600); the body does not carry the "why this is one unit of work" line CLAUDE.md asks for. Fixing B1/B2 adds `desktop-geometry.spec.ts` → 8 code files, before the regenerated PNG baselines that will need the recorded hard-ceiling exception line.

## Shared-component consumers (checked, no further regression found)

`grep -rn "VfoPanel" src`: the only mounts are `semantic/VfoSurface.svelte:854` and `components-v2/vfo/LegacyVfoPanelAdapter.svelte`, the latter reached from `components-v2/layout/VfoHeader.svelte:163` and `components-v2/panels/vfo/DualVfoDisplay.svelte:50,64`. `VfoHeader` renders only in `RadioLayout.svelte:843`'s `{:else}` arm, which needs `semanticDeck === false`, i.e. a manifest declaring no `vfo` surface; `presentation/layouts/declarations.ts:99,166-168` and `desktop-declarations.ts:37` declare it, matching the standing note at `layout/keyboard-map.ts:387`. So that path — and the new `.radio-layout:not(.semantic-deck) > .receiver-deck` rule written for it — is unreachable on the shipping skins; no live regression, but nothing verifies the legacy path either. `ActiveReceiverToggle`'s change is guarded by `.segment.embedded`, and `embedded` is passed only by `semantic/VfoOperationGroup.svelte:132`; `components-v2/vfo/VfoOps.svelte:59` mounts it without the prop and is untouched. `VfoOperationGroup`'s 24px floor is scoped to `[data-vfo-operation-appearance='sdr'|'standard']`, so the `semantic` appearance keeps its metrics. The `@media (pointer: coarse)` neighbour at `ActiveReceiverToggle.svelte:254` still wins with 44px. Design-language reachability and every vitest consumer are green at this head (`quick` step 16).

MOR-2514 is unaffected: `handleFocusOut` is bound on the readout element itself (`primitives/frequency/StandardFrequencyReadout.svelte:77`), with no ancestor dependency, and the readout's own markup, `role`, `aria-label` and `tabindex` are unchanged. Tab order is unchanged — the mode wrapper is a non-focusable `<div>` in both layouts.

## Mutations run (tree restored after each; `git status --short` clean)

Every pin discriminates. Each mutation reddened `src/semantic/__tests__/VfoSurface.panel-rows.test.ts` and nothing else was needed to detect it.

| Mutation | Result |
|---|---|
| meter moved out of `.display-row`, above the frequency | **2 failed** (both parametrised rows), "meter inside the same main row" |
| `flex-wrap: wrap` removed from `.control-strip` | **1 failed**, "control strip declares wrapping" |
| `.bridge` given a literal `margin-block: 6px` | **1 failed**, "bridge block inset … read one custom property" |
| 24px floor dropped in `VfoOperationGroup` + `ActiveReceiverToggle` | **1 failed**, "every bridge control declares a 24px min-height floor" |
| `white-space: nowrap` re-added to `.active-receiver` | **1 failed**, "the caption wraps" |
| `overflow: hidden` reinstated on `.receiver-deck` | **1 failed**, "no receiver-deck rule that can apply to the semantic deck hides overflow" |
| **control**: legitimate nearby edits (`--vfo-control-strip-gap` default 4px→5px, `.bridge` gap 10px→8px) | **16 passed** — the pins do not redden on correct work |

The 16 tests mount the real `VfoSurface` with no `vi.mock`. Coverage gap worth stating: the DOM rows are pinned for `2/main_sub` and `1/ab` only — not `1/single` or `2/ab_shared`.

## UNVERIFIED — needs a browser measurement that was not done

The recorded measurement covers `2/main_sub`, default language, dark theme, `studioline`. Not covered anywhere:
- `1/single` and `2/ab_shared` at `appearance='standard'`, 1024/1200/1440 — `1/single` is one of the rows that went red in B1a.
- `appearance='sdr'` (skin `sdr-test`): `.bridge` gains `margin-block: 6px` where it previously had none; no viewport measured.
- The generic semantic root (`peer-split`, `unified-instrument`, `panadapter-first`, `dual-receiver-cockpit`) at a narrow width, where horizontal overflow now scrolls instead of clipping.
- Languages `fieldline` and `segmentline`, and the LCD / mobile faces.

## MANDATORY SQUASH-BODY CORRECTIONS

The squash body is written by hand at merge; these sentences are false of the net change and must not land on `main`.

1. **`0efeadb2` — drop subject and body entirely.** "the VFO bridge is never taller than the panels; one-line caption, 24 px controls", "Compact the Standard non-pair bridge … (padding 4px, gap 3px) and grow its flex-basis to 152px … (white-space: nowrap)". None of it is in the net diff: `.bridge` is still `gap: 10px; padding: 10px` and `[data-vfo-appearance='standard'] .bridge { flex-basis: 136px }`. Only "24px min-height floor on bridge controls" survives; fold that one clause into the retained text.
2. **`f01b7af8` — "compact 2x3 control block" / "the operation group's own two-column grid applies (2x3 with six controls)" is false.** The pair bridge, the only grid whose 3-column override this diff removed, holds **five** `[data-dual-action]` controls — pinned by this PR's own test, `VfoSurface.panel-rows.test.ts`, "1/ab: the pair bridge grid keeps its controls and names". The `2/main_sub` bridge holds **seven**, with MAIN/SUB spanning both columns (`ActiveReceiverToggle.svelte: .active-receiver-toggle.embedded { grid-column: 1 / -1 }`), so it is 2×4, not 2×3. Replace with "two-column bridge grid" and no count.
3. **`f01b7af8` — narrow "so the semantic deck is never clipped".** On `desktop-v2` and `sdr-test` the deck is `display: contents` (`RadioLayout.svelte:1105`) and generated no box to clip with; the reachable change is horizontal overflow on the generic semantic root.
4. **`f01b7af8` — "Fixed row-height tokens no longer size the panel grid"** is true, but add that `vfo-layout-tokens.ts` still emits five of them and nothing now reads them (N1).

## REQUIRED BEFORE MERGE (PR body — editable in place, no commit needed)

1. "This PR intentionally changes the VFO deck, so the pixel-diff run is expected to fail" — incomplete and currently misleading: `quick` is also red, from 10 hard DOM assertions, not pixel diffs. Remove or correct once B1 lands.
2. "About a hundred test files that mount these components were run one at a time, all green" — true of the vitest files (step 16 passed at this head); the Playwright i18n geometry suite mounts the same components and was not run. Narrow the sentence to vitest.
3. Bullet 2, "so it no longer clips the semantic VFO deck" — same correction as squash item 3.
4. Add the soft-threshold justification for 7 files / 606 changed lines (N5), and state the hard-ceiling file count as `N code + M regenerated baselines = K files` when the baselines land.

## CI as found

`quick` **fail** · `visual` **fail** (expected for this change) · `Agent Review Gate` fail (no directive yet) · `Agent Review Gate v2 observe` fail · `quick-v2-observe` pending · `changes`, `classify`, `grep-gate`, the metadata/observation publishers and `publish-docs-only-quick-status` pass.

morozsm:
Agent Review: PASS 7da8c360d5a539d40b3c6ad8fcac0615f509dabe

Delta review of `2ee3e140..7da8c360` (5 commits, 26 files). Head confirmed equal to `gh pr view 3554 --json headRefOid`. Working tree byte-identical throughout; `git status --short` empty. No vitest/pytest was run on this machine.

## My previous BLOCKED findings are resolved

- **B1a/B1b** — `desktop-geometry.spec.ts` now reaches the rows through `data-vfo-row="identity|main|chips"`. **`quick` passes at this head** (`gh pr checks 3554`, 5m26s), and so does `visual` (1m12s).
- **B2** — the clipped-descendant scan moved to `[data-vfo-row], [data-vfo-row] *`. I checked this is a strict superset of the old list, not a swap: `.panel-header` sits in the identity row, `.panel-meter` and `.vfo-freq` in the main row, `.control-strip` **is** the chips row (`VfoPanel.svelte`, the three row wrappers). Nothing the old list reached has been dropped, and the identity row it was blind to is now covered.
- **Orphan tokens (N1)** — swept. `git grep` over the whole tracked tree for `panelHeaderHeight`, `panelMeterHeight`, `panelBodyHeight`, `displayRowHeight`, `controlStripHeight`, `TOP_ROW_INNER_HEIGHT`, `vfoMeterScale`, `meterScale` (`-w`) and the five `--vfo-*-height` spellings returns **zero** hits outside the unrelated `meterScaleLabel` in the design-language renderers.
- **Guardrails (N5)** — the exception line is present, in the required form, and correct (below).

## Attack on the `getClientRects().length === 0` skip — it cannot hide a real clip

An element that can be clipped generates a box; an element that generates a box has at least one fragment and therefore at least one client rect. The two cases with zero rects — `display: contents` and `display: none` — generate no box of their own and cannot be clipped; their children are scanned separately and do have boxes. So the skip removes exactly the `0,0,0,0`-rect false positives that the broadened `[data-vfo-row] *` selector introduced (`.frequency-readout-content` and friends), and nothing else. The scan is not hollow: it still reports any boxed descendant that scrolls or escapes the card, and `quick` exercised it at 900/1024/1200/1700 on this head. The pre-existing `style.display !== 'none'` test is now redundant but harmless.

## Discrimination proof for the new pins — 7 mutations, 2 controls

Vitest was not run. Instead I extracted `styleBlock`, `rulesFor` and `rule` **verbatim** from `VfoSurface.panel-rows.test.ts` plus the six new assertions into a scratch script, and ran them over `git archive 7da8c360 | tar -x` copies — the tree under review was never touched.

| Mutation (on the isolated copy) | Result |
|---|---|
| drop `display: flex; flex-direction: column` from `.receiver-instrument` | FAIL `wrapper-is-column` |
| drop `flex: 1 1 auto` from `… .receiver-instrument > :where(.vfo-tile)` | FAIL `tile-flex` |
| delete the `.receiver-instrument > :global(.indicator-row)` rule | FAIL `rule MISSING` + `indicatorrow-flex` |
| drop `align-content: start` from `.panel` | FAIL `panel-align-start` |
| drop `align-content: start` from `.indicator-row` | FAIL `indicatorrow-align-start` |
| re-add `var(--vfo-instrument-inset-block, 6px)` to the wrapper padding | FAIL `base wrapper fallback-free` |
| re-add the same fallback to `.bridge`'s `margin-block` | FAIL `bridge fallback-free` |
| **control** `.bridge` gap 10px→8px | ALL PASS |
| **control** `.indicator-row` gap 4px→5px | ALL PASS |

Every new pin discriminates, and neither legitimate nearby edit reddens one.

## Findings

**F1 — non-blocking, test hygiene.** `semantic/__tests__/VfoSurface.panel-rows.test.ts`, test *"the bridge block inset and the panel wrapper padding read one custom property"*: the first of its two `css.match(...)` alternatives can never match. In the template literal `` `${declaration}:\\\\s*([^;]*);` `` the four backslashes emit a literal backslash pair, so the constructed pattern requires a real `\` in the CSS. Running the test's own logic printed `firstRegexMatched=false` for all three of base wrapper, standard wrapper and bridge; the `?? second` alternative carries the whole assertion. No coverage is lost — the mutations above prove the pin still reddens — so this is not a block, but the dead branch reads as if it were doing the work. Fix when the file is next touched: delete the first `css.match` (or write `\\s`).

**F2 — REQUIRED BEFORE MERGE (PR body, editable in place).** The body says: *"In the standard appearance an instrument panel that holds the bridge uses a 6 px block inset (8 px before); one without a bridge keeps 8 px."* Both halves are wrong.
- The discriminator in `VfoSurface.svelte` is `:not(:has(> .standard-pair-bridge))`, i.e. the **A/B pair** bridge specifically. A `2/main_sub` standard panel does hold a `.bridge` (the `showRadioWideFacts` branch) and still matches the `:not(...)`, so it gets **8 px**, not 6.
- *"(8 px before)"* is false for the case that does get 6 px: on `b49e3858`, `[data-vfo-appearance='standard'] .standard-receiver[data-standard-vfo-slot] { padding: 6px }` already gave the pair slots 6 px. The value is preserved, not changed.

Accurate wording: *"a standard instrument panel holding the A/B pair bridge keeps its 6 px block inset; every other standard panel keeps 8 px."* Commit `2ee3e140` already states this correctly (*"8px for the non-pair Standard deck"*, *"the Standard pair slots inherit the base 6px"*), so only the body needs editing.

**F3 — REQUIRED BEFORE MERGE (PR body), carried over.** Bullet 2 still reads *"so it no longer clips the semantic VFO deck"*. On both shipping semantic faces `.receiver-deck` is `display: contents` (`RadioLayout.svelte`, `.desktop-control-face > .receiver-deck`), so it generated no box and clipped nothing. Unchanged by this delta; prose only.

**F4 — reported, not to be fixed here.** Five properties `vfo-layout-tokens.ts` still emits have **zero** `var()` consumers anywhere in the tracked tree: `--vfo-control-badge-height`, `--vfo-control-badge-min-height`, `--vfo-control-badge-padding-x`, `--vfo-ops-secondary-margin-top`, `--vfo-ops-stack-gap`. They are mentioned only by the emitter and its own tests. All five were already orphaned on `b49e3858` (measured there too; the same measurement returns 3 and 1 for `--vfo-ops-badge-height` and `--vfo-panel-pad-x`, so it can report non-zero). Not caused by this PR and outside this slice — but see the squash-body item below, because a commit message asserts the opposite.

## Baseline provenance — every number checked

- Four production-root PNGs: `shasum -a 256` equals the README's two tables **exactly**, all four. All are 1280×800, PNG colour type 2 (truecolour RGB) — the "RGB PNGs at 1280×800" line holds.
- Cited runs, via `gh run view`: **35655644308** = `visual.yml`, `workflow_dispatch`, head `40a01126`, success; **35654800077** = visual, pull_request, same head, failure; **35654800078** = quick, pull_request, same head, failure. All three at the claimed head. `visual.yml` uploads an artifact named `regenerated-baselines` only under `inputs.regenerate == 'true'`, and that run produced one — the `regenerate=true` claim holds.
- I downloaded that artifact and compared it byte-for-byte with `cmp`: **all eleven committed PNGs and `manifest.json` are byte-identical to the regeneration output**, and all eleven differ from their pre-delta versions — so the comparison is demonstrably able to report inequality, not silently equal. "copied byte-for-byte" is a measurement here, not a claim.
- Exactly **six** artifact files differ from the committed tree and were not copied: the six named. `topology-1-single--desktop.png` regenerated identical, so "six" is exact rather than an undercount. None of the six appears among the eleven failures, so each did pass.
- Run 35654800077 reported **"11 failed, 28 passed"**, and its eleven failure outputs are exactly the eleven captures the README names and the eleven PNGs the commit changes.
- Job `106516106686` belongs to run 35654800078 (`quick`, head `40a01126`); its *Frontend i18n visual smoke (RP-ML-006)* step reported **91 passed / 4 failed**, and the four are precisely the four production-root comparisons. Its vitest step was green (485 files / 12462 tests).
- `manifest.json`'s `commit`, `commitShort`, `branch` and `generatedAt` are self-consistent and byte-identical to the regenerated file.

I read `topology-2-ab-shared--desktop.png` (before and after), `dual-main-sub--desktop.png` and `studioline--dark--production-root.png`. The three-row panel renders as intended — identity row, frequency beside the meter on one line, chip row, with slack below it; the bridge caption wraps to three lines without clipping; the A/B-shared fixture differs from its predecessor only by a small vertical shift with no clipping or overlap. The `?`/`—`/`unit unknown` placeholders are the separately ticketed MOR-2520 matter, and `AF LEVEL 12800%` belongs to the RX-audio panel, not this slice. Nothing in the images is a defect of this change.

## PR body counts — verified

`gh pr view 3554 --json files,additions,deletions`: **30 files, +566/−220 = 786 changed lines**; splitting the file list gives **12 code + 18 baseline** (15 PNGs + `manifest.json` + 2 READMEs). The exception line matches, is the first line of the body, and uses the required `N code + M regenerated baselines = K files` form. 786 is under the 1000 ceiling and over the 600 soft threshold, and the body carries its one-unit-of-work justification. The *"at 1024 px the meter wraps below the frequency"* note is true and pre-existing: `VfoPanel.svelte` has `@media (max-width: 1024px) { .display-row { flex-wrap: wrap } }`, unchanged from `b49e3858`.

## MANDATORY SQUASH-BODY CORRECTIONS

Carried over from the previous head, still applicable:
1. **`0efeadb2`** — drop subject and body; none of the compacting, 152 px basis or `white-space: nowrap` survives. Keep only the 24 px min-height floor.
2. **`f01b7af8`** — *"compact 2x3 control block"* / *"(2x3 with six controls)"* is false: the pair bridge holds five `[data-dual-action]` controls and the `2/main_sub` bridge seven, with MAIN/SUB spanning both columns. Say "two-column bridge grid", no count.
3. **`f01b7af8`** — narrow *"so the semantic deck is never clipped"* (same as F3).

New, from this delta:
4. **`2120b17e`** — *"every remaining token still has a reader"* is false. Five of the emitted properties have zero `var()` consumers (F4). They were already orphaned before this PR, so the fix is to the sentence, not the code: say "the removed tokens' last readers are gone; five unrelated tokens were already reader-less before this change and are left alone".

`fef439b8`, `06af9aca`, `40a01126` and `7da8c360` read true against the diff — including `40a01126`'s *"The bridge is unchanged"* (that commit touches no `.bridge` declaration; the `margin-block` fallback removal belongs to `2120b17e`). `fef439b8`'s *"false for a real zero-width box"* is a Chromium behaviour statement I did not execute; the safety property the scan depends on is weaker and holds by construction, as argued above.

## UNVERIFIED — for a later browser measurement

The `sdr` appearance (skin `sdr-test`) inherits both the `.receiver-instrument` column-flex change and the bridge's `margin-block`, and no `sdr-test` scene is among the fifteen re-pinned captures; likewise the `segmentline` design language. Worth one pass at 1280×720 and 1920×1080 on `sdr-test` before the next slice. `1/single`, `2/ab_shared`, mobile portrait, StudioLine and FieldLine are now covered by the passing geometry spec and the re-pinned baselines, so I withdraw them from my previous unverified list. No defect is suspected in either remaining case — the compensating `align-content: start` is unscoped and applies there too.

## CI as found

`quick` **pass** (5m26s) · `visual` **pass** (1m12s) · `changes`, `classify`, `grep-gate`, `publish-docs-only-quick-status`, `Update Agent Review Gate status`, and the three observation publishers **pass** · `quick-v2-observe` **pending** (waiting on canonical legacy quick) · `Agent Review Gate` and `Agent Review Gate v2 observe` **fail**, both reporting only "Missing trusted standalone exact-head Agent Review PASS directive" — which this comment supplies.

PR #3555, opened 2026-09-21
fix(MOR-2515): default AF-level and RF-gain shortcuts move to Alt/Option+Arrow, which macOS does not reserve

morozsm:
MOR-2515. On macOS, Ctrl+ArrowUp / Ctrl+ArrowDown are system shortcuts (Mission Control / App Exposé) and never reach the browser, so the shipped default for AF level did nothing on a stock Mac and the panel hint advertised a dead key (found by the owner during the MOR-2512 keyboard check, 2026-09-21).

Owner ruling: `adjust_af_level` → Alt/Option+ArrowUp / ArrowDown, `adjust_rf_gain` → Alt/Option+Shift+ArrowUp / ArrowDown, one set for all platforms.

## What this changes

- `rigs/_keyboard-default.toml`: the four default chords move from Ctrl / Ctrl+Shift to Alt / Alt+Shift.
- `components-v2/layout/keyboard-map.ts`: `formatShortcut` names the key as the viewer's platform does — "Option" on macOS, "Alt" elsewhere (`isOptionNamingPlatform()`: `navigator.userAgentData.platform`, falling back to `navigator.platform`; the existing `detectPlatform` does not distinguish macOS) — and displays Alt/Option before Shift.
- New `tests/test_keyboard_default_reserved_chords.py`: the default profile may not bind a chord from one named list of OS/browser-reserved chords (Ctrl+any arrow, Meta+any arrow, Alt+Left/Right, Ctrl/Meta/Alt+Tab).
- `frontend/src/lib/runtime/adapters/__tests__/fixtures/ic7300-capabilities.json`: the four keyboard entries replaced by the output of the real `_serialize_keyboard_config(load_rig(rigs/ic7300.toml))` — a 4-line diff. This is a splice of serializer output, NOT a live bench re-capture (`capture-profile.mjs` needs a server and a radio); the serializer's 64 keyboard entries all match the fixture value for value. Three older surplus entries in that fixture (`mode-psk`, `mode-pskr`, `swap-vfo`) are untouched and predate this change.
- `docs/guide/web-ui.md`: the keyboard paragraph names the Alt/Option chords.
- Pins: Alt+ArrowUp with a focused AF slider fires the global action once and does not step the slider (next to the existing Ctrl case; `hasCommandModifier` already covered `altKey`); Alt+ArrowUp inside a text input is not dispatched (Option+Arrow is word navigation on macOS); literal hint strings `Option+ArrowUp` / `Alt+ArrowUp` for both platforms.

## Evidence (tests run on the build host through `focused.yml`, not locally)

RED run 35655044503 on `1d94de8b`: pytest 2 failed (`binds OS/browser-reserved chords: CTRL+ArrowDown, CTRL+ArrowUp`; `assert ['CTRL'] == ['ALT']`), vitest 3 failed (`expected 'ALT+ArrowUp' to be 'Option+ArrowUp'`). GREEN run 35655498849 on head `627a1342`: pytest 2 passed, ruff clean, vitest 4 files / 146 tests passed. `npm run check` 0 errors.

## Size

10 files, +234 / -31, measured with `git diff --stat $(git merge-base origin/main HEAD)` — at the file ceiling, over the 6-file soft threshold: the default chords are one contract spread over the profile, the serializer fixture, the hint renderer, their pins and the user guide. Left out to stay at 10: one stale example line in a comment in `SpectrumPanel.component.test.ts`.

Not verified on hardware: the owner's check on a real Mac (Option+↑ with the AF slider focused; the hint text) follows the merge.

Linear: MOR-2515.


morozsm:
Agent Review: PASS 627a1342c604aae8ce59b9166092aff98f12432b

Independent review of PR #3555 at head `627a1342`. Verified `git rev-parse HEAD` == `gh pr view 3555 --json headRefOid`. Working tree `git status --short` clean before and after; nothing edited, nothing pushed.

## CI state as found

- `quick` **pass** (6m40s), run `35655740149`, `head_sha` confirmed `627a1342` via `gh api .../actions/runs/35655740149` — frontend 484 test files passed.
- `visual` **pass** (1m30s). `changes`, `classify`, `grep-gate`, publishers **pass**.
- `Agent Review Gate` / `Agent Review Gate v2 observe` **fail**: "Missing trusted standalone exact-head Agent Review PASS directive" — this comment supplies it.
- `quick-v2-observe` **pending**, waiting on canonical legacy quick.
- Nothing red is caused by this PR.

I did **not** dispatch a second `focused.yml` run. `focused.yml` is `runs-on: [self-hosted, linux, build]` — the same single runner `quick`/`visual` serialize on — and the exact-head `quick` run above already executed the whole frontend suite at `627a1342`, which strictly supersedes a 4-file focused subset. I verified the builder's GREEN focused run `35655498849` by reading its own log rather than the report: `FOCUSED_REVISION: 627a1342...`, pytest 2 passed, ruff "All checks passed", vitest 4 files / 146 tests passed.

## Pin discrimination (established from the RED runs' own logs, not from the diff's reading)

- `tests/test_keyboard_default_reserved_chords.py: test_default_bindings_avoid_os_reserved_chords` and `test_af_and_rf_defaults_live_on_alt_arrows` — **CONFIRMED discriminating**. Run `35655044503` (`FOCUSED_REVISION: 1d94de8b`) failed for exactly the right reason: `binds OS/browser-reserved chords: [(['CTRL'], 'ArrowDown'), (['CTRL'], 'ArrowUp')]` and `assert ['CTRL'] == ['ALT']`. This is a guard, not a decoration: it would have caught the original defect.
- `keyboard-map.test.ts: formatShortcut platform naming` (3 cases) — **CONFIRMED discriminating** on `modifierDisplayName`: same run, `expected 'ALT+ArrowUp' to be 'Option+ArrowUp'` / `to be 'Alt+ArrowUp'`.
- The `MODIFIER_ORDER` reordering (third commit) is pinned by the `'Option+Shift+ArrowUp'` / `'Alt+Shift+ArrowDown'` assertions, and I found the observed red/green pair for it that the PR body does not mention: run **`35655303061`** at `FOCUSED_REVISION: e69eb077` (the commit immediately before) **failed** with `expected 'Shift+Option+ArrowUp' to be 'Option+Shift+ArrowUp'` and `expected 'Shift+Alt+ArrowDown' to be 'Alt+Shift+ArrowDown'`. That is a real one-line mutation reproduced end to end.
- `KeyboardHandler.test.ts` — the two new cases (`Alt+ArrowUp on an owning widget…`, `does not dispatch the Alt+ArrowUp default when a text input is focused`) were **GREEN at `1d94de8b`** (that file reported "57 tests" with no failures in the RED run). They do **not** discriminate this change: both build their own inline `modifiers: ['ALT']` binding and resolve through the same `hasCommandModifier` / `shouldIgnoreEvent` branches their existing Ctrl and bare-arrow siblings already cover. This is disclosed honestly in commit `1d94de8b` ("Alt composition and input-skip pins land green as required confirmations") and in the PR body, so it is not a false claim — noted so nobody later reads them as coverage this change added.

## Fixture reproduction (item 2)

Reproduced independently, not read: `_serialize_keyboard_config(load_rig(rigs/ic7300.toml))` compared entry by entry against `ic7300-capabilities.json`.

- 64 serializer bindings, 67 fixture bindings, 64 ids in common, **0 differing values** — the four spliced entries match the serializer exactly, as do the other 60 and `leaderKey`/`leaderTimeoutMs`/`altHints`/`helpTitle`.
- The 3 fixture-only entries are `mode-psk`, `mode-pskr`, `swap-vfo` — pre-existing, untouched, as claimed.
- **Negative control run first**: forcing `af-level-up.modifiers` back to `["CTRL"]` in the in-memory serializer copy made the comparison report exactly 1 difference, so the "identical" result is a measurement, not a silently broken compare.
- `conformance/fixture-provenance.guard.test.ts` checks state/capabilities **field presence** against the parsed `state.ts`/`capabilities.ts` interfaces and the provenance sidecar; it does not inspect keyboard binding values, so the splice neither breaks nor is covered by it. It was in the GREEN focused target list and passed.

## Collisions (item 3)

Enumerated every binding that uses ALT today: `git grep -n "modifiers" -- rigs/` returns **one** file, `rigs/_keyboard-default.toml`, with `["SHIFT"]` ×5 and the four new `["ALT"]` / `["ALT","SHIFT"]` entries. No rig ships a `[keyboard]` section of its own (`git grep -ln "\[keyboard" -- rigs/`), so there is no per-rig override to collide with. `DEFAULT_KEYBOARD_CONFIG` in `keyboard-map.ts` (the no-server fallback) has no AF/RF bindings at all. No collision with any existing binding.

- **Established.** Matching is safe: `keyboard-map.ts: modifiersMatch` sorts both sides alphabetically, `getEventModifiers` yields `['SHIFT','ALT']` for Option+Shift which sorts to `['ALT','SHIFT']` and equals the declared `['ALT','SHIFT']`. Arrow keys keep `event.key === 'ArrowUp'` under Option on macOS, and no default binds ALT+letter, so the Option-dead-key concern does not arise here.
- **Established, and the one thing nobody named:** `KeyboardHandler.svelte: handleKeydown` does `if (event.key === 'Alt' && keyboardConfig.altHints) { document.body.dataset.shortcutHints = 'true'; return; }`, and `alt_hints = true` in `rigs/_keyboard-default.toml`. Holding Alt/Option is already the app's own gesture for revealing the inline shortcut badges (the help card literally says "Hold Alt to reveal inline shortcut hints on controls"). The new chords therefore pop the whole-UI hint overlay for as long as the user holds Option to repeat an AF/RF step. The binding still fires correctly — the `Alt` keydown returns early, the subsequent `ArrowUp` keydown reaches `resolveAction` — so this is not a defect in the code, but it is a visible behaviour change the ticket, the PR body, the docs and the new guard all miss. The owner's post-merge Mac check (Option+↑ on the AF slider) will see it; flagging it so it is a decision, not a surprise.
- **Established, scope limits of the new guard:** `RESERVED_DEFAULT_CHORDS` matches on an exact modifier set, so `CTRL`+`SHIFT`+Arrow (the pre-change RF-gain chord) was never flagged — consistent with the RED run finding only the two AF collisions. `_default_chords()` reads only `rigs/_keyboard-default.toml`, so a future per-rig `[keyboard]` override is outside it. The three `Tab` rows are already unreachable at runtime: `handleKeydown` bails on `event.key === 'Tab'` before binding resolution (MOR-1449), so they guard the profile file only.
- **Obvious reserved chords the list misses:** browser tab/window chords (`CTRL`/`META` + `W`,`T`,`N`,`L`,`R`) and, most pointedly, **bare function keys** — the default profile binds `F1`–`F10` to modes, and `F1` (help), `F3` (find), `F5` (reload), `F6`, `F7`, `F10` are browser/OS keys on at least one platform, none of which this guard models. Pre-existing; this PR neither worsens nor claims to fix it.
- **UNVERIFIED** (no Windows/Linux/Firefox host here, no browser permitted): that `Alt`+Up/Down is free in Firefox/Chrome on Windows and Linux; that GNOME/KDE do not bind `Alt`/`Alt+Shift`+Arrow; that Windows' `Alt+Shift` layout switch (which fires on release with no other key, so an intervening arrow should suppress it) does not interfere. The guard's `Alt`+Left/Right entry correctly reserves the browser Back/Forward pair.

## Completeness sweep

- `git grep -n -i -E "ctrl\+arrow|control\+arrow|Ctrl\+Shift\+Arrow"` over the whole tree (no `head`): the only statement left false by this change is `frontend/src/components/spectrum/__tests__/SpectrumPanel.component.test.ts` line 1024, `// (Ctrl+ArrowUp/Down adjust_af_level)`. **The PR body already names it** and explains it was left out to stay at 10 files — that is the class-sweep-and-report the repo rule asks for, so it is not a block. Every other `Ctrl+Arrow` mention I checked is still true (`DualParamRenderer`/`ValueControl` "ignores Ctrl+Arrow, Alt+Arrow, Meta+Arrow"; the `Ctrl+ArrowRight` widget test names; `KeyboardHandler.test.ts:473`, which declares its own CTRL bindings).
- No remaining `'CTRL'` token in `frontend/src` outside `KeyboardHandler.test.ts`'s own inline bindings and `getEventModifiers`. No capabilities fixture other than `ic7300` carries a `keyboard` block (`ftx1-capabilities.json` has none). No Playwright e2e or visual spec presses a modified arrow (`grep -rn -E "keyboard\.press|Arrow|Control|Alt\+" frontend/tests/e2e`); the only arrow press is the spectrum separator's bare `ArrowDown`. No locale string or skin hardcodes a chord.
- `docs/` has no other statement of these bindings (`git grep -i -E "AF gain up|adjust_af_level|adjust_rf_gain" -- docs/` → one unrelated i18n-inventory mention of the action name).

## Display-order change, measured over the whole space

The input space is 15 non-empty modifier subsets, so I diffed all of them between the old `['CTRL','SHIFT','ALT','META']` and the new `['CTRL','ALT','SHIFT','META']` rather than sampling. Exactly the 4 subsets containing both ALT and SHIFT change (`ALT+SHIFT`, `CTRL+ALT+SHIFT`, `ALT+SHIFT+META`, `CTRL+ALT+SHIFT+META`); `CTRL+SHIFT`, `CTRL+ALT`, `SHIFT+META`, `ALT+META` and every single modifier are byte-identical. No binding anywhere in the tree used ALT+SHIFT before this PR, so **no other hint's rendering changes** and the in-code comment "no pre-MOR-2515 chord combined the two" is true.

## Prose and platform detection

- `isOptionNamingPlatform()` in `keyboard-map.ts` is a justified second helper, not a duplicate: `controls/install-prompt-utils.ts: detectPlatform` takes a UA *string* and returns `'ios' | 'android' | 'desktop'` — it genuinely cannot distinguish macOS, exactly as claimed. Behaviour I checked by reading: no `navigator` → `'Alt'` (safe in node/test/SSR); iPadOS reports `MacIntel` → `'Option'`, which is right for an iPad hardware keyboard; `userAgentData.platform` `'macOS'` matches `/mac/i`; `userAgentData` present without `platform` falls through to `navigator.platform`.
- The test comment "panel `shortcutHint` props and the help overlay both render its output" is true: `layout/shortcut-hints.ts: getShortcutHint` → `formatShortcut` feeds `RxAudioPanel`/`RfFrontEnd`/`FilterPanel`/`ModePanel`/`RitXitPanel`/`BandSelector`, and `KeyboardHandler.svelte` renders `formatShortcut(binding)` in the `<kbd>` of the help card and in the leader pill.
- `docs/guide/web-ui.md`: both added sentences check out. macOS does reserve Ctrl+Up/Down (Mission Control / App Exposé), and "inside a focused text input the global map stays silent" is enforced by `shouldIgnoreEvent` + `IGNORED_TAGS = {INPUT, TEXTAREA, SELECT}` and pinned by the new KeyboardHandler case.
- Static checks I ran locally: `uv run ruff check tests/test_keyboard_default_reserved_chords.py rigs` → "All checks passed!"; `ruff format --check` → already formatted; `npm run check` → **4867 files, 0 errors, 0 warnings**, confirming that PR-body claim.

Commit `627a1342`'s body says: "The first GREEN focused run showed `['ALT','SHIFT']` rendering as Shift+Option+Arrow". That run is `35655303061` at revision `e69eb077`, and its conclusion is **failure**, not green — it failed on exactly those two assertions. This PR squash-merges with `COMMIT_MESSAGES`, so the sentence would land on `main`. When writing the squash body, replace it with something like: "Focused run 35655303061 on e69eb077 failed with `['ALT','SHIFT']` rendering as Shift+Option+Arrow; MODIFIER_ORDER is display-only (matching sorts both sides alphabetically), so Alt/Option now precedes Shift."

## REQUIRED BEFORE MERGE (PR body edits — head SHA unaffected, this PASS stays valid)

1. "parity of the fixture's **61** keyboard bindings with the serializer was checked" — the number is **64**. Established by running `_serialize_keyboard_config(load_rig(rigs/ic7300.toml))` and comparing against the fixture: 64 serializer entries, 67 in the fixture, 64 in common, 3 fixture-only. Correct the figure, or state it as "64 of the fixture's 67".
2. "`docs/guide/web-ui.md`: the shortcut table says Alt/Option and why" — the "Representative bindings from the shared default profile" table in that section has no AF/RF row and is untouched; the edit is to the prose paragraph below it. Reword to "the modified-arrow paragraph".

## Notes (no action required)

- `mor1563-keyboard-fanout-conformance.isolated.test.ts` rewrites the sentence containing "`modifiersMatch()` (line ~206)"; that function is now at line 237 (it was 211 at the merge-base, so the hint was already approximate and this change widened the drift to ~31). Not blocking: the repo's line-number rule is scoped to `docs/`, the reference is pre-existing and marked `~`, and the same sentence already carries the correct symbol name. Worth deleting the parenthetical whenever that file is next touched.
- Guardrails confirmed with `gh pr view 3555 --json files,additions,deletions`: **10 files, +234 / -31**. At the 10-file hard ceiling, not over it; over the 6-file / 600-line soft threshold, and the PR body gives its one-unit-of-work justification. No new abstraction or layer; `isOptionNamingPlatform` is a scoped private helper with a single caller.

PR #3556, opened 2026-09-21
fix(MOR-2518): FTX-1 mode codes come from the profile; NB, NR and data-mode decisions read the radio

morozsm:
MOR-2518. One unit of work (7 files, above the 6-file soft threshold): the profile key, its loader validation, its one consumer and the pins for each land together because none of them has a caller without the others; the seventh file is a one-line schema row for the new key.

## What changes

**Mode codes.** `YaesuCatRadio.__init__` derived each `MD` code from the mode's position in the profile list (`format(i + 1, "X")`), so the 16th and 17th modes got the two-character strings `10` and `11`, which the single-character `MD` field can never carry. Codes now come from a new profile list `[modes].codes` in `rigs/ftx1.toml`, parallel to `[modes].list`: `1`–`9`, `A`–`F`, `H` (C4FM-DN), `I` (C4FM-VW). `rig_loader.py: load_rig` rejects a list of the wrong length, a multi-character entry and a duplicate. A code the profile does not assign raises `CatParseError` in `read_mode` and `get_if_status` (new `YaesuCatRadio._mode_name`) instead of decoding to the placeholder name `UNKNOWN(<code>)`.

**Decisions read the radio, not the legacy mirror.** `set_nb` and `set_nr` (for profiles where the level doubles as the on/off switch) chose the level to restore from `self._state.main.nb_level` / `nr_level`; they now call `read_nb_level(receiver)` / `read_nr_level(receiver)`. `get_data_mode` derived its answer from `self._state.main.mode`; it now calls `read_mode(0)`.

## Evidence

Tests ran on the build host through `focused.yml` against pushed commits:
- RED, tests only, commit `8d4c9737`: run 35655407915 (failed).
- GREEN, head `7a6af85f`: run 35655831074 — 1066 passed, 1 skipped.

Locally: `uv run ruff check`, `uv run ruff format --check`, `uv run lint-imports`.

## Not verified

No live-radio check yet: real `MD0H;` / `MD1I;` answers, `NL` / `RL` level restore, and the width refusal in C4FM are to be checked receive-only on the bench.


morozsm:
Agent Review: PASS 7a6af85fac2d7b6302c7db3e77bba31e8673f71f

Independent review. I did not write this change. Head confirmed equal to `gh pr view 3556 --json headRefOid`. Worktree `git status --short` clean before and after; no edits, commits or pushes.

## Mode table — checked against the manual, name by name

Read `FTX-1_CAT_OM_ENG_2508-C.pdf`, **page 20**, `MD OPERATING MODE` (`pdftotext -f 20 -l 20 -layout`). P2 as printed:

`0:-  1:LSB 2:USB 3:CW-U 4:FM 5:AM 6:RTTY-L 7:CW-L 8:DATA-L 9:RTTY-U A:DATA-FM B:FM-N C:DATA-U D:AM-N E:PSK F:DATA-FM-N G:- H:C4FM-DN I:C4FM-VW J:-`

All 17 pairs in `rigs/ftx1.toml: [modes].codes` match position-for-position against `[modes].list`. The comment "The manual marks 0, G and J unused; H = C4FM-DN and I = C4FM-VW" is CONFIRMED verbatim against that page.

NL/RL answer formats: page 22 `NL NOISE BLANKER LEVEL` — P2 `000: OFF, 001-010`, three digits; page 24 `RL NOISE REDUCTION LEVEL (DNR)` — P2 `00: "OFF", 01-10`, two digits. Profile templates `NL{P1}{level:03d};` and `RL{P1}{level:02d};` match. The third commit's "fix the SUB NR pin to the two-digit RL level format" corrected the pin **toward the manual**, not toward the code: `RL102;` is what page 24 specifies.

## Exhaustive resolver diff (not a sample)

`uv run python -c` loading the real profile through `load_rig`, comparing the pre-change `format(i + 1, "X")` map against the new profile map over **every** printable single character:

- Encode: only `C4FM-DN "10"→"H"` and `C4FM-VW "11"→"I"` change. Codes `1`-`9`, `A`-`F` are byte-identical.
- Decode: `H` and `I` change from `UNKNOWN(H)`/`UNKNOWN(I)` to the correct names; every other character changes only from a placeholder to a raise.

No silent remapping of any working mode. `set_mode("C4FM-DN")` previously formatted `MD010;` — a malformed frame; it now sends `MD0H;`.

## Pin discrimination

I could not mutate the tree (read-only, no push), so I established discrimination from the builder's RED run **35655407915** (tests-only commit `8d4c9737`) and from the production line each pin depends on. 22 pins failed there, each for the right reason, not incidentally:

- `test_get_mode_c4fm_dn_main` — `assert 'UNKNOWN(H)' == 'C4FM-DN'`; `test_read_mode_c4fm_vw_sub` — `assert 'UNKNOWN(I)' == 'C4FM-VW'`. Exactly the removed bug.
- `test_read_mode_rejects_unused_code`, `test_get_if_status_rejects_unassigned_mode_code` — `DID NOT RAISE CatParseError`.
- `test_set_nb_on_keeps_radio_level`, `test_set_nb_sub_reads_and_writes_sub`, `test_set_nr_on_keeps_radio_level`, `test_set_nr_sub_reads_and_writes_sub` — the query mock was *called 0 times*: proof the pre-change path issued no live read at all.
- `test_get_data_mode_reads_live_mode` / `..._false_for_live_voice_mode` — `assert False is True` / `assert True is False`: the mirror, not the radio, decided.
- Four loader pins — `DID NOT RAISE RigLoadError`.

The four loader rows each resolve through a **distinct** branch, so none is hollow: `["2","1"]` → length branch (isinstance passes); `[2,1,3]` → type branch (length is equal, so it cannot be the length branch); `["10","1","2"]` → single-character branch; `["1","1","2"]` → uniqueness branch.

The IF-frame pin is well-formed: `IF00000014228000+000000G00003` is `MEASURED_FRAME` with only the P6 character changed (`2`→`G`), same length, so it cannot pass for a framing reason — `CatParseError` can only come from `_mode_name`.

`test_mode_map_codes_are_single_characters_and_round_trip` is largely tautological over the map, as suspected — but its `len(code) == 1` assertion is not: it failed in the RED run with `AssertionError: C4FM-DN`. It adds the single-character guarantee and nothing else.

## The central prose claim

"the legacy `self._state` mirror, which the observation pipeline never writes (MOR-2518)" appears three times. **CONFIRMED, and it is load-bearing.** `yaesu_cat/observations.py` contains no write to `_state`/`radio_state` at all (grep, no `head`). The only writers of `RadioState.main.nb_level`/`nr_level` anywhere in `src/` are `yaesu_cat/poller.py: YaesuCatPoller._poll_slow`, and they sit **after** `if await self._emit_slow_control_observations(): return` — i.e. in the legacy fallback branch only. So on the observation path the mirror's NB/NR levels are permanently 0, the old `set_nb(True)`'s "keep existing level" branch was unreachable, and every NB toggle clobbered the operator's level with the midpoint 5. The sentence is narrow and true; deleting it would remove the reason the change exists.

`get_if_status` and `get_mode` do write `_state`, but neither is the observation pipeline: `get_if_status` is called only from `connect()`, and the poller's `get_mode(0)` is in the same post-early-return legacy branch.

## Claims checked and confirmed

- Every `YaesuCatRadio(...)` construction in `src/`, `tests/`, `.github/scripts/` (grep without `head`, 60+ sites): none builds a `RigConfig` by hand — there is exactly one `RigConfig(` literal in the repo, inside `load_rig`. Every site resolves to `rigs/ftx1.toml`, the only `yaesu_cat` profile, which now carries codes. `replace(config, mode_codes=None)` in the new pin is the only `None` path, and it is the test that asserts the raise. `mode_codes=None` remains reachable for every Icom profile, so the optional field is not an orphan.
- `read_nb_level(1)`/`read_nr_level(1)` resolve to `get_nb_level_sub = "NL1;"` and `get_nr_level_sub = "RL1;"` — both real in the profile. `receiver` reaches both branches.
- No deadlock from the added read: `YaesuCatTransport.query` and `.send` each acquire the exchange gate for one exchange; `set_nb` performs two sequential exchanges, not a nested one.
- `get_data_mode` is MAIN-only because `core/radio_protocol.py: get_data_mode` takes no `receiver` — the protocol's signature, not a dropped parameter.
- `mode_codes` is not serialised anywhere: grep across `*.py`, `*.ts`, `*.json`, `*.md` finds it only in the loader, the backend and the new tests. No capabilities payload or frontend fixture is affected.
- No other profile carries a `codes` key (`grep -rn "^codes" rigs/*.toml`).
- Guardrails: 7 files / 408 changed lines — soft threshold crossed on files, and the PR body's first line states the one-unit-of-work justification, as required. Hard ceiling not approached.
- `uv run lint-imports`: 5 contracts kept, 0 broken. `uv run ruff check` and `ruff format --check` clean on all five Python paths. Nothing in `backends/` imports `web/` or `rigctld/`.
- PR-body evidence verified independently: GREEN run **35655831074** checked out `7a6af85f...` (`git checkout --progress --force 7a6af85fac2d7b6302c7db3e77bba31e8673f71f` in its log) and reported `1066 passed, 1 skipped`. Both numbers as claimed.
- All three commit messages are true of the net change. No squash-body correction is required for their technical content.

## CI state as found

`gh pr checks 3556`: `quick` **pass** (1m42s) at head `7a6af85f` — confirmed via `gh run view 35656498819 --json headSha`. `classify`, `grep-gate`, `publish-docs-only-quick-status` and the four observation publishers pass. `quick-v2-observe` pending (waits on canonical legacy quick). `Agent Review Gate` and `Agent Review Gate v2 observe` fail with "Missing trusted standalone exact-head Agent Review PASS directive" — that is this comment's absence, and it should clear on the next gate run. Nothing red is caused by this PR's code.

## Notes (not blocking, no commit needed)

1. **Latency on the rigctld `get_mode` fallback.** `rigctld/handler.py` reads the mode via `get_mode_reader` and then calls `get_data_mode()` on the next line; for Yaesu both now issue `MD0;`, so that branch does two identical CAT reads per poll where it previously did one. It only runs when the state store has no `main_state`, and correctness is unchanged, so it is not a regression — but it is free to eliminate later by deriving the flag from the mode string already in hand.
2. **Unassigned code escalates a cycle, not a field.** A `CatParseError` from `read_mode` reaches `observations.py: _safe_read` with declared paths, which raises `DeclaredCommandDefect` and records a startup defect — the whole poll cycle, not one field. This matches the file's existing contract for any declared read that answers in an unknown shape, and `0`/`G`/`J` are unassigned on this firmware, so it is unreachable today. Worth remembering if Yaesu ever assigns one.
3. **NB/NR toggle now fails on a read timeout** where it previously wrote the default. Defensible (a radio that will not answer `NL0;` will not honour `NL0005;` either), and it is a deliberate consequence of the change.
4. **Lower-case and whitespace codes pass the loader.** `[modes].codes = ["a"]` validates; the radio answers upper-case, so such a profile would decode nothing. Profile-authoring hazard only; `rigs/_schema.md` does not claim case validation, so its new row is not false.
5. **Pre-existing doc drift, untouched and not worsened:** `docs/api/rig-loader.md` § Validation Rules lists 14 of the loader's 188 `raise RigLoadError` sites and its `RigConfig` snippet already omits several real fields (`rx_audio_channel`, `filter_width_min`, …). Omitting the new key there is consistent with that table's existing practice, and the key is documented where the schema lives, in `rigs/_schema.md`.
6. **Dead branch in `get_data_mode`,** pre-existing: `mode.endswith("-D")` cannot fire for any FTX-1 mode name — `DATA-L`, `DATA-U`, `DATA-FM`, `DATA-FM-N` all resolve through `"DATA" in mode.upper()`, and the only test that exercised `-D` (`radio_state.main.mode = "USB-D"`) was replaced by a live-read pin. `PSK` answers `False`; if the operator expects PSK to count as a data mode, that is a separate ticket.
7. **Attribution to confirm before writing the squash body.** All three commits carry `Co-Authored-By: Claude Fable 5.1`. I have no way to verify authorship from the repository, and the dispatch that commissioned this review names a different model. Whoever merges should confirm the trailer matches the actual author before it lands in the squash body, since this repository squash-merges with `COMMIT_MESSAGES`.

## Suggested receive-only bench check (no transmit)

Switch the radio by hand and read back: `MD0;` in C4FM-DN → `MD0H;` and in C4FM-VW → `MD1I;`/`MD0I;`; one non-C4FM mode as a control. Confirm `IF;` carries the same P6 character. Set the NB level to a non-midpoint value (e.g. 3), toggle NB off and on in the UI, and confirm the level returns to 3 rather than 5 — that is the operator-visible fix. Repeat on SUB to confirm `NL1;`/`RL1;` are addressed, not MAIN. Confirm the width control refuses in C4FM-DN instead of sending a raw index. Confirm a WSJT-X-style `get_mode` still reports `PKTUSB` in DATA-U.

PR #3557, opened 2026-09-21
perf(MOR-2521): the S-meter updates attributes instead of creating and removing nodes

morozsm:
MOR-2521, first part. One unit of work (9 files, above the 6-file soft threshold): one component changes, and the other eight files are tests that asserted on the presence or absence of the nodes it used to create and remove.

## What changes

`components-v2/meters/LinearSMeter.svelte` created and removed SVG nodes as the reading moved: each lit segment of the main row and of the lower row was a `<rect>` inside `{#if}`, and the peak line was a `<line>` inside `{#if showPeak}`. A 90-second observation of the live page earlier today counted 327–352 add/remove pairs of S-meter nodes.

Every fill rect and the peak line are now permanent nodes; a reading changes only their `width`, `fill` and `visibility` attributes. A hidden rect paints nothing and the width and colour expressions are the ones the conditional markup used, including the `> 0.01` partial-segment guard, so no pixel is intended to change and no visual baseline is re-pinned.

## Tests

`__tests__/LinearSMeter.test.ts` sweeps the value range and pins the node counts as constant. Consumer tests that looked for an absent `[data-meter-peak]` / fill rect now look at `visibility`.

Tests ran on the build host through `focused.yml`: RED run 35656116345 (tests only), GREEN run 35656903066 at head `ff67f4a9` — 348 passed. Locally `npm run check`: 0 errors.

## Not in this PR

- The readout-slot work (value text that does not shift) moves pixels; it is kept on branch `codex/mor-2521-readout-slots` and will land with the MOR-2509 batch, which re-pins baselines once.
- `BarGauge.svelte` has the same create/remove pattern.
- Not re-measured in a browser yet.

Linear: MOR-2521 (remains open).


morozsm:
Agent Review: PASS ff67f4a9f1ba8a4a18ad4cf39d782b244391ec84

Independent review of the net diff (`git diff $(git merge-base origin/main HEAD)`, merge-base `b49e3858`): 9 files, +249/-88 = 337 changed lines — matches `gh pr view 3557 --json files,additions,deletions`. Over the 6-file soft threshold, under the 10-file/1000-line hard ceiling; the PR body carries the one-unit-of-work justification. No new imports, so `components-v2` layering is untouched.

## Pixel equivalence — CONFIRMED

`LinearSMeter.svelte: segLit`/`segWidth` reproduce the old two-arm markup exactly: `i < full` → `SEG_W`; `i === full && frac > 0.01` → `max(1, SEG_W*frac)`; everything else → `visibility="hidden"`. Enumerating the reachable states of `fullSegs`/`fracSeg` (`signalProjection.motionFraction === null` forces both to 0; over-range gives `full >= SEG_COUNT`; a negative or NaN `smoothedSegs` makes both arms false) every previously-absent rect is now hidden and every previously-present rect keeps its old x/width/fill. A *lit* rect can never get a width that is zero, negative or NaN: `segLit` requires `frac > 0.01`, and `segWidth` floors at 1. `mainPresent`/`relevant`/`DIM_OPACITY`, compact, `sdr-screen` and the `lowerActiveColor()` fault path are all outside the changed expressions.

Nothing selects the affected nodes from CSS. `grep -rn "visibility" frontend/src --include=*.css --include=*.svelte` and a grep of `frontend/src/presentation/languages/**`, `frontend/src/skins/**` and `styles/semantic-controls.css` for `[data-meter-fill]`/`[data-lower-fill]`/`[data-meter-peak]` return no rule at all — no `display`, `opacity`, `transition`, `animation` or `visibility` override, so item 2 (permanent nodes now animating between states) cannot arise. `styles/animations.css`'s `meter-pulse` comment names the legacy `SMeter.svelte` `.meter-fill` **class**, not this attribute. The rects carry no role, no aria and no pointer handler; the `<svg>` keeps `role="img"` + `aria-label`, so item 4 is a non-issue.

Empirical support: the `visual` check is green at this exact head — "Pixel-diff vs approved baselines", 39 comparisons passed in 30.4s, including `dual-main-sub--desktop`, `unified-instrument--desktop`, `peer-split-chassis--desktop` and the six `TX lower state clearance` cases that drive the lower row.

## Runtime consumers of the node set (item 3) — none outside tests

`grep -rn "meter-fill\|lower-fill\|meter-peak"` over the whole repo, no `head`:

- `frontend/tests/**` (e2e, visual, i18n) — **0 hits**. `visual-baselines.spec.ts` mounts `LinearSMeter` directly but queries `[data-main-relevant] text` / `[data-lower-row-label]` and measures bounding boxes; the permanent `<line>` is not a `<text>` and does not change any element's position among its siblings. **No impact.**
- `frontend/scripts/**`, `frontend/fixtures/capture.mjs`, `capture-ptt.mjs`, `capture-profile.mjs` — **0 hits**. `capture.mjs`'s only `LinearSMeter` mention is a comment about `requestAnimationFrame` counting. **No impact.**
- `frontend/src/**` non-test — only `LinearSMeter.svelte` itself. **No impact.**
- Counters that bypass the data attributes were swept separately (`querySelectorAll('rect'|'line')`, `line[stroke-width="2"]`, `outerHTML`, `childElementCount`): `design-language-wiring.component.test.ts: tileSegmentFills` and `MetersSurface.test.ts:849` count rects inside `BAR_FIELDS` = power/alc/drainCurrent/drainVoltage/compression (BarGauge tiles, unchanged) — **no impact**; `LinearSMeter.test.ts: denseTickXs`/`majorTickXs` filter on `y2="38"` / `y1="18"`, while the peak line is at `y1=TRACK_Y`/`y2=TRACK_Y+TRACK_H` — **no impact**; `LinearSMeter.test.ts:361` (`line` count 1, unmodified) holds because that raw-domain case armed the peak before and shows it now — **no impact**; `meter-ballistics.test.ts:183` pins source text about `createSignalMeterMotion` only — **no impact**; `BarGauge.peakhold.svelte.test.ts` rect counts are BarGauge's — **no impact**.

## Pin discrimination, per row (item 5) — CONFIRMED for all three

`LinearSMeter.test.ts` "MOR-2521 — the S-meter never adds or removes nodes across a value sweep" mounts the real component (`mount(LinearSMeter, …)` over a `proxy` state, `flushSync` per step), sweeps real `frame` props, and re-derives everything from the DOM — no re-implementation of `segLit`/`segWidth`. Discrimination is established from the builder's RED run `35656116345`, whose revision is the tests-only commit `93e619f7` (`FOCUSED_REVISION` in the log): the mutant there is the pre-change production markup itself, which is the exact bug this change removes.

- **Main row** — RED `× renders normal and SDR geometry from every supplied host-frame field` (the `fills` `toHaveLength(20)` / 10-visible assertion) and `× default variant: one node count for every sweep step`. Re-adding `{#if}` around `data-meter-fill` moves `fill` 0→20 and `rect` 22→42 across the sweep, breaking both `toEqual(steps[0].counts)` and the literal `{ rect: 42, line: 82, fill: 20 }`. **FAILS.**
- **Lower row** — RED `× MOR-2521: keeps all 20 lower fill slots present (hidden, not absent) at valueFraction 0` and `× with lowerScale: one node count for every valueFraction step`. The fractions `[0, 0.53, 1, 0.0004]` would give rect 63/74/83/63. **FAILS.**
- **Peak line** — RED `× suppresses motion, peak, and geometry for unprojectable domain {"kind":"engineering","unit":"db"}` and `{"kind":"unknown"}`, which are exactly the `[data-meter-peak]` `visibility === 'hidden'` assertions; with `{#if showPeak}` the element is absent and `?.getAttribute` yields `undefined`. Independently, `line: 82` would drop to 81 at sweep step 4 and `peakVisible` would be all-false. **FAILS.**
- The `sdr-screen` row stayed green in the RED run. That is expected, not hollow: the SDR markup was never conditional, so the mutation cannot reach the branch that row covers. It is a stability guard for the SDR path, not a pin on this change.

Reverse direction: the same ten target files are green at this head (run `35656903066`, `FOCUSED_REVISION: ff67f4a9…`, 10 files / 348 tests passed), so the checks do not redden on correct work.

Retargeted consumer tests preserve the behaviour they asserted — presence became `visibility`, never "any node will do". `MetersSurface.test.ts: visibleSlotCount` and the `ReceiverInstrumentHost` `:not([visibility="hidden"])` selectors are behaviour-preserving for `[data-gauge-fill]` too, because BarGauge rects carry no `visibility` attribute and so still count. `lower-scale.test.ts`'s `lit()` moved from `rect:not([data-segment]).length - 2` to an explicit visible-`[data-meter-fill]` count — tighter, not weaker. The `.every(line => hidden)` rewrites in `MetersSurface.test.ts:844` and `ReceiverInstrumentHost.isolated.test.ts:459` are vacuous on an empty match, but the previous assertion was `toHaveLength(0)`, which also passes on an empty match, so nothing was lost; a *visible* regression still fails both. No weakened test found.

## Removed readout-slot work — fully absent

`git diff <merge-base>` grepped for `slot`/`text-anchor`/`READOUT_`/`reserve`/`min-width` returns only the identifier `visibleSlotCount` and the words "permanent slots" in comments. `semantic-controls.css` is byte-identical to `origin/main` (not in the diff). The four RED-run tests that pinned the readout slots (`every {displaySUnit} text uses a fixed anchor, never middle`, `the default variant anchors both readouts at slot left edges reserved by char count`, `.meter-native-value reserves 6ch right-aligned…`) are gone at this head. Clean.

## Prose in the diff — checked, all true

`segLit`/`segWidth` header comment: "the node-count sweep … fails if a value step adds or removes a node" — true (RED evidence above). "The `frac > 0.01` arm keeps the sub-1% partial guard … pinned by the same sweep" — true: step `0.15025` (→ `fracSeg 0.005`) expects 3 visible fills, and dropping the guard makes it 4; the lower-row `0.0004` case expects 0 and would become 1. The literal totals are derivations I re-checked against the source, not just against the green test: `<rect>` sites in the markup give container background + bar track + 20 dim + 20 fill = 42, and + lower track + 20 + 20 = 83; `smeter-scale.ts: scaleTicks` emits one major plus nine subdivisions per interval, so the 9-row `IC7610_LIKE_CAL` (8 intervals) yields 80 + the closing major = 81 ticks, + 1 peak line = 82.

## Class sweep

The conditional-lit-node shape occurs in exactly two components: `grep -rln "{#if i <" frontend/src --include=*.svelte` returns `LinearSMeter.svelte` and `BarGauge.svelte` (`BarGauge.svelte:231,238` plus `{#if peakPct !== undefined}` at `:249`). The PR body names `BarGauge.svelte` as the remaining instance — the class is enumerated and reported, not silently left.

This repository squash-merges with `COMMIT_MESSAGES`, so all four branch commit messages are concatenated into the squash body and land on `main`. Two of them describe work the net change no longer contains. The hand-written squash body must **not** carry:

1. From `e4da7ce1` — the subject clause "**; its value text no longer shifts**", the whole paragraph beginning "Both SVG readout texts move from a value-dependent middle anchor to a start anchor…", and the whole paragraph beginning "The bottom dock's .meter-native-value / .meter-native-secondary spans get the same treatment: right-aligned reserves of 6ch / 7ch…". None of it is in the diff: the readouts are unchanged at `x={READOUT_CX}` with `text-anchor="middle"`, and `semantic-controls.css` is not in the PR's file list.
2. From `93e619f7` — the subject clause "**and fixed readout slots**" and the second bullet "source pins for the readout slot declarations (start anchors at fixed slot left edges reserved by char count) and the dock spans' reserved right-aligned widths". Those tests were dropped at `ff67f4a9`.

The squash subject should stay the PR title. This is a body-at-merge-time correction, not a code defect: it needs no commit, and it is not a reason to block.

## Notes (no action required)

- `gh pr checks 3557` as found: `visual` **pass** (1m35s, 39 pixel comparisons), `changes`, `classify`, `grep-gate`, `publish-docs-only-quick-status` and the observers **pass**; `quick` and `quick-v2-observe` **pending**; `Agent Review Gate` / `Agent Review Gate v2 observe` red only for "Missing trusted standalone exact-head Agent Review PASS directive" — this comment supplies it. Nothing red is caused by this PR.
- `npm run check` reproduced locally at this head: 4867 files, 0 errors, 0 warnings — the PR body's claim is accurate.
- `git merge-tree --write-tree HEAD origin/codex/mor-2509-vfo-skeleton` exits 0 with a single tree id and no conflict block: **no conflict with open PR #3554**. Reported only, not resolved.
- Not verified here, for a later browser measurement: the 90-second live count of 327–352 add/remove pairs (PR body, no browser in this review) and whether the approved baselines capture the S-meter in a *lit* state — the 39 green comparisons prove the comparison ran and covers the meter's composites, but a regression confined to lit-segment rendering could in principle sit in a scenario whose meter is unlit. `quick` will close the remaining unit-test surface (the focused run covered 10 files; the untouched `LinearSMeter`-adjacent suites were cleared by grep, not by execution).

PR #3558, opened 2026-09-21
fix(MOR-2513): unobserved public state is null, and the frontend treats null as unknown

morozsm:
Guardrail exception (file count only, granted by the coordinating session under the owner's delegation of 2026-09-03): 22 code + 0 regenerated baselines = 22 files. Changed lines are 715 + 279 = 994, under the line ceiling; a large share of them is the two mechanical type files (`src/rigplane/web/state_schema.py` and the generated `frontend/src/lib/types/state.ts`). This is one unit because the payload change and the frontend consumers that must tolerate it cannot land separately: the payload alone makes `npm run check` fail in ten frontend files.

MOR-2513. Owner rulings, 2026-09-21: every public state leaf that has not been observed is published as `null` with its key present (option A), for all such leaves; and the operator's screen never prints `null` or a placeholder for it.

## What changes

**Server.** `web/runtime_helpers.py: _null_unobserved_public_leaves`, called inside `_build_public_state_payload_from_dict`, replaces the value of every public leaf whose `fieldStatus` says it was never observed with `null`. A stale leaf keeps its last value. Leaves that carry no status entry are untouched. `txTarget` has a status entry and is excepted explicitly: it expresses absence in-band, as an unknown-status object, not as `null`. `web/state_schema.py` makes those leaves nullable; `frontend/src/lib/types/state.ts` is regenerated from it (`scripts/gen_state_types.py --check` passes). `docs/api/web.md` and `docs/internals/radio-state-pipeline-validation.md` state the contract.

**Frontend consumers.** A `null` leaf is treated as unknown: never printed, never replaced by a default.
- `adapters/display-observation.ts`: `null` maps to the existing unknown observation, which the panel adapters already render as unavailable.
- `adapters/scope-passband-display.ts`: a `null` leaf means no passband is drawn; a `null` mode means no PBT lattice.
- `props/panel-props.ts`: ATT / PRE badges unlit; no filter config or PBT reading from a `null` mode.
- `adapters/radio-view-model-adapter.ts`: `currentMode` becomes an unknown reading instead of failing the view-model validator.
- `commands/panel-commands.ts`: no tuning step, band select, data-mode cycle, keyboard target or PBT write is emitted from a `null` value.
- `components-v2/layout/RadioLayout.svelte`: a `null` meter is passed as absent, so its tile is not drawn; three unused derivations that substituted `0` / `''` are deleted.

## Left for MOR-2520 (not in this PR)

`toVfoProps.sValue ?? 0` still substitutes zero for the VFO strip's S value; `AudioRoutingControl`, `LcdLayout`, `AmberCockpit`, `AmberScope` and `MemoryPanel` were not changed and their behaviour with a `null` leaf has not been checked on screen. No visual suite and no browser measurement has been run on this branch.

## Evidence

Tests ran on the build host through `focused.yml` against pushed commits. Server part: RED 35654696018, GREEN 35655534113 (288 passed). Frontend part: RED 35657442828 (the render pin failed in the view-model validator), GREEN 35657571294 (487 passed); 35657869902 (96 passed, ruff clean); final head `1dd0d074`: 35659780356 (124 passed, including `tests/test_state_type_codegen.py`). Locally at the final head: `npm run check` 0 errors, `uv run mypy --strict src/rigplane/web`, `scripts/gen_state_types.py --check`.

Consumer-visible API change: clients of `/api/v1/state` and the WebSocket state stream now receive `null` where they previously received a dataclass default for a never-observed leaf.


morozsm:
Agent Review: BLOCKED 3ce8e882030b6361812db5e8a583440c853c0010

Independent review. The design is right and the server contract is measurably correct; two blocking defects were introduced by the last commit.

## BLOCKING 1 — generated `state.ts` is stale; `quick` and `state-types-gate` are red (IN THE DIFF)

`frontend/src/lib/types/state.ts:1-322` no longer matches `src/rigplane/web/state_schema.py`.

- Reproduced locally: `uv run python scripts/gen_state_types.py --check` → `ERROR: frontend/src/lib/types/state.ts is stale.` exit 1.
- CI at this head: `state-types-gate` **fail** (run 35658079068) with the same message; `quick` **fail** (run 35658079088) — `1 failed, 16774 passed, 217 skipped, 13 xfailed`, the single failure being `tests/test_state_type_codegen.py::test_state_type_generation_is_deterministic_and_current`. `visual` passed. Both failures have this one cause and this PR causes them.
- Cause: commit `3ce8e882` trimmed the docstrings in `state_schema.py` without re-running the generator. I diffed the committed block against `gen_state_types.py --stdout`: **the drift is docstring lines only, no type changes** — so the regeneration cannot invalidate the frontend review already done here, and the changed-line total stays at 999.

Required fix: run `uv run python scripts/gen_state_types.py` and commit. See BLOCKING 2 before choosing which text to regenerate from.

## BLOCKING 2 — a false claim about `txTarget`, in four places (IN THE DIFF)

Measured with the real builder (`uv run python -c` over `StateStore().snapshot()` → `build_public_state_payload_from_snapshot`, `receiver_count=2`):

```
fieldStatus entries: 199
txTarget in fieldStatus: True {'storePath': 'global.tx_state.tx_target', 'observed': False,
                               'freshness': 'unknown', 'availability': 'missing'}
```

`txTarget` **does** have a `fieldStatus` entry — that is precisely why `_null_unobserved_public_leaves` needs its `or public_path == "txTarget"` special case. Four sentences added or touched here deny it:

1. `src/rigplane/web/runtime_helpers.py:963` — `_null_unobserved_public_leaves` docstring: "Paths without a status entry (connection, health, revisions, ``wsClients``, ``txTarget``) are not radio observations." False for `txTarget`. The pre-trim wording at `61eae4b1` was correct; the trim deleted the explanation and kept the wrong premise.
2. `src/rigplane/web/state_schema.py:287` — `ServerStatePublic` docstring: "Leaves with a ``fieldStatus`` entry publish ``null`` while unobserved." `txTarget` has one and does not.
3. `frontend/src/lib/types/state.ts:13-18` — the stale generated copy asserts both halves and so contradicts itself on `txTarget`.
4. `docs/api/web.md:981` — "A field whose observation has not arrived ... publishes `null`, never a fabricated default." Unobserved `txTarget` publishes `{"status": "unknown", "reason": "not-observed"}`, not `null`.

Correct wording already exists in this PR and should be reused — `tests/test_web_runtime_helpers.py`, in `test_unobserved_leaves_publish_null_with_keys_present`: "``txTarget`` encodes absence fail-closed in-band (an unknown-status object), not as a dataclass default; it is not nulled."

Also in this class, and contradicted by its own test body ~20 lines below: `tests/test_web_runtime_helpers.py` — `test_ftx1_empty_store_nulls_every_unobserved_status_leaf` docstring now reads "FTX-1 unobserved: **every published leaf is null**", while the body asserts `connection`, `radioDetail`, `wsClients` and `revision` keep their values. Narrow or delete.

Required fix: correct 1, 2 and 4, then regenerate so 3 follows. Restoring the pre-trim `runtime_helpers.py` sentence and adding the `txTarget` clause to `ServerStatePublic` fixes the class in one pass.

Checks to run after the fix: `uv run python scripts/gen_state_types.py --check`, then the natural `quick` run at the new head. No frontend re-run is needed for the regeneration itself (comment-only drift).

## What I verified and could not refute

**Server contract (ruling A) — CONFIRMED by measurement, not by reading.** I built the public payload from a fresh `StateStore` on both the merge base (`b49e3858`, extracted with `git archive` into scratch) and this head, and diffed every leaf:

- 155 leaves changed, **every one of them to `null`**; "changed to something NOT null: []". Old defaults were `0` (94), `False` (44), `'USB'` (6), `128` (4), `'A'` (2), `'MAIN'`, `True`, `0.0`, `1`, `[]`.
- Repeating it with three observations applied (`freq_hz`, `mode`, `power_on`): **observed leaves whose value changed: none; fieldStatus entries changed: none.** This independently reproduces the claim in `test_observed_values_match_the_prewire_contract_byte_for_byte`'s docstring. The comparison demonstrably reports inequality (it reported 155 and 152 differences respectively), so "no change" here is a measurement.
- `UNOBSERVED-BUT-NOT-NULL: []` — no status-carrying leaf escapes, except the deliberate `txTarget`.

**Status-less leaves — 21, not 25, and all genuinely non-observations:** `connection.{controlConnected,radioReady,rigConnected}`, `revision`, `stateRevision`, `freshnessRevision`, `observationSeq`, `healthRevision`, `updatedAt`, `radioDetail.status`, `radioHealth.{lastError,likelyCause,radioLink,readiness,serverReachable,sinceMs}`, `wsClients.{scope,control,audio}`, `txTarget.{status,reason}`. None is a radio observation, so ruling A is met. `txTarget` is the one judgement call — its unobserved encoding is the in-band unknown object rather than `null`; that is honest and the PR body says so plainly, but it is an exception to "no exceptions list" and is the owner's call, not mine.

**Single choke point — CONFIRMED.** `_null_unobserved_public_leaves` is called once, inside `_build_public_state_payload_from_dict`, which both `build_public_state_payload` and `build_public_state_payload_from_snapshot` funnel through; `server.py` routes HTTP (`build_public_state`), the initial WS full and the WS delta (`_build_public_state_for_delivery` → `_delta_encoder.encode`) through the same `_build_public_state_payload_from_snapshot_impl`. The builder `deepcopy`s its input, so the canonical `RadioState` is not mutated. **rigctld is unaffected**: `rigctld/handler.py` reads `state.sub.mode` off the dataclass and never builds or parses the public payload. **Vendored consumer contracts are unaffected**: `contracts/consumer-expectations/` covers only the `discovery` and `info` artifacts, not `/api/v1/state` (note: `consumer-contracts-gate.yml` would not have run here anyway — its path filter lists `server.py`/`discovery.py`/`api_contract.py`, not `runtime_helpers.py`).

**Stale keeps its value (ruling D) — CONFIRMED** by `test_observed_value_stays_published_after_the_freshness_tick_marks_it_stale` and by reading the helper: it keys only on `observed is False`.

**Frontend, ruling B/C — no category (c) found.** Svelte 5.55.8 renders `null` in a text position as `''`, not `"null"` (`svelte/src/internal/client/render.js`: `value == null ? '' : ...`). `npm run check`: `4867 FILES 0 ERRORS`. The decisive argument is the before/after table above: for the 138 leaves whose old default was `0`/`False`, every `?? 0` / `?? false` / `!== undefined` consumer produces exactly what it produced before, so those sites cannot be new defects. I then swept the 17 leaves whose old default was something else:

- `txBandEdges` (`[]`→null) and `scopeControls.fixedEdge` (object→null): **no production consumers at all** in `frontend/src`. No `.map()` on null, no throw.
- `active` (`'MAIN'`→null): every read is `=== 'SUB' ? sub : main` or `?? 'MAIN'` — null behaves identically to the old `'MAIN'`.
- `activeSlot` (`'A'`→null): every consumer gates on `observedAvailableField` / `activeSlot !== 'A' && activeSlot !== 'B'`.
- `powerOn` (`True`→null): `lib/stores/radio.svelte.ts` already mapped it through `isFieldAvailable`, yielding `null` before and after.
- `txAntenna` (`1`→null): `panel-props.ts: toAntennaProps` `?? 1` — unchanged.
- `pbtInner`/`pbtOuter` (`128`→null): `toAudioSpectrumProps` `?? 128` unchanged; `toFilterProps` now yields no reading instead of a fabricated centre — an improvement.
- `mode` (`'USB'`→null): `RadioLayout`/`LcdLayout` `if (activeMode)` skip (previously ran `applyModeDefault('USB')` on a fabricated value); `toMemoryPanelProps` returns the pre-existing `'---'` sentinel, but `MemoryPanel.storeVfoToChannel` is a click handler and `panel-commands.ts: makeMemoryHandlers().onStore` refuses when the snapshot is null, so nothing renders or persists.

`RadioLayout.svelte`'s new comment — "MetersDockPanel's own priority order leaves that tile not drawn — never a fabricated zero" — is **true for all eight props passed**: `MetersDockPanel.svelte` gates every tile on `!== undefined` (`compMeter` additionally on `compressorOn === true`). This is a real improvement over the previous seven zero-valued tiles.

**Commands.** The guards are tightened, not weakened. The one substantive behaviour change: before, `cycle_data_mode` passed `Number.isSafeInteger(0) && 0 >= 0 && 0 < count` on the fabricated default and **dispatched a command computed from a value the radio never reported**; it now refuses. The keyboard fan-out test is extended (+48), not weakened — the new case asserts refusal for fourteen actions that all DISPATCH on the live fixture, and the fixture arithmetic gained `throw`-guards rather than `?? 0` fallbacks.

**Pin discrimination**, judged from the builder's RED runs as instructed (no local pytest/vitest, no pushes available):
- Server, run 35654696018: eight pins FAILED against pre-fix code with genuine assertions (`assert 0 is None`, `AssertionError: main.freqHz`). They discriminate.
- Frontend, run 35657442828: the `RadioLayout` render pin failed with `TypeError: Invalid radio view model at $.modeFilter.currentMode.reading.value: expected a string` — a null mode crashed the view-model validator, i.e. exactly the panel-blanking exception the fix removes. NARROWED: this proves the mount path reddens, but not that the `not.toContain('null')` arm discriminates — and since Svelte renders null as `''`, that particular arm can essentially never fire. The `'NaN'` and `'0.000'` arms can. Not blocking; the test's name over-claims slightly against what it can detect.

**Test quality.** Server pins use the real `StateStore`, real `FieldPath`/`Observation`, real profile payload builder and real `WebServer` — no MagicMock. The `3ce8e882` "compaction" is a **faithful** refactor: the three parity assertions each still assert their own dict with their own `storePath` and `lastObservedMonotonic`, so a failure still localises; nothing was deleted or merged. Its "No behavior change" claim is true of `runtime_helpers.py` and `panel-props.ts` (comments only). The `tests/test_mor2338_web_compatibility.py` one-line edit (`== 0` → `is None`) is the honest restatement of the intended contract and still pins key presence and `fieldStatus`; it is not a weakening.

**Guardrails and layering.** 22 files, 720+279 = 999 changed lines — file count over the hard ceiling, exception recorded in the required format on the PR body's first line; lines under 1000. `uv run mypy --strict src/rigplane/web` clean (27 files); `uv run ruff check` on all changed Python clean; `uv run lint-imports` 5 contracts kept, 0 broken.

**Open-PR interaction.** #3554 and #3557 merged to `main` at 21:37 today, so the brief's branch names are stale. `git merge-tree --write-tree HEAD origin/main` → exit 0, clean, despite both sides touching `RadioLayout.svelte`. Semantically clean too: #3554's `VfoPanel.svelte` is already null-aware (`sValue === null` → `"S meter unknown"` aria-label, and it passes `null` through to `LinearSMeter`), and #3557's added lines read `main.sMeter` only as a `fieldStatus` path string. No category (c) appears once both are on main. #3559 is unrelated.

## Notes, not blockers

- `props/panel-props.ts: toVfoProps` still does `sValue: rx.sMeter ?? 0`, so the VFO strip keeps showing S0 — but `sMeter`'s old default was also `0`, so this PR does not worsen it. Correctly left to MOR-2520, as the PR body says.
- Ruling D, for the live stand: the bottom meter dock previously drew seven tiles at zero from connect and now draws none until observed, so tiles will **pop in and reflow the grid** as observations arrive. That is the honest behaviour and I do not think it blocks, but it is the one thing a browser measurement should watch — together with the LCD/amber faces (`lcd-cockpit`, `lcd-scope` are live entries in `skins/registry.ts`, not dead legacy) and the mobile layout, none of which this PR's pins or `visual` cover.
- PR body, REQUIRED BEFORE MERGE: "Locally at the final head: ... `scripts/gen_state_types.py --check`" is false at this head. `npm run check` 0 errors and `mypy --strict` are true (I reproduced both). Once BLOCKING 1 is fixed the sentence becomes true again; if the body is not regenerated, correct it.

Left byte-identical: `git status --short` clean at `3ce8e882030b6361812db5e8a583440c853c0010`.

morozsm:
Agent Review: PASS 1dd0d074cd9e61fa1d6eb56901779e349ef2524b

Delta review of `3ce8e882..1dd0d074` (one commit, 6 files, +24/-29). Both BLOCKING findings from my review of `3ce8e882` are fixed. No new defect that a commit must fix.

## BLOCKING 1 (stale generated types) — RESOLVED

- `uv run python scripts/gen_state_types.py --check` → `state.ts generated block is up to date.`, exit 0.
- `state-types-gate` is now **pass** (run 35659776238).
- The `state.ts` delta is **comment-only**, proven rather than eyeballed: stripping JSDoc blocks and blank/`//` lines from both revisions leaves 249 lines on each side and `diff` reports them identical. I first confirmed that comparison can report inequality (a one-token mutation of the new side is flagged) — an earlier attempt of mine returned a false "identical" on two empty files and was discarded.

## BLOCKING 2 (false `txTarget` prose) — RESOLVED, and the new prose holds

All four sites I named are corrected, plus the two the coordinator listed. I re-measured the underlying fact through the real builder: an unobserved `txTarget` carries `{'status': 'unknown', 'reason': 'not-observed'}`, and its `fieldStatus` entry is `{'storePath': 'global.tx_state.tx_target', 'observed': False, 'freshness': 'unknown', 'availability': 'missing'}`.

Attacking the replacements as their own pass:

- `runtime_helpers.py: _null_unobserved_public_leaves` — "``txTarget`` excepted: it encodes absence in-band (an unknown-status object), not as null". TRUE, and the summary line now matches the code's `or public_path == "txTarget"` guard exactly. The false "paths without a status entry (… ``txTarget``)" sentence is deleted rather than softened, per the owner ruling.
- `state_schema.py: ServerStatePublic` — same clause. TRUE.
- `frontend/src/lib/types/state.ts` JSDoc — carries the corrected schema docstring; the self-contradiction is gone.
- `docs/api/web.md` — "`txTarget` excepted: it encodes absence fail-closed in-band (an unknown-status object), not as a dataclass default." "Unknown-status object" is literally what the payload carries. "Not as a dataclass default" is TRUE: the value is a literal set by `_build_public_state_payload_from_dict` via `state.setdefault("tx_target", {"status": "unknown", "reason": "not-observed"})`, not a `RadioState` field default. "Fail-closed" is fair — a consumer reads `status: "unknown"` rather than a usable target.
- `test_ftx1_empty_store_nulls_every_unobserved_status_leaf` docstring — "txTarget (in-band unknown object) and paths without an entry keep their delivered values" now matches the body, which asserts exactly `connection`, `radioDetail`, `wsClients` and `revision`. The previous "every published leaf is null" over-claim is gone.
- `RadioLayout.isolated.test.ts` comment — now describes what the fixture loop does, including the `txTarget` skip. TRUE.

## No assertion was thinned — proven, not read

I parsed both revisions of each changed Python file, stripped every module/class/function docstring, and compared the ASTs:

```
SAME src/rigplane/web/runtime_helpers.py
SAME src/rigplane/web/state_schema.py
SAME tests/test_web_runtime_helpers.py
```

The same comparison against the merge base reports `DIFFERENT` for two of those files, so it is capable of detecting a real change. For `RadioLayout.isolated.test.ts`, stripping comments leaves 848 lines on each side, identical, with the pin and its `not.toContain` assertions still present; that comparison also detects an appended line. So the delta touched **prose and the generated file only** — no test body, no production logic.

Static checks at this head: `uv run ruff check` clean, `uv run ruff format --check` 3 files already formatted, `uv run mypy --strict src/rigplane/web` clean (27 files). `npm run check` was 0 errors at `3ce8e882` and the delta changes only comments in `.ts` files, so it cannot have regressed — I did not re-run it.

Guardrails: 22 files, 715 + 279 = **994** changed lines, confirmed with `git diff --shortstat` against the merge base. File-count exception recorded in the required format on the PR body's first line.

## CI as found at this head

| Check | State |
|---|---|
| `state-types-gate` | **pass** (1m0s) |
| `visual` | **pass** (1m13s) |
| `quick` | **pending** |
| `Agent Review Gate` | fail — "Missing trusted standalone exact-head Agent Review PASS directive" (this comment is that directive) |
| `Agent Review Gate v2 observe` / `Trusted review observation publisher` | fail |
| `quick-v2-observe` | fail |

The three v2 observer failures share one cause, which I read from the failed log: `Agent Review Gate v2 observation evaluation failed closed: pull request base is not the current main head`. `main` has moved twice since this branch's merge base `b49e3858` — #3554 and #3557 landed at 21:37, and #3559 landed as `f9de95fa` while I was reviewing. **Not caused by this PR's code**; it is a stale-base observation. Whoever merges should expect these to stay red until the branch is brought onto the live `main`, and should confirm the required set (`quick`, `visual`, `Agent Review Gate`) at the exact head immediately before merging.

I did not wait for `quick`. Its previous failure at `3ce8e882` was the single test `tests/test_state_type_codegen.py::test_state_type_generation_is_deterministic_and_current` (`1 failed, 16774 passed`), which is exactly what this delta fixes and which now passes on the build host.

## PR body

Verified true at this head: the 994 line count; "22 code + 0 regenerated baselines = 22 files"; "`scripts/gen_state_types.py --check` passes"; "`txTarget` has a status entry and is excepted explicitly"; "Leaves that carry no status entry are untouched"; `npm run check` 0 errors; `mypy --strict`.

I also checked the newly cited evidence run rather than taking it on trust. `gh run view 35659780356 --json headSha` reports `f9de95fa`, which is **not** this head — but that field is the workflow ref (`main`), not the tested revision. The log shows `git checkout --progress --force 1dd0d074cd9e61fa1d6eb56901779e349ef2524b`, `FOCUSED_REVISION: 1dd0d074…`, `test_state_type_generation_is_deterministic_and_current PASSED`, `124 passed`. The claim is TRUE.

**REQUIRED BEFORE MERGE (PR body edit, no commit needed):** the body still says "447 of them are the two mechanical type files". At this head those two files are 224 + 212 = **436** changed lines (`git diff --numstat b49e3858..HEAD -- src/rigplane/web/state_schema.py frontend/src/lib/types/state.ts`). The regeneration shrank them and the figure was not updated with the 720→715 / 999→994 edits. Replace 447 with 436.

**MANDATORY SQUASH-BODY CORRECTION:** none. I re-read all seven commit messages against the net change and found nothing false of it. Commit `d54b6720`'s "regenerate frontend/src/lib/types/state.ts" was briefly falsified by the trim commit and is true again at this head. Note for the hand-written squash body: as the coordinator already knows, drop any "status-less leaves" count — the figure is 21, not 25, measured on a fresh store at `receiver_count=2`.

## Note, not blocking

`docs/api/web.md:981` opens the bullet with "State keys are always present." Measured against the real builder, two keys can be absent: `sub` is popped for `receiver_count=1`, and `main.unselectedVfo` is absent on a fresh store even though `fieldStatus['main.unselectedVfo.freqHz']` exists — the normal state at connect for a relative-VFO radio.

I do not treat this as blocking, and I want to be explicit about why rather than leave the calibration implicit. Unlike the `txTarget` sentence, this one is not contradicted by a fact it asserts: the bullet's subject is the nulling contract, and of that mechanism the sentence is true — `_null_unobserved_public_leaves` assigns `None` and never deletes a key. Both counterexamples come from separate, pre-existing projection rules, and the `sub` one is already documented two bullets above in the same list. It is imprecise about a scope it does not address, not false about the scope it does. It is also a sentence I passed over at `3ce8e882`, so raising it as a blocker now would buy a third round for a wording nuance after the substantive defects are fixed.

Still worth a one-line narrowing whenever this file is next touched, because `unselectedVfo` is not mentioned anywhere in `docs/api/web.md` and a client that discovered the key from `fieldStatus` paths could reasonably expect it always to be there: "Nulling never removes a key: a field whose observation has not arrived …". That states what the mechanism guarantees without claiming anything about conditional groups.

Everything I cleared at `3ce8e882` — the measured server contract (155 leaves, all to null; zero observed-value or `fieldStatus` drift), the single choke point shared by HTTP and WebSocket, rigctld's independence from the public payload, the frontend consumer sweep with no category (c), and the pin discrimination evidence from the RED runs — is untouched by a delta that changes only comments, so I did not re-verify it.

Left byte-identical: `git status --short` clean at `1dd0d074cd9e61fa1d6eb56901779e349ef2524b`.

PR #3559, opened 2026-09-21
fix(MOR-2523): shortcut hints appear only when Alt/Option is held alone, never during a chord

morozsm:
MOR-2523. Found by the independent review of #3555: `KeyboardHandler.svelte: handleKeydown` raised the shortcut hints (`body[data-shortcut-hints]`) on the keydown of Alt itself, so since the default AF-level / RF-gain chords moved to Alt/Option+Arrow the hints appeared on every control during every keyboard adjustment.

## What changes

- The hints appear only after Alt/Option has been the only key down for `ALT_HINTS_DELAY_MS` (400 ms, one constant in `keyboard-map.ts`).
- Any other keydown during the hold — including a modifier-only one such as the RF chord's Shift — cancels the pending hints, removes them if already shown, and keeps them hidden until Alt is released.
- Releasing Alt, window blur, the document becoming hidden and component destroy all clear the timer and the attribute.
- `altHints: false` still shows nothing. Which actions fire is unchanged; no CSS changed.

The delay is a frontend constant rather than a keyboard-profile key by coordinator decision recorded in the ticket.

## Tests

`__tests__/KeyboardHandler.test.ts` on the real component with fake timers: Alt alone past the delay shows and keyup hides; Alt+ArrowUp before the delay never shows and dispatches `adjust_af_level` once; a key pressed after the hints are up removes them for the rest of the hold; Alt, Shift, ArrowUp never shows; release before the delay; blur and visibility change; `altHints: false`. The former test of immediate appearance is replaced.

Ran on the build host through `focused.yml`: RED 35657909515 at the tests-only commit `97f7cc27` (6 failed, 59 passed), GREEN 35658052525 at head `92909c62` (65 passed). Locally `npm run check`: 0 errors.

## Not verified

Not measured on a live page yet, including whether showing the hints moves any element.


morozsm:
Agent Review: PASS 92909c62556e5516d3d3ec83a5d21d45706c5dfd

Independent review of the code at this head. Head confirmed equal to `gh pr view 3559 --json headRefOid`; net diff taken against `git merge-base origin/main HEAD` = `cf6e90e3` (3 files, +250/-10 — within the 6-file/600-line soft threshold; the only new import is `ALT_HINTS_DELAY_MS` from the sibling `keyboard-map.ts`, no layer crossing).

## Discrimination of the new pins

Checked from the two focused runs' own logs, not from the PR body. RED 35657909515 checked out revision `97f7cc27` (tests-only) — 6 failed / 59 passed; GREEN 35658052525 checked out revision `92909c62` (log line `HEAD is now at 92909c6`, `FOCUSED_VITEST_TARGETS` = the KeyboardHandler spec) — 65 passed. The six that died on unfixed production are exactly the six that depend on the new code (`AssertionError: expected 'true' to be undefined` in each): alone-past-the-delay, chord-before-the-delay, chord-after-the-hints, the Shift chord, blur, visibility. So the core behaviour is pinned by a real pre/post comparison, and the delay itself is pinned by the `ALT_HINTS_DELAY_MS - 1` step asserting absence.

Three of the nine were green on unfixed code, and I checked what each still buys: *releasing Alt before the delay* (passes on old code and on any delay value — regression pin only, discriminates nothing about this change); *unmount mid-hold* (first half old behaviour, second half genuinely pins `onDestroy -> clearAltHold` cancelling a pending timer, which cannot exist on old code); *`altHints: false`* (preserved-behaviour pin).

## Sequence walk of the state machine

- Alt alone, no repeat: arm at keydown, attribute at the delay, `clearAltHold` at keyup. Pinned.
- Alt auto-repeat while waiting: `altHintTimer !== null` short-circuits, no re-arm, no restart of the delay.
- Alt auto-repeat after the hints are up: `dataset.shortcutHints === 'true'` short-circuits, no flicker.
- Alt -> other key -> that key's keyup -> Alt still held: `handleKeyup` only acts on `key === 'Alt'`, and `altHoldBlocked` plus the absent timer keep it hidden for the rest of the hold. Correct.
- Left Alt + right Alt, one released: the keyup clears the whole hold while the other Alt is still physically down; a subsequent repeat keydown re-arms and shows hints 400 ms later. Old code re-showed them immediately, so this is not a regression.
- `AltGraph` (and the Control that Windows layouts deliver with it): both take the `key !== 'Alt'` path, so AltGr never reveals hints and suppresses an Alt hold in progress. Consistent with "any other key".
- Typing in an INPUT: the Alt arm sits after `shouldIgnoreEvent`, so Alt over a text field never starts the timer — same as the old code, which also set the attribute only past that guard. No hints over someone typing.
- Stuck `altHoldBlocked`: it is set only inside `suppressAltHold`, which is reached only when `altHoldActive()` is already true (timer, attribute, or an existing block), i.e. only inside a real hold; it is cleared by any Alt keyup, window blur, `visibilitychange` to hidden, and destroy. I could not construct a path that sets it outside a hold, so no "hints silently dead until the next Alt keyup".
- `enabled === false` mid-hold: an already-scheduled timer is not cancelled and would still set the attribute. Unreachable as shipped — `LcdLayout.svelte`, `RadioLayout.svelte` (both branches) and `MobileRadioLayout.svelte` (both branches) mount `KeyboardHandler` without an `enabled` prop; only tests pass it.
- Destroy/remount with the attribute set: `onDestroy -> clearAltHold` removes attribute and timer; pinned.
- Mouse click during a hold: not a keydown, hints stay. Matches the stated rule.

## Cleanup and listeners

`onblur` on `<svelte:window>` compiles to a non-capturing listener (`is_capture_event` in svelte's `utils.js` keys off a `capture` suffix, which `blur` does not have), so an element blur — which does not bubble — cannot reach it; only a real window blur clears the hold. That matters: a capturing binding would have wiped the hints on every focus move. `<svelte:document onvisibilitychange>` is the first `svelte:document` in `frontend/src`, and it both typechecks and functions (see below, plus the two visibility assertions in the GREEN run). Exactly one handler instance is mounted at a time — the two mounts in each of `RadioLayout.svelte` and `MobileRadioLayout.svelte` sit in mutually exclusive `{#if}/{:else}` branches — so no cross-instance race over the shared body attribute.

## Actions unchanged

`suppressAltHold` touches only the timer, the body attribute and the flag: no `return`, no `preventDefault`, no focus or event mutation, so placing it above the early returns cannot change which action fires or what gets defaulted. The evidence beyond reading: all 56 pre-existing tests in that file are green at this head (65 = 56 + 9), including the Tab guard, the digit-routing/`data-vfo-freq` swallow, `data-owns-arrows`, the leader sequences and the Alt+ArrowUp default chord; and the new chord tests assert `onAction` fired exactly once with the expected action while the hints stayed absent between steps, not only at the end.

## Consumers swept

`grep -rn` (no `head`) over `frontend/src`, `frontend/tests` and `docs` for `shortcutHints` / `shortcut-hints` / `data-shortcut-hint`: 70 hits, all of them either this component, the `data-shortcut-hint` producers (panels, controls, `lib/Button/types.ts`) or two planning documents — nothing reads the body attribute. `frontend/tests/e2e` has no spec that presses Alt at all (the only `alt` matches are `radioHealth`, `globalTeardown`, `healthRevision`). So no browser spec can be reddened by the delay; `visual` is in fact green at this head.

## Reproduced claims

`npm run check` in `frontend/`: `COMPLETED 4867 FILES 0 ERRORS 0 WARNINGS` — reproduced locally once. "No CSS changed" is true of the diff. No test suite was run on this machine.

## Non-blocking notes

1. **Unpinned guard branch.** Removing `altHoldBlocked = true` from `KeyboardHandler.svelte: suppressAltHold` would leave all 65 tests green: no test delivers a second Alt keydown between a suppression and the Alt keyup (the one in the third test comes after `keyup('Alt')`, which by then has cleared the flag). The flag is nevertheless load-bearing where a held modifier auto-repeats its keydown — without it, a repeat Alt keydown after suppression re-arms the timer and the hints reappear mid-chord. Worth one test: Alt down, ArrowUp, then an Alt keydown with `repeat: true`, asserting the attribute stays absent past the delay. Analysis of the nine tests, not a run — the mutation avenue was unavailable to this session (no push, no local test execution).
2. **"Only key down" is "no other keydown seen".** The Alt branch does not inspect `event.shiftKey/ctrlKey/metaKey`, so a modifier already held when Alt goes down does not prevent arming: Shift held, then Option, then a pause longer than the delay reveals the hints, and the Arrow then removes them. Same shape as Option-alone-then-pause, which the chosen 400 ms threshold deliberately accepts, so this is a design edge rather than a defect; checking the modifier flags on the Alt keydown would close it cheaply.
3. A live-page check should watch a control with `data-shortcut-hint` while doing Option+Arrow and Option+Shift+Arrow adjustments at normal speed (nothing should appear or shift), and separately hold Option alone for a second (hints appear once, no jitter, and the PR records that no live measurement of layout movement was taken).

CI as found at this head: `visual` pass, `quick` pending, `quick-v2-observe` pending behind it, `changes`/`classify`/`grep-gate`/publishers pass, `Agent Review Gate` red only for the absent exact-head directive this comment supplies. Nothing red is caused by this PR.

PR #3560, opened 2026-09-21
fix(MOR-2513): the page accepts a state payload with null leaves; a rejected envelope is logged

morozsm:
MOR-2513 follow-up — fixes a regression that #3558 put on `main`. One unit of work (8 files, above the 6-file soft threshold): the ingestion fix, the live-payload fixture it is proven against, and the fixture's registration in the provenance guard belong together — the fixture is the evidence.

## What was broken

After #3558 the server publishes `null` for every never-observed state leaf. On the live bench (FTX-1) the default desktop page then rendered no radio surfaces at all: only the status bar, BANDS and MEMORY. No exception, empty console.

Cause: `frontend/src/lib/stores/radio.svelte.ts: validReceiver` (and `isValidServerState`, `matchesCurrentCapabilityTopology`) still required non-null primitives for leaves the regenerated `state.ts` types as nullable — in the live payload `main.dataMode`, `sub.dataMode`, `sub.att`, `sub.preamp`, `scopeControls.receiver` among others. The whole state envelope was rejected; `frontend/src/lib/transport/ws-client.ts: applyDeltaEnvelope` dropped it with `return null` and no log, before capabilities were refreshed, so both stores stayed empty, the view model was never built and no surface mounted.

## What changes

- `radio.svelte.ts`: the ingestion validators accept `null` for every leaf `state.ts` types as nullable.
- `ws-client.ts`: a rejected full or delta state envelope logs one `console.error`; the rejection itself stays.
- New fixture `fixtures/ftx1-state-unobserved-leaves.json`: the `/api/v1/state` body captured from the bench FTX-1 at `86187ed3`, values as captured (scanned for addresses, host names, device paths, serial numbers, call signs: none found), registered with the fixture-provenance guard through an `ftx1` conformance profile.
- Tests: the live payload is accepted by the store, the view model validates, the default desktop composition mounts with `[data-vfo-row]` for MAIN and SUB and no rendered `null` / `NaN` / `undefined`; and the same with every nullable leaf set to `null`.

## Evidence

- `focused.yml` on the build host: RED 35662269867 at the tests-only commit `aed61ad2`; GREEN 35662629725 at head `603b6a14` — 252 passed.
- Live bench, by the coordinator, before opening this PR: bench checkout at `603b6a14`, frontend rebuilt, server restarted; `/api/v1/state` carries `powerOn: null` and `main.filter: null`; the page shows 6 `[data-vfo-row]` elements and its text contains no `null`, `NaN` or `undefined`. At `86187ed3` the same check gave 0 rows.


morozsm:
Agent Review: PASS 603b6a14a358a6eb8187af53fb47376788e341aa

Independent verification at head `603b6a14`. I did not write this change; I tried to refute it. No blocking defect found. No `REQUIRED BEFORE MERGE` and no `MANDATORY SQUASH-BODY CORRECTION`: every claim I checked in the PR body and both commit messages is true of the net diff.

## 1. Over-loosening — mechanically enumerated, not sampled

Method: parsed the generated block of `frontend/src/lib/types/state.ts` (everything above `// END GENERATED`) with a throwaway script, extracting each interface's fields with their optionality and whether the type union contains `null`; diffed that set against the leaves the validators in `frontend/src/lib/stores/radio.svelte.ts` actually check.

- Schema-nullable **required** leaves of `ServerStatePublic`: exactly `active, ptt, split, dualWatch, tunerStatus`. `isValidServerState` (radio.svelte.ts:168) relaxes exactly those five. Exact match.
- Schema-nullable **required** leaves of `ReceiverStatePublic`: exactly `freqHz, mode, filter, dataMode, att, preamp, nb, nr, afLevel, rfGain, squelch, sMeter`. `validReceiver` (radio.svelte.ts:132) checks exactly those twelve and relaxes all twelve. Exact match.
- `ScopeControlsPublic.receiver` is `number | null`; `matchesCurrentCapabilityTopology` (radio.svelte.ts:204) relaxes exactly that one.

**Over-loosened list: empty. Under-fixed list: empty.** The type of each relaxation matches the schema type (`stringOrNull` only on `mode`; `boolOrNull` only on `nb/nr/ptt/split/dualWatch`; `numberOrNull` on the numerics). `undefined`/missing still fails every one of them (`numberOrNull(undefined) === false`), a wrong type still fails, and a delta that `removed` a required key still fails.

One behaviour change that is not a relaxation: `filter` previously passed `typeof === 'number'` with no finiteness check, so `NaN` was accepted; `numberOrNull` rejects it. That is a tightening, unreachable over JSON, and harmless. `scopeControls.receiver` keeps its safe-integer and `0..hasSub` range rejection for every non-null value.

## 2. Null `ptt` / `split` — every acting site is fail-closed

- `frontend/src/lib/runtime/adapters/tx-adapter.ts: canRestoreModAfterConfirmedOff` requires `observation.ptt === false`; `null === false` is `false`, so a null PTT **blocks** the MOD restore rather than enabling it. Conservative.
- `frontend/src/lib/runtime/commands/panel-commands.ts: onSplitToggle` and the `toggle_split` / `toggle_monitor` / `switch_active_vfo` keyboard cases all require `typeof … === 'boolean'` (or an observed-field gate) before dispatching. A null leaf disables the command; it never inverts it.
- `frontend/src/lib/audio/audio-manager.ts` gates TX frames on the local `_txEnabled`, explicitly not on the store's `ptt`.
- Nothing reads `ptt` as a permission to act. `radio.svelte.ts: getIsTransmitting` coerces `ptt ?? false`, but `git grep` finds no production caller — only `stores/__tests__/radio.test.ts`. Same for `getFrequency`'s `?? 0`.

**No safety misread found. This is not a blocking finding.**

Finding A (low, display-only, pre-existing coercion newly reachable): `frontend/src/lib/runtime/props/panel-props.ts: toTxProps` / `toMetersProps` set `txActive: state?.ptt ?? false`, and `toReceiverProps` uses `if (state.split)` / `(state.tunerStatus ?? 0) > 0` for the SPLIT/ATU/TUNE badges. With a null leaf these read as a definite "off" rather than unknown, which brushes the operator-never-sees-a-fabricated-reading rule. `toTxProps` already carries the separate `txActiveAvailable` / `atuAvailable` flags, and the live desktop path builds its facts through `radio-view-model-adapter.ts` (`boolFact`, `txAuxField`), which degrade to an explicit unknown — the badge map in `toReceiverProps` is the one place with no availability companion. Follow-up ticket, not this PR.

## 3. Pin discrimination — reproduced from the builder's RED run, not taken on trust

Run `35662269867` checked out `aed61ad2` (tests only, production untouched = the exact bug still in place) and reported `Tests 11 failed | 85 passed (96)`. The 11 are all six store pins in `mor2513-null-live-payload.test.ts`, all three mount pins in `RadioLayout.isolated.test.ts`, and two adapter tests. Run `35662629725` checked out `603b6a14` — I confirmed the checked-out SHA in its own log, not from the run's `headSha` field — and reported `7 files, 252 passed`. The pins therefore redden on the unfixed code and green on the fixed code.

I checked the builder's claim that two of the eleven were defects of his own test, and specifically whether the second commit *loosened* them:

- `currentMode` → `currentMode.reading`: a **tightening**. `semantic/radio-view-model.ts` declares `currentMode: ModeFilterField<string>`, and `ModeFilterField<T> = TxAuxField<T>` (line 316), so the original `toEqual({status:'known',value:'USB'})` compared a wrapper to a reading and could never pass. The corrected form asserts a literal reading.
- the meters test: the expectation `{ status: 'known', value: 13.8 }` is **unchanged**; only the caps input changed. The RED log shows it failed with `expected { status: 'unknown' } to deeply equal { status: 'known', value: 13.8 }` — the capabilities capture named a different `providerGeneration` than the state capture, `validIdentity` failed, and the meters degraded. Aligning the pair mirrors production (the server publishes both from one generation, and `matchesCurrentCapabilityTopology` requires the match), and it is applied to the caps object only — the state capture is untouched. It does not mask an identity failure the live page could hit; removing the alignment turns the assertion red again, so the row still discriminates.

The five adapter tests that stayed green at `aed61ad2` do so because `toRadioViewModel` is downstream of ingestion and the mutation cannot reach it — expected, not hollow.

## 4. Fixture

- Identifier scan: I enumerated every distinct **string value** in `ftx1-state-unobserved-leaves.json` (not just keys) plus a regex pass for IPv4/IPv6, MAC, `/Users/`, `/dev/`, `usbserial`, host names and call-sign shapes. The only strings are capability paths, native command ids, enum words, `"serial"` as a transport kind, and one ISO timestamp. Every `sessionId` is `null`. Clean.
- Provenance registration: `conformance/fixture-provenance.guard.test.ts` iterates `Object.entries(PROFILES)`, so adding `ftx1` to `conformance/profiles.ts` is the registration that guard requires. It went from 7 to 13 cases and passes at head. I independently reproduced its two substantive checks against the fixture: no top-level or receiver key the current schema no longer defines, and no required `ServerStatePublic` field missing. No pre-existing conformance expectation was weakened or skipped — the eight `mor14xx/mor15xx` conformance files all pin `PROFILES.ic7300` explicitly, so adding `ftx1` does not silently re-run them against a second radio.
- `ftx1-profile.ts` is not a second copy: the file already existed for MOR-1679 with `FTX1_CAPABILITIES`; this change adds two exports to it.
- I measured the "108 null leaves" claim: counting nulls in the capture outside the `fieldStatus` subtree gives exactly 108.
- I also replicated the `FTX1_STATE_FULLY_UNOBSERVED` derivation and checked the result against the parsed schema: every schema-nullable field present in the object is `null` (only the `sub` subtree itself stays an object, correctly — it is a container, not a leaf). The docstring claim holds as measured.

Finding B (low): that derivation enumerates leaves from **the capture's own `fieldStatus` map**, not from the generated types — so it is data-driven but not schema-driven, and the commit message's "per the generated state.ts contract" describes the outcome I measured rather than the mechanism. Nothing fails today if `state_schema.py` gains a nullable leaf that a validator rejects. The risk is genuinely small, because the validators are a whitelist: an unchecked new leaf is accepted by default, and a new *required* leaf missing from the fixture already reddens the provenance guard. Smallest self-updating alternative, for a follow-up: `fixture-provenance.guard.test.ts` already parses `state.ts` through the TS compiler API and already loops `PROFILES`; adding one case there that, for each required leaf whose parsed type union contains `null`, sets it to `null` on the profile state and asserts `isValidServerState` still accepts, is roughly fifteen lines and cannot rot.

Finding C (low): the sidecar cross-check at the bottom of that same guard is hand-scoped to `ic7300-provenance.json` by name, so a newly registered profile gets no provenance-sidecar check. `ftx1`'s provenance lives in the loader header prose instead. Same hand-maintained-list shape as Finding B; fold into the same follow-up.

## 5. `ws-client.ts` logging and ordering

- The two `console.error` calls (ws-client.ts:950, ws-client.ts:996) log a **fixed string**. No payload, no field values, no user data. No leak.
- No rate limit and no once-per-reason guard. A persistently rejected delta stream can emit one line per envelope, so of the order of ten per second — console only, no network, nothing retained. Finding D (low): the message also does not name the failing leaf, so it says a contract violation happened but not which one. A follow-up that logs the first failing path once per provider generation would fix both. Not a blocker, and not something this PR must carry.
- The ordering claim in the new comment is **true**, and I verified it rather than reading it: `git grep` finds `setCapabilities` called from exactly one production site, inside `refreshCapabilities` (ws-client.ts:826), and `refreshCapabilities` is reached only from ws-client.ts:977 — after a full envelope is accepted — and from its own retry timer, whose `isCurrent()` guard requires `_hasReceivedFullState && _fullState !== null`. So a rejected full genuinely leaves both stores empty. That ordering is **not** changed here, and I agree it should not be: letting capabilities load off a contract-invalid envelope is a fail-open direction that needs its own design. Finding E (low, follow-up): a connected-but-no-valid-state condition still renders as a blank page rather than an explicit fault.

## 6. Tests, layering, guardrails

- `mor2513-null-live-payload.test.ts` drives the real `radio.svelte`/`capabilities.svelte` modules through `vi.resetModules()` + dynamic import, with no mock of the code under test, and asserts through the real `setRadioState`. Assertions are literal.
- The `RadioLayout.isolated.test.ts` addition mounts the real `desktop-v2` composition and asserts over `t.textContent` of the **whole mounted tree**, not a fragment — so "no null/NaN/undefined" is a whole-page claim. It also pins `[data-receiver-instrument]` counts for MAIN and SUB and six `[data-vfo-row]`, which matches what the coordinator observed on the bench.
- Tests sit beside what they test; the stores test reaches across to the adapters' fixtures directory, which is test-only and fine. `radio.svelte.ts` imports no component; `ws-client.ts` → stores is the pre-existing direction.
- Guardrails: `gh pr view` reports 8 files, +634/-34 = 668 changed lines. Above the 6-file/600-line soft threshold, below the 10-file/1000-line hard ceiling, and the PR body's first line states why this is one unit of work. Compliant.
- Prose: I read every comment the diff adds as its own pass. The three new comments in `radio.svelte.ts` and `ws-client.ts` each state something I could check against the code, and each checked out.

## 7. CI as found at `603b6a14`

`quick` **pass** (5m1s), `visual` **pass**, `classify` / `changes` / `grep-gate` / the metadata publishers all **pass**. `quick-v2-observe` is **pending**, waiting on the canonical legacy quick observer. `Agent Review Gate` and `Agent Review Gate v2 observe` are **fail** for the only reason that this directive did not exist yet. Nothing red is caused by this PR. Focused evidence at the exact head: run `35662629725`, 7 files / 252 tests passed.

Working tree left byte-identical; `git status --short` is empty.

PR #3561, opened 2026-09-21
feat(MOR-2524): MAIN/SUB gains and RF GAIN/SQUELCH use the fader look of AF LEVEL

morozsm:
Guardrail exception (file count only, granted by the coordinating session under the owner's delegation of 2026-09-03): 12 code + 0 regenerated baselines = 12 files; 237 + 37 = 274 changed lines.

MOR-2524, slice 1. One unit of work: two production hosts and the surface that seats one of them change, two snippet types follow, and the other seven files are tests and a test fixture that pinned the replaced markup or the changed snippet signatures. No visual baseline is re-pinned: at head `3c5a118a` the `visual` check and the `quick` check (which compares the four production-root scenes) both pass, and a full Linux regeneration (run 35664037312) differs from the approved set only in the six captures that differ on every regeneration (`panadapter-first`, `unified-instrument`, `peer-split-chassis`, each at two sizes).

Owner, 2026-09-21: "We have sliders of three different kinds. That is no good. I like the first variant — it should look like an expensive fader." The fader is the `hardware-illuminated` variant of the shared `ValueControl` hbar that AF LEVEL already uses.

## What changes

- `semantic/RxAudioInstrumentHost.svelte`: the MAIN / SUB channel-gain pair was a raw `<input type="range">` reading `value ?? 0`; it is now `ValueControl` hbar bound through `createContinuousScalar` like AF LEVEL (domain −60…+12 dB, step 1, same `onChannelGainChange(channel, value)` request, same accessible names). The look follows the same `hardware` flag AF LEVEL uses (`hardware-illuminated` on the Standard face, `modern` elsewhere). An unknown gain is an unlit, disabled control with no value text; the value `<output>` keeps a reserved width and tabular digits and is now visible on the Standard face (it was screen-reader-only).
- `semantic/RfFrontEndInstrumentHost.svelte` and `semantic/RfFrontEndSurface.svelte`: the RF GAIN / SQUELCH level handles (separate and combined form) take a `hardware` flag; the surface passes it from the same signal RX AUDIO uses — a Standard finite layout is present → fader, otherwise (`sdr-test`, deck, fallback) → `modern`, next to a `modern` AF LEVEL. Before this PR the combined RF+SQL form was already a fader beside a `modern` AF LEVEL on `sdr-test`; that is fixed by the same flag.
- `components-v2/layout/RadioLayout.svelte`: the two rules that hid the gain outputs are removed.

## Behaviour of the gain pair, before → after

Arrow keys: native → one step (Shift: same step, `fineStepDivisor` 1). Home / End: kept. **PageUp / PageDown: lost** (ValueControl has none). Wheel: none → armed wheel. Commit: every `input` event → the shared continuous-scalar binding AF LEVEL uses. `data-owns-arrows`: none → set while operable. Focus: no new focus frame added.

## Evidence

`focused.yml` on the build host: RED 35661527869 at the tests-only commit `22ddf171` (8 failed — the new pins); GREEN 35666061790 at head `f427cefb` (422 passed, 6 files). `npm run check` 0 errors. After merging `main` (with #3560) into the branch, the coordinator loaded head `3c5a118a` on the live bench: the page mounts (6 `[data-vfo-row]`), the two gain rows draw as unlit faders with no text (the bench FTX-1 does not report channel gains), 7 `role="slider"` controls, 3 raw range inputs remain (RIT/XIT, filter — later slices).

## Not in this PR

Raw range inputs in `RitXitScanSurface`, `FilterSurface`, `CwKeyerSurface`, `DspSurface`; the `modern` line sliders in `DspScalarHost`, `CwKeyerInstrumentHost`, `TxAuxScalarHost`; mobile panels; design-language tokens for the fader. The approved captures therefore do not picture these controls; no pixel measurement of them exists yet.

Linear: MOR-2524 (remains open).


morozsm:
Agent Review: BLOCKED 3c5a118ac7e4819900373f33609216bb0183bed6

Independent review at the exact head (`git rev-parse HEAD` = `gh pr view 3561 --json headRefOid` = `3c5a118a`). Net diff `git diff origin/main...HEAD`: 9 files, +181/-25. Tree clean (`git status --short` empty). No tests run locally.

## BLOCKING — the change creates the ticket's own defect on a second live face

**`frontend/src/semantic/RfFrontEndInstrumentHost.svelte:509`** sets `variant="hardware-illuminated"` unconditionally on the separate-form RF GAIN / SQUELCH sliders. The PR body defends this with "this host has no face flag and is mounted only by the desktop control faces". Read at this head, that justification does not hold:

- `frontend/src/components-v2/layout/RadioLayout.svelte:760-762` — the `desktop-control-face` branch is entered for **both** `desktop-v2` and `sdr-test`.
- On `sdr-test` the same branch renders `{@render instruments.rfFrontEnd()}` (`RadioLayout.svelte:781`) and `{@render instruments.rxAudio()}` (`RadioLayout.svelte:810`) with **no** `finiteLayout` — `hostedRfFrontEnd` / `hostedRxAudio` take `finiteLayout` as their *second* positional parameter (`components-v2/wiring/SemanticRadioSurfaces.svelte:2668-2679`), so both are `undefined` here. Only the Standard branch passes `rxAudioFiniteLayout` (`RadioLayout.svelte:704-706`).
- Without `finiteLayout`, `semantic/RxAudioSurface.svelte:87-97` renders AF LEVEL through `handles.afLevel()`, which is `{#snippet afLevel()}{@render afLevelControl(false)}{/snippet}` (`semantic/RxAudioInstrumentHost.svelte:357`) → `variant="modern"`.
- `semantic/RfFrontEndSurface.svelte:55-56` renders `rfGain()` / `squelch()` on that same path regardless of `finiteLayout`, i.e. now hardware-illuminated.
- `ValueControl.svelte:114` defaults `variant = 'modern'`, so **before** this PR both AF LEVEL and RF GAIN/SQUELCH were `modern` on `sdr-test`: consistent. **After** it they differ.
- `sdr-test` is operator-reachable, not a query-param QA face: `skins/registry.ts:226` resolves it from a normalised `CanonicalLayoutMode` layout preference (contrast `flagship-probe`/`dual-receiver-cockpit`, gated at `registry.ts:220-224` before normalisation). The generic deck at `RadioLayout.svelte:838` → `semanticDeckContent` (`:740-741`) has the same shape.

Risk: MOR-2524 exists to remove "sliders of three different kinds on one face". This PR removes one mismatch on `desktop-v2` and creates a new one on `sdr-test` — AF LEVEL thin-line next to an expensive fader — which is the defect, not a leftover from a later slice.

This is a half-swept class inside the PR itself: commit `4f64d08b` ("one hardware-flag decision for the gain variant") established that the variant must follow a per-face flag and threaded `hardware` through `mainGain`/`subGain` (`RxAudioInstrumentHost.svelte:479-480`, `rx-audio-instruments.ts:86-89`), then left the sibling host hardcoded.

**Required fix.** Give `rfGain`/`squelch` the same treatment as `mainGain`/`subGain`: a `hardware` parameter on the handles (`Snippet<[hardware?: boolean]>` in `semantic/rf-front-end-instruments.ts`), the `hardware ? 'hardware-illuminated' : 'modern'` ternary at `RfFrontEndInstrumentHost.svelte:509`, and the hardware call made only where AF LEVEL is also hardware — i.e. through `RfFrontEndSurface`'s `finiteLayout`-bearing path — or, equivalently, drive AF LEVEL and RF GAIN/SQUELCH from one face decision so no face can show both looks. Either direction is acceptable; what is not is a face where the two disagree. Pin it: extend `semantic/__tests__/RfFrontEndSurface.test.ts` with a non-hardware case asserting `.vc-hbar` has no `hw-illum` and has `.vc-thumb`, mirroring the pair already added in `semantic/__tests__/RxAudioInstrumentHost.isolated.test.ts:284-318`.

**Focused targets for the fix head** (`gh workflow run focused.yml --ref main -f revision=<new sha> -f vitest_targets=...`):
`frontend/src/semantic/__tests__/RfFrontEndSurface.test.ts`, `frontend/src/semantic/__tests__/RfFrontEndInstrumentHost.isolated.test.ts`, `frontend/src/semantic/__tests__/RxAudioInstrumentHost.isolated.test.ts`, `frontend/src/components-v2/wiring/__tests__/semantic-rx-audio-wiring.component.test.ts`, `frontend/src/components-v2/wiring/__tests__/semantic-rf-front-end-wiring.component.test.ts`, `frontend/src/components-v2/layout/__tests__/semantic-desktop-migration.component.test.ts`.

## Prose in the diff — correct with the same commit

`frontend/src/components-v2/wiring/__tests__/semantic-rx-audio-wiring.component.test.ts:498-500`: "the dB readout reserves the width of its longest value ('-60 dB')". `min-width: 6ch` reserves six *zero-glyph* widths; `-60 dB` is six characters of which `-`, space, `d` and `B` are not zero-width, and `min-width` is a floor, not a clamp — the box still grows if the content is wider, and `.rx-audio-level :global(.vc-hbar) { flex: 1 1 auto }` (`RxAudioInstrumentHost.svelte:573`) means the fader absorbs the difference. Nothing in the change measures this, and no test fails if it is wrong. Per CLAUDE.md § Testing, delete the width claim or state only what the test checks (two CSS declarations are present).

## What I checked and could not refute

- **Command path unchanged.** `request: (next) => onChannelGainChange?.(channel, next)` (`RxAudioInstrumentHost.svelte:314`) → `SemanticRadioSurfaces.svelte:1885` → `lib/runtime/commands/panel-commands.ts: makeAudioRoutingHandlers.onChannelGainChange` → `audioManager.setAudioConfig({ main_gain_db | sub_gain_db })`, integer dB, unchanged by the `origin/main` merge (`git diff 4f64d08b..HEAD -- frontend/src/lib/runtime/commands/panel-commands.ts` touches only PBT/VFO null-handling). The wiring pin exercises a real `ArrowLeft`/`ArrowRight` on `[role="slider"]` and asserts the literal payloads `{main_gain_db:-7}` / `{sub_gain_db:3}` from readings `-6`/`2` — `handleKeyboardStep` with `step 1`, `fineStepDivisor 1` gives exactly that.
- **Authority.** The gains are a browser-local mixer setting, not a radio command; the reading is `runtime.audioRouting` echoed by `audioManager`. `authorityOf` for reading evidence keys on `[…domain…, enabled, ownerKey, reading.status]` (`primitives/scalar/continuous-scalar.svelte.ts:430-447`), and a stale draft is cleared by `reconcile` as soon as the reading moves. Not cancelling these two on `subscribeControlAuthority` is defensible; an in-flight optimistic gain survives an authority flip because radio authority does not own it.
- **Never-confirming source:** the optimistic draft sticks (no snap-back, no drift) — `viewOf`'s `displayed` keeps `draft` until `canonicalOf(input)` differs from `draftCanonical`. Same as AF LEVEL.
- **`enabled` parity:** `known && operational === true && handler !== undefined` matches the removed `disabled={value === undefined || !operational || handler === undefined}`, plus a `Number.isFinite` guard. No regression.
- **`defaultValue: null`:** the host overrides the base policy's `reset: (domain) => domain.defaultValue ?? domain.min` with `reset: (domain) => domain.defaultValue` (`RxAudioInstrumentHost.svelte:296-299`), shared by all three bindings, so double-click writes nothing.
- **Unknown reading:** `aria-valuenow` absent, `aria-disabled="true"`, `.disabled` (opacity .4, `pointer-events: none`), `<output>` empty, no digits. The `hil-thumb` is still drawn, at `left: var(--vc-fill-percent)` = `0%` (`HBarRenderer.svelte:613-615`, `:92-97`) — the far-left end. Identical to AF LEVEL's unknown state and strictly better than the old `value ?? 0`, which put a real `0` in the DOM; noting it, not blocking on it.
- **Fill** is unipolar from `min` (`primitives/scalar/value-control-core.ts: valueToPosition`), so `0 dB` sits at 83 %. Consistent with a physical fader; not a bipolar-fill bug.
- **Handle consumers** (`grep -rn "mainGain\|subGain" frontend/src frontend/tests`, no `head`): the only production call sites are `RadioLayout.svelte:597-598` in `rxAudioFiniteLayout` — the Standard seats, where AF LEVEL is `afLevelRow` → `afLevelControl(true)`. The default `hardware = true` therefore matches AF LEVEL everywhere it is reachable today; the grouped `RxAudioSurface` never renders these handles. No two-look defect from *that* default.
- **Browser specs:** `frontend/tests/e2e` has no `input[type=range]` or slider-count assertion touching this face; its only `role=slider` use is the combined RF/SQL control (`v2-ui-interactive.impl.ts:697`), unchanged. `[aria-label="MAIN gain in decibels"]` in `semantic-rx-routing-restoration.component.test.ts:148` queries a standalone `AudioRoutingControl` mount, which keeps its raw range inputs. Slider/range counts in `FilterPanel`, `TxAuxFiniteHost`, `RfFrontEndInstrumentHost.isolated`, `AudioRoutingControl.isolated` are all scoped elsewhere.
- **Test discrimination:** run `35661527869` (tests-only `22ddf171`) reddened 8 of the new/changed pins, including the payload pin, the fader pin, the unknown pin, the CSS pin, both RF variant pins and the inverted source pin. The two `channel-gain appearance (MOR-2524)` cases arrived with `4f64d08b` and so have no red evidence; read against the code they do discriminate (removing or inverting the ternary at `RxAudioInstrumentHost.svelte:471` fails the `hw-illum` assertion in each direction). The `.not.toContain` inversion in `semantic-desktop-migration.component.test.ts:2593-2594` is the inverse of the pre-existing source-text idiom in that file and did go red — not a weakening.
- **CSS:** removing `.rx-audio-gain input` and the two sr-only rules leaves no orphaned selector that now matches the new markup; `input[type='range']` rules in `skins/desktop-v2/semantic-controls.css` no longer match this row, and `label > output` (`:112`) places the readout exactly as it already places AF LEVEL's.
- **Guardrails/layering:** 9 files / 206 changed lines — over the 6-file soft threshold, under the hard ceiling; the body gives its one-unit rationale. No transport import enters either semantic host.

## Notes (not blocking, no commit required)

1. `RxAudioInstrumentHost.svelte:378-386`: AF LEVEL's own `<output>` still renders `UNKNOWN_TEXT` (`—`) when unread, and it is *not* sr-only on the Standard face. The gain rows next to it now render nothing. After this PR the two rows of the same block disagree, and the placeholder that survives is the one the operator-facing rule forbids. Worth a follow-up in this ticket's series.
2. `__tests__/control-feedback-debt-inventory.test.ts:133-144` keeps an exemption for `src/semantic/RxAudioInstrumentHost.svelte`'s raw browser-local gain input, which this PR deletes. The inventory's shrink-only assertion still passes; the exemption is now unreachable and should go with a later slice.
3. Lost behaviour beyond the body's declared PageUp/PageDown: the row is a `<label>` whose only labelable descendant is now the `<output>`, so clicking the "MAIN"/"SUB" text no longer reaches the control. AF LEVEL already has this shape, so it is a consistency consequence rather than a new pattern.

## Corrections outside the code

REQUIRED BEFORE MERGE (PR body, editable in place, head SHA unaffected):
- "this host has no face flag and is mounted only by the desktop control faces" — true as stated but used as a safety argument it does not support; `sdr-test` is a desktop control face and renders AF LEVEL `modern`. Replace with the outcome of the blocking fix.
- "Visual baselines are re-pinned in a follow-up commit on this branch from this PR's own Linux runs; the file-count line is added at the top of this body then." `visual` is green at this head and no baseline changed, so no re-pin commit and no file-count exception line are owed. Delete or restate.

## CI as found (`gh pr checks 3561`)

`visual` pass (1m33s), `classify` pass, `changes` pass, `grep-gate` pass, `publish-docs-only-quick-status` pass, metadata/observer jobs pass; **`quick` pending**, `quick-v2-observe` pending (waiting on quick); `Agent Review Gate` / `Agent Review Gate v2 observe` fail — "Missing trusted standalone exact-head Agent Review PASS directive", expected before this comment. No red check is attributable to this PR at the time of writing; `quick` had not finished and was not waited on.

morozsm:
Agent Review: PASS f427cefb1ea67ede32bff4d8e20487ceacafd5b9

Delta review of `git diff 3c5a118a..HEAD` (commits `5640ccaa`, `f427cefb`; 7 files, +68/−24). Head confirmed: `git rev-parse HEAD` = `gh pr view 3561 --json headRefOid` = `f427cefb`. Tree clean (`git status --short` empty). Nothing run locally.

## The blocking finding is resolved, and its class was swept

`RfFrontEndInstrumentHost.svelte:523-524` now take `hardware` and pass it into `scalar(field, hardware)` (`:487`), whose `ValueControl` carries `variant={hardware ? 'hardware-illuminated' : 'modern'}` (`:510`). `RfFrontEndSurface.svelte:56-59` supplies `finiteLayout !== undefined` at all three call sites.

Re-enumerated every call site of the RF level handles (`grep -rn "rfSql(\|rfGain(\|squelch(" frontend/src frontend/tests`, no `head`):

| Call site | flag | AF LEVEL on that face |
|---|---|---|
| `RfFrontEndSurface.svelte:56` (`rfSql`) | `finiteLayout !== undefined` | — |
| `RfFrontEndSurface.svelte:58-59` (`rfGain`, `squelch`) | `finiteLayout !== undefined` | — |
| `__tests__/fixtures/RfFrontEndInstrumentHostFixture.svelte:130,132,133` (non-surface path) | default `false` | test fixture |
| `components-v2/layout/__tests__/fixtures/HostedRadioLayoutFixture.svelte:112,114,115` | default `false` | test fixture |

Production has exactly one consumer, and the two surfaces cannot disagree on any face reachable today:

- **Standard** — `RadioLayout.svelte:665` passes `rfFrontEndFiniteLayout` and `:704-706` passes `rxAudioFiniteLayout`, both inside `standardServicePanels`, each panel independently gated on `owner.order.includes(...)`; an absent panel renders nothing, so no half-state. Both faders.
- **`sdr-test`** — `RadioLayout.svelte:781`, `:810`: neither surface gets a layout. Both `modern`.
- **Generic deck / fallback** — `:838` → `semanticDeckContent` (`:740-741`): the `allowBare` argument is positional slot 1, `finiteLayout` slot 2 (`SemanticRadioSurfaces.svelte:2668-2679`), so both undefined. Both `modern`.
- **desktop-v2 without drag owners** — falls through to the same bare `{:else}` renders. Both `modern`.

Pinned in both directions in the same file, over the same code path, so removing or inverting the ternary necessarily reddens one of the pair: `__tests__/RfFrontEndSurface.test.ts:363-378` (seated → `hw-illum` + `hil-thumb`; grouped → no `hw-illum`, `.vc-thumb` present) and `:586-594` for the combined form. The source pin at `:582` was updated with the expression, not weakened.

## Item 2 — the combined RF/SQL form

In scope: it is the same class my BLOCKED named, and `DualParamRenderer.svelte:40` defaults `variant = 'hardware-illuminated'` with no caller previously passing one, so that control *was* a fader beside a `modern` AF LEVEL on `sdr-test` before this PR. The body says so and it checks out.

`modern` is a real look there, not a missing one. `DualParamRenderer.svelte:246-259` renders `.vc-track`, `.vc-track-base`, `.vc-fill-rf`, `.vc-fill-sql` and `.vc-thumb` — **both lanes**, not a single-lane degradation — and every one of those selectors has rules in the component's own style block (`.vc-track`, `.vc-track-base`, `.vc-fill-rf`, `.vc-fill-sql`, `.vc-thumb`, plus the shared `.vc-dual`, `.vc-track-container`, `.vc-axis`). No unstyled control.

Also refuted a concern of my own: `RfFrontEndInstrumentHost.svelte:511` now passes `accentColor` explicitly, and the non-hardware value `var(--v2-accent-cyan)` is byte-identical to `ValueControl.svelte:107`'s default, so the grouped path's colour is unchanged from before the PR.

## Item 3 — the new prose

The overclaiming width comment is gone; `min-width` is `7ch` and the rewritten comment ("jsdom computes no layout; this checks the two declarations are written, not that they guarantee no shift") says exactly what the assertion does and no more. The renamed case matches. No stale `6ch` remains in the changed area (`grep -rn 6ch frontend/src` finds only `FilterSurface.svelte:512`, untouched). The other added sentences — `rf-front-end-instruments.ts:71-73`, `RfFrontEndSurface.svelte:52-54`, the fixture doc and the test comments — are true as of this head against the call-site table above.

## Findings (all notes; none needs a commit to be safe today)

1. **Asymmetric defaults, medium.** `mainGain`/`subGain` default `hardware = true` (`RxAudioInstrumentHost.svelte:479-480`); `rfSql`/`rfGain`/`squelch` default `false` (`RfFrontEndInstrumentHost.svelte:456,523,524`). Two sibling handle families with opposite silent defaults is a trap in both directions for the next caller. Only the `true` default is exercised — `RadioLayout.svelte:597-598` calls `mainGain()`/`subGain()` bare and the wiring pin asserts `hw-illum`. The `false` default is unreachable in production (`RfFrontEndSurface` always passes explicitly; `render()` in `RfFrontEndSurface.test.ts:110-121` always sets `renderSurface: true`, so even the "grouped" cases exercise the explicit `false`, not the default). Cleanest follow-up: drop both defaults, making `hardware` a required snippet parameter so no call site can inherit the wrong look silently.
2. **Nothing pins cross-surface agreement, low.** The two `finiteLayout` props are distinct props on distinct surfaces; they coincide only because `RadioLayout` supplies both from `standardServicePanels`. No test would fail if a future face passed one and not the other. A single pin on the Standard composition asserting AF LEVEL and RF GAIN carry the same frame class would close it.
3. **First production reachability, low.** `DualParamRenderer`'s `modern` branch has never rendered on a real face before this PR. It is complete and styled per the reading above, and `visual`/`quick` are green, but no browser has looked at `sdr-test` at this head.

## Corrections outside the code

REQUIRED BEFORE MERGE (PR body, editable in place, head unaffected):
- "Commit: every `input` event → ValueControl's optimistic, debounced commit." The host builds its policy with `debounceMs: 0` (`RxAudioInstrumentHost.svelte:296`), and `createHBarContinuousScalarPolicy` maps that to `'immediate'` for every source. Drop "debounced".

Verified and left alone: the guardrail exception line matches `gh pr view 3561 --json files,additions,deletions` exactly (12 files; 237 + 37 = 274). `quick` at `3c5a118a` (run 35664044513) is `completed/success`, so the body's baseline sentence is not an unfinished claim. Run 35664037312 is `Visual regression (pixel-diff)` at `3c5a118a`, `completed/success`; I did not re-measure its six-capture delta and do not rely on it — `visual` is green at this head. Run 35666061790 checks out `f427cefb` (`ref: f427cefb…`, `FOCUSED_REVISION: f427cefb…`) with `Test Files 6 passed`, so the GREEN citation is at the real head despite the run's `headSha` reading `8271e6e1` (`focused.yml` dispatches from `main` and checks out the `revision` input). Every bench and evidence sentence is explicitly attributed to `3c5a118a`, and the Standard face is unchanged by this delta, so none of them now reads as a claim about the current head.

Commit messages `5640ccaa` and `f427cefb` describe the net change accurately. No squash-body correction.

## CI as found (`gh pr checks 3561`)

`quick` **pass** (5m6s), `visual` **pass** (1m26s), `classify` pass, `changes` pass, `grep-gate` pass, `publish-docs-only-quick-status` pass, metadata/observer publishers pass; `quick-v2-observe` pending ("waiting for canonical legacy quick"); `Agent Review Gate` and `Agent Review Gate v2 observe` fail with "Missing trusted standalone exact-head Agent Review PASS directive" — expected before this comment, and no other check is red. Confirming required statuses at the exact head immediately before merge remains the coordinator's step.

